Issue #8, opened 2023-06-15
Vue: `forceViteRestart` function does not always work.


He110te4m:
[`forceViteRestart`](https://github.com/gioboa/vue-microfrontend-demo/blob/vue2_implementation/shared/forceViteRestart.js) uses a fixed delay to restart vite, but this does not always work.

Sometimes I need to rerun `npm run dev` several times.

Can it be integrated into the plugin automatically reload like [`vite-plugin-restart`](https://github.com/antfu/vite-plugin-restart/blob/main/src/index.ts#L78)?

gioboa:
Hi, yes sure.
You can open a PR here https://github.com/gioboa/vue-microfrontend-demo
Then we can also change the examples in module-federation/examples

He110te4m:
> Hi, yes sure. You can open a PR here https://github.com/gioboa/vue-microfrontend-demo Then we can also change the examples in module-federation/examples

No, I think there's been a misunderstanding.
In the example project, I found that the Dev script needs to manually restart each time to make the plug-in's modified configuration entry take effect.
Perhaps we can build this restart logic into the `@module-federation/Vite' .

gioboa:
This could be a good solution

He110te4m:
I've written a custom plugin to do this.

When using Vite serve, there is no need to execute an additional restart script.

Also, when I modify the Federation, Vite is automatically restarted.

The code is as follows:

```typescript
import { readFileSync, writeFileSync } from 'node:fs'
import { type Plugin } from 'vite'
import ViteRestart from 'vite-plugin-restart'

const reloadedSet = new Set<string>()
function reload() {
  const filename = 'vite.config.ts'
  setTimeout(() => {
    writeFileSync(filename, readFileSync(filename, 'utf8'))
  }, 2000)
}

/** 检查 CSS 是否被污染的 vite 插件 */
export function autoReloadPlugin(): Plugin[] {
  return [
    {
      name: 'vite-plugin-micro-service-auto-reload',
      apply: 'serve',
      enforce: 'post',
      buildStart() {
        const service = process.cwd()
        if (!service || reloadedSet.has(service)) {
          return
        }
        reloadedSet.add(service)
        reload()
      },
    },
    ViteRestart({
      restart: [
        'module-federation/federation.config.js',
      ],
    }),
  ]
}
```

gioboa:
Great, thanks for that. Are you using this package for a Vue project?

gioboa:
[...]

PR #9, opened 2023-07-05
Refactor federation init he110

He110te4m:
Reference https://github.com/module-federation/vite/issues/8#issuecomment-1620914897.

Fixed the vite module parsing error caused by the error call timing.

close: #8


gioboa:
Thanks for this PR. I will test it soon

gioboa (approved):
Thanks for your amazing help

PR #15, opened 2023-07-17
Feat base url he110

### LINKED ISSUE #14: When vite configures base, the plugin does not work properly.
More information can be found on [this issue](https://github.com/module-federation/vite/issues/10)

Problem recurrence can be viewed in [this project](https://github.com/He110te4m/module-federation-issue-10-reproduction-repo).

The initial determination is that the contents of dist do not contain the contents of base, causing the URL processed in server not to be added correctly.

The relevant code is on [this line](https://github.com/module-federation/vite/blob/main/src/index.ts#L51).

He110te4m:
close #14 

The problem has two parts:

1. When base is configured, the address of `remoteEntry.json` does not match. The reason is that [this line of code](https://github.com/module-federation/vite/blob/main/src/index.ts#L51) splices the address directly, but Vite's base doesn't affect the path in dist, so the urls don't match.
2. `baseUrl` is not supported when external imports are replaced [in the plugin](https://github.com/module-federation/vite/blob/main/src/dev-externals-mixin.ts#L22). The reason is that the replacement rule only matches `/@id/${libraryName}`. When `baseUrl` is configured, the import path is: `/${baseUrl}/@id/${libraryName}`. So we need to modify the rule to match the quotes directly.

### DIFF
--- a/src/dev-externals-mixin.ts
+++ b/src/dev-externals-mixin.ts
@@ -16,10 +16,16 @@ export const devExternalsMixin = {
   },
   configResolved(resolvedConfig) {
     const VALID_ID_PREFIX = `/@id/`;
-    const reg = new RegExp(`${VALID_ID_PREFIX}(${federationBuilder.externals.join('|')})`, 'g');
+    const reg = new RegExp(
+      `(?<quote>["\'])[^\'"]*?${VALID_ID_PREFIX}(${federationBuilder.externals.join(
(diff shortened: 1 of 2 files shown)


He110te4m:
Sorry, there may be something wrong with this code.

He110te4m:
This change now completely overrides my usage scenario.

PR #27, opened 2024-07-08
Use ESBuild Module Federation instead of Native Federation

renansoares:
This MR contains a draft plugin and a test app to use Module Federation ESBuild instead of Native Federation.

The goal is to investigate if it is possible to leverage ESBuild MF to create a Vite plugin.


renansoares:
I'm facing some issues with the dependencies, ESBuild isn't happy when try to compile/bundle it again.

Vite already uses ESBuild to bundle, using MF ESBuild Plugin is an extra bundle step. I'm starting to wonder if create a brand new plugin and reuse some of the code wrote for the ESBuild plugin is a better approach than to reuse it because of the dependencies issues.

gioboa:
Hi @renansoares thanks for your help.
We can basically rewrite from scratch this Vite plugin to use the esBuild implementation. 💪

> I'm facing some issues with the dependencies, ESBuild isn't happy when try to compile/bundle it again.

How can I reproduce this problem locally?

renansoares:
Closing this in favor of new implementation using `module-federation-vite`.

PR #37, opened 2024-08-06
fix: remove default export

husayt:
default export  is not really needed and a bad practice in general to have both named and default exports of the same module.

Also this is causing number of warnings during build


gioboa (approved):
Thanks @husayt :rocket:

Issue #42, opened 2024-08-09
Multiple runtime loaded


smelukov:
Just use one https://github.com/module-federation/vite/tree/new-config/examples/vite as a host and one another as a remote. In this case, the module federation runtime chunk is loaded multiple times. Sometimes this leads to a race condition and error with the message "Call init first". If we have 10 vite remotes, then 10 runtime chunks will be loaded. 
We need to load and use only one runtime, especially if we use different bundlers (e.g. webpack host, 2 vite remotes, 3 webpack remotes).

@zhangHongEn

![image](https://github.com/user-attachments/assets/8126e9ce-93ef-4ed4-b93f-7fd91f78b561)

gioboa:
Thanks @smelukov  for your feedback. Are you using the latest alpha version?

smelukov:
> Are you using the latest alpha version?

yes

zhangHongEn:
#39 Let me solve this more serious problem first, and then look at this。

Does the "Call init first" error appear randomly after multiple refreshes, or does it appear once every time you start dev for the first time and disappear after a refresh?

smelukov:
@zhangHongEn It seems like "Call init first" appears on the first refresh after the build has started

zhangHongEn:
OK, the dynamic module has a minor flaw when building in vite dev. You need to initiate a request first, that is, refresh the page once and it will be normal.

zhangHongEn:
> Just use one https://github.com/module-federation/vite/tree/new-config/examples/vite as a host and one another as a remote. In this case, the module federation runtime chunk is loaded multiple times. Sometimes this leads to a race condition and error with the message "Call init first". If we have 10 vite remotes, then 10 runtime chunks will be loaded. We need to load and use only one runtime, especially if we use different bundlers (e.g. webpack host, 2 vite remotes, 3 webpack remotes).
> 
> @zhangHongEn
> 
[...]

PR #48, opened 2024-08-11
fix: windows remoteEntry error

zhangHongEn:
fix https://github.com/module-federation/vite/issues/42#issuecomment-2282324217


gioboa (approved):
It looks good to me

PR #50, opened 2024-08-11
updated examples and dependencies

husayt:
* update examples with more features (SASS, named export, etc)
* update dependencies 
* small fixes


gioboa (approved):
Thanks @husayt 
It looks great to me

PR #55, opened 2024-08-15
improved  examples with MUI and etc

husayt:
* added MUI examples
* used vite preview instead of http-server
* more logging
* using react 18.2 on remote and 18.3 on server to show MF in effect
*  more smaller improvements


gioboa (approved):
It looks good to me @husayt 💪👏

Issue #56, opened 2024-08-15
Can't use styled-components as shared


smelukov:
@zhangHongEn
Bug reproduction is [here](https://github.com/smelukov/vite/tree/repro) (repro-branch)

Just start vite-host and vite-remote

```
App1.jsx:4 Uncaught TypeError: styled is not a function
    at App1.jsx:4:16
```

If styled-components is not shared, then all is fine

zhangHongEn:
@smelukov ok fix it tonight

husayt:
@smelukov can you also try that in latest. I added MUI demo with styled components. it will be useful to see how this issue reveals itself in that case.

smelukov:
Reproduction example uses the latest version of plugin

husayt:
what I mean, there is already an example with styled components. Can you try do reproducation with them. Is it just adding styled-component to shared, or there is something else as well?

zhangHongEn:
thanks, I probably know the cause of this problem

smelukov:
@husayt just add styled components into shared

husayt:
@zhangHongEn I am not sure if this is fixed. I still get 
![image](https://github.com/user-attachments/assets/f325d1d3-d51b-464f-8efc-189ecf62cdce)
in the latest, when i set
```
     "@emotion/styled": { singleton: true, },
```

zhangHongEn:
@husayt I tried but couldn't reproduce it. It may not be the same issue

husayt:
> @husayt I tried but couldn't reproduce it. It may not be the same issue

i have created a reproduction in a separate issue here #61. Hopefully, this will be straightforward

Issue #41, opened 2024-08-07
default exports from pathed dependencies


husayt:
Until [this fix](https://github.com/module-federation/vite/commit/7c3ba59d2581ee713b02aee85f5854dbe8e0e114) i had to add `.default`
to use any default exports from shared libraries.
```
import Lib from "sharedLib";
```

[this fix](https://github.com/module-federation/vite/commit/7c3ba59d2581ee713b02aee85f5854dbe8e0e114) had resolved it.

But if my shared library is like this
```
  "@material-ui/core/Grid": { singleton: true },
 ```  
and I import it like this
```
import Grid from "@material-ui/core/Grid";
```
then i still have to add `.default` to get default out
```
 return (
 <Grid.default> </Grid.default>
 return (
```

zhangHongEn:
The need for explicit .default is a bug

zhangHongEn:
I understand that what you shared is not a complete npm package. I have never done this before. This scenario really needs to be enhanced.

PR #58, opened 2024-08-15
fix: #56

zhangHongEn:
fix: #56  
fix: #41

### DIFF
--- a/examples/vite-vite/vite-host/package.json
+++ b/examples/vite-vite/vite-host/package.json
@@ -18,7 +18,6 @@
     "@swc/core": "~1.7.10",
     "@vitejs/plugin-react": "^4.3.1",
     "vite": "^5.4.0",
-    "vite-plugin-top-level-await": "^1.4.4",
-    "http-server": "^14.1.1"
+    "vite-plugin-top-level-await": "^1.4.4"
   }
 }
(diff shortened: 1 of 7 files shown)


examples/vite-vite/vite-host/package.json
```diff
@@ -18,7 +18,6 @@
     "@swc/core": "~1.7.10",
     "@vitejs/plugin-react": "^4.3.1",
     "vite": "^5.4.0",
-    "vite-plugin-top-level-await": "^1.4.4",
-    "http-server": "^14.1.1"
+    "vite-plugin-top-level-await": "^1.4.4"
```
husayt, line 21: @zhangHongEn do we still need vite-plugin-top-level-await? Vite allows async config like this:

```
export default defineConfig(async ({ command, mode }) => {
// do await here
return { }
}
```
can we not use that?

zhangHongEn: This is browser compatibility
https://v8.dev/features/top-level-await

husayt: I see.

husayt (approved):

gioboa (approved):
:rocket: @zhangHongEn

Issue #61, opened 2024-08-16
v1.0.0: error when federating dependencies in prod build


husayt:
this pr #62 shows this error

uncomment any of the following in vite-config. in dev it works. In build preview it doesn't load. 
```
        // "ag-grid-react": {},
        // "@emotion/react": {},
        // "styled-components": { singleton: true },
        // "@emotion/styled": {},
        // "@mui/material": {},
```

No error messages in console

gioboa:
Closed by #63

PR #63, opened 2024-08-19
fix: #61

zhangHongEn:
Modify the "prod-mode: alias.replacement" configuration to enable commonjs-plugin to correctly handle cjs modules


gioboa (approved):
:clap: @zhangHongEn

Issue #40, opened 2024-08-07
v1.0.0: error in shared dependencies ending with "/"


husayt:
Some [examples](https://github.com/module-federation/core/blob/60be41eccf8cd6481e0aa7b633b095f60dfa2166/packages/nextjs-mf/README.md?plain=1#L84) show using dependencies ending with `/`, but when I try the same here I get error:

```
error when starting dev server:
Error: Missing "./" specifier in "react" package
```
 when

```
shared: {
         react:  {  },
         "react/": {  },
...
}
```

zhangHongEn:
I am not familiar with this feature, it can be added to the roadmap

zhangHongEn:
If there is a document, please send it to me. I seem to have missed the document and did not see the function of this feature.

husayt:
you can see that in official next-js MF examples

https://github.com/module-federation/core/blob/60be41eccf8cd6481e0aa7b633b095f60dfa2166/packages/nextjs-mf/README.md?plain=1#L84

smelukov:
Is there any update?

husayt:
@smelukov is it not fixed for you? I thought previous version had a fix. Can you try latest, please

zhangHongEn:
This feature was recently developed

smelukov:
> @smelukov is it not fixed for you? I thought previous version had a fix. Can you try latest, please

nope, still does'n work

smelukov:
Still actual problem

zhangHongEn:
I've been too busy recently, so I'll prioritize it this weekend.

hi-ogawa:
> vite modulePreload will bring "shared module" forward and execute it before init. I can't disable this behavior.

@zhangHongEn This looks like Rollup's `output.hoistTransitiveImports` https://rollupjs.org/configuration-options/#output-hoisttransitiveimports Can you try disabling it?

zhangHongEn:
My external is a dynamic import, but it has been modified to synchronous import. I will provide an example this afternoon


---- Replied Message ----
| From | Hiroshi ***@***.***> |
| Date | 08/26/2024 07:58 |
| To | module-federation/vite ***@***.***> |
| Cc | zhn ***@***.***>,
Mention ***@***.***> |
| Subject | Re: [module-federation/vite] v1.0.0: error in shared dependencies ending with "/" (Issue #40) |

vite modulePreload will bring "shared module" forward and execute it before init. I can't disable this behavior.

[...]

PR #72, opened 2024-08-31
fix: monorepo

zhangHongEn:
Fixes an extreme monorepo scenario where multiple packages share a node_modules directory and multiple `shared` configurations conflict.
In this [demo](https://github.com/module-federation/vite/issues/40#issuecomment-2317363004), there is no conflict between two shared files.


gioboa (approved):
You rock @zhangHongEn

PR #75, opened 2024-09-03
feat: add Nuxt example

David-Sellen:


gioboa (approved):
Thanks @David-Sellen that's amazing. A remote example will be amazing as well, you can create a PR for that if you want.

Issue #78, opened 2024-09-03
v1.0.0: typeError: factory is not a function (after last update)


smelukov:
@zhangHongEn

- Pull https://github.com/smelukov/mf-vite-nx-bug
- run `npm start`
- go to http://localhost:4201/ and see that all is ok
- edit `packages/my-app/src/main.tsx`
```diff
--- a/packages/my-app/src/main.tsx
+++ b/packages/my-app/src/main.tsx
@@ -8,6 +8,6 @@ const root = ReactDOM.createRoot(
 );
 root.render(
   <StrictMode>
-    <App />
+    123
   </StrictMode>
 );
```
- stop `npm start` started on step 2
- run `npm start` again
- go to http://localhost:4201/ and see the error

Workaround to fix:
- stop `npm start`
- remove node_modules/.vite
- run `npm start` again
- go to http://localhost:4201/ and see that all is ok

![image](https://github.com/user-attachments/assets/124830e6-35c9-404d-af7e-82ef545e00cb)

zhangHongEn:
`vite dev --force` needs to add `--force` option

smelukov:
@zhangHongEn 
I have the same error event on module-federation-vite@1.1.0

![image](https://github.com/user-attachments/assets/00867901-6f38-473b-8d2b-220c8ce6c746)

zhangHongEn:
![image](https://github.com/user-attachments/assets/4f6c33e4-d619-4425-9266-ee2849d1e2ea)
I can't reproduce it anymore

zhangHongEn:
![image](https://github.com/user-attachments/assets/d11cd15d-f590-4bdd-9e53-424f3e84af7a)
This totally works for me

zhangHongEn:
> @zhangHongEn I have the same error even on module-federation-vite@1.1.0 (it includes `optimizeDeps.force = true` in node_modules)
> 
[...]

PR #79, opened 2024-09-04
fix: #78

zhangHongEn:
I think this priority is not very high, and we will consider compatibility with pre-built caches in the future
fix: #78

### DIFF
--- a/src/utils/normalizeOptimizeDeps.ts
+++ b/src/utils/normalizeOptimizeDeps.ts
@@ -12,6 +12,7 @@ export default {
       config.optimizeDeps = {};
       optimizeDeps = config.optimizeDeps;
     }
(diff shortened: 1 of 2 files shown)


src/utils/normalizeOptimizeDeps.ts
```diff
@@ -12,6 +12,7 @@ export default {
       config.optimizeDeps = {};
       optimizeDeps = config.optimizeDeps;
     }
+    optimizeDeps.force = true
```
smelukov:
```suggestion
    // todo: fix this workaround
    optimizeDeps.force = true;
```

gioboa (approved):
You are doing great @zhangHongEn thank you

Issue #80, opened 2024-09-04
v1.0.0: add more examples


husayt:
This is notes for me

```
import { useState } from 'react';
import './App.css';
import reactLogo from './assets/react.svg';
// import viteLogo from './assets/vite.svg';

function App() {
  const [count, setCount] = useState(0);

  return (
    <div className="App">
      <div>
        <a href="https://vitejs.dev" target="_blank">
          <img src="/vite.svg" className="logo" alt="Vite logo" />
        </a>
        <a href="https://reactjs.org" target="_blank">
          <img src={reactLogo} className="logo react" alt="React logo" />
        </a>
      </div>
      <h1>Vite + React</h1>
      <div className="card">
        <p>
          Edit <code>src/App.jsx</code> and save to test HMR
        </p>
      </div>
      <p className="read-the-docs">Click on the Vite and React logos to learn more</p>
    </div>
  );
}

export default App;

```

css 
```
.logo {
  height: 6em;
  padding: 1.5em;
  will-change: filter;
}

.logo:hover {
  filter: drop-shadow(0 0 2em #646cffaa);
}

.logo.react:hover {
  filter: drop-shadow(0 0 2em #61dafbaa);
}

@keyframes logo-spin {
  from {
    transform: rotate(0deg);
  }

  to {
    transform: rotate(360deg);
  }
}

@media (prefers-reduced-motion: no-preference) {
  a:nth-of-type(2) .logo {
    animation: logo-spin infinite 20s linear;
  }
}

.card {
  padding: 2em;
}

.read-the-docs {
  color: #888;
}
```

husayt:
```
[...]

Issue #70, opened 2024-08-29
v1.0.0: update MF to v0.6


husayt:
see changes here:
https://github.com/module-federation/core/releases/tag/v0.6.0

New Features 🎉
feat(enhanced): add support for virtual runtime entrypoints by @Coly010 in https://github.com/module-federation/core/pull/2860
Bug Fixes 🐞
fix: fix router alias issue for react-bridge by @danpeen in https://github.com/module-federation/core/pull/2887
Other Changes
Release v0.5.2 by @zhoushaw in https://github.com/module-federation/core/pull/2886
chore: remove useless schemas by @2heal1 in https://github.com/module-federation/core/pull/2888

PR #81, opened 2024-09-04
update module federation and more examples

husayt:
- update @module-federation/runtime to 0.6.0
- add default export to remote
- example with image in public folder
- renderBuiltUrl  instead of explicit base folder


husayt:
solves #70 #80

husayt:
@zhangHongEn also here there is a issue regarding images in public folder. One of the images is in public folder. It works fine in remote standalone, but I couldn't get it loading from host.

image

gioboa (approved):
Thanks @husayt 🚀

PR #105, opened 2024-09-25
feat: manage shareStrategy option

### LINKED ISSUE #103: On-Demand Loading of remoteEntry.js Files
I am currently evaluating a migration from the @originjs plugin. One of the things I liked about their approach was, that they only try to load the remoteEntry.js once a module is required in the code.

Our use case involves many modules behind feature flags. As a result, not all modules are active in a given environment, and pre-loading all remoteEntry.js files can lead to 404 errors or application crashes if a particular remoteEntry.js file is not available when the site initially loads.

Question: Is there a way to configure this plugin to load the remoteEntry.js of a module only when that module is accessed for the first time at runtime? This would significantly improve resilience and performance in environments with varying feature flags.

I have seen https://module-federation.io/guide/basic/runtime.html but since our config does not really change, i would like to keep the config within our vite config

gioboa:
Fixes #103

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -1,4 +1,4 @@
-import { SharedConfig } from '@module-federation/runtime/types';
+import { SharedConfig, ShareStrategy } from '@module-federation/runtime/types';
 
 export type RemoteEntryType =
   | 'var'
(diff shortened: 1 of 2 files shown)


zhangHongEn:
🚀

src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -232,36 +224,31 @@ export type ModuleFederationOptions = {
...
+          }
+      >
+    | undefined;
   runtimePlugins?: string[];
   getPublicPath?: any;
   implementation?: any;
   manifest?: ManifestOptions | boolean;
   dev?: any;
   dts?: any;
+  shareStrategy: ShareStrategy;
```
JP250552, line 251: Should this be optional?

gioboa: Yup, nice catch. Would you like to drop a PR for that? you deserve it

jrparish: Certainly! https://github.com/module-federation/vite/pull/130

PR #110, opened 2024-09-25
feat(example): add vite, webpack and rspack example

RussellCanfield:
Add a simple example of:

Vite - host
Vite - remote
Webpack - remote
RSPack - remote


gioboa (approved):
Thanks @RussellCanfield 
It looks amazing!

image

Issue #112, opened 2024-09-26
feat: support Type Declerations


econneely:
Does this support the hotreloading of types mentioned here and if so is there any examples of how to implement it https://module-federation.io/guide/basic/type-prompt.html

Does this have something to do with it not working https://github.com/module-federation/vite/blob/main/src/plugins/pluginMFManifest.ts#L376-L377

`react refresh:6 [ dynamic-remote-type-hints-plugin ] err: [object Event]` this is also one of the errors on the host side and the remote is using webpack and its manifest has a reference to the .d.ts and zip

gioboa:
Hi @econneely, can you open a PR with a failing scenario in [this example](https://github.com/module-federation/vite/tree/main/examples/vite-webpack-rspack) pls? With your help we will solve the problem sooner. Thanks

econneely:
@gioboa that example fails on the host example with this error `Failed to resolve entry for package "@module-federation/vite". The package may have incorrect main/module/exports specified in its package.json. [plugin externalize-deps`

econneely:
I ran pnpm install then pnpm run multi-example

gioboa:
> I ran pnpm install then pnpm run multi-example

I cloned the repo from scratch and you need to run `pnpm run build`  after the install

econneely:
Update the readme with tha that fixed it will get the pr up asap

gioboa:
> Update the readme with that, it fixed it will get the pr up asap

Fixed, thanks

econneely:
Hi any timeframe on this want to know for planning some work with a Vite project thanks

gioboa:
I'm on it, I need few days to figure out a solution.

econneely:
Thanks much appreciated

gioboa:
The idea is to use the [@module-federation/dts-plugin]( https://www.npmjs.com/package/@module-federation/dts-plugin) because it's always up to date. I'm trying to feed it with a generic compiler. Let's see

Haprog:
The issue title is "Hotreloading Type Declerations" but isn't it so that not even loading once is supported yet (even without hot reload)? Maybe the issue should be renamed to "Support loading Type Declarations (DTS)" or similar?

I would be happy to initially have even a manual way to generate the `@mf-types` directory (either with build plugin or even a separate command I can run) and hot relaoding could be added later (if that is more complicated than implementing loading once).

tduforet:
Does a branch already exist for this issue?

[...]

PR #113, opened 2024-09-26
example: Hotreloading Type Declerations

econneely:
Repo example for #112
Docs here https://module-federation.io/configure/dts.html

Expected a the host to generate a `@mf-types/` as described in here https://module-federation.io/guide/basic/type-prompt.html

### DIFF
--- a/examples/vite-webpack-rspack/host/src/App.jsx
+++ b/examples/vite-webpack-rspack/host/src/App.jsx
@@ -1,3 +1,4 @@
+import { loadRemote } from '@module-federation/runtime';
 import { lazy, StrictMode, Suspense, useState } from 'react';
 import { createRoot } from 'react-dom/client';
 import './App.css';
@@ -6,7 +7,7 @@ import './index.css';
 import viteLogo from '/vite.svg';
 
 const RemoteButton = lazy(() => import('remote/Button'));
-const WebpackButton = lazy(() => import('webpack/Image'));
+const WebpackButton = lazy(() => loadRemote('webpack/Image'));
 const RspackImage = lazy(() => import('rspack/Image'));
 
 function App() {
--- a/examples/vite-webpack-rspack/host/vite.config.js
+++ b/examples/vite-webpack-rspack/host/vite.config.js
@@ -4,6 +4,10 @@ import { defineConfig } from 'vite';
 
 const mfConfig = {
   name: 'host',
+  dts: {
+    consumeTypes: true,
+  },
+  dev: true,
   remotes: {
     remote: {
       entry: 'http://localhost:4001/remoteEntry.js',
--- a/examples/vite-webpack-rspack/webpack/src/Image.tsx
+++ b/examples/vite-webpack-rspack/webpack/src/Image.tsx
@@ -1,9 +1,10 @@
-import React from 'react';
-import { useState } from 'react';
+import React, { useState } from 'react';
 import logo from './assets/webpack-logo.png';
 
-const WebpackImage = () => {
+const WebpackImage = ({ foo }: { foo: number }) => {
   const [count, setCount] = useState(0);
+  console.log(foo);
+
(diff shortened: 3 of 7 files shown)


examples/vite-webpack-rspack/host/src/App.jsx
```diff
@@ -6,7 +7,7 @@ import './index.css';
 import viteLogo from '/vite.svg';
 
 const RemoteButton = lazy(() => import('remote/Button'));
-const WebpackButton = lazy(() => import('webpack/Image'));
+const WebpackButton = lazy(() => loadRemote('webpack/Image'));
```
econneely, line 10: https://module-federation.io/static/image/hot-types-reload.c4031555.gif autocomplete missing

examples/vite-webpack-rspack/webpack/src/Image.tsx
```diff
@@ -1,9 +1,10 @@
-import React from 'react';
-import { useState } from 'react';
+import React, { useState } from 'react';
 import logo from './assets/webpack-logo.png';
 
-const WebpackImage = () => {
+const WebpackImage = ({ foo }: { foo: number }) => {
```
econneely, line 4: missing types on the component in host

econneely: I know its jsx but it can still use and should have the @mf-types/ folder extracted

examples/vite-webpack-rspack/webpack/webpack.config.js
```diff
@@ -69,6 +69,14 @@ module.exports = {
       name: 'webpack',
       filename: 'remoteEntry.js',
       remotes: {},
+      dts: {
+        generateTypes: true,
```
econneely, line 73: with this you can see it creates a dist folder with the types that should be synced to the folder in host

examples/vite-webpack-rspack/webpack/src/Image.tsx
econneely: The foo: number is added for example

gioboa: Thanks 👍 it helps a lot

gioboa:
The @module-federation/enhanced is doing 
```
if (options.dts !== false) {
  new DtsPlugin(options).apply(compiler);
}
```
so I think we need to do the same

examples/vite-webpack-rspack/host/vite.config.js
```diff
@@ -4,6 +4,10 @@ import { defineConfig } from 'vite';
 
 const mfConfig = {
   name: 'host',
+  dts: {
+    consumeTypes: true,
+  },
+  dev: true,
```
econneely, line 10: https://module-federation.io/configure/dev.html some relevent options for types hot reloading too

ScriptedAlchemy:
Yes, but the DTS manager itself is agnostic and could be implemented in other bundlers.

ScriptedAlchemy:
@2hea1 is on leave this week, but he understands the DTS system best. It's quite intricate, but there's a DTS manager that should create a separate server for handling type requests and builds. We can also register dynamic remote types from the browser by sending the data back to the DTS server, which then generates the discovered types.
It shouldn't be tightly coupled to any bundler, but it still has a few moving parts.

ScriptedAlchemy:
Perfect. You can send a pr back to core if additional changes or packages need to be created for dts to support other tools

danielBence:
Hello does closing this branch mean that the vite tool wont be getting remote type hints?

gioboa:
> Hello does closing this branch mean that the vite tool wont be getting remote type hints?

Hope, I really would like to support this feature but I need an help from the official mf team.
You can drop a PR for this if you like, I'm more than happy to collaborate. Thanks

PR #118, opened 2024-09-28
chore: add types for dev and dts properties

IrishBruse:
Add types for dev and dts


gioboa (approved):
Thanks for your help @IrishBruse 
It looks great to me

PR #122, opened 2024-09-29
fix: Support empty entry JS in build mode.

zhangHongEn:
Delete the main.js code of any remote, run preview, and you can reproduce this issue.


gioboa (approved):
Thanks for this improvement @zhangHongEn 👏

PR #130, opened 2024-10-03
fix: changed shareStrategy property to optional in ModuleFederationOptions

jrparish:
`shareStrategy` was implemented as optional but was required in the typing.

Ref: https://github.com/module-federation/vite/pull/105#discussion_r1786523283


gioboa (approved):
Thanks @jrparish

PR #138, opened 2024-10-04
chore: use cjs extension in temporary file

mshima:
cjs/esm depends on package.json's type, would be better to use non ambiguous extension.

Issue #74, opened 2024-09-02
v1.0.0: Question: re `.__mf__win`


husayt:
@zhangHongEn I see new folder `.__mf__win` created when running latest demo. What is it for?

zhangHongEn:
The syntax `shared: "react/"` can only know which sub-paths are imported during the build execution process, and it is unknown at the beginning.
I need to trigger the vite pre-build again after the build is completed, which is achieved by modifying the .__mf__win/localSharedImportMap.js file
Due to this [bug](https://github.com/vitejs/vite/issues/8619), I had to create it outside of node_modules on Windows

husayt:
@gioboa thank you, I forgot to close

@zhangHongEn thanks for clarification

PR #145, opened 2024-10-05
fix: Increase compatibility

zhangHongEn:
fix: #138


gioboa:
Can we move mf temp directory under node_modules?

zhangHongEn:
> Can we move mf temp directory under node_modules?

https://github.com/module-federation/vite/issues/74

PR #148, opened 2024-10-06
docs: add dynamic remote to example

### LINKED ISSUE #146: Examples for Dynamic Remotes
Module Federation has a method to load dynamic remotes: https://module-federation.io/guide/basic/runtime.html 

The webpack plugin has the following application example: https://github.com/module-federation/module-federation-examples/tree/master/advanced-api/dynamic-remotes

It would be great if the examples here could be updated to demonstrate how to use this approach with vite.

judithhartmann:
Closes #146

I also improved the standalone rendering for the remotes in the example to easier work on them in isolation


judithhartmann:
@gioboa i seem to encounter a problem where the `pnpm multiexample` only executes 4 scripts at the same time, we have  5 projects in this example now (the tests work if i start all the remotes manually). 

is there a way to circumvent this pnpm limit?

judithhartmann:
ok i found the `--parallel` flag that seems to disregard all limits

gioboa:
Nice, I didn't know about the 4 script limit. I'm glad you found a solution

gioboa (approved):
@judithhartmann you rock 👏
Thanks for this amazing contribution

jofmi:
Thank you!

PR #153, opened 2024-10-10
fix: increase compatibility during build process

zhangHongEn:


gioboa (approved):
Thanks @zhangHongEn :clap:

PR #167, opened 2024-10-29
fix: issue with multiple init calls in remote

zhangHongEn:
hotfix:
If the remote is configured as `exposes: {".": "./src/main"}`, the remote will automatically call runtime.init(), which may cause intermittent errors in certain network environments.


gioboa (approved):
It looks great to me @zhangHongEn Thanks you rock

Issue #189, opened 2024-11-22
Adding hostInit does not work if there is no html file (SvelteKit)


RobinBomkamp:
If you want to use module-federation/vite in combination with SvelteKit. The hostInit file does not get included in the generated output. The file is compiled, but the reference is missing.

It used to work in version 1.1.3 of this package. 

In version 1.1.4 there was a fix for the issue #167. This caused the problem.
After the fix the hostInit will be referenced directly in the html <head> element in a script tag. Therefore, the plugin check every file ending with '.html' (see [pluginAddEntry.ts](https://github.com/module-federation/vite/blob/main/src/plugins/pluginAddEntry.ts#L117))

SvelteKit does not generate any HTML files. The <head> element is generated on the server side due to ssr. A template of the base html file is part of the generated internal.js file.

```typescript
...addEntry({
      entryName: 'hostInit',
      entryPath: getHostAutoInitPath(),
      inject: 'html',
    }),
``` 

The `inject: 'html'` causes the problem. If you are setting this value to 'entry' or simply remove it the hostInit will be referenced as before the issue #167.

Should this field be made configurable? Or can it be injected as an entry for svelte kit projects as before?

zhangHongEn:
I noticed another issue remains unresolved. Can version 1.1.3 be successfully used in SvelteKit?

zhangHongEn:
If the SSR environment can be detected, use inject: 'entry'.

RobinBomkamp:
Issue can be closed :)
It is part of version 1.1.8.

gioboa:
Thanks @RobinBomkamp

PR #193, opened 2024-11-26
fix: Inject hostInit as entry if there are no html entry files

RobinBomkamp:
fix:
#189

### DIFF
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -9,6 +9,10 @@ interface AddEntryOptions {
   inject?: 'entry' | 'html';
 }
 
+function getFirstHtmlEntryFile(entryFiles: string[]): string | undefined {
+  return entryFiles.find((file) => file.endsWith('.html'));
+}
+
 const addEntry = ({
   entryName,
   entryPath,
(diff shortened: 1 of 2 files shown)


src/plugins/pluginAddEntry.ts
```diff
@@ -127,7 +141,7 @@ const addEntry = ({
         }
       },
       transform(code, id) {
-        if (inject === 'entry' && entryFiles.some((file) => id.endsWith(file))) {
+        if (injectEntry() && entryFiles.some((file) => id.endsWith(file))) {
```
gioboa, line 143: Thanks for your help @RobinBomkamp 
Here we are changing the behaviour for the non SSR scenario.
Am I missing something?

RobinBomkamp: It should not change the non SSR scenario. 

There are 3 times the add-entry-plugin will be called two time with 'inject:entry' and one time with 'inject:html'. 

If the value is 'entry' there is everything normal. Only if the value is 'html' then there will be a different behavior if your build is an ssr build.

gioboa:
```
function injectEntry() {
  return inject === 'entry' || _ssrBuild; <-- so we can remove this I guess
}
```

RobinBomkamp: No, that is the difference needed to fix the issue. 

Html should be injected like an 'entry' in ssr scenarios.

RobinBomkamp:
Is it correct that the tests fail?

I couldn't get it running on my local machine.

src/plugins/pluginAddEntry.ts
RobinBomkamp: It seems like there is a bug in my ssr detection. I will fix it.

RobinBomkamp: The merge request is now ready.

I have removed the ssr detection. The underlying problem was that there is no html file in a ssr build. However the htmlFilePath value was set to the first entry file. But there was no check if the first entry file is an html file.

More over if there is no html file injections with the type html will be injected like an entry.

gioboa (approved):
Great :+1: Thanks @RobinBomkamp 
I'm really happy to merge your first PR in this project, kudos :partying_face:

koga73:
Is there a way to instruct federation to import into entry instead of html? For my use case 3rd party applications consume my js and I want it bundled instead of having to instruct them to add a separate import for hostInit.

PR #195, opened 2024-11-26
fix: vitest should be a devDependency

yuvalkarif:


gioboa (approved):
Thanks @yuvalkarif 
Kudos for your first contribution into this project :partying_face:

PR #131, opened 2024-10-03
fix: don't try to discover shared item version in case of custom version or invalid package name

mshima:
Exposing a local share causes lots of errors in console:
```js
Error: Cannot find module '@/package.json'
Require stack:
- /Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1140:15)
    at Module._load (node:internal/modules/cjs/loader:981:27)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at normalizeShareItem (/Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:296:15)
    at /Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:335:21
    at Array.forEach (<anonymous>)
    at normalizeShared (/Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:334:25)
    at normalizeModuleFederationOptions (/Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:379:13)
    at federation (/Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:1236:17) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs'
  ]
}
Error: Cannot find module '@/package.json'
Require stack:
- /Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1140:15)
    at Module._load (node:internal/modules/cjs/loader:981:27)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at normalizeShareItem (/Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:296:15)
    at /Users/mshima/git/jhipster-samples/vite-mf/notification/node_modules/@module-federation/vite/lib/index.cjs:335:21
[...]

PR #216, opened 2024-12-07
fix: try to find shared package.json

### LINKED ISSUE #152: share `chart.js` package: error Package subpath './package.json' is not defined by "exports"
In the federation plugin when adding a dependency which has multiple exports paths, as shared dependency in remote like this:

```
federation({ 
      name: "remoteapp",  
      filename: "remoteEntry.js",  
      remotes: {},  
      exposes: {. 
        './RemoteApp: './src/bootstrap',  
      },  
      shared: {. 
        "some-package-name":"1.0.0". 
     } 
})
```

getting following error:
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in \node_modules\<some-package>\package.json
    at exportsNotFound (node:internal/modules/esm/resolve:304:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:651:9)
    at resolveExports (node:internal/modules/cjs/loader:592:36)
    at Module._findPath (node:internal/modules/cjs/loader:669:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1131:27)
    at Module._load (node:internal/modules/cjs/loader:986:27)
    at Module.require (node:internal/modules/cjs/loader:1233:19)
    at require (node:internal/modules/helpers:179:18)
    at normalizeShareItem ()
    at C:\Users\\Desktop\vite\UI\node_modules\@module-federation\vite\lib\index.cjs:320:21 {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```
Vite version : 5.4.8
Node.js: 20.16.0

nicotu01:
Fix #152

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -21,7 +21,7 @@ export type RemoteEntryType =
   | 'system'
   | string;
 
-import { existsSync, readFileSync } from 'fs';
+import { existsSync } from 'fs';
 import * as path from 'pathe';
 import { warn } from './logUtils';
 
@@ -122,10 +122,13 @@ function findPackageJson(moduleName: string) {
   const mainFilePath = require.resolve(moduleName);
 
   let currentDir = path.dirname(mainFilePath);
-  while (currentDir !== path.parse(currentDir).root) {
+  while (
+    path.parse(currentDir).base !== 'node_modules' &&
+    currentDir !== path.parse(currentDir).root
+  ) {
     const potentialPackageJsonPath = path.join(currentDir, 'package.json');
     if (existsSync(potentialPackageJsonPath)) {
-      return JSON.parse(readFileSync(potentialPackageJsonPath, 'utf8'));
+      return require(potentialPackageJsonPath);
     }
     currentDir = path.dirname(currentDir);
   }
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -195,13 +195,13 @@ function normalizeShared(
   const result: NormalizedShared = {};
   if (Array.isArray(shared)) {
     shared.forEach((key) => {
-      result[key] = normalizeShareItem(key, key);
+      result[cleanShareItem(key)] = normalizeShareItem(key, key);
     });
     return result;
   }
   if (typeof shared === 'object') {
     Object.keys(shared).forEach((key) => {
-      result[key] = normalizeShareItem(key, shared[key] as any);
+      result[cleanShareItem(key)] = normalizeShareItem(key, shared[key] as any);
     });
   }
 
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -107,17 +107,6 @@ export interface ShareItem {
   shareConfig: SharedConfig;
 }
 
-function removePathFromNpmPackage(packageString: string): string {
-  // 匹配npm包名的正则表达式，忽略路径部分
-  const regex = /^(?:@[^/]+\/)?[^/]+/;
-
-  // 使用正则表达式匹配并提取包名
-  const match = packageString.match(regex);
-
-  // 返回匹配到的包名，如果没有匹配到则返回原字符串
-  return match ? match[0] : packageString;
-}
-
 function findPackageJson(moduleName: string) {
   const mainFilePath = require.resolve(moduleName);
 
@@ -151,7 +140,7 @@ function normalizeShareItem(
 ): ShareItem {
   let version: string | undefined;
   try {
-    version = findPackageJson(removePathFromNpmPackage(key)).version;
+    version = findPackageJson(key).version;
   } catch (e) {
     console.log(e);
   }
@@ -332,9 +321,7 @@ export function getNormalizeModuleFederationOptions() {
 
 export function getNormalizeShareItem(key: string) {
   const options = getNormalizeModuleFederationOptions();
-  const shareItem =
-    options.shared[removePathFromNpmPackage(key)] ||
-    options.shared[removePathFromNpmPackage(key) + '/'];
+  const shareItem = options.shared[key] || options.shared[key + '/'];
   return shareItem;
 }
 
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -107,6 +107,10 @@ export interface ShareItem {
   shareConfig: SharedConfig;
 }
 
+function cleanShareItem(key: string) {
+  return key.replace(/^\//, '').replace(/\/$/, '');
+}
+
 function findPackageJson(moduleName: string) {
   const mainFilePath = require.resolve(moduleName);
 
@@ -139,14 +143,15 @@ function normalizeShareItem(
       }
 ): ShareItem {
   let version: string | undefined;
+  const shareName = cleanShareItem(key);
   try {
-    version = findPackageJson(key).version;
+    version = findPackageJson(shareName).version;
   } catch (e) {
     console.log(e);
   }
   if (typeof shareItem === 'string') {
     return {
-      name: shareItem,
+      name: shareName,
       version,
       scope: 'default',
       from: '',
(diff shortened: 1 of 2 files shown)


nicotu01:
Hi @irwinarruda 
Thanks for the review. 
I will take your first remark into account.  
Concerning the second, the use of require to read the package.json is precisely the subject of this ticket. If the package.json is not exported, then it is not possible to import it like this. Hence the use of readFile.  

Finally, for the third point. Do you use module federation in node contexte ?

irwinarruda:
The problem we had was because `removePathFromNpmPackage(key)` only returned the name of the package. The end result would be something like `require("packageName/package.json")`, causing an error.

In your implementation, the `potentialPackageJsonPath` is an absolute path, so `require` should be fine (eg `require("/Users/.../packageName/package.json")`). However, I'm not that confident to say `require` is much better than `JSON.parse` in this case.

For some reason, I had those Browserify packages for emulating Node Apis in the browser. I don't think I even use them. Either way, for those libs, the `package.json` returned by `findPackageJson` is the project one and not from the `node_modules/packageName`.

irwinarruda:
I also think #131 will affect this PR

gioboa:
@irwinarruda It looks like you've carefully debugged the change, thanks

> I also think #131 will affect this PR if it is merged.

This PR is stale, we can merge this one an than try to complete that one too.

nicotu01:
Hi @irwinarruda 
I understand your remarks and take them in a new commit, Thanks.
I don't understand to why browserify doesn't work

src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -117,6 +118,24 @@ function removePathFromNpmPackage(packageString: string): string {
...
+    currentDir !== path.parse(currentDir).root
+  ) {
+    const potentialPackageJsonPath = path.join(currentDir, 'package.json');
+    if (existsSync(potentialPackageJsonPath)) {
+      return require(potentialPackageJsonPath);
+    }
+    currentDir = path.dirname(currentDir);
+  }
+
+  throw new Error(`Unable to find package.json for the module "${moduleName}"`);
```
gioboa, line 129: I like it @nicotu01 :+1:
A good error message is always valuable

irwinarruda:
**It's also working for me. Thank you!**

> I don't understand to why browserify doesn't work

I'm sure it's because of the `require.resolve(moduleName)` function. Since Vite runs in a Node environment, it already has those native packages.
```ts
console.log("browserify", require.resolve("browserify"));
console.log("process", require.resolve("process"));
console.log("buffer", require.resolve("buffer"));
// browserify /Users/irwinarruda/Documents/project/node_modules/.pnpm/browserify@17.0.1/node_modules/browserify/index.js
// process process
// buffer buffer
```
The good thing now is that the code throws an error instead of getting the wrong `package.json`.
```
Error: Unable to find package.json for the module "buffer"
```
I think if we try to fit those packages into this implementation, we'll have to change everything... In my case, those were used by Webpack and should've been in devDependencies.

irwinarruda:
After I sent the previous message, I tested the code in another MF project. The `firebase` package is now throwing an error.

```bash
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/irwinarruda/Project/node_modules/firebase/package.json
```

Firebase does not have a main entry file. From what I understand, it's only meant to be used by its exports, such as `firebase/app` and `firebase/auth`, but it only has one package.json. I'm not sure if sharing Firebase was working properly before this change since I'm not that familiar with this code base.

nicotu01:
Yes, there's something wrong.

After some investigations on webpack federation plugin.

- we can't share `firebase` because there is no default import.
- we can share `firebase/app` : the version will be firebase package version
- `firebase/app` point on `@firebase/app`. We can also share `@firebase/app` : the version will be `@firebase/app` package version which is different from `firebase/app`

I think the solution is find package.json version from something that can be imported (resolved). But the function `removePathFromNpmPackage` removes the path :  

```
console.log(removePathFromNpmPackage('firebase/app'));
// firebase
console.log(removePathFromNpmPackage('@firebase/app'));
// @firebase/app
```

So, can we remove the usage of the function `removePathFromNpmPackage` ?

nicotu01:
> it's appropriate to throw an error when trying to share firebase

Yes, I think too.
But actually, we can't share `firebase/app` because of `removePathFromNpmPackage`.

I think `removePathFromNpmPackage` is not suitable. Can we remove this function ?

gioboa:
> I think `removePathFromNpmPackage` is not suitable. Can we remove this function ?

In that file is used only there, so yes.

zhangHongEn:
1.	Please test the Nx multi-package scenario where node_modules is not located under the current project.
	2.	removePathFromNpmPackage cannot be removed, as it is used for these two cases:
shared: ['react/', 'react/xxxxx.js'].

zhangHongEn:
Make sure the tests pass, and you’re welcome to merge.

nicotu01:
Hi @zhangHongEn 
I clean the shared item name. 

So, You can share `react` and `react/`

You can also share `react/jsx-runtime` because it is exported by react.

But you can't share `react/jsx-runtime.js` because it is **not exported** by react.

Here is the exports from react package: 

```
// extract of package.son from react
{
  "name": "react",
  "exports": {
    ".": {
      "react-server": "./react.shared-subset.js",
      "default": "./index.js"
    },
    "./package.json": "./package.json",
    "./jsx-runtime": "./jsx-runtime.js",
    "./jsx-dev-runtime": "./jsx-dev-runtime.js"
  },
 
```

irwinarruda:
For my use case, it's all working fine... However, there is an error using `react/`:
```ts
    federation({
      name: 'host',
      manifest: true,
      filename: 'remoteEntry.js',
      shared: {
        'react/': { singleton: true },
      },
    }),
```
```
[module-federation-manifest] Cannot read properties of undefined (reading 'version')
```
I believe it's an error in the `pluginMFManifest.ts` file line `266`.

irwinarruda:
This issue is starting to get more complex than I've imagined...

I don't know if it's the best one, but the easiest solution is to keep `removePathFromNpmPackage` and change the `getNormalizeShareItem` to include every possibility.

```ts
function getNormalizeShareItem(key) {
  var options = getNormalizeModuleFederationOptions();
  var shareItem =
    options.shared[cleanShareItem(key)] ??
    options.shared[removePathFromNpmPackage(key)] ??
    options.shared[removePathFromNpmPackage(key) + '/'];
  return shareItem;
}
```

gioboa:
@irwinarruda thanks for your commitment, I think we can ship a solution and then iterate

nicotu01:
@irwinarruda I push a new commit for your usecase.
Is it ok for you ?

irwinarruda:
Hi @nicotu01, I've tested it and there still is a problem with the `package/` way of sharing modules. It does not error, but the end result is not right.

@gioboa, correct me if I'm wrong, but when we share something like:

```ts
      shared: {
        'react/': { singleton: true },
        'firebase/': { singleton: true },
      },
```
The goal is to automatically share every submodule if it is used in the app.

Let's say I'm using `react/jsx-runtime`. If I share `react/`, I expect to see a build chunk with the `jsx-runtime` code. When I tested your implementation with this example the chunk was not created (you can check the `remoteEntry.js` file for confirmation). It only works if you explicitly pass the module name and don't rely on the trailing slash for it.

This implementation already solves the issue we started with, but it's sometimes breaking default behavior.

gioboa:
Yep we need to respect the [official behaviour](https://webpack.js.org/plugins/module-federation-plugin/)

irwinarruda:
Hey @nicotu01, did you understand why there is still a bug in the code? I think the best solution is:

> ```ts
> function getNormalizeShareItem(key) {
>   var options = getNormalizeModuleFederationOptions();
>   var shareItem =
>     options.shared[cleanShareItem(key)] ??
>     options.shared[removePathFromNpmPackage(key)] ??
>     options.shared[removePathFromNpmPackage(key) + '/'];
>   return shareItem;
> }
> ```

If you have any problems, feel free to reach out, maybe I can help you.

irwinarruda:
@nicotu01, no problem. I'll give it a go. Tomorrow I should start and maybe have a solution.

gioboa (commented):
Thanks @nicotu01 for your help 🙏
@irwinarruda I close this one. Feel free to open a new one.
Thanks for your help as well 👏

Issue #214, opened 2024-12-06
vite build fail,when use vite-plugin-top-level-await


menghany:
When building my project, I want to ensure compatibility with older browsers, so I modified the build configuration in the demo. However, after completing the build, it fails to run.
![image](https://github.com/user-attachments/assets/e6798c54-c624-4069-b401-c38a1ce25f86)
my build command success
`cd /examples/vite-vite/vite-remote`
`pnpm run build`
When i run 
`pnpm run preview`
is error
![image](https://github.com/user-attachments/assets/31a88292-b77e-4052-a811-ec1fde796a7f)
![image](https://github.com/user-attachments/assets/04d4556b-3460-40ac-8450-ddb961094561)

gioboa:
Hi @menghany,
I reproduced the problem and hacked the compiled like this

from 
```
i()
```

to 
```
setTimeout(() => i(), 0);
```

and now it's working. So I'm woundering if it's a specific problem of `vite-plugin-top-level-await` :thinking:

deizianens:
I have the exact same error on my app, also using `vite-plugin-top-level-await ` and running `vite preview `command. Did you find any solution to this?

gioboa:
You can use the build -> target Vite config.
Which browser do you need to support?

deizianens:
@gioboa The most used browsers: chrome, safari, edge...
Think I can use build target: esnext, right?

Issue #210, opened 2024-12-03
TypeError (analogjs): Cannot read properties of undefined (reading 'push') when launching shell


Seifenn:
When i launch my application using module federation i get the following error

TypeError: Cannot read properties of undefined (reading 'push')

When i enter inside the code of the federation module i noticed that this part is causing the error inside index.cjs

```javascript

     name: 'watchLocalSharedImportMap',
    apply: 'serve',
    config: function config(_config3) {
      _config3.optimizeDeps = defu.defu(_config3.optimizeDeps, {
        exclude: [getLocalSharedImportMapPath()]
      });
      _config3.server = defu.defu(_config3.server, {
        watch: {
          ignored: []
        }
      });
      var watch = _config3.server.watch;
      watch.ignored = [].concat(watch.ignored);
      watch.ignored.push("!**" + getLocalSharedImportMapPath() + "**");
    }
  }];
```
when i log the value of the watch variable i get false but it should be   

```javascript
watch: {
          ignored: []
        }
```
So i initialized it to this value and its working i can run the shell and the remote loaded in the shell

```javascript

       name: 'watchLocalSharedImportMap',
    apply: 'serve',
    config: function config(_config3) {
      _config3.optimizeDeps = defu.defu(_config3.optimizeDeps, {
        exclude: [getLocalSharedImportMapPath()]
      });
      _config3.server = defu.defu(_config3.server, {
        watch: {
          ignored: []
        }
      });
      var watch = _config3.server.watch;
      console.log(watch)
      watch =  {
        ignored: []
      }
      watch.ignored = [].concat(watch.ignored);
      watch.ignored.push("!**" + getLocalSharedImportMapPath() + "**");
    }
  }];
}
```
here the value of watch inside _config3.server is false and the code should change this value

```javascript
 _config3.server = defu.defu(_config3.server, {
        watch: {
          ignored: []
        }
      });
```

 But its not overriding the value of watch . I dont know if this error is happening for a specific case in my project  since i'm using analog js as a host or it should be corrected when the value is false it should be initialized if the defu function does not update the value my project can be found on github [vite-module-federation-angular-test](https://github.com/Seifenn/vite-module-federation-angular-test).

gioboa:
Hi @Seifenn I didn't look into the code yet 
What's the difference between your example and the [official one](https://github.com/module-federation/module-federation-examples/tree/master/module-federation-vite-angular)?

Seifenn:
[...]

PR #222, opened 2024-12-15
fix: #210 #214 hack vite-plugin-top-level-await

zhangHongEn:
* fix: #214 
* fix: #210


gioboa (approved):
Nice :clap: 
Thanks @zhangHongEn

PR #246, opened 2025-01-30
fix: create workaround for error in packages with wrong exports

irwinarruda:
Solves #152

After studying the code, I took a different approach. Instead of changing the default way `package.json` was being fetched, I only execute the workaround solution if there is an error with the lib (i.e., there is no `package.json` inside `exports`).

Though it executes the "same operation" twice, it only happens sometimes and this implementation makes the code more maintainable because there are no side effects.

Besides that, I found some other minor issues that don't have anything to do with this issue... should I commit them to this PR? I'll give you an example:

- The way `normalizeShareItem(key: string, shareItem: ....)` is being used, `shareItem` can be the name or the version of the package causing some inconsistencies. I don't think this "bug" is affecting anything, but it is there.


gioboa (commented):
Thanks @irwinarruda I will check your PR :clap:

gioboa (commented):
Thanks @irwinarruda I will check your PR :clap:

gioboa (commented):
Thanks @irwinarruda I will check your PR :clap:

gioboa (commented):
It looks great
Thanks @irwinarruda 🎉
we can add few libraries eg.
`chalk , chart.js, ramda, react-chartjs-2, react-dropzone, react-easy-crop, sass`
in our multi-example app.
That app is under e2e test, a second page with all these libraries can be great to certify the implementation I guess

irwinarruda:
Hey @gioboa, can you check if my latest commit is what you had in mind?

I decided only to download `ramda` (which I've confirmed was throwing an error before) because I didn't want to make the example app too complex for people only looking to learn Module Federation. I've also used a function from `ramda` to confirm it was actually being bundled and I've incremented the e2e test suit to fit the `remote-vite` component.

I could've added some of the other libraries, but all of them seemed too complex for the example app which is meant to be simple. Please let me know if you have another idea in mind.

Should I create an issue reporting this inconsistency?
> Besides that, I found some other minor issues that don't have anything to do with this issue... should I commit them to this PR? I'll give you an example:
> 
>     * The way `normalizeShareItem(key: string, shareItem: ....)` is being used, `shareItem` can be the name or the version of the package causing some inconsistencies. I don't think this "bug" is affecting anything, but it is there.

gioboa (commented):
> Should I create an issue reporting this inconsistency?

Thanks, let's do another PR for this

gioboa (commented):
Thanks for the update on the e2e app and I understand your concern about the basic example but I really would like to cover few problematic packages with tests.
We have a GH action for unit tests, maybe we can use that action rather than the e2e one.
What do you think? Sorry for the back and forth

irwinarruda:
That's cool, I have no problem adding those other libraries to the example app. If you think it's good, we can even create a page or another remote to add those types of tests. It would be something like Tanstack Table's "Kitchen Sink" example, where we test all sorts of things.

Unit tests are difficult because `vitest` runs the main code, so we don't want to download unnecessary dependencies. Because we use a lot of `require` and `require.resolve`, the only way I see it working is by mocking those functions, which is hard. Do you think this is the way?

No problem with the back and forth, I'm still learning the codebase. 
I'm happy to help either way. What approach should I take?

gioboa (commented):
That's a good idea, let's create another remote with these libraries. I think it's good to cover those special libraries with tests.

irwinarruda:
Hey @gioboa, I've created the new testing remote and made sure it is used inside the muiti-example app so we can e2e test it! 

From the list you gave me, I only didn't download `sass` and `chalk` because those are node libraries and shouldn't be a shared dependency.

Here is how it looks.

image
image

gioboa (commented):
I reviewed it briefly from my mobile and it looks great.
I will test it in few hours.
Thanks for this improvement 💪

gioboa (approved):
That's perfect :clap:
Thanks you so much for this amazing job @irwinarruda :partying_face:

Issue #156, opened 2024-10-11
vite with SSR: require is not defined


omridevk:
when using SSR, I am getting the following error:
```
11:02:24 AM [vite] Error when evaluating SSR module /node_modules/__mf__virtual/host_mf_2_page_mf_2_app__loadRemote__remote_mf_1_remote_mf_2_app__loadRemote__.js:
|- ReferenceError: require is not defined
```

see stackblitz for reproduction:
https://stackblitz.com/edit/github-ad1c93-e4uuvj?file=vite.config.ts,src/main.jsx,test.js

omridevk:
run 
`npm run test` 
to see the error

gioboa:
Thanks @omridevk for your feedback. We still need to focus into the SSR story, would you like to tackle the problem?

omridevk:
sure, it was quite easy to resolve this by using 
```
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);
```
and using
```default export```
instead of the CJS module.exports

**Comment by gioboa** [comment]
@omridevk awesome :clap: I'm looking forward to testing your PR :rocket:

**Comment by RobinBomkamp** [comment]
We are also facing the same problem in SvelteKit. It was working in version 0.2.8. see #163

**Comment by omridevk** [comment]
Yes sorry, had to shift my focus on something else, hopefully will get to this, this week

**Comment by veeramarni** [comment]
I couldn't find a workaround for this issue. I'm testing this with Remix app.

**Comment by omridevk** [comment]
@gioboa  I was able to get the preview working, but I can't seem to be able to run tests, it is running into some memory issues, i'll add the logs later today

**Comment by omridevk** [comment]
`virtualRemotes.ts`

```
import VirtualModule from '../utils/VirtualModule';
import { virtualRuntimeInitStatus } from './virtualRuntimeInitStatus';

const cacheRemoteMap: {
  [remote: string]: VirtualModule;
} = {};
export const LOAD_REMOTE_TAG = '__loadRemote__';
export function getRemoteVirtualModule(remote: string, command: string) {
  if (!cacheRemoteMap[remote]) {
    cacheRemoteMap[remote] = new VirtualModule(remote, LOAD_REMOTE_TAG, '.js');
    cacheRemoteMap[remote].writeSync(generateRemotes(remote, command));
  }
  const virtual = cacheRemoteMap[remote];
  return virtual;
}
const usedRemotesMap: Record<string, Set<string>> = {
[...]

PR #250, opened 2025-02-06
fix: #156 ssr: require is not defined

poenneby:
Further to this discussion: https://github.com/module-federation/vite/issues/156#issuecomment-2428952385

### DIFF
--- a/src/plugins/pluginProxyRemotes.ts
+++ b/src/plugins/pluginProxyRemotes.ts
@@ -17,7 +17,7 @@ export default function (options: NormalizedModuleFederationOptions): Plugin {
           find: new RegExp(`^(${remote.name}(\/.*|$))`),
           replacement: '$1',
           customResolver(source: string) {
-            const remoteModule = getRemoteVirtualModule(source, _command);
+            const remoteModule = getRemoteVirtualModule(source, _command, remote.type === 'module');
             addUsedRemote(remote.name, source);
             return remoteModule.getPath();
           },
--- a/src/virtualModules/virtualRemotes.ts
+++ b/src/virtualModules/virtualRemotes.ts
@@ -5,10 +5,10 @@ const cacheRemoteMap: {
   [remote: string]: VirtualModule;
 } = {};
 export const LOAD_REMOTE_TAG = '__loadRemote__';
-export function getRemoteVirtualModule(remote: string, command: string) {
+export function getRemoteVirtualModule(remote: string, command: string, esm = false) {
   if (!cacheRemoteMap[remote]) {
     cacheRemoteMap[remote] = new VirtualModule(remote, LOAD_REMOTE_TAG, '.js');
-    cacheRemoteMap[remote].writeSync(generateRemotes(remote, command));
+    cacheRemoteMap[remote].writeSync(generateRemotes(remote, command, esm));
   }
   const virtual = cacheRemoteMap[remote];
   return virtual;
@@ -23,14 +23,14 @@ export function addUsedRemote(remoteKey: string, remoteModule: string) {
 export function getUsedRemotesMap() {
   return usedRemotesMap;
 }
-export function generateRemotes(id: string, command: string) {
+export function generateRemotes(id: string, command: string, esm = false) {
   return `
-    import { createRequire } from 'node:module';
-    const require = createRequire(import.meta.url);
+    import {createRequire} from 'node:module'
+    const require = createRequire(import.meta.url)
     const {loadRemote} = require("@module-federation/runtime")
     const {initPromise} = require("${virtualRuntimeInitStatus.getImportId()}")
     const res = initPromise.then(_ => loadRemote(${JSON.stringify(id)}))
     const exportModule = ${command !== 'build' ? '/*mf top-level-await placeholder replacement mf*/' : 'await '}initPromise.then(_ => res)
-    export default exportModule;
+    ${esm ? `export default exportModule` : `module.exports = exportModule`}
   `;
 }
--- a/src/virtualModules/virtualRemotes.ts
+++ b/src/virtualModules/virtualRemotes.ts
@@ -25,10 +25,12 @@ export function getUsedRemotesMap() {
 }
 export function generateRemotes(id: string, command: string) {
   return `
+    import { createRequire } from 'node:module';
+    const require = createRequire(import.meta.url);
     const {loadRemote} = require("@module-federation/runtime")
     const {initPromise} = require("${virtualRuntimeInitStatus.getImportId()}")
     const res = initPromise.then(_ => loadRemote(${JSON.stringify(id)}))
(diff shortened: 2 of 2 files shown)


gioboa (commented):
Thanks @poenneby for your help 🤗

poenneby:
@gioboa Running the multi-example with these changes correctly `export default exportModule` but at runtime I get "unexpected token: default"
Tracing this back in the code the `export default` gets incorrectly identified as a named export in `src/plugins/pluginDevProxyModuleTopLevelAwait.ts`

The transpiled code that causes the error:
```js
// node_modules/__mf__virtual/host__loadRemote__remote_mf_1_Product__loadRemote__.js
var import_node_module = __toESM(require_node_module());
var { loadRemote } = require2("@module-federation/runtime");
var { initPromise } = require2("__mf__virtual/host__mf_v__runtimeInit__mf_v__.js");
var res = initPromise.then((_) => loadRemote("remote/Product"));
var exportModule = (
  /*mf top-level-await placeholder replacement mf*/
  initPromise.then((_) => res)
);
var require2 = (0, import_node_module.createRequire)(import.meta.url);
var host_loadRemote_remote_mf_1_Product_loadRemote_default = exportModule;

              const __mfproxy__awaitdefault = await default(); <-- HERE
              const __mfproxy__default = () => __mfproxy__awaitdefault;
            
export { __mfproxy__default as default };
```

Any ideas?

gioboa:
I see, we can remove the plugin and use the Vite build configuration [here is the code](https://github.com/module-federation/vite/blob/main/README.md?plain=1#L119)

gioboa:
I see, I will check it. Thanks for the deep investigation

gioboa:
I did look at it yet. But eventually we can continually change the code if an SSR env is detected

poenneby:
Hey @gioboa ! Are you looking at this. Would be happy to help but I am struggling to understand the inner workings. Thanks!

poenneby:
> Not yet but I will look at it. Can you share your repository to test this change please?

Here you go! https://github.com/poenneby/mf-vanilla-react-router

tiagobnobrega:
Another idea is to use a plugin (https://github.com/originjs/vite-plugins/tree/main/packages/vite-plugin-commonjs) to convert CJS to ESM on the host doing SSR and requesting the module. 
I'll try out a few more things today, but I'm not sure how long I can spend on this ATM.

tiagobnobrega:
ok... this is a slightly modified code, that doesn't seem to affect top level await logic:

```typescript
export function generateRemotes(id: string, command: string) {
  return `
    const runtimePromise = import('@module-federation/runtime');
    const virtualRuntimePromise = import("${virtualRuntimeInitStatus.getImportId()}");
    const res = Promise.all([runtimePromise, virtualRuntimePromise])
      .then(([runtime, virtualRuntime]) => {
        const initPromise = virtualRuntime.initPromise;
        return initPromise.then(() => runtime.loadRemote(${JSON.stringify(id)}));
      });
    const exportModule = ${command !== 'build' ? '/*mf top-level-await placeholder replacement mf*/' : 'await '}res;

    // Dual exports for CJS/ESM compatibility
      module.exports.default = exportModule;
    // ESM export fallback
      const _exports = exports || {};
      Object.defineProperty(_exports, '__esModule', { value: true });
      exports.default = exportModule;
  `;
}
```

but the weird thing is that now the build fails, from what it seems like we want the top level await transform to do.

```
error during build:
[commonjs--resolver] node_modules/__mf__virtual/host__loadRemote__remote_mf_1_Product__loadRemote__.js (9:31): await isn't allowed in non-async function
file: /Users/tiago/ws/other/vite/examples/vite-webpack-rspack/host/node_modules/__mf__virtual/host__loadRemote__remote_mf_1_Product__loadRemote__.js:9:31

 7:         return initPromise.then(() => runtime.loadRemote("remote/Product"));
 8:       });
 9:     const exportModule = await res;
                                   ^
10:
11:     // Dual exports for CJS/ESM compatibility

RollupError: await isn't allowed in non-async function
```
Not sure what's happening and why we need the top level await logic exactly.

tiagobnobrega:
@poenneby I think I have a slightly better understanding of it now. We were making it harder than it had to be. We don't have to make the exports esm compatible. Rollup will process the file before writing it, and wrap the code into a `__commonJS` call. But we can't use the `require` calls as those are not available on esm mode.

So, I think we only need to get rid of the `require` calls.

I've got a fork for this here :https://github.com/tiagobnobrega/mf-vite/tree/interop
It's on the interop branch.

Haven't tested on SSR yet. I'll try to make progress on that if I find the time.

pcfreak30:
> @poenneby I think I have a slightly better understanding of it now. We were making it harder than it had to be. We don't have to make the exports esm compatible. Rollup will process the file before writing it, and wrap the code into a `__commonJS` call. But we can't use the `require` calls as those are not available on esm mode.
> 
> So, I think we only need to get rid of the `require` calls.
> 
> I've got a fork for this here :https://github.com/tiagobnobrega/mf-vite/tree/interop It's on the interop branch.
> 
> Haven't tested on SSR yet. I'll try to make progress on that if I find the time.

im running into a related issue that may require the use of esm import() vs require. It seems that there is an edge case that can cause an import cycle, and as cjs gets rewritten to esm, the page hangs b/c a module waits on a promise that loads a chunk that goes and ends up requesting itself...

So far rewriting `writeLoadShareModule` may be needed.

tiagobnobrega:
Sorry. I had to move focus elsewhere. I'm still interested in this, but it is more complex than I anticipated. I found issues with top-level await in ESM modules. I was looking at the [runtime packages](https://github.com/module-federation/core/tree/main/packages/runtime), maybe this could leverage that ecosystem more.

vekunz:
> > > this plugin will be superseded by the one rolldown is making
> > 
> > 
> > What does this mean for this plugin? Currently, this plugin cannot be used. Do we have to wait for a new plugin to use Vite with Module Federation? Is there a timeline or other information about this new plugin?
> 
> I use it but i made a fork and added a patch. But I will be going to rolldowns MF plugin as soon as im able to...

I guess "rolldown" means the [new bundler for Vite](https://github.com/rolldown/rolldown)? I never heard of it before, I just read about it for the first time. And it seems that rolldown will have first class support for Module Federation.

pcfreak30:
> > > > this plugin will be superseded by the one rolldown is making
> > > 
> > > 
> > > What does this mean for this plugin? Currently, this plugin cannot be used. Do we have to wait for a new plugin to use Vite with Module Federation? Is there a timeline or other information about this new plugin?
> > 
> > 
> > I use it but i made a fork and added a patch. But I will be going to rolldowns MF plugin as soon as im able to...
> 
> I guess "rolldown" means the [new bundler for Vite](https://github.com/rolldown/rolldown)? I never heard of it before, I just read about it for the first time. And it seems that rolldown will have first class support for Module Federation.

yes, its in rust. https://github.com/rolldown/rolldown/tree/main/crates/rolldown_plugin_module_federation

gioboa:
I close this one because it's stale. I will be more than happy to review a new PR for this issue.

PR #257, opened 2025-02-14
refactor: code cleanup and support for css files in manifest

pcfreak30:
It seems css files do not get processed at ALL.

So I did a code cleanup and I had it support css files by collecting any found in the graph/tree and any in module ID's. However none may be in the graph, and I found something is manipulating the CSS assets so the filenames change. 

In my case I have a generated css asset but a moduleId reference in a js file to `tailwind.css`.

Since there is a relational disconnect I did the best thing I could and collected all found css files and added them to any export that was found to have a css file in its moduleId list.

This works well enough for now to ensure its added to the manifest for my purposes.

If anyone knows how to do better, I welcome feedback b/c I couldn't find the data in the vite graph data.


gioboa (commented):
Thanks @pcfreak30 for your help. I'll look at it in the next days

pcfreak30:
Due to further experimenting, I now just have every css asset be added to every export. the metadata is not properly there to track it otherwise.

zhangHongEn:
Please try to comment this line of code to see if it works properly

pcfreak30:
> is this PR still valid or is it abandoned?

I can't say. ill be getting back to this likely in my work soon. I am on other priorities atm in my project and will review what @zhangHongEn mentioned So guess hold off on doing anything for now.

pcfreak30:
> if (/\.css$/.test(source)) return;

I am working on this again now and checked. no this doesn't help anything. no css files ever run though any of the customResolvers in any of the hooks.

Overall there seems to be no good way to link a css import to the module that imported it... so keeping them all in every export seems to be the only path? I haven't found anything obvious yet in the rolldown api for this.

pcfreak30:
Ive done more digging and was able to create an experiment that computes the source imports mapped to the js file importing. I can also get the final css asset outputs.

However, at-least when running in lib mode, there is no metadata being tracked to link a source css file to the generated css asset that that is in.

It appears a PR to vite would be needed to add this to the css plugin in core.

So as it stands, tracking every css asset output and putting in every manifest export is the best current solution still :/.

PR #259, opened 2025-02-22
fix: solve issue with filename !== remoteEntry

Xeikim:
If filename !== remoteEntry, the metaData of manifest.json cannot get the correct name because the name of the vite bundle does not contain the suffix

### DIFF
--- a/examples/vite-webpack-rspack/host/vite.config.js
+++ b/examples/vite-webpack-rspack/host/vite.config.js
@@ -6,7 +6,7 @@ const mfConfig = {
   name: 'host',
   remotes: {
     remote: {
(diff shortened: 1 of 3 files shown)


gioboa (commented):
Thanks for your help @Xeikim 
Can you change one application in the multi example to check this fix?
That example is under e2e test

Xeikim:
> Thanks for your help @Xeikim Can you change one application in the multi example to check this fix? That example is under e2e test

I've adjusted the use case. Previously, changing the remote application's filename to something other than `remoteEntry` would cause an exception.

gioboa (approved):
It looks awesome to me :clap: 
Thanks for your help @Xeikim

PR #262, opened 2025-02-27
fix(manifest): public path

underfin:
The pr fixed some issues.
- The mainifest should generate `publicPath` as key
- `getPublicPath` is not working at dev

### DIFF
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -67,7 +67,7 @@ const Manifest = (): Plugin[] => {
                   types: { path: '', name: '' },
                   globalName: name,
                   pluginVersion: '0.2.5',
-                  publicPath,
+                  ...(!!getPublicPath ? { publicPath: getPublicPath } : { publicPath }),
                 },
               })
             );
@@ -326,7 +326,7 @@ const Manifest = (): Plugin[] => {
         },
         globalName: name,
         pluginVersion: '0.2.5',
-        ...(!!getPublicPath ? { getPublicPath } : { publicPath }),
+        ...(!!getPublicPath ? { publicPath: getPublicPath } : { publicPath }),
       },
       shared,
       remotes,


DanilAgafonov:
@gioboa I think PR broke `getPublicPath` functionality. Now if you set `getPublicPath` in vite config, it will end up in `publicPath` field in manifest file.

Like that:
```json
{
  "id": "@app/remote",
  "name": "@app/remote",
  "metaData": {
    "name": "@app/remote",
    "type": "app",
    "buildInfo": { "buildVersion": "1.0.0", "buildName": "@app/remote" },
    "remoteEntry": {
      "name": "remoteEntry-BsIUFaZS.js",
      "path": "",
      "type": "module"
    },
    "ssrRemoteEntry": {
      "name": "remoteEntry-BsIUFaZS.js",
      "path": "",
      "type": "module"
    },
    "types": { "path": "", "name": "" },
    "globalName": "@app/remote",
    "pluginVersion": "0.2.5",
    "publicPath": "function() {return \"/\" }"
  },
  // ... more
}
```

It makes all consumers to use `function() {return "\" }` as a literal public path and they end up loading .js files from URLs like that: `http://localhost:55186/function()%20%7Breturn%20%22/%22%20%7Dmodule1-CyXhgF5G.js`

Please revert this change ASAP.

gioboa:
Yes, I can revert it. You can use 1.2.2 in the meantime

DanilAgafonov:
@gioboa Yes, but I only can prepare a revert PR.

Here it is: https://github.com/module-federation/vite/pull/265

A proper fix for original issue probably should be discussed separately, but so far I don't think there is anything to fix for dev env. Current version already prints warnings in dev saying:
> We are ignoring the getPublicPath options because they are natively supported by Vite with the "experimental.renderBuiltUrl" configuration https://vitejs.dev/guide/build#advanced-base-options

So I think people who want to get `getPublicPath` working in dev should use `experimental.renderBuiltUrl`.

Issue #261, opened 2025-02-27
[Bug]: remoteEntryExports is undefined


StevenLangbroek:
### Describe the bug

Hey folks! I haven't used Module Federation since experimenting with it in its early days, and have a use-case now, so I thought I'd start prototyping. I created 2 apps, both with the same stack (TanStack Query + Router, Vite):

- host (offering an app shell + routing, host application, runs on port 5175)
- hello-world (a simple hello-world component, a remote, runs on port 4173, exposes mf-manifest.json)

I'm not using runtime or dynamic stuff, just a straight up `import HelloWorld from 'hello-world/HelloWorld';`. When I try to vier my app in browser, it seems as though Module Federation is confused:

![Image](https://github.com/user-attachments/assets/614f04e4-0ee2-432a-b306-67899b54954b)

It tries to load the `assets/remoteEntry-CdYN5o5R.js` from 5175 instead of 4173, so it 404s, which then leads to the remoteEntryExports being undefined.

If I change my host to load `assets/remoteEntry-CdYN5o5R.js` instead of `mf-manifest.json` instead, it kind of works, but I still get the following errors:

![Image](https://github.com/user-attachments/assets/32076a77-d57c-4ace-b71c-b84d0b3207e5)

This is also kind of undesirable, as I'd have to keep track of hashes somehow, rather than just pointing to mf-manifest & letting that figure it out for me... 

Am I missing something? I've upgraded everything in host & hello-world to latest versions of all dependencies (with `npx npm-check-updates`).

### Version

6.2.0

### Reproduction

It's literally the most basic example I can think of...

### Relevant log output

```shell

```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
- [x] Check that there isn't [already an issue](https://github.com/module-federation/vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

StevenLangbroek:
On a sidenote, I tried setting `path` in mf-manifest.json manually to `http://localhost:4173`, but that 404'ed as well to `http://localhost:5135http://localhost:4173/asset/remoteEntry-[hash].js` (so it seemed they were just naively joined, rather than parsed into a proper URL with a single protocol, host & port)

gioboa:
[...]

Issue #251, opened 2025-02-08
How can I set publicPath to auto


Xeikim:
The generation rule for publicPath in pluginMFManifest is to set it to auto only when base is empty
https://github.com/module-federation/vite/blob/665233db683704117c568dca75c6a09fa7b0d62e/src/plugins/pluginMFManifest.ts#L104

But the default value of vite's base option is '/', and when I set base to empty, vite overrides it to the default value '/'

There seems to be no way to set the value of publicPath to 'auto'

gioboa:
Thanks @Xeikim for your report.
Would you like to drop a PR to fix it?
We can try to figure out if there is a way of retrieve the original base information before the default overwrite

Etroid:
Would it make sense to expose `publicPath` as an option in the federation plugin?

Xeikim:
I am not sure if I need to set publicPath to `auto`. In my project, my remote is based on vite and my host is based on webpack. In the development environment, when I load the remote module through manifest.json in the host, the publicPath of the requested entry file is set to the address of the host. So I tried to use the publicPath `auto` feature of webpack to make it load correctly.
Do you have any better suggestions? Thank you very much

pcfreak30:
@gioboa this issue can likely be closed

PR #267, opened 2025-02-28
fix: publicPath with 'auto' value.

zeratul1:
Get original base and set 'auto' to publicPath properly.

Refs: #251 , #261

### DIFF
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -28,7 +28,7 @@ const Manifest = (): Plugin[] => {
   let remoteEntryFile: string;
   let publicPath: string;
   let _command: string;
-  let _base: string;
+  let _base: string | undefined;
   let viteConfig: any;
   return [
     {
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -28,6 +28,7 @@ const Manifest = (): Plugin[] => {
   let remoteEntryFile: string;
   let publicPath: string;
   let _command: string;
+  let _base: string;
   let viteConfig: any;
   return [
     {
@@ -85,6 +86,7 @@ const Manifest = (): Plugin[] => {
         if (!config.build.manifest)
           config.build.manifest = config.build.manifest || !!manifestOptions;
         _command = command;
+        _base = config.base;
       },
       configResolved(config) {
         root = config.root;
@@ -101,7 +103,11 @@ const Manifest = (): Plugin[] => {
         if (_command === 'serve') {
           base = (config.server.origin || '') + config.base;
         }
-        publicPath = base ? base.replace(/\/?$/, '/') : 'auto';
+        publicPath = _base === ''
+          ? 'auto'
+          : base
+            ? base.replace(/\/?$/, '/')
+            : 'auto';
       },
       async generateBundle(options, bundle) {
         if (!mfManifestName) return;
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -103,11 +103,7 @@ const Manifest = (): Plugin[] => {
         if (_command === 'serve') {
           base = (config.server.origin || '') + config.base;
         }
-        publicPath = _base === ''
-          ? 'auto'
-          : base
-            ? base.replace(/\/?$/, '/')
-            : 'auto';
+        publicPath = _base === '' ? 'auto' : base ? base.replace(/\/?$/, '/') : 'auto';
       },
       async generateBundle(options, bundle) {
         if (!mfManifestName) return;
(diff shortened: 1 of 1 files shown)


gioboa (commented):
Thanks @zeratul1 for your help :clap: 
@StevenLangbroek can you try this change on your local environment please?

gioboa (commented):
Yeah, the change is pretty minimal so you can change your node_modules files by hand

src/plugins/pluginMFManifest.ts
```diff
@@ -28,6 +28,7 @@ const Manifest = (): Plugin[] => {
   let remoteEntryFile: string;
   let publicPath: string;
   let _command: string;
+  let _base: string;
```
DeveshSapkale: I think it should be,

let _base:  string | undefined;

It is one of the reasons for build failure.

src/plugins/pluginMFManifest.ts
```diff
@@ -85,6 +86,7 @@ const Manifest = (): Plugin[] => {
         if (!config.build.manifest)
           config.build.manifest = config.build.manifest || !!manifestOptions;
         _command = command;
+        _base = config.base;
```
DeveshSapkale: If you don't want to make changes as previous comment where I suggested

let  _base : string | undefined; 


On this line we can make it
 _base = config.base ?? ""

Hopefully this should also resolve the build issue.
I haven't tested it

gioboa (commented):
Thanks @DeveshSapkale for your review.
let's wait @zeratul1 for the updates, otherwise I can do these changes too.

src/plugins/pluginMFManifest.ts
zeratul1: Yes, let's take this one.

src/plugins/pluginMFManifest.ts
zeratul1: I prefer the previous one 😄

StevenLangbroek:
Hey @zeratul1 and @gioboa! From the issue I opened:

> We won't know the path at build time for our use-case. The host app can add remotes, and ideally, the paths are relative to the manifest... Is that not possible?
> Our use-case is that in the host application, there's a configuration screen where admins can add remotes to the application's config, which should expose predetermined entrypoints for navigation, screens & configuration forms, which are dynamically loaded at runtime... We don't always know how our customers deploy stuff, so setting the base at build time isn't possible. 

Ideally, it can be configured at the host, or assumed that the base is the base of the mf-manifest file... Does that make sense?

gioboa:
Did you try to load mfe dinamically?

StevenLangbroek:
> Did you try to load mfe dinamically?

Not yet, wanted to start with simple static, build time loading before moving onto dynamic... Should I try that first? See how that behaves?

gioboa:
I guess so, if you don't know the paths at build time that's the only way

zeratul1:
Yes.
You will get a `mf-manifest.json` like this.

```
{
    "id": "vite_provider",
    "name": "vite_provider",
    "metaData": {
        "publicPath": "auto"
    }
}
```
And the `auto` value of `publicPath` should work properly during dynamic loading.

Such as serve provider at `http://localhost:3000/dist` or `http://127.0.0.1/`.

zeratul1:
The problem here is this plugin can not generate a `mf-manifest.json` with a `auto` value of `publicPath`.
As the `base` config reading at `configResolved` hook can not be falsy.

StevenLangbroek:
So, that sort of works, in that my routes are registered and their components are rendered. There still does seem to be some 404s happening:

image

Not sure where that's coming from, but the initiator of the request suggests it's to do with route preloading... Any thoughts?

zeratul1:
For assets loading, there is another config(`getpublicpath`).
Maybe you should config it properly to get rid of those 404s under assets path.

https://module-federation.io/configure/getpublicpath.html

StevenLangbroek:
I have:

image
image

DeveshSapkale:
TBH I am not sure but this won't help right. Because we won't know what the url of remotes will be because the same build will be deployed to different servers with different urls.
I am not sure how the RSbuild plugin handles this automatically.. we don't need to specify much.
Is it this plugin who makes this decision or its bundler which makes this decision of chunk's origin .
Or is it the host application who makes this decision?

DeveshSapkale:
> The problem here is this plugin can not generate a `mf-manifest.json` with a `auto` value of `publicPath`.
> As the `base` config reading at `configResolved` hook can not be falsy.

I think if this point can be somehow implemented and mf-manifest have auto as public path it might resolve the issue

StevenLangbroek:
The 404 is being triggered by the preload-helper of the remote, not the host :thinking: 
 
image

StevenLangbroek:
Oh huh... I fixed it by setting `build.modulePreload` to `false` in the remote's `vite.config.ts` :thinking:

gioboa (approved):
Thanks @zeratul1 for your help.
Kudos for your first contribution :clap:

StevenLangbroek:
Thanks @zeratul1! (and the rest who helped me troubleshoot)

PR #275, opened 2025-03-14
fix: add shareScope attribute to usedRemotes in virtualRemoteEntry

crijke:
Without this attribute, share scope isolation does not work as every remote will fall back to the 'default' share scope.

### DIFF
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -96,7 +96,7 @@ export function generateLocalSharedImportMap() {
                   name: ${JSON.stringify(remote.name)},
                   type: ${JSON.stringify(remote.type)},
                   entry: ${JSON.stringify(remote.entry)},
-                  shareScope: ${JSON.stringify(remote.shareScope) ?? "default"},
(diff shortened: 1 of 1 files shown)


crijke:
> Hi @crijke thanks for your help. do you have an example to use for testing? I would like to verify the behavior with and without your PR. Thanks 🚀

Hi @gioboa Yes, I published an example repository https://github.com/crijke/module-federation-sharescope-vite-example

The example has 3 apps running in 2 share scopes. The react versions should be isolated between share scopes, but they are shared across share scopes when using the vite plugin. 

There's a branch 'rspack' for comparison with an rspack build. With rspack, react is shared only within the share scopes.

gioboa:
Thanks @crijke I will check the PR in few hours 👏

gioboa (approved):
Thanks for your help @crijke 
Kudos for your first PR in this project :clap:

crijke:
> Thanks for your help @crijke Kudos for your first PR in this project 👏

Thanks for your review! 🚀

gioboa:
> Thanks for your review! 🚀

👏you can find this fix in the new 1.2.6 release

Issue #274, opened 2025-03-10
[Bug]: NodeJS vm code + 'eval' needlessly included in bundle (Causing CSP errors)


SunsetFi:
### Describe the bug

We have an in-house microfrontend framework, and we are in the process of adopting @module-federation/vite to deduplicate our shared libraries and reduce the network overhead and browser memory usage.

However, on our trial projects, we are encountering an odd issue: The process of using @module-federation/vite is causing our browser code to gain an `eval()` statement.

This is specifically called out by rollup in a warning:

`node_modules/@module-federation/sdk/dist/index.cjs.js (670:83): Use of eval in "node_modules/@module-federation/sdk/dist/index.cjs.js" is strongly discouraged as it poses security risks and may cause issues with minification.`

Even weirder, on further investigation, this `eval()` statement seems only relevant for NodeJS, and is never touched or invoked by our web project.

A few things stand out here:

## Why is rollup consuming the CJS bundle?
@module-federation/sdk has the proper entries in its package.json to specify ESM builds, which are far more tree-shakable.  But, vite seems to be consuming the cjs build instead...

## If we force rollup to consume the mjs bundle, does it tree-shake away the NodeJS code?

...No.  We tried this using a resolve alias:

```
resolve: {
    alias: {
      "@module-federation/sdk": import.meta.resolve("@module-federation/sdk"),
    },
  },
```

Which properly resolves to the esm build, but we still get the same error, only pointing at the esm this time:

`node_modules/@module-federation/sdk/dist/index.esm.mjs (668:83): Use of eval in "node_modules/@module-federation/sdk/dist/index.esm.mjs" is strongly discouraged as it poses security risks and may cause issues with minification`

Here's that reproduction: https://github.com/SunsetFi/vite-module-federation-createScriptNode/tree/with-esm

## Is it really being included in the bundle?

It's hard to tell with minify on, but yes:
[...]

PR #276, opened 2025-03-17
feat: move main package into sub-package of monorepo

SunsetFi:
This is step 1 of 3-ish for #274 

This moves the main package into a workspace of the monorepo.

The reason for this is that, in the next PR, we will introduce a new package to serve as the proxy to @module-federation/runtime in order to tree-shake it.

I did my best to find everything affected by the change in file paths, paying particular attention to the github actions CI.  However, I am not too fluent in that system, so I would really appreciate a close review to ensure everything is working and the pipelines still work.

Note that I switched from invoking prettier-quick directly to using lint-staged for the husky commit hook, as this allows each workspace package to provide its own rules for commit hooks and they all get discovered and ran automatically.


gioboa:
Thanks @SunsetFi for your commitment 
I created a new branch from main called `monorepo`, let's iterate on it

gioboa (approved):

PR #277, opened 2025-03-17
feat: Vite runtime

SunsetFi:
This creates the runtime package pre-tree-shaken for @module-federation/vite

It has some small improvements over my prototype:
- It modifies `@module-federation/sdk` `isBrowserEnv` directly, and allows the ESM code analysis to flow-through and tree-shake everything, which it does wonderfully well.  This is much more elegant than replacing calls to that function, but it does mean we need to ensure this library is always working with the ESM build, not the CommonJS one.
- It generates the appropriate source maps, squelching a warning we were getting on build
- It has some basic tests on the compiled output.


SunsetFi:
It looks like my changes to use lint-staged don't play nice with prettier-quick.  I was wondering about that... Fix incoming.

SunsetFi:
Fix to the linter in #278

SunsetFi:
Turns out you cant specify working directory on anything other than a `run` step.  I fixed that for the new E2E tests, and also fixed it for the previous MR / plugin library.

gioboa (approved):
Thanks for this amazing help @SunsetFi

PR #278, opened 2025-03-17
fix: use prettier directly in lint-staged

SunsetFi:
prettier-quick looks at git for what files have changed, but lint-staged is already running in a loose file environment.  However, lint-staged passes a list of all changed files as command arguments.

As such, lint-staged should run prettier directly, as it will receive the list of changed files and restrict what it formats.

### DIFF
--- a/packages/module-federation-vite/.lintstagedrc
+++ b/packages/module-federation-vite/.lintstagedrc
@@ -1,3 +1,3 @@
 {
-    "*": "pnpx pretty-quick"
+    "./packages/module-federation-vite/src/**/*.ts": "pnpm fmt"
 }


gioboa:
Thanks @SunsetFi 
In this PR there are few changes and these are not related to prettier.
Can you fix up the PR please? I will be more than happy to merge this improvement. Thanks

SunsetFi:
Sorry about that, I got some things tangled up when fixing this.

This fix should target the monorepo branch, as that's where lint-staged was introduced.

gioboa (approved):
That's great @SunsetFi 👏

PR #281, opened 2025-03-29
feat: Add a dev option to always read origin from config

sabov:
Hello everyone! Thank you for your great work on this project!

This PR adds a dev option named `alwaysReadOriginFromConfig: boolean` to make sure the origin is never read from `window.origin` in dev mode.

**Why is it even needed?**
When working in the "prod-like" dev environment*, the vite dev server is serving its content via localhost but the web application is accessed via a deployed proxy (think of it as an app deployed to a public domain that loads its content from localhost). 

When the proposed above option is enabled, it ensures that all assets are consistently fetched from one origin rather than from two (localhost and the origin of the proxy) and the module federation will work as expected. It's hard to come up with an example in this repository but I could give it a try if it would help to illustrate the use case. 

*"Prod-like" dev environment -- an environment that provides access to other services like auth, APIs, etc. in the same way as it works in prod.

Please let me know if you have any concerns or you need more details. I'm happy to add docs, test, etc. if needed. Thank you!

### DIFF
--- a/src/plugins/pluginProxyRemoteEntry.ts
+++ b/src/plugins/pluginProxyRemoteEntry.ts
@@ -60,8 +60,11 @@ export default function (): Plugin {
             typeof viteConfig.server?.host === 'string' && viteConfig.server.host !== '0.0.0.0'
               ? viteConfig.server.host
               : 'localhost';
+          const originFromConfig = `//${host}:${viteConfig.server?.port}`;
+          const alwaysReadOriginFromConfig =
+            typeof options.dev === 'object' && options.dev.alwaysReadOriginFromConfig;
           return `
-          const origin = window ? window.origin : "//${host}:${viteConfig.server?.port}"
+          const origin = ${alwaysReadOriginFromConfig ? '' : 'window ? window.origin : '}"${originFromConfig}"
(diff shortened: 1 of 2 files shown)


sabov:
Hi @gioboa, thanks for checking this PR so quickly! 

I tried to create a vite plugin with `enforce: 'post'` option, but since the execution order of post plugins is not enforced I can't transform the code that assigns `origin`. I could also use something like [patch-package](https://www.npmjs.com/package/patch-package), but this is the last resort. Do you have any alternative ideas off the top of your head? 

If it will be helpful I could make a quick demo that would show this issue without proxies (e.g. configure everything to use `127.0.0.1` and then use `localhost` instead).

gioboa:
Yep, an example will be amazing to figure out an possible solution too. Thanks

gioboa (commented):
Yep, an example will be amazing to figure out an possible solution too. Thanks

sabov:
I was not able to reproduce the issue with localhost/127.0.0.1 on a clean setup. Looks like it has something to do with one of custom plugins on my end. I'll close this PR for now, thanks for taking a look!

gioboa:
Thank you for your commitment 💯 appreciate

gioboa (commented):
Thank you for your commitment 💯 appreciate

PR #290, opened 2025-04-21
feat: add ignoreOrigin option

lelemm:
I have a use case that I need to run the remote in dev mode but keep attaching it to the host.
This original line
`const origin = window ? window.origin : "//${host}:${viteConfig.server?.port}"`
prevents me from using the remote this way, because it tries to load the remote from the host url.

Adding this new parameter makes it possible to do this use case and keep the original behavior.

### DIFF
--- a/src/plugins/pluginProxyRemoteEntry.ts
+++ b/src/plugins/pluginProxyRemoteEntry.ts
@@ -61,7 +61,7 @@ export default function (): Plugin {
               ? viteConfig.server.host
               : 'localhost';
           return `
-          const origin = window ? window.origin : "//${host}:${viteConfig.server?.port}"
+          const origin = (window && ${!options.ignoreOrigin}) ? window.origin : "//${host}:${viteConfig.server?.port}"
           const remoteEntryPromise = await import(origin + "${viteConfig.base + options.filename}")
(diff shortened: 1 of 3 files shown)


gioboa (commented):
Thanks @lelemm for your help
Is it similar to #281 ?

lelemm:
> Thanks @lelemm for your help Is it similar to #281 ?

Yes. it's similar. they both achieve the same goal.

lelemm:
I created a basic example for this:

https://github.com/[lelemm/example](https://github.com/lelemm/example)

`npm run dev` on both apps (host and remote)

Press the button:
image


Check your network tab:
image

It shows the remote with the URL of the host, not the remote as expected

lelemm:
The way I'm using module-federation is to dynamic load remotes to use as a plugin system for the react host app. This plugin system gets the default of the remote and call it passing some parameter to the plugin initialize.

lelemm:
@gioboa , do you think this is a mergeable change? I'm waiting on this to be able to release the actualbudget.org plugin system so it starts with a nice developer experience (I'm a maintainer of AB)

gioboa (approved):
Thanks @lelemm it looks good to me

gioboa:
@lelemm A new release with this feat is out now. :clap:

PR #292, opened 2025-04-24
fix: getting share item when package is only partially shared (fixes #288)

### LINKED ISSUE #288: [Bug]: Partial shared package
### Describe the bug

   We try to override a shared package partially (sub path). In our case it is
```
  shared: {
      react: {
        singleton: true,
      },
      'react-dom': {
        singleton: true,
      },
      '@eschbach/app-sdk/extensions/shared/v1': {
        singleton: true,
      },
      ...mfOptions.shared,
    },
```

The problem looks like it's here:

Image
It is removing the path of the provided package using removePathFromNpmPackage and cannot find the key anymore.

https://github.com/module-federation/core/issues/3682#issuecomment-2799878077

### Version

.

### Reproduction

Company internal code, I'm sorry.

### Relevant log output

```shell

```

### Validations

danieldaeschle:
Fixes #288 

@gioboa I hope this fix is good. I have no clue how it's supposed to work. I also did not test this change yet, but I think it should work logically wise.

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -341,6 +341,7 @@ export function getNormalizeModuleFederationOptions() {
 export function getNormalizeShareItem(key: string) {
   const options = getNormalizeModuleFederationOptions();
   const shareItem =
+    options.shared[key] ||
     options.shared[removePathFromNpmPackage(key)] ||
     options.shared[removePathFromNpmPackage(key) + '/'];
   return shareItem;


danieldaeschle:
I will try to provide something next week.

gioboa (commented):
Thanks 💪

danieldaeschle:
Sorry, didn't find time yet.

gioboa (commented):
@danieldaeschle Thanks again for your help :rocket: 
I will be more than happy to merge it. Can you provide me a basic example to validate this fix please?

gioboa:
> https://github.com/danieldaeschle/module-federation-vite-bug-partial-shared-package

Thanks for it. I'll check it out in soonish

src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -341,6 +341,7 @@ export function getNormalizeModuleFederationOptions() {
 export function getNormalizeShareItem(key: string) {
   const options = getNormalizeModuleFederationOptions();
   const shareItem =
+    options.shared[key] ||
```
gioboa, line 344: I tested the solution and It's working :clap: 
Thanks @danieldaeschle 
image

PR #299, opened 2025-05-14
feat: support manifest's query params

vkalinichev:
### Problem
My host application uses parametrized manifest URLs (`mf-manifest.json?v=1.2.3`) for caching reasons

Current implementation fails to load manifests with query params in dev mode (returns HTML instead of JSON)

### Solution
Ignore all query parameters when loading `mf-manifest.json`

### Testing
✅ Versioned manifests (`?v=1.2.3`)
✅ Standard manifests (no params)
✅ Both development and production modes

### Impact
Enables proper manifest loading during development

### DIFF
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -43,7 +43,9 @@ const Manifest = (): Plugin[] => {
             next();
             return;
           }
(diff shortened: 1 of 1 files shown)


gioboa (approved):
np for the test in the multi-example scenario.
Thanks for your help @vkalinichev 
Kudos for your first PR in this project 🎉

PR #301, opened 2025-06-02
fix: improve package resolution for pnpm environments (fixes #213)

### LINKED ISSUE #213: no support for sharing monorepo aliases
Hey,
Recently, when trying to shift from originjs's plugin to this one, I stumbled across an issue.
Im using nx, and Im trying to share from my app, a "core" library I also have in the monorepo.
In nx, I have alias for the library, and instead of importing 'lib-name' I use '@monorepo-name/lib-name'.
(The reason for that is that it lets me not needing to publish and install the packages that I maintain inside the monorepo)
The package Im sharing this to (lets name it 'federated-lib') is in the repo, and it is too, using the core library with the aliasing.

I'm currently using originjs's federation plugin and it works great but when trying to switch to this plugin I get an error of something along the lines of 'module @monorepo-name/lib-name/package.json not found'.

I tried whatever I can, fiddling around with the aliasing in the tsconfig but had no luck.
Would be glad to hear from you what I should be doing because this blocks me from upgrading to this plugin.

Also, unfortunately I wouldn't be able to share the code as it is company private. 
Thanks for any help.

sstewart-cs:
fixes #213 
The Module Federation plugin was failing to resolve package versions in pnpm
monorepo environments due to pnpm's unique node_modules structure. This commit
adds an additional resolution strategy that checks the local node_modules
directory before falling back to the more expensive searchPackageVersion
function.

This change:
- Maintains the original resolution behavior as the first attempt
- Adds a direct path check to local node_modules for pnpm compatibility
- Preserves the existing fallback mechanism
- Keeps the same error handling pattern

Fixes issue where shared dependencies like 'lit' couldn't be resolved in pnpm
monorepos, resulting in "Cannot find module 'lit/package.json'" errors.

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -170,7 +170,12 @@ function normalizeShareItem(
       version = require(path.join(removePathFromNpmPackage(key), 'package.json')).version;
     } catch (e1) {
       try {
-        const localPath = path.join(process.cwd(), 'node_modules', removePathFromNpmPackage(key), 'package.json');
+        const localPath = path.join(
+          process.cwd(),
+          'node_modules',
+          removePathFromNpmPackage(key),
+          'package.json'
+        );
         version = require(localPath).version;
       } catch (e2) {
         version = searchPackageVersion(key);


gioboa (approved):
Thanks @sstewart-cs for your help.
Kudos for your first PR in this project :clap:

gioboa (approved):

PR #305, opened 2025-06-18
feat: add support for custom virtual module directory (fixes #306)

### LINKED ISSUE #306: [Bug]: Cannot run parallel builds due to shared virtual module directory conflict
### Describe the bug

# Issue Description
## Problem
When running multiple Module Federation builds in parallel (e.g., in CI/CD pipelines or monorepos), builds fail with the following error:

```
Error: EEXIST: file already exists, mkdir '/tmp/jenkins/workspace/_X_pull-requests_PR-4025/working-dir/apps/package/node_modules/__mf__virtual'
```

This occurs because the virtual module directory (`__mf__virtual`) is hardcoded and shared across all builds, causing race conditions when multiple processes try to create or modify this directory simultaneously.

## Root Cause
The Module Federation plugin currently uses a hardcoded directory name (`__mf__virtual`) for storing virtual modules, with no way to customise it per application or build process.

## Proposed Solution
Add a new configuration option `virtualModuleDir` that allows specifying a custom directory name for virtual modules. This would enable:

- Running parallel builds by using unique directories for each application
- Better isolation between applications in monorepo setups
- Preventing collisions in shared environments

Example usage:
```javascript
federation({
  name: 'my-app',
  virtualModuleDir: '__mf__virtual__my_app',
  // other options...
})
```

Note: The directory name should not contain slashes to ensure proper module resolution. A catch has been added for this, and tests also.

This enhancement maintains backward compatibility by defaulting to the original directory name when no custom path is provided.

### Version

1.4.1

### Reproduction

https://no-repo-required.com

### Relevant log output

```shell

```

### Validations

sstewart-cs:
### **PR Description**
**Overview**

This PR adds the ability to customise the virtual module directory used by Module Federation. Previously, the directory was hardcoded to `__mf__virtual`, which caused conflicts when running multiple builds in parallel or in monorepo setups.

**Changes**
- Added a new virtualModuleDir option to the federation plugin configuration
- Updated the virtual module resolution to use the custom directory name
- Added validation to prevent path-like values with slashes that would cause module resolution issues
- Added tests to verify the new functionality and validation
- Updated documentation to explain the new option

**Example Usage**

```javascript
// vite.config.js
federation({
  name: 'my-app',
  virtualModuleDir: '__mf__virtual__my_app',
  // other options...
})
```

Related Issue
Fixes #306  - Cannot run parallel builds due to shared virtual module directory conflict

### DIFF
--- a/src/utils/VirtualModule.ts
+++ b/src/utils/VirtualModule.ts
@@ -30,11 +30,6 @@ function getNodeModulesDir() {
   return cachedNodeModulesDir;
 }
 
-export function getVirtualPackageName(): string {
-  const options = getNormalizeModuleFederationOptions();
-  return options?.virtualModuleDir || '__mf__virtual';
-}
-
 const patternMap: {
   [tag: string]: RegExp;
 } = {};
@@ -69,7 +64,8 @@ export default class VirtualModule {
    */
   static ensureVirtualPackageExists() {
     const nodeModulesDir = getNodeModulesDir();
-    const virtualPackagePath = resolve(nodeModulesDir, getVirtualPackageName());
+    const { virtualModuleDir } = getNormalizeModuleFederationOptions();
+    const virtualPackagePath = resolve(nodeModulesDir, virtualModuleDir);
 
     if (!existsSync(virtualPackagePath)) {
       mkdirSync(virtualPackagePath);
@@ -77,7 +73,7 @@ export default class VirtualModule {
       writeFileSync(
         resolve(virtualPackagePath, 'package.json'),
         JSON.stringify({
-          name: getVirtualPackageName(),
+          name: virtualModuleDir,
           main: 'empty.js',
         })
       );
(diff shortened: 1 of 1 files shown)


src/utils/VirtualModule.ts
```diff
@@ -30,7 +30,10 @@ function getNodeModulesDir() {
   return cachedNodeModulesDir;
 }
 
-export const virtualPackageName = '__mf__virtual';
+export function getVirtualPackageName(): string {
+  const options = getNormalizeModuleFederationOptions();
+  return options?.virtualModuleDir || '__mf__virtual';
```
gioboa: we already have the fallback [here](https://github.com/module-federation/vite/pull/305/files#diff-abc6cd67fc307307789c9cacee7765bf773360586663d0bc333eeb55766f8830R410) so this is not necessary I guess

gioboa (commented):
it looks great to me :clap: 
I added just a little note

src/utils/VirtualModule.ts
sstewart-cs: Thanks for the review @gioboa that should be the PR updated now 🤞

gioboa (approved):
it looks great to me :clap: 
Thanks for your help @sstewart-cs

Issue #321, opened 2025-07-24
[Bug]: Scopes containing dots break the functionality


rasimx:
### Describe the bug

When specifying dependencies with dots in the namespace (e.g., @company.x.y/library) inside the shared section of the Module Federation config, the library fails to resolve them correctly. It appears that the dot (.) in the namespace is mistakenly treated as a file extension separator. This leads to incorrect resolution behavior — for example, attempting to load @company.x.y/library.js as if it were a file path, rather than a valid npm package name.

### Version

v.1.6.0

### Reproduction

-

### Relevant log output

```shell

✗ Build failed in 50ms
error during build:
[commonjs--resolver] ENOENT: no such file or directory, open '/Volumes/Data/work/mf-core-interactive-assistant/node_modules/__mf__virtual/mf_mf_2_core_mf_2_interactive_mf_3_assistant__prebuild___mf_0_rshbintech_mf_3_rprul_mf_3_ecp_mf_1_core_mf_2_frontend_mf_2_helper_mf_2_lib__prebuild__.ecp/core-frontend-helper-lib'
    at Object.openSync (node:fs:596:3)
    at Object.writeFileSync (node:fs:2322:35)

```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
- [x] Check that there isn't [already an issue](https://github.com/module-federation/vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #323, opened 2025-07-24
fix: correctly detect suffix for scoped packages with dots

rasimx:
Fixes an issue where `VirtualModule` failed to detect the correct suffix for scoped packages that contain dots in the namespace (e.g., `@company.x.y/lib.name`).

This caused the suffix to be incorrectly split from the namespace, resulting in broken imports.

Fixes: https://github.com/module-federation/vite/issues/321

### DIFF
--- a/src/utils/VirtualModule.ts
+++ b/src/utils/VirtualModule.ts
@@ -30,6 +30,22 @@ function getNodeModulesDir() {
   return cachedNodeModulesDir;
 }
 
+function getSuffix(name: string): string {
+  const isScoped = name.startsWith('@');
+  const slashIndex = name.indexOf('/');
+
+  const namespaceEnd = isScoped && slashIndex !== -1 ? slashIndex : -1;
+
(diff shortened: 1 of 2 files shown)


src/utils/VirtualModule.ts
```diff
@@ -30,6 +30,22 @@ function getNodeModulesDir() {
   return cachedNodeModulesDir;
 }
 
+function getSuffix(name: string): string {
```
gioboa: It looks great to me, can you cover this function with a test with few different test cases?

rasimx: Added unit tests for the getSuffix function to verify behavior with scoped packages and nested paths (e.g., .bin/@scope/pkg).

While writing tests, identified additional edge cases where the suffix was incorrectly inferred. Updated the implementation to handle these cases properly.

gioboa (approved):
Thanks, that's great @rasimx 
Kudos for your fist PR in this project :partying_face:

PR #328, opened 2025-07-27
fix: prevent multiple processing of await plugin output

pcfreak30:
I ran into an edge case recently during some experimentation.

If you try running multiple module federation plugin instances in 1 vite config, the top await rewriter plugin will try to process itself multiple times. Around the 3rd time it created a duplicate const and the AST parse failed.

This is a simple fix that prevents that edge case.

This is part of a series of PR's to upstream my R&D.

### DIFF
--- a/src/plugins/pluginDevProxyModuleTopLevelAwait.ts
+++ b/src/plugins/pluginDevProxyModuleTopLevelAwait.ts
@@ -8,10 +8,15 @@ import { Plugin } from 'vite';
 
 export function PluginDevProxyModuleTopLevelAwait(): Plugin {
   const filterFunction = createFilter();
+  const processedFlag = '/* already-processed-by-dev-proxy-module-top-level-await */';
+
   return {
     name: 'dev-proxy-module-top-level-await',
     apply: 'serve',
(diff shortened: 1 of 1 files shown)


pcfreak30:
> Thanks for your help @pcfreak30 Can you shared a reproduction repository for this? I would like to double check it. It looks good to me though

no. I don't have anything simple ready.

what  I can say is it would ideally crash the vite server if you stick a host and a remote such that you have 2 federation() plugin objects in 1 vite config.

You can view a dev env I recently made to get an idea of some of what I was doing (https://github.com/LumeWeb/web/blob/b1dd36f3307f59a7039f40156c860c6829c064fc/libs/portal-framework-core/src/vite/plugin.ts), but while im no longer trying to stuff multiple federation builds in 1 vite instance... I figured i can submit the fix anyways. I was using a custom virtual folder as well for the remote.

gioboa (commented):
`pnpm fmt` will solve the error in the pipeline 
Thanks

gioboa (approved):
It looks straightforward and this will add an extra check in the process.
It's a good addition @pcfreak30 👌

Issue #132, opened 2024-10-03
Build is failing with Unfinished hook action(s) on exit: (proxyRemoteEntry) load "virtual:mf-REMOTE_ENTRY_ID"


mshima:
Is there a way to debug the error?

```
x Build failed in 2.20s
error during build:
Unexpected early exit. This happens when Promises returned by plugins cannot resolve. Unfinished hook action(s) on exit:
(proxyRemoteEntry) load "virtual:mf-REMOTE_ENTRY_ID"
```

gioboa:
With VSCode you can run scripts in debug mode and you can use the breakpoints to proceed step by step.

zhangHongEn:
> Is there a way to debug the error?
> 
> ```
> x Build failed in 2.20s
> error during build:
> Unexpected early exit. This happens when Promises returned by plugins cannot resolve. Unfinished hook action(s) on exit:
> (proxyRemoteEntry) load "virtual:mf-REMOTE_ENTRY_ID"
> ```

“Could you provide a reproducible example? I have a general idea of the cause of this error, but I don’t know how to reproduce it.“​ @mshima

mshima:
@zhangHongEn I've found the error.
The exposed modules was not found.
Execution `vite` correctly shows the error message.

Happened when migrating from Webpack which does not require vue extension in exposes.

gioboa:
@mshima Can we add a warning in the terminal for this?

mshima:
> @mshima Can we add a warning in the terminal for this?

Would be nice to fail with correct error message.

gioboa:
Can we throw the correct one?

BanShan-Alec:
![Image](https://github.com/user-attachments/assets/3219d717-f426-4358-ae4f-1317714ba489)
build fail in linux，but work well in windows10

BanShan-Alec:
> ![Image](https://github.com/user-attachments/assets/3219d717-f426-4358-ae4f-1317714ba489) build fail in linux，but work well in windows10

`https://github.com/module-federation/vite/issues/245` same issue in docker env

pcfreak30:
I just ran into this and debugged the code

```
? {
                  find: new RegExp(`(.*${PREBUILD_TAG}.*)`),
                  replacement: function ($1: string) {
                    const pkgName = (VirtualModule.findModule(PREBUILD_TAG, $1) as VirtualModule)
                      .name;
                    return pkgName;
                  },
                }
              : {
[...]

PR #330, opened 2025-08-22
feat: add support for disabling imports for shared modules

pcfreak30:
- Add import flag to shareConfig interface and normalization
- Handle import: false modules in virtual remote entry generation

---

Looking to get feedback on this as a draft before merging.


pcfreak30:
> It looks great to me 👏👏 Can we recreate the import scenario in the multi example e2e test please?

I can try though that may take a bit, will be lower priority for me.

src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -289,6 +293,7 @@ export type ModuleFederationOptions = {
             singleton?: boolean;
             requiredVersion?: string;
             strictVersion?: boolean;
+            import?: boolean;
```
gioboa: In the [mfe docs](https://module-federation.io/configure/shared#import) is a string or false 🤔

pcfreak30: updated to use the shared plugin interface type `sharePlugin.SharedConfig['import']`

gioboa: Thanks 🙏

src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -38,16 +38,22 @@ export function writeLocalSharedImportMap() {
 export function generateLocalSharedImportMap() {
   const options = getNormalizeModuleFederationOptions();
   return `
+    import {loadShare} from "@module-federation/runtime";
```
gioboa, line 41: Nice catch, you moved the import outside the loop 👏

pcfreak30: Was a bit worse then that. Code was trying to do a require() before un-neededly, and that wasn't getting pre-processed either. The es import is getting processed by vite properly now.

gioboa (commented):
> Is there any specific real test you would like for this?

No one in particular, just one valid to activate the new code.

pcfreak30:
@gioboa this is ready now.

I also in the processed have reproduced https://github.com/module-federation/vite/issues/132 https://github.com/module-federation/vite/blob/1e466bf8a17089a0bffcb17b3a16941d5c944d3a/src/plugins/pluginProxySharedModule_preBuild.ts#L86

If you have a shared dependency but its not included in package json, it breaks. You can likely reprod by deleting lodash from `dynamic-remote` and try to run it. I did a lot of troubleshooting before I noticed the build for it failing.

How do you think that should be handled since we know the likely cause. Would be a separate PR.

gioboa (commented):
> How do you think that should be handled since we know the likely cause. Would be a separate PR.

with that check the PR is self contained, so let's add it please. Thanks for your commitment :muscle:

gioboa (commented):
I'm sorry, let's add the check to prevent this error
`Cannot read properties of null (reading 'id') [plugin vite:dep-pre-bundle]`

pcfreak30:
> I'm sorry, let's add the check to prevent this error `Cannot read properties of null (reading 'id') [plugin vite:dep-pre-bundle]`

That should be in its own separate PR b/c its not actually connected, I just found the error by chance, so I want to know what you want the behavior to be. Silently ignore, throw an error? `console.error`?

I can submit a PR for that once you clarify and thats a very simple change.

gioboa:
> > I'm sorry, let's add the check to prevent this error `Cannot read properties of null (reading 'id') [plugin vite:dep-pre-bundle]`
> 
> That should be in its own separate PR b/c its not actually connected, I just found the error by chance, so I want to know what you want the behavior to be. Silently ignore, throw an error? `console.error`?
> 
> I can submit a PR for that once you clarify and thats a very simple change.

Options:
- Silently ignore: is not good because the developer can remove the shared dependency.
- throw an error: will stop you to build and deploy
- console.error: will fill your browser console and most of the time this kind of errors are ignored by devs

So based on that, throw an error should be fine. What do you think?

gioboa (commented):
As mentioned, the new code is currently unused, so I think we should find a different way to support the `import false` configuration.

ScriptedAlchemy:
> I run the multi-example with your code on my local env.
> I added a console log here to verify the new code
> 
> Image
> 
> but I can't see that log. 🤔 

Can't we just not render that template string part

ScriptedAlchemy:
no, you deleted the entire template object. Im saying have a get method return a throw else return the original. 

```ts
export function generateLocalSharedImportMap() {
  return `
    import {loadShare} from "@module-federation/runtime";
    const importMap = {
      ${Array.from(getUsedShares())
        .sort()
        .map((pkg) => {
          const shareItem = getNormalizeShareItem(pkg);
          return `
        ${JSON.stringify(pkg)}: async () => {
          ${
            shareItem?.shareConfig.import === false
              ? `throw new Error(\`Shared module '\${${JSON.stringify(pkg)}}' must be provided by host\`);`
              : `let pkg = await import("${getPreBuildLibImportId(pkg)}");
            return pkg;`
          }
        }
      `;
        })
        .join(',')}
    }
      const usedShared = {
      ${Array.from(getUsedShares())
        .sort()
        .map((key) => {
          const shareItem = getNormalizeShareItem(key);
          if (!shareItem) return null;

          const getBody =
            shareItem.shareConfig.import === false
              ? `
              const shared = await loadShare(${JSON.stringify(key)});
              if (shared) return () => shared;
              throw new Error(\`Shared module '\${${JSON.stringify(key)}}' must be provided by host\`);
              `
              : `
              usedShared[${JSON.stringify(key)}].loaded = true
              const {${JSON.stringify(key)}: pkgDynamicImport} = importMap
              const res = await pkgDynamicImport()
              const exportModule = {...res}
              // All npm packages pre-built by vite will be converted to esm
              Object.defineProperty(exportModule, "__esModule", {
                value: true,
                enumerable: false
              })
              return function () {
                return exportModule
              }
              `;

          return `
          ${JSON.stringify(key)}: {
            name: ${JSON.stringify(key)},
            scope: [${JSON.stringify(shareItem.scope)}],
            loaded: false,
            from: ${JSON.stringify(options.name)},
            async get () {${getBody}
            },
            shareConfig: {
              singleton: ${shareItem.shareConfig.singleton},
              requiredVersion: ${JSON.stringify(shareItem.shareConfig.requiredVersion)},
              ${shareItem.shareConfig.import === false ? 'import: false,' : ''}
            }
          }
        `;
        })
        .filter((x) => x !== null)
        .join(',')}
    }
      const usedRemotes = [${Object.keys(getUsedRemotesMap())
        .map((key) => {
          const remote = options.remotes[key];
          if (!remote) return null;
          return `
                {
                  entryGlobalName: ${JSON.stringify(remote.entryGlobalName)},
                  name: ${JSON.stringify(remote.name)},
                  type: ${JSON.stringify(remote.type)},
                  entry: ${JSON.stringify(remote.entry)},
                  shareScope: ${JSON.stringify(remote.shareScope) ?? 'default'},
                }
          `;
        })
        .filter((x) => x !== null)
        .join(',')}
      ]
      export {
        usedShared,
        usedRemotes
      }
      `;
}
```

package.json
```diff
@@ -47,6 +47,7 @@
   "packageManager": "pnpm@9.1.3",
   "dependencies": {
     "@module-federation/runtime": "^0.17.1",
```
ScriptedAlchemy: the federation ecosystem must remain synced to avoid catastrophic issues . 
bump runtime to match sdk version, all the versions must remain in sync

pcfreak30:
@gioboa @ScriptedAlchemy done.

pcfreak30:
@gioboa following up.

gioboa:
> @gioboa following up.

I'm waiting the feedback from ScriptedAlchemy

pcfreak30:
@ScriptedAlchemy can you please review.

pcfreak30:
@gioboa ping.

PR #331, opened 2025-08-24
fix: check if html file exists before reading it

pcfreak30:
In remotes, im not actually making use of `index.html` but the plugin currently assumes it exists b/c its always set due to https://github.com/module-federation/vite/blob/43fa399ab4ea37e6db9d82a72aa954befe3304b1/src/plugins/pluginAddEntry.ts#L96

This is a simple fix for that.


gioboa (approved):
It looks great to me 👏💪

PR #332, opened 2025-08-25
feat: add publicPath option

pcfreak30:
- Add publicPath option to ModuleFederationOptions interface
- Implement resolvePublicPath utility for consistent path resolution
- Support explicit publicPath override, Vite base, and 'auto' fallback
- Integrate with pluginMFManifest and pluginProxyRemoteEntry

---

I have an edge case use case where I need to have an empty base path for localhost development which would trigger the `auto` mode in the federation path lookup. However this causes a problem with vites base which if empty creases the html assets at `./` not `/` which breaks my app.

So I need the ability to override and separate them, keeping MF publicPath auto, but changing the vite base for the HTML.

### DIFF
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -17,6 +17,7 @@ import {
   processModuleAssets,
   trackAsset,
 } from '../utils/cssModuleHelpers';
+import { resolvePublicPath } from '../utils/publicPath';
 
 // Helper to build share key map with proper context typing
 interface BuildFileToShareKeyMapContext {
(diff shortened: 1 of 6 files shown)


gioboa (commented):
Thanks for this improvement @pcfreak30

gioboa (approved):
Thanks, that's great to me @pcfreak30 👏

pcfreak30:
> is it solving a specific issue?

Can you please clarify?

gioboa (commented):
is it solving a specific issue?

gioboa (commented):
Can we link and close any issues with this PR ?

PR #343, opened 2025-10-15
feat: host init inject option parameter

szebest:
With the changes made in https://github.com/module-federation/vite/pull/167 the hostInit script is injected into the html file without the option to change this behavior.
In some cases the host application is not served via the html file, but by the entry point js file.
There was a discussion before about SvelteKit https://github.com/module-federation/vite/issues/189, but in my opinion it would be still a good idea to let the programmer have the option to decide if we should include it in the entry or html file.

The default is still html file, so it is backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,11 @@ export default defineConfig({
       },
       filename: "remoteEntry.js",
       shared: ["vue"],
+      // Optional parameter that controls where the host initialization script is injected.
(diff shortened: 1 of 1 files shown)


gioboa (commented):
It sounds reasonable to me, thanks for this PR @szebest 
Can you fix it the tests please?

PR #344, opened 2025-10-15
fix: resolve warnings when building with sourcemaps enabled

szebest:
When vite is configured to generate sourcemaps we are getting warnings about missing sourcemaps from the plugins that use the transform hook. To address this issue we should return both the code and the map in the transform hook.

image


gioboa (approved):

gioboa (commented):
Thanks for your help @szebest 👏

PR #346, opened 2025-10-29
feat: opt to disable manifest css inject

### LINKED ISSUE #337: [Bug]: CSS exposure changed from component-scoped to project-wide
### Describe the bug

Starting from version 1.7.2, something changed how CSS assets are bundled and exposed. The previous behavior of selective CSS inclusion has been replaced with a unified approach that includes all project styles causing unexpected behavior.  If this change was intentional, maybe it could be considered a breaking change and released as a major version to avoid breaking existing behavior and allow proper migration planning for downstream projects since the fix kinda makes sense in a way.

### Version

1.7.2

### Reproduction

build using any version above 1.7.1

### Relevant log output

```shell

```

### Validations

songlairui:
`Add all CSS assets to every export` will lead unused global style loaded

skip this by opt

Close #337

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -255,6 +255,7 @@ function normalizeLibrary(library: any): any {
 
 interface ManifestOptions {
   filePath?: string;
+  /** if enable, will skip auto inject global css */
   disableAssetsAnalyze?: boolean;
   fileName?: string;
 }
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -29,11 +29,13 @@ const Manifest = (): Plugin[] => {
   const { name, filename, getPublicPath, manifest: manifestOptions } = mfOptions;
 
   let mfManifestName: string = '';
+  let disableMainfestCssInject: boolean | undefined;
   if (manifestOptions === true) {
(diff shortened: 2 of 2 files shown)


src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -255,6 +255,7 @@ function normalizeLibrary(library: any): any {
 
 interface ManifestOptions {
   filePath?: string;
+  /** if enable, will skip auto inject global css */
```
gioboa, line 258: In the #337 issue we decided to use the opposite logic. Skip the inject global by default.

gioboa (changes requested):
Thanks for your help @songlairui

src/utils/normalizeModuleFederationOptions.ts
gioboa: cc @pcfreak30

gioboa:
In the #337 issue we decided to use the opposite logic. Skip the inject global by default. 347 is doing the same :(
I don't want to be in this situation, decide to merge one and close the other one :(

pcfreak30:
> In the #337 issue we decided to use the opposite logic. Skip the inject global by default. 347 is doing the same :( I don't want to be in this situation, decide to merge one and close the other one :(

and im doing opt-in by default for the previous behavior (dont process css at all)?

PR #347, opened 2025-10-29
feat(css): add bundleAllCSS option to control CSS asset inclusion in exposed modules

### LINKED ISSUE #337: [Bug]: CSS exposure changed from component-scoped to project-wide

pcfreak30:
* Added `bundleAllCSS` option to `ModuleFederationOptions` and `NormalizedModuleFederationOptions` interfaces
* Updated manifest plugin to conditionally include all CSS assets in exports based on `bundleAllCSS` setting
* Added tests to verify default and explicit `bundleAllCSS` option behavior
* Updated `cssModuleHelpers.test.ts` to import `normalizeModuleFederationOptions` and mock `bundleAllCSS` option
* Added new test suite in `cssModuleHelpers.test.ts` to validate `bundleAllCSS` option handling
* Updated `normalizeModuleFederationOption.test.ts` to include `bundleAllCSS`, `getPublicPath`, and `publicPath` in expected default values

Close #337

### DIFF
--- a/src/plugins/pluginMFManifest.ts
+++ b/src/plugins/pluginMFManifest.ts
@@ -163,7 +163,7 @@ const Manifest = (): Plugin[] => {
         }
 
         // Second pass: Collect all CSS assets
-        const allCssAssets = collectCssAssets(bundle);
+        const allCssAssets = mfOptions.bundleAllCSS ? collectCssAssets(bundle) : new Set<string>();
 
         const exposesModules = Object.keys(mfOptions.exposes).map(
           (item) => mfOptions.exposes[item].import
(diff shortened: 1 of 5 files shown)


src/plugins/pluginMFManifest.ts
gioboa: Do we need to ship this one too?

pcfreak30: @gioboa huh?

pcfreak30: ive fixed `collectCssAssets` as it was overlooked

src/plugins/pluginMFManifest.ts
```diff
@@ -163,7 +163,7 @@ const Manifest = (): Plugin[] => {
         }
 
         // Second pass: Collect all CSS assets
-        const allCssAssets = collectCssAssets(bundle);
+        const allCssAssets = bundleAllCSS ? collectCssAssets(bundle) : [];
```
gioboa:
```suggestion
        const allCssAssets = mfOptions.bundleAllCSS ? collectCssAssets(bundle) : new Set<string>();
```

pcfreak30:
@gioboa tests are passing now.

Issue #247, opened 2025-01-30
library type var


Vladislav-Lukyanuk:
Hi! How I can build a vite project exposes a code as a global variable?
Similar to webpack
```
library: { type: 'var', name: MODULE_FEDERATION_APPLICATION_NAME },
```

Vladislav-Lukyanuk:
@gioboa Great! But I got the error `Cannot use import statement outside a module`. 
I thought it resolve the problem. 
I changed a vite and a package configs, but in a production build I still see imports inside.
Unfortunately the host app I can't chage.

vite.config.ts
```
  build: {
    target: 'es2015',
    outDir: './dist',
    modulePreload: false,
    minify: false,
    cssCodeSplit: true,
    assetsDir: '',
  },
```

package.json
```
  "type": "commonjs",
```

If I try to change tsconfig.json to `target: 'es2015'` I got a error I can't use `import.env.meta`

gioboa:
I see, can you share a basic reproduction to clone please?

Vladislav-Lukyanuk:
It's going to be problematic. But in general can we build the code in commonJS wo imports?

gioboa:
> It's going to be problematic. But in general can we build the code in commonJS wo imports?

Can you give me an example please?

Vladislav-Lukyanuk:
Solved using this code
```
build: {
    outDir: './dist',
    modulePreload: false,
    minify: false,
    cssCodeSplit: true,
    assetsDir: '',

    rollupOptions: {
      output: {
        format: 'commonjs',
      }
    }
  },
```

gioboa:
Nice 👍 Thanks for the update 
Can we close this issue?

crutch12:
@Vladislav-Lukyanuk I'm not able to chage the host app too 🥴
So I workaround it with `varRemoteEntry.js`
See: https://github.com/module-federation/core/pull/4199

PR #353, opened 2025-11-10
feat: new varFilename option - allow generate additional "var" remoteEntry.js file (#247)

crutch12:
### Problem

`@module-federation/vite` generates esm only `remoteEntry.js` file, thus you are not able to load it in "old fashion way"

Problem example:
```js
// host
plugins: [
  federation({
    remotes: {
      remote: 'remote@http://localhost:3000/remoteEntry.js' // doesn't work with esm files
    },
  }),
]
```

### Solution

We can generate additional `varRemoteEntry.js` file, 

This file will just `import()` our esm `remoteEntry.js` file and call `init/get` methods. Example:

```js
// varRemoteEntry.js
var remote; // name of remote app
remote = (function () {
  function getScriptUrl() {
    return document.currentScript.src
  }

  const entry = `${getScriptUrl()}/../remoteEntry.js` // path to esm remoteEntry.js

  return {
    get: (...args) => import(entry).then(m => m.get(...args)),
    init: (...args) => import(entry).then(m => m.init(...args)),
  };
})();
```

Thus `host` application may use it:

```diff
// host
plugins: [
  federation({
    remotes: {
-      remote: 'remote@http://localhost:3000/remoteEntry.js'
+      remote: 'remote@http://localhost:3000/varRemoteEntry.js'
    },
  }),
]
```

Moreover: I've added new property into `manifest.json` - `varRemoteEntry`, I hope it's okay...

### Current problems
- [ ] Handle `publicPath`/`getPublicPath` options
- [ ] Write tests

### Related issues:
- https://github.com/module-federation/vite/issues/247
- https://github.com/module-federation/core/pull/4199
- https://github.com/originjs/vite-plugin-federation/issues/368


src/plugins/pluginMFManifest.ts
```diff
@@ -304,6 +316,7 @@ const Manifest = (): Plugin[] => {
         },
         remoteEntry,
         ssrRemoteEntry: remoteEntry,
+        varRemoteEntry,
```
crutch12, line 319: @ScriptedAlchemy wdyt about additional `varRemoteEntry` field in `manifest.json`?

Or maybe we should generate the second item in `manifest.json` array. Sry, I couldn't find the manifest specification...

src/plugins/pluginVarRemoteEntry.ts
```diff
@@ -0,0 +1,132 @@
...
+   * @returns Complete globalVar remoteEntry.js file source
+   */
+  function generateVarRemoteEntry(remoteEntryFile: string) {
+    const options = getNormalizeModuleFederationOptions();
+
+    const { name, varFilename } = options;
+
+    const isValidName = isValidVarName(name);
+
+    // @TODO: implement publicPath/getPublicPath support
```
crutch12: I don't handle `publicPath`/`getPublicPath` here. Should I? I rely on generated `remoteEntry.js` relative path

src/plugins/pluginVarRemoteEntry.ts
```diff
@@ -0,0 +1,132 @@
...
+    return `
+  ${isValidName ? `var ${name};` : ''}
+  ${isValidName ? name : `globalThis['${name}']`} = (function () {
+    function getScriptUrl() {
+      const currentScript = document.currentScript;
+      if (!currentScript) {
+        console.error("[VarRemoteEntry] ${varFilename} script should be called from sync <script> tag (document.currentScript is undefined)")
```
crutch12, line 114: I'm not sure about this logic. Maybe there is another way to indicate path of original `remoteEntry.js` file?
I've stolen it from `__webpack_public_path__` (`'auto'`) implementation...

src/plugins/pluginVarRemoteEntry.ts
```diff
@@ -0,0 +1,132 @@
...
+       * @param options - Rollup output options
+       * @param bundle - Generated bundle assets
+       */
+      async generateBundle(options, bundle) {
+        if (!varFilename) return;
+
+        const isValidName = isValidVarName(name);
+
+        if (!isValidName) {
+          warn(
```
crutch12: Should we fail if provided `name` is not valid js name?
In this case it's not possible to load this remote entry in old fashion way, it won't work:
```js
remotes: {
  '@app/remote': '@app/remote@http://localhost:3000/varRemoteEntry.js',
}
```

But, as my example shows, this still works with dynamic `loadRemote`, therefore I've decided to handle it with `globalThis` and showing the warn message

crutch12:
@gioboa I've removed `warn`, check it out, please

e2e/vite-webpack-rspack/tests/index.spec.ts
```diff
@@ -139,7 +152,11 @@ test.describe('Dynamic remote', () => {
     // Check that lodash version is displayed in SpecialPromo banner
     await showAdToggle.check({ force: true });
 
-    const specialPromoBanner = page.getByRole('heading', { level: 2, name: 'Up to 50% off!', exact: true });
+    const specialPromoBanner = page.getByRole('heading', {
```
crutch12, line 155: prettier fixed on commit

PR #355, opened 2025-11-17
fix: removed a warning that falsely warns that `getPublicPath` is ignored

vergilfromadyen:
It looks like the warning was introduced in [this commit](https://github.com/module-federation/vite/pull/108/files).  
`getPublicPath` was restored a day later in [this commit](https://github.com/module-federation/vite/commit/2670d79c79ebaa5c6d163b52826aef0e93827acd#diff-abc6cd67fc307307789c9cacee7765bf773360586663d0bc333eeb55766f8830R344-R345), but the warning is still in the code.

Nowadays [`publicPath` and `getPublicPath` are still in the code](https://github.com/module-federation/vite/blob/b65d5762301dbb50f0e09172d2b70c0c1b9d0eeb/src/utils/normalizeModuleFederationOptions.ts#L435-L436), so these options don't appear to actually be ignored.

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -24,7 +24,6 @@ export type RemoteEntryType =
 
 import * as fs from 'fs';
 import * as path from 'pathe';
-import { warn } from './logUtils';
 
 interface ExposesItem {
   import: string;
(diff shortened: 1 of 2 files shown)


src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -24,7 +24,6 @@ export type RemoteEntryType =
 
 import * as fs from 'fs';
 import * as path from 'pathe';
-import { warn } from './logUtils';
```
gioboa, line 27: We can remove the warn function as well.

vergilfromadyen: Removed :)

Issue #354, opened 2025-11-12
[Bug]: Remote Component Fails to Render When Using Slots (Vue3 + Vite + Module Federation)


wangjingyangmaster:
### Describe the bug

I have also encountered this problem and it has not been resolved. 
https://github.com/module-federation/vite/issues/282

### Version

1.9.0

### Reproduction

https://github.com/neversleeppy/mf-vite-slot

### Relevant log output

```shell

```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
- [x] Check that there isn't [already an issue](https://github.com/module-federation/vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

gioboa:
Closed by #358

Issue #282, opened 2025-03-31
[Bug]: Remote Component Fails to Render When Using Slots (Vue3 + Vite + Module Federation)


neversleeppy:
## Description
When loading a remote Vue component containing slots via Module Federation, the component fails to render. Console shows error:
 TypeError: Cannot read properties of null (reading 'ce')
![Image](https://github.com/user-attachments/assets/a2284511-89d8-4f42-b34c-3f62624242db)

![Image](https://github.com/user-attachments/assets/9203a2ae-9df4-44bb-928b-9256578f29fa)

### Steps to Reproduce
1. Configure a remote app to expose a component with slots, e.g.:
```vue
<template>
<!-- remote -->
  <button
    style="
      border: 0 solid #e2e8f0;
      margin-top: 10px;
      background-color: rgb(246, 179, 82);
      border-radius: 0.25rem;
      font-weight: 700;
      padding: 0.5rem 1rem 0.5rem 1rem;
      color: rgb(24, 24, 24);
    "
    @click="store.increment"
  >
  Remote counter: {{ count }}
  <p>
    <p>this is slot</p>
    <slot></slot>
  </p>
  </button>
</template>
```
2. use the component in host application
```vue
<script setup lang="ts">
import { defineAsyncComponent } from "vue";
import Counter from "./components/Counter.vue";
const RemoteMFE = defineAsyncComponent(
  // @ts-ignore
  () => import("remote/remote-app")
);
</script>

<template>
  <RemoteMFE v-if="!!RemoteMFE" />
</template>
```
### Reproduction Repo
[https://github.com/neversleeppy/mf-vite-slot]([](https://github.com/neversleeppy/mf-vite-slot))

### Environment
Framework: Vue3 + Vite
Module Federation: "@module-federation/vite": "1.2.5",

### Version

1.2.5

### Reproduction

https://github.com/neversleeppy/mf-vite-slot

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
- [x] Check that there isn't [already an issue](https://github.com/module-federation/vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

gioboa:
Hi @neversleeppy thanks for your feedback, this is a good first issue.
Feel free to drop a PR to solve this problem. Thanks

[...]

PR #358, opened 2025-11-18
chore: warn if shared dependency is resolved to somewhere (#354)

CaptainWang98:
Related to: #354 #282 
In vite, `resolve.alias` has higher priority than MF.
Accidently, user may point shared dependency to somewhere else:
https://github.com/neversleeppy/mf-vite-slot/blob/main/host/vite.config.ts
This may cause unexpected runtime behavior.
This PR will detect conflict between `resolve.alias` and `MF.shared`, and then show warn in build-time.

### DIFF
--- a/src/plugins/__tests__/pluginCheckAliasConflicts.test.ts
+++ b/src/plugins/__tests__/pluginCheckAliasConflicts.test.ts
@@ -64,7 +64,7 @@ describe('pluginCheckAliasConflicts', () => {
 
     plugin.configResolved!(mockConfig as any);
 
-    expect(mockLogger.warn).toHaveBeenCalledTimes(5);
+    expect(mockLogger.warn).toHaveBeenCalledTimes(4);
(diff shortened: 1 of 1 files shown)


src/plugins/__tests__/pluginCheckAliasConflicts.test.ts
```diff
@@ -0,0 +1,199 @@
...
+        ],
+      },
+    };
+
+    plugin.configResolved!(mockConfig as any);
+
+    expect(mockLogger.warn).not.toHaveBeenCalled();
+  });
+
+  it('should not warn when shared is empty', () => {
```
gioboa, line 111: Can you add a test case with undefined alias?

CaptainWang98: Done, 'undefined' should always be handled

gioboa (approved):
Thanks @CaptainWang98 
LGTM 👏

PR #359, opened 2025-11-21
fix: generated alias should not trigger warning

CaptainWang98:
There are some alias created by `proxySharedModule`. They should not trigger conflict warnings.


gioboa (approved):
Thanks it looks great to me 🎉

Issue #362, opened 2025-11-28
[Bug]: Cannot load shared dependencies for the first time


wojtask9:
### Describe the bug

```
npm i
npm run dev
```
open in browser http://localhost:5010/app-1

in console.log there be an error.

After pressing F5 everything is OK and everything works correct.

To reproduce bug you must kill `npm run dev` and launch again.

This issue is also reproduced with lastest versions (everything including vite, vue, module-federation, etc).

Not sure where the problem is but if you remove `shared` from remote apps (app1, app2)  then issue doesn't exists.
Also porting to non `deprecated  init` to use `registerRemotes` and checking if `getInstance() != null` (always is not null) issue still exists.

Issue exists in firefox and chrome. I'm using linux env.

This repo is taken from https://dev.to/lmlonghuynh/building-a-micro-frontend-architecture-with-vue-3-vite-and-module-federation-1bb1

### Version

probably all

### Reproduction

https://github.com/lmlong-huynh/micro-frontend-sample

### Relevant log output

```shell
[MF] Error loading remote module: Error: [ Federation Runtime ]: Please call createInstance first. #RUNTIME-009
View the docs to see how to solve: https://module-federation.io/guide/troubleshooting/runtime/RUNTIME-009
    error2 logger.ts:22
    assert2 logger.ts:10
    loadShare index.ts:71
    res chunk-DW7XOL6G.js:17
    promise callback*node_modules/__mf__virtual/app1__loadShare__vue__loadShare__.js chunk-DW7XOL6G.js:17
    __require chunk-ZKAWKZG5.js:11
    <anonymous> chunk-DW7XOL6G.js:31
```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
- [x] Check that there isn't [already an issue](https://github.com/module-federation/vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

wojtask9:
super cool :)

PR #363, opened 2025-12-01
fix: use already-initialized runtime instance to avoid mixing ESM and CJS imports

sabov:
Issue: https://github.com/module-federation/vite/issues/362

Multiple instances of `@module-federation/runtime` are used because:
- virtualRemoteEntry.ts imports via ESM import
- virtualShared_preBuild.ts imports via CommonJS require

The fix is to get loadShare/loadRemote methods directly from the runtime instance that was passed through initPromise.

An alternative solution could be turning `@module-federation/runtime` into a singleton but there are multiple edge cases to take care of like: it could be pre-bundled in one of the remotes, multiple versions could be imported, etc.

### DIFF
--- a/src/virtualModules/virtualRemotes.ts
+++ b/src/virtualModules/virtualRemotes.ts
@@ -25,12 +25,9 @@ export function getUsedRemotesMap() {
 }
 export function generateRemotes(id: string, command: string) {
   return `
-    // Use dynamic import() - works in both browser (dev) and Rollup (build)
-    const initModulePromise = import("${virtualRuntimeInitStatus.getImportId()}")
-    const res = initModulePromise
-      .then(({initPromise}) => initPromise)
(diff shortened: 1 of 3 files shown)


sabov:
Looks like the `runtime.loadShare()` instance method returns false when a share isn't found, while the global `loadShare()` function has fallback behavior to look across different share scope. This might be a breaking change for users.

sabov:
> Looks like the runtime.loadShare() instance method returns false when a share isn't found, while the global loadShare() function has fallback behavior to look across different share scope. This might be a breaking change for users.

Ignore this, I forgot to use `lib: () => RefToLib` in my other example so it looks good on my end now. 

EDIT: In fact we can't ignore this, before the change the page would load eventually (after a couple of reloads) but now there is an error unless you specify the `lib`. So indeed a breaking change. 

> We should definitely try with the latest module federation runtime.

I can update the package and give it a try.

PR #256, opened 2025-02-13
fix: solve issue with location of __mf__virtual in monorepo

Xeikim:
https://github.com/module-federation/vite/issues/254

PR #370, opened 2025-12-11
feat: Yarn PnP support

### LINKED ISSUE #155: Yarn support: Yarn PnP can't find __mf__virtual
When using Yarn PnP + Vite I get this error message:
```
Failed to resolve dependency: __mf__virtual, present in 'optimizeDeps.include'
```
And then after the server is started:
```
4:42:25 PM [vite] Pre-transform error: Failed to resolve import "__mf__virtual/test__mf_v__runtimeInit__mf_v__.js" from "virtual:mf-REMOTE_ENTRY_ID". Does the file exist?
4:42:25 PM [vite] Internal server error: Failed to resolve import "__mf__virtual/test__mf_v__runtimeInit__mf_v__.js" from "virtual:mf-REMOTE_ENTRY_ID". Does the file exist?
```

ha1fstack:
closes #155

## Problem

Yarn PnP can not resolve arbitary qualified resolutions (or packages) whose are not declared explicitly.

However this plugin relies on on-the-fly generated __mf__virtual package inside physical node_modules folder - so it was not compatible

## Tries

I first tried to use Vite's virtual module (in-memory) instead of real fs. But it did not work well since this plugin seems to be relying on (the intend is not documented so this is a guess) Vite's dependency pre-bundling - but virtual modules cannot be pre-bundled.

## Solution

The root cause was not fs, so I focused on the qualified / unqualified resolution part. Instead of making the generated code as a package (with package.json and importing it with name), using unqualified resolution (real path) + creating alias to that path to keep things simple solved the issue.

Also fixed the problematic searchPackageVersion util too since the packages needs to be resolved from the host package (the service) root not the vite-mf's package root.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,11 +3,13 @@ import { Plugin } from 'vite';
 import addEntry from './plugins/pluginAddEntry';
 import { checkAliasConflicts } from './plugins/pluginCheckAliasConflicts';
 import { PluginDevProxyModuleTopLevelAwait } from './plugins/pluginDevProxyModuleTopLevelAwait';
+import pluginDts from './plugins/pluginDts';
 import pluginManifest from './plugins/pluginMFManifest';
 import pluginModuleParseEnd from './plugins/pluginModuleParseEnd';
 import pluginProxyRemoteEntry from './plugins/pluginProxyRemoteEntry';
 import pluginProxyRemotes from './plugins/pluginProxyRemotes';
 import { proxySharedModule } from './plugins/pluginProxySharedModule_preBuild';
+import pluginVarRemoteEntry from './plugins/pluginVarRemoteEntry';
 import aliasToArrayPlugin from './utils/aliasToArrayPlugin';
 import {
   ModuleFederationOptions,
@@ -44,6 +46,7 @@ function federation(mfUserOptions: ModuleFederationOptions): Plugin[] {
     aliasToArrayPlugin,
     checkAliasConflicts({ shared }),
     normalizeOptimizeDepsPlugin,
+    ...pluginDts(options),
     ...addEntry({
       entryName: 'remoteEntry',
       entryPath: REMOTE_ENTRY_ID,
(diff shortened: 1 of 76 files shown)


src/utils/VirtualModule.ts
```diff
@@ -1,32 +1,24 @@
...
+  if (cachedNodeModulesDir) {
+    return cachedNodeModulesDir;
   }
+  const targetRoot = rootDir || process.cwd();
+  const nodeModulesPath = join(targetRoot, 'node_modules');
+  if (!existsSync(nodeModulesPath)) {
+    mkdirSync(nodeModulesPath, { recursive: true });
+  }
+  cachedNodeModulesDir = nodeModulesPath;
   return cachedNodeModulesDir;
```
ha1fstack, line 22: We are using vite project's root instead of traversing for existing folder because pnp doesn't generate node_modules and this seems more natural

src/index.ts
```diff
@@ -90,9 +90,12 @@ function federation(mfUserOptions: ModuleFederationOptions): Plugin[] {
           },
         });
         const virtualDir = options.virtualModuleDir || '__mf__virtual';
+        const virtualDirPath = VirtualModule.getVirtualModuleDirPath();
+        (config.resolve as any).alias.push({
+          find: new RegExp(`^${virtualDir}/`),
+          replacement: `${virtualDirPath}/`,
+        });
```
ha1fstack, line 105: using alias to map path to __mf__virtual

src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -126,7 +126,7 @@ function removePathFromNpmPackage(packageString: string): string {
  */
 function searchPackageVersion(sharedName: string): string | undefined {
   try {
-    const sharedPath = require.resolve(sharedName);
+    const sharedPath = require.resolve(sharedName, { paths: [process.cwd()] });
```
ha1fstack, line 129: It would be better to use resolved config root instead of cwd but this seems fine for now

src/utils/VirtualModule.ts
gioboa: I see, I'm not sure if this is working on monorepo projects. cc @Xeikim you added this fix long time ago for monorepos.

ha1fstack: The original problem was that rootDir find traversal was using process.cwd not the actual vite project's root
We are still using vite's project rootDir (fixed by #256) but only omitting the traversal part - so it seems fine to me.

ha1fstack:
Hi @gioboa , can you review my reply?

src/utils/VirtualModule.ts
gioboa: `findNodeModulesDir` is traversing upwards from a starting directory to find the correct path though.

CXZ7720:
@ha1fstack Thanks for your contribute. Is there any chance to re-open or continue this task?

PR #371, opened 2025-12-12
fix: Add peerDependencies for vite

zhangHongEn:
The plugin currently does not support Vite 7.


gioboa (approved):
Thanks for your help @zhangHongEn

PR #373, opened 2025-12-15
fix: add-entry with renderBuiltUrl [372]

### LINKED ISSUE #372: [Bug]: `hostinit` is not prepended with `renderBuiltUrl`
### Describe the bug

When building with vite's experimental `renderBuiltUrl`, the `hostinit` file does not correctly render the intended path.

With the linked repro, `cd host`, then `pnpm build`, and inspect the file located at `./host/dist/index.html`. You'll see something like:
```html
<!doctype html>
<html lang="en">
	<head>
          <script type="module" src="/assets/hostInit-rD8F1DyE.js"></script>
```

`src="/assets/hostInit-rD8F1DyE.js"` is not prepended with the result from `renderBuiltUrl`.

### Version

v6.4.1

### Reproduction

https://stackblitz.com/edit/github-qxeypwge?file=host%2Fdist%2Findex.html

### Relevant log output

```shell

```

### Validations

BiscuiTech:
Linked is a small fix to added support to `renderBuiltUrl` inside add-entry's plugin.
Fixes #372

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Vite plugin for Module Federation",
   "type": "module",
   "source": "src/index.ts",
-  "main": "./lib/index.ts",
+  "main": "./lib/index.cjs",
   "types": "lib/index.d.ts",
   "files": [
     "lib/**/*"


gioboa (approved):
Thanks for this great fix @BiscuiTech

gioboa:
> this seems to break when defining it with the runtime option

Thanks @deanohyeah for you report. Can you create an issue for it please?

PR #393, opened 2026-01-29
fix: adjust renderBuiltUrl resolution for relative pathing

### LINKED ISSUE #384: [Bug]:
### Describe the bug

 experimental.renderBuiltUrl in the add-entry plugin is called with an incomplete signature that breaks compatibility with Vite's RenderBuiltAssetUrl type and libraries that depend on it.

  In lib/index.cjs line 156, the plugin calls:
  renderBuiltUrl(file)

  However, Vite's RenderBuiltAssetUrl type expects:
  renderBuiltUrl(filename, { type?: 'public' | 'asset' | 'js', ... })

  This causes two issues:

  1. Missing options parameter: Functions expecting options like { hostType } fail with "Cannot destructure property of undefined"
  2. Unhandled runtime functions: When renderBuiltUrl returns { runtime: "function code" }, it gets stringified to [object Object] in the HTML instead of being processed correctly

### Version

1.9.5 

### Reproduction

https://github.com/deanohyeah/vite-mf-renderbuildurl-bug

### Relevant log output

```shell

```

### Validations

ryok90:
# Fix renderBuiltUrl API signature in pluginAddEntry

Fixes #384

## Problem

PR #373 added `renderBuiltUrl` support but had two critical bugs:

1. **Incomplete API signature**: Called `renderBuiltUrl(file)` with only one parameter, but Vite expects `renderBuiltUrl(filename, options)`
2. **Missing return type handling**: Didn't handle the three possible return types (string, `{runtime}`, `{relative}`)

This broke compatibility with any `renderBuiltUrl` implementation that depends on the `options` parameter, causing "Cannot destructure property of undefined" errors.

## Solution

Updated `src/plugins/pluginAddEntry.ts` (lines 135-191) to:

1. **Call with correct signature**: Pass both `filename` and `options` object with `hostId`, `hostType`, `type`, and `ssr` parameters
2. **Handle all return types**:
   - `string` → Use directly
   - `{relative: true}` → Use filename without base prefix
   - `{runtime: "code"}` → Warn and fallback (incompatible with HTML `<script src="">`)
   - `undefined` → Fallback to default `base + file`
3. **Per-file resolution**: Resolve path for each HTML file with correct `hostId`

## Testing

- ✅ Build succeeds
- ✅ All unit tests pass (61 tests)
- ✅ vite-vite example builds successfully
- ✅ Tested with reproduction from #384 - build completes without errors
- ✅ Backward compatible (works without `renderBuiltUrl` configured)

## Preserves PR #373

The fix maintains the original functionality from #373 (issue #372) - `renderBuiltUrl` transformations are correctly applied to hostInit script tags.

### DIFF
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -135,21 +135,55 @@ const addEntry = ({
       generateBundle(options, bundle) {
         if (!injectHtml()) return;
         const file = this.getFileName(emitFileId);
-        const path = viteConfig.experimental?.renderBuiltUrl
-          ? viteConfig.experimental?.renderBuiltUrl(file)
-          : viteConfig.base + file;
-        const scriptContent = `
-          <script type="module" src="${path}"></script>
-        `;
+        // Helper to resolve path with proper renderBuiltUrl handling
+        const resolvePath = (htmlFileName: string): string => {
+          if (!viteConfig.experimental?.renderBuiltUrl) {
+            return viteConfig.base + file;
+          }
 
+          const result = viteConfig.experimental.renderBuiltUrl(file, {
+            hostId: htmlFileName,
+            hostType: 'html',
+            type: 'asset',
+            ssr: false,
+          });
+
+          // Handle return types
+          if (typeof result === 'string') {
+            return result;
+          }
+
+          if (result && typeof result === 'object') {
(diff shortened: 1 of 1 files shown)


gioboa (approved):
Thank you @ryok90 nice fix 💪

PR #395, opened 2026-01-30
docs: add varFilename description

crutch12:
Just a little mention about `varFilename` option usage


gioboa (approved):
Thanks @crutch12 you rock 🚀

PR #402, opened 2026-02-12
fix(css): correctly extract CSS from the vite pipeline

joshuaellis:
### Description
In Vite environment builds, chunk.viteMetadata.importedCss is not populated even when the chunk contains CSS modules (e.g. from vanilla-extract). This means CSS assets are emitted to disk but never appear in `mf-manifest.json` under `exposes[].assets.css`

### Solution
The fix adds a fallback: when importedCss is empty, check if the chunk's modules contain any CSS files. If they do, associate the bundle's CSS assets with that chunk.

### Notes
Environments was technically added in V6, but better finalised in V7, so this probably helps towards #320

### DIFF
--- a/.changeset/sharp-eagles-jog.md
+++ b/.changeset/sharp-eagles-jog.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': patch
+---
+
+fix: css loaded in environments was not linked to exposed modules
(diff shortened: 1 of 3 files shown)


gioboa (approved):
Thanks it looks great to me 👏 
Nice fix @joshuaellis

PR #404, opened 2026-02-12
fix(virtual-modules): transitive dependency breaks PNPM

joshuaellis:
## Description

When using @module-federation/vite with pnpm (strict mode, no shameful hoisting), the dev server fails with:

```
Failed to resolve import "@module-federation/dts-plugin/dynamic-remote-type-hints-plugin" from "virtual:mf-REMOTE_ENTRY_ID". Does the file exist?                                                                                                                                                            
```

This fix intercepts imports originating from the virtual remote entry and re-resolves them from this package's own location, where the transitive dependencies are available.

## Notes

There weren't any unit tests nor e2e tests that I could write to be more confident in the change other than running on my local instance – open to suggestions though.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,9 @@
 {
   "compilerOptions": {
     "outDir": "./dist",
-    "module": "commonjs",
-    "target": "es6",
+    "module": "preserve",
+    "moduleResolution": "bundler",
+    "noEmit": true,
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,


src/plugins/pluginProxyRemoteEntry.ts
```diff
@@ -35,6 +35,21 @@ export default function (): Plugin {
...
+      if (
+        importer === REMOTE_ENTRY_ID &&
+        !id.startsWith('.') &&
+        !id.startsWith('/') &&
+        !id.startsWith('\0') &&
+        !id.startsWith('virtual:')
+      ) {
+        const resolved = await this.resolve(id, __filename, { skipSelf: true });
+        if (resolved) return resolved;
+      }
```
gioboa, line 55: It looks great to me 👏 Thanks @joshuaellis 

I double checked with AI and __filename is CJS-only, in ESM we can use fileURLToPath(import.meta.url)

Let's add a fallback
```
const importerPath = typeof __filename !== 'undefined' ? __filename : fileURLToPath(import.meta.url);
const resolved = await this.resolve(id, importerPath, { skipSelf: true });
if (resolved) return resolved;
```

gioboa: Fixed.

PR #409, opened 2026-02-13
chore: refactor bundler to be tsdown instead of microbundle

joshuaellis:
### Description

Moves the repo over to use tsdown over microbundle, the later hasn't been updated in ~3 years and was stopping important TS upgrades we could be utilising. I've checked the type exports by running `npx --yes @arethetypeswrong/cli --pack .` and are happy with the output. I've also used `yalc` locally to connect it to a project I have and it all seems well. I got AI to do some fake ESM/CJS tmp runs to test it and the imports were also successful.

### DIFF
--- a/.github/publish-ci/node-esm/.gitignore
+++ b/.github/publish-ci/node-esm/.gitignore
@@ -1,0 +1,35 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# testing
+/coverage
+
+# production
+/build
+
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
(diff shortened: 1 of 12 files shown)


joshuaellis:
> Nice, can you add the changeset please?

Sure can, just so I understand, do you prefer changesets for all PRs or those that would require the package to be published for instance I wouldn't have done one for this because it isn't necessarily public facing - no strong preference ofc!

joshuaellis:
> We added changeset few weeks ago and we didn't create a release yet with this new flow. I would like to be transparent and highlight all the changes we are doing in the repo.

SGTM, thank you for clarifying!

joshuaellis:
_I do_ have a CI workflow to test artifacts that I borrowed from the redux gang – https://github.com/pmndrs/react-spring/blob/2ff6de7a3b295a79475113824b0962ebf3ca5249/.github/workflows/tests.yml#L185

_if_ you want it implemented?

gioboa (approved):
Thanks @joshuaellis

PR #412, opened 2026-02-15
fix: inline usedShared generation to prevent empty remoteEntry in preview builds

cipsys:
## Summary

Fixes the "factory is not a function" error that occurs in Vite preview mode when Module Federation tries to load shared dependencies.

## Problem

When running `vite preview`, the generated `remoteEntry.js` had empty `usedShared = {}` and `usedRemotes = []`, causing Module Federation to fail with:

```
TypeError: factory is not a function
```

This only occurred in preview/production builds, not in dev mode.

## Root Cause

The `generateRemoteEntry()` function in `virtualRemoteEntry.ts` was importing `usedShared` and `usedRemotes` from the `localSharedImportMap` virtual module:

```typescript
import {usedShared, usedRemotes} from "${getLocalSharedImportMapPath()}"
```

During Rollup bundling, this import statement was being stripped/resolved incorrectly, resulting in the empty objects being generated in the final bundle.

## Solution

Instead of importing from a virtual module, the fix inlines the complete generation of both `importMap` and `usedShared` directly in `generateRemoteEntry()`. Key changes:

1. **Generate importMap inline**: Creates async import functions for each shared module
2. **Generate usedShared inline**: Creates complete shared module entries with `get()` functions
3. **Use options.shared**: Uses ALL configured shared modules instead of `getUsedShares()`, because the HOST application needs to provide all configured shared modules, not just the ones it dynamically imports

## Changes

- Modified `packages/module-federation-vite/src/virtualModules/virtualRemoteEntry.ts`:
  - Added inline generation of `importMap` with dynamic imports
  - Added inline generation of `usedShared` with complete `get()` functions
  - Changed from using `getUsedShares()` to `options.shared` to include all configured shared modules

## Testing

Tested with a production Vite build in preview mode:

- ✅ No "factory is not a function" errors
- ✅ No "Can not get shared" errors  
- ✅ Module Federation correctly loads shared React, MUI, Emotion, and other dependencies
- ✅ Generated `remoteEntry.js` contains populated `usedShared` object with working `get()` functions

## Example

**Before (broken):**
```javascript
const usedShared = {}
const usedRemotes = []
```

**After (fixed):**
```javascript
const importMap = {
  "react": async () => {
    let pkg = await import("./assets/index-DY0B4gru.js");
    return pkg;
  },
  // ... other modules
}
const usedShared = {
  "react": {
    name: "react",
    version: "19.0.0",
    scope: ["default"],
    loaded: false,
    from: "mygpt_host",
    async get () {
      usedShared["react"].loaded = true
      const {"react": pkgDynamicImport} = importMap
      const res = await pkgDynamicImport()
      const exportModule = {...res}
      Object.defineProperty(exportModule, "__esModule", {
        value: true,
        enumerable: false
      })
      return function () {
        return exportModule
      }
    },
    shareConfig: {
      singleton: true,
      requiredVersion: "^19.0.0"
    }
  },
  // ... other modules
}
```

## Impact

This fix ensures Module Federation works correctly in production builds, not just in dev mode. It resolves a critical issue that would prevent applications using Module Federation + Vite from deploying to production.

## Related Issues

This may be related to other reports of Module Federation not working in production builds with Vite.


src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -137,14 +141,98 @@ export function generateRemoteEntry(options: NormalizedModuleFederationOptions):
...
+        entry: ${JSON.stringify(remote.entry)},
+        shareScope: ${JSON.stringify(remote.shareScope) ?? 'default'},
+      }`;
+    })
+    .filter((x) => x !== null)
+    .join(',');
+
+  // Dev mode: Use virtual module imports (avoids circular dependencies in Vite's on-demand serving)
+  // Production: Use inline code (works with Rollup bundling)
+  const sharedImportsCode = isDevMode
```
gioboa, line 222: I'm not sure about this, dev mode works differently than preview.
This means that all configured shared modules are included in the remoteEntry, not just the ones actually used. 
What do you think @Nsttt ?

PR #416, opened 2026-02-17
feat: enable all of vite@7

### LINKED ISSUE #320: ✨ Vite 7 Support

joshuaellis:
### Description

Off the back of #415 which refactored the virtual modules to be ESM during build, this PR looks to resolve the locking issue which stops builds. This PR also aims to ensure `vite-rolldown` works correctly.

### Related Issues
* resolves #320
* closes #369

### Notes
This PR requires: 
* #415 to be merged first

### DIFF
--- a/.changeset/perky-mangos-say.md
+++ b/.changeset/perky-mangos-say.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': minor
+---
+
+feat: allow all v7 version of vite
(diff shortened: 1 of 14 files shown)


gioboa (commented):
Thanks for your help, I created #425 and it's addressing the issue.
So I'm closing this one.

PR #418, opened 2026-02-18
fix: prevent 504 Outdated Optimize Dep errors in dev mode

AlonMiz:
## Problem

When using `@module-federation/vite` in dev mode, Vite frequently returns **504 (Outdated Optimize Dep)** errors for virtual module files under `node_modules/__mf__virtual/`. This causes broken page loads, missing shared dependencies, and a poor developer experience.

The issue is particularly noticeable when using [AnalogJS](https://analogjs.org/) (Vite-based Angular meta-framework) with Module Federation, but affects any Vite project using shared dependencies.

### Root Cause

The plugin creates virtual module files (physical `.js` files in `node_modules/__mf__virtual/`) that represent shared dependencies. These files are created in the `configResolved` hook, which runs **after** Vite has already started its dependency optimization phase.

When Vite's optimizer encounters these "new" files mid-optimization, it triggers a re-optimization cycle. During this re-optimization, any in-flight requests for the old optimized modules receive a **504 Outdated Optimize Dep** response.

The existing workaround (`optimizeDeps.force = true`) forces Vite to re-optimize on every dev server start, but this creates a race condition: the plugin writes files while Vite is simultaneously trying to optimize them.

## Solution

This PR fixes the timing issue by ensuring virtual module files exist **before** Vite's optimization phase begins.

#### 1. `src/index.ts` — Early virtual module creation plugin
Added `createEarlyVirtualModulesPlugin()` that runs in the `config` hook (which executes before `configResolved` and before optimization):
- Creates the `__mf__virtual` directory structure
- Writes all shared module virtual files (loadShare + preBuild)
- Adds virtual module paths to `optimizeDeps.include` so Vite knows about them upfront

#### 2. `src/plugins/pluginProxySharedModule_preBuild.ts` — Eager state initialization
Added eager population of `usedShares` in `configResolved` to ensure the plugin's internal state (which tracks which shared modules are active) is correctly initialized even when Vite uses cached optimization results and skips the `customResolver` callbacks.

#### 3. `src/utils/normalizeOptimizeDeps.ts` — Remove force workaround
Removed `optimizeDeps.force = true` (which had a `// todo: fix this workaround` comment). This is no longer needed since virtual files now exist before optimization starts.

#### 4. `src/utils/VirtualModule.ts` — Infrastructure helper
Added `initVirtualModuleInfrastructure()` to create the virtual module directory and `package.json` before the `VirtualModule` class is fully initialized.

#### 5. `src/virtualModules/virtualShared_preBuild.ts` — Defensive null check
Added null check in `writeLoadShareModule()` to handle the case where it's called before `getLoadShareModulePath()` has populated the cache map.

## Vite Plugin Lifecycle Context

```
config hook        ← NEW: createEarlyVirtualModulesPlugin creates files here
    ↓
configResolved     ← Existing: VirtualModule.setRoot(), eager usedShares population
    ↓
optimizeDeps       ← Vite optimizes deps (files already exist, no 504s)
    ↓
dev server ready   ← All modules available
```

## Testing

Tested with an AnalogJS (Angular) host + AnalogJS (Angular) remote + React remote Module Federation setup:
- ✅ Host page: zero 504 errors, zero network errors
- ✅ Remote Angular (AnalogJS) page: zero 504 errors, zero network errors
- ✅ Remote React page: zero 504 errors, zero network errors
- ✅ Host MFE page loading both remotes: zero 504 errors, zero network errors
- ✅ Consistent across multiple consecutive test runs

### DIFF
--- a/.changeset/bright-sheep-boil.md
+++ b/.changeset/bright-sheep-boil.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': patch
+---
+
+fix: fix up imports originating from the virtual remote entry
--- a/.changeset/loose-donuts-help.md
+++ b/.changeset/loose-donuts-help.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': minor
+---
+
+feat: add target option to configuration for better tree-shaking
--- a/.changeset/ripe-cloths-grin.md
+++ b/.changeset/ripe-cloths-grin.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': patch
+---
+
+chore: update bundler from microbundle to tsdown
--- a/.changeset/sharp-eagles-jog.md
+++ b/.changeset/sharp-eagles-jog.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': patch
+---
+
+fix: css loaded in environments was not linked to exposed modules
--- a/.changeset/violet-garlics-allow.md
+++ b/.changeset/violet-garlics-allow.md
@@ -1,0 +1,5 @@
+---
+'@module-federation/vite': patch
+---
+
+fix: apply bundler specific defaults
--- a/.github/publish-ci/node-esm/.gitignore
+++ b/.github/publish-ci/node-esm/.gitignore
@@ -1,0 +1,35 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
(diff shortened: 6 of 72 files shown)


gioboa (commented):
> 3. Open http://localhost:4200 — you'll see 504 errors in the browser network tab

I tested your ☝️ project with PR #425 and no 504 errors, the app it's working fine.
So I'm closing this PR.
Thanks for your commitment @AlonMiz

PR #423, opened 2026-02-21
fix(examples): enable HMR for vite-vite example host and remotes

sabov:
Add `reactRefreshHost` to the remote's React plugin config to route HMR updates through the host's WebSocket. Name anonymous default exports so React Fast Refresh can track component boundaries.


gioboa (approved):
Thanks @sabov for your help, it looks great to me 👏

PR #425, opened 2026-02-25
feat: support rolldown-vite + vite 8

### LINKED ISSUE #320: ✨ Vite 7 Support

### LINKED ISSUE #376: [Bug]: 504 Outdated Optimize Dep errors when using shared dependencies in dev mode
## Describe the bug

When using `@module-federation/vite` with shared dependencies in dev mode, Vite throws "504 Outdated Optimize Dep" errors.

**Root Cause**: The issue is in [`src/utils/normalizeOptimizeDeps.ts` line 17](https://github.com/module-federation/vite/blob/4307b7aa2f1b01b7f5b11cf1ece5fa7f9b8c83bc/src/utils/normalizeOptimizeDeps.ts#L17):

```typescript
// todo: fix this workaround
optimizeDeps.force = true;
```

The plugin sets `optimizeDeps.force = true`, which forces Vite to re-optimize dependencies on every server start. When combined with shared dependencies, this causes a race condition where the federation plugin receives a 504 error because the optimized dependency it was loading becomes "outdated".

**Current Workaround**: Disable shared dependencies in dev mode:
```typescript
shared: isDev ? {} : { /* shared deps */ }
```

## Version

- `@module-federation/vite`: 1.9.4
- `vite`: 7.3.1

## Reproduction

https://github.com/user/analog-federation (Nx monorepo with Angular + React remotes)

Steps:
1. Configure shared dependencies in federation config
2. Run `vite dev`
3. Observe 504 errors in console

## Relevant log output

```
504 Outdated Optimize Dep
```

## Suggested Fix

Instead of always forcing re-optimization, the plugin should use a smarter invalidation strategy:

### Option 1: Hash-based invalidation
Only force re-optimization when the federation config actually changes:

```typescript
// In normalizeOptimizeDeps.ts
import { createHash } from 'crypto';

function getFederationConfigHash(config: FederationConfig): string {
  const configString = JSON.stringify({
    name: config.name,
    remotes: config.remotes,
    exposes: config.exposes,
    shared: config.shared,
  });
  return createHash('md5').update(configString).digest('hex');
}

// Instead of: optimizeDeps.force = true;
// Check if config changed since last run
const federationHash = getFederationConfigHash(federationConfig);
if (hasFederationConfigChanged(federationHash)) {
  optimizeDeps.force = true;
  saveFederationConfigHash(federationHash);
}
```

### Option 2: Use optimizeDeps.entries instead of force
Pre-specify federation modules without forcing full re-optimization:

```typescript
// Don't force, instead explicitly include federation entries
// optimizeDeps.force = true; // REMOVE THIS

optimizeDeps.entries = [
  ...(optimizeDeps.entries || []),
  '**/remoteEntry.js',
];

// Include shared deps explicitly
if (federationConfig.shared) {
  optimizeDeps.include = [
    ...(optimizeDeps.include || []),
    ...Object.keys(federationConfig.shared),
  ];
}
```

## Validations

- [x] Read the docs.

gioboa:
Implement conditional virtual module generation (ESM/CJS) based on Vite version and Rolldown detection.

Close #320 
Close #369
Close #376

### DIFF
--- a/.changeset/bright-sheep-boil.md
+++ b/.changeset/bright-sheep-boil.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-fix: fix up imports originating from the virtual remote entry
--- a/.changeset/loose-donuts-help.md
+++ b/.changeset/loose-donuts-help.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': minor
----
-
-feat: add target option to configuration for better tree-shaking
--- a/.changeset/misty-hounds-laugh.md
+++ b/.changeset/misty-hounds-laugh.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-chore: add changesets configuration
--- a/.changeset/ripe-cloths-grin.md
+++ b/.changeset/ripe-cloths-grin.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-chore: update bundler from microbundle to tsdown
--- a/.changeset/sharp-eagles-jog.md
+++ b/.changeset/sharp-eagles-jog.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-fix: css loaded in environments was not linked to exposed modules
--- a/.changeset/some-chefs-occur.md
+++ b/.changeset/some-chefs-occur.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-ci: bump version of workflows packages
--- a/.changeset/tiny-pets-judge.md
+++ b/.changeset/tiny-pets-judge.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-chore: update changesets config
--- a/.changeset/violet-garlics-allow.md
+++ b/.changeset/violet-garlics-allow.md
@@ -1,5 +1,0 @@
----
-'@module-federation/vite': patch
----
-
-fix: apply bundler specific defaults


package.json
```diff
@@ -59,7 +59,7 @@
   "homepage": "https://github.com/module-federation/vite#readme",
   "packageManager": "pnpm@10.28.2",
   "peerDependencies": {
-    "vite": "<=7.1.7"
+    "vite": ">=5.0.0"
```
joshuaellis:
```suggestion
    "vite": "^5.0.0 || ^6.0.0 || ^7.0.0"
```

Is a bit more explicit?

src/utils/viteVersion.ts
```diff
@@ -0,0 +1,21 @@
...
+
+function _isEsmSupported(): boolean {
+  try {
+    _require.resolve('rolldown-vite');
+    return true;
+  } catch {
+    const major = parseInt(_require('vite/package.json').version, 10);
+    return major >= 8;
+  }
+}
```
joshuaellis: would it be better to follow these guidelines? https://vite.dev/guide/rolldown#detecting-rolldown-vite

gioboa: fixed

package.json
gioboa: fixed

src/utils/viteVersion.ts
gioboa: I tested with `!!(vite as any).rolldownVersion` as mentioned in the docs but it's not working fine 😢

gioboa: let's keep the actual method, I tested it and it's working fine.

gioboa: I used this.meta.rolldownVersion 
it's better 💪

Nsttt:
Throw me a reproduction and I'll dig deeper

wojtask9:
@gioboa probably I'll try tomorrow if you can't reproduce issue :)

PR #443, opened 2026-03-05
fix(pluginModuleParseEnd): replace fixed build timeout with idle timeout

### LINKED ISSUE #437: [Bug]: some usedRemotes may be missing during long builds
### Describe the bug

I propose reworking the protection against long-running module parsing. Currently, there is an option to set the `moduleParseTimeout` parameter, but it has a certain catch. If the build time exceeds this parameter's value, some usedRemotes may be missing from the remoteEntry, and the application crashes during initialization with the error `Failed to locate remote`. It's clear that protection against hanging is necessary, but with this kind of overall timeout, as the codebase grows (and consequently the build time increases), users will inevitably run into this issue. I suggest implementing not an overall build timeout, but a timeout starting from the last module resolution (any module, not just from Module Federation). This way, nothing will break during long builds, but the protection against hanging will still be in place.

### Version

1.11.1

### Reproduction

Try to build huge codebase using default `moduleParseTimeout`

### Relevant log output

```shell

```

### Validations

DemonStore:
Close #437

## Problem

`moduleParseTimeout` was a fixed timer that started counting from the very beginning of the build. On large codebases where the total build time exceeds the timeout value, the `parsePromise` resolved prematurely — before all modules were parsed. This caused some `usedRemotes` to be missing from `remoteEntry`, leading to **"Failed to locate remote"** errors at runtime.

As codebases grow (and build times increase), users inevitably hit this issue even with a correctly configured codebase.

## Solution

Replace the one-shot timeout with an **idle timeout** that resets on every `moduleParsed` event. The timer now only fires when there has been **no module activity for the configured duration**, meaning the build is genuinely stuck — not just slow.

Changes in `pluginModuleParseEnd.ts`:
- `setParseTimeout` (called once at plugin init) → `resetIdleTimeout(timeout)` (called on every `moduleParsed`)
- Added `buildStart` hook to arm the initial idle timer when the build actually starts
- `resetIdleTimeout` is called for **every** module in `moduleParsed` (including excluded ones), since any activity means the build is still progressing
- Improved warning message to clarify the cause and suggest a remedy

## Test plan

- [ ] Build a large project where build time exceeds the default `moduleParseTimeout` of 10s — `remoteEntry` should now contain all remotes
- [ ] Simulate a genuinely hung build (e.g. a module that never resolves) — timeout should still fire after N seconds of inactivity
- [ ] Verify dev mode is unaffected (promise resolves immediately via `config` hook)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "dev-nv": "pnpm clean && pnpm -filter 'examples-nuxt-vite-host' -filter 'examples-vite-vite-remote' run dev",
     "preview-vv": "pnpm clean && pnpm -filter 'examples-vite-vite*' --parallel run preview",
     "multi-example": "pnpm clean && pnpm --filter \"multi-example-*\" --parallel run start",
-    "test": "vitest run src",
+    "test": "vitest run --dir src",
     "test:integration": "vitest run integration",
     "e2e": "playwright test",
     "changeset": "changeset",
@@ -58,6 +58,9 @@
   },
   "homepage": "https://github.com/module-federation/vite#readme",
   "packageManager": "pnpm@10.28.2",
+  "overrides": {
+    "koa": "3.1.2"
+  },
   "peerDependencies": {
     "vite": "^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0"
   },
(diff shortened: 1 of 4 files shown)


gioboa (commented):
can you update the readme.md and add a changeset please?

DemonStore:
I’ve put a quick fix for the error above and verified it - works as expected. Now even a small idleTimeout is enough to keep all remotes from dropping during long builds.

PR #448, opened 2026-03-06
fix: fix up path resolution in monorepo projects

### LINKED ISSUE #446: [Bug]: plugin with createServer incorrectly set rootDir, causing node_modules resolution to break
### Describe the bug

Using the module-federation/vite (federation) plugin in a setup where Vite is started via createServer() (middleware mode) results in incorrect URLs for shared/virtual modules in the transformed index.html.

In our case, the generated <script src="..."> paths for Module Federation virtual modules point to the server package node_modules directory instead of the client package node_modules, causing runtime 404s.

### Setup
We have a monorepo with two packages:

- `packages/client` (Vite + React app)
- `packages/server` (Express server that hosts Vite in middleware mode)

Client `vite.config.ts`:
```typescript

export default defineConfig({
  // config has "root" property value
  root: resolve(__dirname, './client'),
  plugins: [
    react(),
    // mf plugin
    federation({
      name: 'MyClient',
      remotes: {...},
      shared: {
        react: {
          singleton: true,
          requiredVersion: '^19.2.3',
        },
        'react-dom': {
          singleton: true,
          requiredVersion: '^19.2.3',
        },
        'react-dom/client': {
          singleton: true,
          requiredVersion: '^19.2.3',
        },
      },
    })
...
});
```

Server (`server.ts`) creates a Vite dev server:
```typescript
import { createServer } from 'vite';

if (config.isDevelopment) {
  vite = await createServer({
    configFile: <path-to-client-"vite.config.ts">,
    server: {
      middlewareMode: true,
    },
    appType: 'custom',
  });
}

// expressjs app
app.use(vite.middlewares);
```

We then transform `index.html` via:
```typescript
template = await vite.transformIndexHtml(url, template);
```

### Outcome

After `transformIndexHtml()`, the `federation` plugin generates a script tag for a virtual/shared module with an incorrect path.

Generated (incorrect):
```
/app/my-app/packages/**server**/node_modules/__mf__virtual/MyClient__H_A_I__hostAutoInit__H_A_I__.js
```

Expected:
```
/app/my-app/packages/**client**/node_modules/__mf__virtual/MyClient__H_A_I__hostAutoInit__H_A_I__.js
```

This leads to a runtime error because the browser attempts to fetch a non-existent file (404) from the wrong node_modules location.

### DEBUG notes

- The plugin configures the `VirtualModule` using `config.root` (see configure in the plugin:
https://github.com/module-federation/vite/blob/1a868defb2deeb1421c21a24491ca6eaf6ec76bc/src/index.ts#L103).
- However, `addEntry` is executed earlier (effectively at federation() call time), before the plugin has the final resolved Vite config root available:
https://github.com/module-federation/vite/blob/1a868defb2deeb1421c21a24491ca6eaf6ec76bc/src/index.ts#L139
- When the virtual module root is not available yet, module-federation/vite falls back to `process.cwd()` inside VirtualModule:
https://github.com/module-federation/vite/blob/1a868defb2deeb1421c21a24491ca6eaf6ec76bc/src/utils/VirtualModule.ts#L34
- In a `createServer()` middleware setup, `process.cwd()` is typically the server package directory, so the plugin resolves virtual module paths under `packages/server/node_modules` instead of `packages/client/node_modules`.

### Workaround (works for us)
Temporarily change the working directory to the client package before calling `createServer()`:

```typescript
import { createServer } from 'vite';

const createViteServer = async () => {
  const originalCwd = process.cwd();

  try {
    process.chdir(resolve(originalCwd, '../client'));

    return await createServer({
      configFile: <path-to-client-"vite.config.ts">,
      server: {
        middlewareMode: true,
      },
      appType: 'custom',
    });
  } finally {
    process.chdir(originalCwd);
  }
};

if (config.isDevelopment) {
  vite = await createViteServer();
   app.use(vite.middlewares);
}
```

With this workaround, the transformed HTML contains the expected path:

```html
<script src="/app/my-app/packages/client/node_modules/__mf__virtual/MyClient__H_A_I__hostAutoInit__H_A_I__.js"></script>
/>
``` 
and the request is served correctly by Vite’s middleware.

### Version

v1.12.0

### Reproduction

-

### Relevant log output

```shell

```

### Validations

gioboa:
Close #446

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,8 +14,8 @@ import pluginVarRemoteEntry from './plugins/pluginVarRemoteEntry';
 import aliasToArrayPlugin from './utils/aliasToArrayPlugin';
 import {
   ModuleFederationOptions,
-  normalizeModuleFederationOptions,
   NormalizedModuleFederationOptions,
+  normalizeModuleFederationOptions,
 } from './utils/normalizeModuleFederationOptions';
 import normalizeOptimizeDepsPlugin from './utils/normalizeOptimizeDeps';
 import VirtualModule, { initVirtualModuleInfrastructure } from './utils/VirtualModule';
@@ -23,22 +23,22 @@ import {
   getHostAutoInitImportId,
   getHostAutoInitPath,
   getLocalSharedImportMapPath,
+  getRemoteEntryId,
   initVirtualModules,
   LOAD_REMOTE_TAG,
   LOAD_SHARE_TAG,
-  getRemoteEntryId,
   writeLocalSharedImportMap,
 } from './virtualModules';
 import { getVirtualExposesId } from './virtualModules/virtualExposes';
-import {
-  writeLoadShareModule,
-  writePreBuildLibPath,
-  getLoadShareModulePath,
-  getPreBuildLibImportId,
-} from './virtualModules/virtualShared_preBuild';
 import { addUsedShares } from './virtualModules/virtualRemoteEntry';
 import { addUsedRemote } from './virtualModules/virtualRemotes';
 import { virtualRuntimeInitStatus } from './virtualModules/virtualRuntimeInitStatus';
+import {
+  getLoadShareModulePath,
+  getPreBuildLibImportId,
+  writeLoadShareModule,
+  writePreBuildLibPath,
+} from './virtualModules/virtualShared_preBuild';
 
 /**
  * Plugin that runs FIRST to create virtual module files in the config hook.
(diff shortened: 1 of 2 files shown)


gioboa (commented):
The pipeline is failing so I am checking

shevdrakon:
I can see that now browser has a request which fails with 404:
```
GET: http://localhost:1234/my-app/static/remoteEntry-[hash]
```
can't remember that `remoteEntry-[hash]` was requested before the changes.

PR #454, opened 2026-03-09
fix: init virtual modules during build config

yashnextgenit1-design:
**Summary**

- Fixes a first‑run CI failure where Vite build can resolve a virtual runtime module before it’s created.
- Ensures virtual module infrastructure and core virtual modules are initialized during build config, not only serve.
- Prevents ENOENT errors like __mf__virtual/...runtimeInit... on clean pipelines.

**Root Cause**

On a cold CI run, the build path could attempt to resolve a virtual runtime file before it was created. A retry succeeded because the file existed from the prior attempt.

**Fix**

Move early virtual module initialization to run for both build and serve in createEarlyVirtualModulesPlugin.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -62,16 +62,16 @@ importers:
         version: 0.21.0(typescript@5.5.3)
       vite:
         specifier: ^7.3.1
-        version: 7.3.1(@types/node@25.3.3)(jiti@2.6.1)(lightningcss@1.31.1)(sass-embedded@1.77.8)(terser@5.31.6)(yaml@2.6.1)
+        version: 7.3.1(@types/node@25.3.3)(jiti@2.6.1)(lightningcss@1.31.1)(sass-embedded@1.77.8)(terser@5.31.6)(yaml@2.8.2)
       vitest:
         specifier: ^4.0.18
-        version: 4.0.18(@types/node@25.3.3)(jiti@2.6.1)(lightningcss@1.31.1)(sass-embedded@1.77.8)(terser@5.31.6)(yaml@2.6.1)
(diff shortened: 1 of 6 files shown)


src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -14,14 +14,73 @@ import { ShareItem } from '../utils/normalizeModuleFederationOptions';
...
+
 function getPackageNamedExports(pkg: string): string[] {
   try {
     // Resolve from the project root (process.cwd()) so that shared packages
     // like react are found even when the plugin is installed in a nested
     // pnpm store location where peer dependencies are not hoisted.
     const projectRequire = createRequire(new URL('file://' + process.cwd() + '/package.json'));
     const mod = projectRequire(pkg);
-    return Object.keys(mod).filter((k) => k !== 'default' && k !== '__esModule');
+    return Object.keys(mod).filter(
```
gioboa, line 24: 🤔  I don't like hardcoded lists.
we moved `if (_command !== 'serve') return;` and it's not clear to me why do we need this filter now.

yashnextgenit1-design: This `delete` syntax error has actually *always* existed when building a shared component like `axios` that exports a JS reserved keyword. It just never surfaced before because the build was crashing with the `ENOENT` error *first* (since the virtual directory wasn't created). By moving the `if (_command !== 'serve') return;`, we fixed the `ENOENT` error, which finally allowed Rollup to start parsing the file and exposed the underlying `delete` keyword syntax error.

Thanks and I agree, hardcoded lists aren't great. We can actually drop the list entirely. Since JavaScript allows exporting any valid identifier as an alias (e.g., `export { _mf_0 as delete }`), we can just safely alias *all* named exports unconditionally. This fixes the issue without needing to maintain a list of reserved words. I have updated the PR to use this cleaner approach.

gioboa: Thanks for the great explanation, can we add a test for this?

PR #461, opened 2026-03-10
fix: preserve proxy import alias after Rollup deconflict  

Nerevar123:
After updating to the latest version I ran into this runtime error:

```
index-BrR8pw7K.js:493 Uncaught ReferenceError: require$$0 is not defined
```

Tracked it down to how non-inlineable bindings get reconstructed in the commonjs-proxy import. Rollup's deconfliction sometimes renames the import alias (e.g. `require$$0 → require$0`) but doesn't update references in the chunk's code body — those still use the original name, so the binding is undefined at runtime.

The fix checks whether `b.local `(the deconflicted alias) actually appears in the code body (excluding the import line). If it doesn't, the code body still references `proxyLocal` from the proxy's export map, so we use that as the alias instead. If `b.local` is used in the code body, we keep it - Rollup's rename was correct in that case (e.g. `commonjsGlobal → commonjsGlobal$1` to avoid a local variable clash).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ import pluginProxyRemotes from './plugins/pluginProxyRemotes';
 import { proxySharedModule } from './plugins/pluginProxySharedModule_preBuild';
 import pluginVarRemoteEntry from './plugins/pluginVarRemoteEntry';
 import aliasToArrayPlugin from './utils/aliasToArrayPlugin';
+import { resolveProxyAlias } from './utils/bundleHelpers';
 import {
   ModuleFederationOptions,
   NormalizedModuleFederationOptions,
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -394,7 +394,14 @@ function federation(mfUserOptions: ModuleFederationOptions): Plugin[] {
                 );
                 inlineable.push({ local: b.local, funcBody: renamedFunc });
               } else {
-                nonInlineable.push(b);
+                // If b.local isn't used in the code body, Rollup's deconflict mangled
+                // the import alias without updating references — use proxyLocal instead.
+                const codeWithoutImport = code.replace(fullImport, '');
+                const localUsedInCode = new RegExp(`\\b${b.local}\\b`).test(codeWithoutImport);
```
gioboa: `b.local` can contain regex-special chars (e.g. $ in require$$0). The new RegExp(\b${b.local}\b) will silently produce wrong matches. 
Needs escaping:                                                                                                      
```                                                                                                                              
  const escaped = b.local.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
  const localUsedInCode = new RegExp(`\\b${escaped}\\b`).test(codeWithoutImport);
```
Otherwise the fix looks good to me.

gioboa (commented):
Thanks for your amazing help @Nerevar123

PR #462, opened 2026-03-10
feat: trigger server.restart for watchShared paths

### LINKED ISSUE #455: [Bug]: HMR does not update a shared local dependency
### Describe the bug

Hello,

I have an issue where HMR does not work for a shared `file:` dependency.

I have a minimal reproduction with two Vite apps using `@module-federation/vite`:

- `apps/host-app`
- `apps/remote-app`

Both apps depend on two local dependencies via `file:` dependencies:

- `pkgs/shared-pkg`
- `pkgs/not-shared-pkg`

The difference is that `shared-pkg` is declared in the Module Federation `shared` config in both apps, while `not-shared-pkg` is not.

In dev mode, HMR behaves differently depending on which local dependency changes:

- Changing `pkgs/not-shared-pkg/index.js` works as expected and the UI updates through HMR.
- Changing `pkgs/shared-pkg/index.js` does not update the UI through HMR.

This makes it look like HMR is broken specifically for local dependencies that are shared through Module Federation.

### Reproduction

1. Clone https://github.com/clemvnt/vite-mf-shared-package-hmr-issue.
2. Install dependencies in both apps.
3. Start `apps/remote-app` with `npm run dev`.
4. Start `apps/host-app` with `npm run dev`.
5. Open both apps in the browser.
6. Edit `pkgs/not-shared-pkg/index.js` and change the exported value from `"A"` to `"B"`.
7. Observe that both apps update correctly.
8. Edit `pkgs/shared-pkg/index.js` and change the exported value from `"A"` to `"B"`.
9. Observe that the apps do not update correctly through HMR.

### Actual behavior

Changes in the non-shared local dependency are reflected immediately, but changes in the shared local dependency are not.

### Expected behavior

When a shared local dependency changes, both apps should receive the update in dev mode, the same way they do for a local dependency that is not shared.

### Note

I suspect the issue may be related to shared dependencies being included in `optimizeDeps`. 

### Validations

gioboa:
Close #455

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,9 +66,9 @@
     "vite": "^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0"
   },
   "dependencies": {
-    "@module-federation/dts-plugin": "^2.0.1",
-    "@module-federation/runtime": "^2.0.1",
-    "@module-federation/sdk": "^2.0.1",
+    "@module-federation/dts-plugin": "2.0.1",
+    "@module-federation/runtime": "2.0.1",
+    "@module-federation/sdk": "2.0.1",
     "@rollup/pluginutils": "^5.3.0",
     "defu": "^6.1.4",
     "estree-walker": "^3.0.3",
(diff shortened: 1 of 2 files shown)


clemvnt:
I tested it and left feedback on the original issue: https://github.com/module-federation/vite/issues/455#issuecomment-4037423718

PR #487, opened 2026-03-16
chore: Vue example for hmr testing

Staremang:


examples/vite-vite/vite-host/src/main.jsx
```diff
@@ -2,12 +2,17 @@ import React from 'react';
...
 const root = ReactDOM.createRoot(document.getElementById('app'));
 root.render(
   <React.StrictMode>
-    <h1>MF HOST Demo</h1>
+    <h1>React host component</h1>
     <App />
   </React.StrictMode>
 );
+
+const vapp = createApp(VueHost)
```
gioboa, line 17: We are mixing React and Vue in the Host and on Remote too.
We already have few examples [here](https://github.com/gioboa/module-federation-vite-examples/commits/main/) to test new features / fixes. 
You can implement the HMR feature and then we can test them there. Thanks 🙏 Sorry for the back and forth

PR #488, opened 2026-03-16
chore(deps): update sdk dep

### LINKED ISSUE #274: [Bug]: NodeJS vm code + 'eval' needlessly included in bundle (Causing CSP errors)

joshuaellis:
### Description

the `sdk` dependency now correctly tree-shakes node-esque code when a `target` is correctly set. 

### Related issues

* closes #274

### DIFF
--- a/integration/proxy-alias-deconflict.test.ts
+++ b/integration/proxy-alias-deconflict.test.ts
@@ -71,4 +71,32 @@ describe('resolveProxyAlias', () => {
 
     expect(result.imported).toBe('myExport');
   });
+
+  it('keeps rewritten locals unique across multiple proxy imports', () => {
+    const firstImport = `import{a as first$1}from"./proxy-a.js"`;
+    const secondImport = `import{b as second$1}from"./proxy-b.js"`;
+    const code = `${firstImport};${secondImport};console.log(app);`;
+    const claimedLocals = new Set(['first$1', 'second$1']);
+
+    claimedLocals.delete('first$1');
+    const first = resolveProxyAlias(
+      { imported: 'a', local: 'first$1' },
+      'o',
+      code,
+      firstImport,
+      claimedLocals
+    );
+    claimedLocals.add(first.local);
+    claimedLocals.delete('second$1');
+    const second = resolveProxyAlias(
+      { imported: 'b', local: 'second$1' },
+      'o',
+      code,
+      secondImport,
+      claimedLocals
+    );
+
+    expect(first.local).toBe('o');
+    expect(second.local).toBe('second$1');
+  });
 });
--- a/src/index.ts
+++ b/src/index.ts
@@ -410,6 +410,7 @@ function federation(mfUserOptions: ModuleFederationOptions): Plugin[] {
             // loadShare-dependent value.
             const inlineable: Array<{ local: string; funcBody: string }> = [];
             const nonInlineable: Array<{ imported: string; local: string }> = [];
+            const claimedLocals = new Set(bindings.map((binding) => binding.local));
 
             for (const b of bindings) {
               const proxyLocal = exportMap[b.imported];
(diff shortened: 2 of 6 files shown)


gioboa (commented):
## multi framework tests:

[E2E Tests](https://github.com/gioboa/module-federation-vite-examples/actions/workflows/e2e.yml)

gioboa (approved):
Looks great to me 👏 thanks @joshuaellis you rock!

PR #491, opened 2026-03-17
fix: solve errors with mixing dev/prod React apps

### LINKED ISSUE #476: [Bug]: React 19 - Mixing dev/prod React builds breaks
### Describe the bug

Bug described perfectly here: https://github.com/module-federation/core/discussions/4103

### Version

7.3.1

### Reproduction

https://github.com/module-federation/core/discussions/4103

Same as previously reported issue: https://github.com/module-federation/vite/issues/297

Key is to create a production build and consume that in a local dev server.

### Validations

gioboa:
Close #476


src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -55,7 +53,7 @@ export function generateLocalSharedImportMap() {
           ${
             shareItem?.shareConfig.import === false
               ? `throw new Error(\`Shared module '\${${JSON.stringify(pkg)}}' must be provided by host\`);`
-              : isVinext && pkg === 'react'
+              : pkg === 'react'
```
wojtask9: out of curiosity. Why different `paths` for `react` ?

wojtask9:
out of curiosity. Why different paths for react ?

src/index.ts
```diff
@@ -50,6 +50,27 @@ import {
...
+  '\f': '\\f',
+  '\n': '\\n',
+  '\r': '\\r',
+  '\t': '\\t',
+  '\0': '\\0',
+  '\u2028': '\\u2028',
+  '\u2029': '\\u2029',
+};
+
+function escapeUnsafeJsSourceChars(str: string): string {
```
devinrhode2, line 68: Just started to read code.. still piecing things together... react 19 src must have some of these chars?

devinrhode2: Seems like a code comment would go a long ways here

examples/vite-vite/vite-remote/package.json
```diff
@@ -5,7 +5,9 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "preview": "vite build && vite preview"
+    "preview": "vite build && vite preview",
+    "mixed:1": "vite build && vite preview",
+    "mixed:2": "MIXED_VV=2 NODE_ENV=production vite"
```
gioboa, line 10: We need to set production to serve the correct bundle.

src/index.ts
devinrhode2: Cleaning up old browser tabs... curious what this change was about

Issue #501, opened 2026-03-19
[Bug]: Vite 8 dev mode crash — shared React exports uninitialized when Rolldown pre-bundles react-dom


falldowngoboone:
### Describe the bug

When using `@module-federation/vite` with Vite 8 in dev mode, the app crashes immediately with:

```
TypeError: Cannot read properties of undefined (reading 'S')
  at react-dom/cjs/react-dom-client.development.js
```

This is `ReactSharedInternals.S` (`__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE`) being `undefined` — React 19's internals are not populated when `react-dom` tries to read them.

The issue occurs in the **host app only**, before any remote is loaded. Production builds work fine.

### Root cause

Rolldown's dependency optimizer pre-bundles `react-dom/client`. Inside the pre-bundled output, `react-dom`'s CJS code accesses React through a `__commonJSMin` wrapper:

```js
// line ~11676 of pre-bundled react-dom_client.js (inside __commonJSMin factory)
var React = (init_host__loadShare__react__loadShare__(), __toCommonJS(host__loadShare__react__loadShare___exports))
```

The `init_...()` call triggers `__esmMin`, which invokes the `__loadShare__react` module's async body. But the return value (a Promise) is **discarded by the comma operator** — it's called but never awaited in the CJS context. `__toCommonJS(exports)` then immediately accesses the exports object, whose getters return the still-`undefined` closure variables (`__mf_0`, `__mf_1`, etc.) because the async `initPromise.then(runtime => runtime.loadShare(...))` hasn't resolved yet.

The hazard is in `writeLoadShareModule()` — it generates `__loadShare__` virtual modules that defer ALL export initialization to after the async `loadShare` resolves:

```js
// Generated __loadShare__react module (simplified)
const exportModule = await initPromise
  .then(runtime => runtime.loadShare("react", {...}))
  .then(factory => typeof factory === "function" ? factory() : factory);
const { useState: __mf_40, ... } = exportModule;
export { __mf_40 as useState, ... };
```

When Rolldown pre-bundles this into `react-dom_client.js`, the CJS wrapper calls `init_...()` synchronously. The `await` inside the async function body can't block the synchronous caller, so all named exports remain `undefined`.

[...]

Issue #466, opened 2026-03-11
[Bug]: MISSING_EXPORT. using syntheticNamedExports is not working in Vite@8


KatSick:
### Describe the bug

here:

https://github.com/module-federation/vite/blob/88f32d213000a95096621452ad2bab6d76fe1854/src/index.ts#L251

we are using `syntheticNamedExports` which is not supported in Rolldown, and build is failing with https://github.com/rolldown/rolldown/issues/8628

however, it looks like, they do not plan to support this according to https://github.com/rolldown/rolldown/issues/819

I'm not sure what is the best way to contribute the fix

### Version

1.12.3

### Reproduction

run `vite build` on `vite@8.0.0-beta.18` using `shared` react and react-dom

### Relevant log output

```shell
> vite build

error during build:
Build failed with 996 errors:

[MISSING_EXPORT] Error: "Badge" is not exported by "node_modules/__mf__virtual/MFE__loadShare___mf_0_org_mf_3_org_mf_3_org_mf_1_org_mf_2_design_mf_2_system_mf_2_components__loadShare__.mjs".
   ╭─[ src/modules/a/components/button/button.tsx:5:10 ]
   │
 5 │ import { Badge } from "@org/design-system";
   │          ──┬──  
   │            ╰──── Missing export

...

...
    at aggregateBindingErrorsIntoJsError (file:///Users/ostapchervak/Projects/project.worktrees/f-vite8-final/node_modules/.pnpm/rolldown@1.0.0-rc.8/node_modules/rolldown/dist/shared/error-tpgdcGrK.mjs:48:18)
    at unwrapBindingResult (file:///Users/ostapchervak/Projects/project.worktrees/f-vite8-final/node_modules/.pnpm/rolldown@1.0.0-rc.8/node_modules/rolldown/dist/shared/error-tpgdcGrK.mjs:18:128)
    at #build (file:///Users/ostapchervak/Projects/project.worktrees/f-vite8-final/node_modules/.pnpm/rolldown@1.0.0-rc.8/node_modules/rolldown/dist/shared/rolldown-build-r9VyalQ4.mjs:3311:34)
[...]

PR #502, opened 2026-03-19
fix: initialize React exports with Rolldown

falldowngoboone:
## Problem

When using this plugin with Vite 8 in dev mode, the host app crashes with `TypeError: Cannot read properties of undefined (reading 'S')` on `ReactSharedInternals` — before any remotes load. Production builds are unaffected.

The root cause is a timing mismatch between Rolldown's CJS interop and the async `__loadShare__` virtual modules:

1. `writeLoadShareModule()` generates virtual modules where `exportModule` and named exports are only initialized **after** an async `loadShare()` resolves.
2. Vite 8's dep optimizer (Rolldown) pre-bundles `react-dom/client` and wraps the loadShare module with `__esmMin`/`__commonJSMin` helpers. The CJS wrapper pattern `(init_...(), __toCommonJS(exports))` calls `init_...()` but cannot await the resulting Promise — the synchronous comma expression discards it and immediately reads exports that are still `undefined`.
3. Additionally, Rolldown does not support `syntheticNamedExports`, so the existing Rollup-only `load` hook strategy doesn't work.

## Changes

**`src/virtualModules/virtualShared_preBuild.ts`** — New early-return branch in `writeLoadShareModule()` for Rolldown dev mode (`isRolldown && command !== 'build'`). Generates a `__loadShare__` virtual module that:
- Imports the prebuild module as a namespace (`import * as __mf_prebuild_ns__`) with a `.default ?? namespace` fallback to handle both CJS and ESM packages
- Synchronously initializes `exportModule` and destructures named exports **before** the async `loadShare` call, so they're available when Rolldown's `__esmMin` helper returns synchronously
- Still calls `loadShare` asynchronously to allow the runtime to provide a different shared instance, and re-destructures after it resolves

**`src/index.ts`** — The `load` hook now branches on Rolldown vs Rollup:
- **Rollup path** (unchanged): strips prebuild imports, uses `syntheticNamedExports: '__moduleExports'`
- **Rolldown path** (new): converts the bare prebuild side-effect import to a default import, rewrites `const exportModule` to `let` with synchronous pre-initialization from the prebuild, and returns plain `{ code }` without `syntheticNamedExports`

**`src/utils/packageUtils.ts`** — Extracted `getIsRolldown(ctx)` utility to centralize the `meta.rolldownVersion` check that was duplicated across 5 call sites in 3 files.

**`src/plugins/pluginProxyRemotes.ts`**, **`src/plugins/pluginProxySharedModule_preBuild.ts`** — Updated to use `getIsRolldown(this)` (separate commit for ease of review).

## Tests

8 new test cases in `virtualShared_preBuild.test.ts` covering the Rolldown dev-mode codegen:
- Namespace import with CJS/ESM fallback
- Synchronous pre-initialization of `exportModule`
- Named exports destructured before and after async `loadShare`
- `let` declarations for reassignability
- `export *` fallback for packages without named exports
- Build command is unaffected (still uses original path)

## Test plan

- [x] `pnpm test` — 102 tests pass (13 files, 0 failures)
- [x] `pnpm build` — CJS + ESM outputs build cleanly
- [x] Manual: `vite dev` with React + react-dom as shared singletons on Vite 8 — app loads without TypeError
- [x] Manual: `vite build` — no `MISSING_EXPORT` errors

## Related issues

- #501 
- #466 — `syntheticNamedExports` not supported in Rolldown
- rolldown/rolldown#3615 — `resolve.alias` doesn't call `resolveId` hooks


gioboa (changes requested):

Screenshot 2026-03-19 at 23 06 25

https://github.com/gioboa/module-federation-vite-examples/actions/runs/23315445654

---

Some tests are failing, can you check it please? That repo is testing simple apps with different frameworks.

gioboa (commented):
Closing this PR for now. The mentioned issue appears resolved in the current code, so this change is no longer needed in its present form. I also have a concern about the Rolldown load-hook changes, those look risky for named-export/default-export interop.

Plus one for `getIsRolldown` refactor, I'm going to create a separate PR for this change and I add you as author. 👏 

Thanks for your help @falldowngoboone

falldowngoboone:
Thanks @gioboa!

PR #536, opened 2026-03-24
chore: add shared-lib to vite-vite example

BernhardtDT:
## Summary
- Add a shared workspace library (`shared-lib`) to the vite-vite example
- Wire it up in both host and remote configs to demonstrate shared dependency handling across federated modules

## Test plan
- [x] `pnpm install` from root
- [x] Run remote and host dev servers
- [x] Verify `SharedCounter` component renders in both apps


gioboa (commented):
LGTM 💪 Would you mind improving e2e tests to verify the visibility of this shared component?

PR #554, opened 2026-03-26
chore: limit manualChunks warn

gioboa:


gioboa (commented):

Screenshot 2026-03-26 at 21 59 41

PR #563, opened 2026-03-30
fix: improve Astro React interop

Nsttt:
## Summary
- keep `ENV_TARGET` undefined for Astro mixed client/server builds unless the target is set explicitly
- avoid proxying React through `loadShare` for Astro the same way we already do for Vinext
- cover the Astro-specific shared/import behavior with regression tests

## Why
Astro hosts can execute mixed client and server federation paths in the same app. Forcing `ENV_TARGET` to `"node"` during SSR builds and routing React through the prebuild/shared-provider path caused the Astro integration to trip over React runtime resolution in those mixed flows.

## Testing
- `pnpm test`

### DIFF
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -30,8 +30,7 @@ export function proxySharedModule(options: {
   const { shared = {} } = options;
   let _config: ResolvedConfig | undefined;
   let _command = 'serve';
-  let isVinext = false;
-  let isAstro = false;
+  let useDirectReactImport = false;
   const savePrebuild = new PromiseStore<string>();
 
   return [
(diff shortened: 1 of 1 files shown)


src/plugins/pluginProxySharedModule_preBuild.ts
```diff
@@ -163,7 +166,7 @@ export function proxySharedModule(options: {
         const isRolldown = getIsRolldown(this);
         Object.keys(shared).forEach((key) => {
           if (key.endsWith('/')) return;
-          if (isVinext && key === 'react') {
+          if ((isVinext || isAstro) && key === 'react') {
```
gioboa: Can we use `useDirectReactImport` here ?

Nsttt: Switched this to `useDirectReactImport` in eedf1ec so the config hook and `configResolved` path share the same gate.

Nsttt: Added !

PR #568, opened 2026-03-30
fix: respect custom cacheDir in dev-await-shared-init transform (#566)

### LINKED ISSUE #566: [Bug]: `module-federation-dev-await-shared-init` skips transform when custom cacheDir is configured, causing shared module race condition
### Describe the bug

When using `@module-federation/vite` with a custom [`cacheDir`](https://vite.dev/config/shared-options#cachedir) in `vite.config.ts`, the `module-federation-dev-await-shared-init` plugin skips processing of pre-bundled dependency files. This causes a race condition where shared modules (e.g. `react`) are not awaited before bundle execution, resulting in runtime errors such as `React is undefined`.

#### Root cause

The transform in `module-federation-dev-await-shared-init` guards on a hardcoded path segment:

https://github.com/module-federation/vite/blob/main/src/index.ts#L638

```ts
if (!id.includes('.vite/deps/')) return;
```

When `cacheDir` is set to a non-default value (e.g. `node_modules/.vite/_myapp_static_`), Vite places pre-bundled dependencies under `.vite/_myapp_static_/deps/` instead of `.vite/deps/`. The hardcoded check never matches, so the critical top-level `await` for `loadShare` is never injected.

#### Expected behavior

The plugin should resolve the actual `cacheDir` from Vite's config (or use `server.config.cacheDir`) and check against that, rather than assuming the default `.vite/deps/` path.

### Reproduction

1. Create a Vite remote with `@module-federation/vite`, sharing `react` as `singleton: true`
2. Add a custom `cacheDir` to the Vite config:
   ```ts
   cacheDir: resolve(__dirname, 'node_modules', '.vite/_myapp_static_'),
   ```
3. Run the dev server (`createServer` or `vite dev`)
4. Load the remote from a Webpack host that also shares `react` as `singleton: true`

**Result:** Runtime errors — `React` is `undefined` due to `loadShare` not being awaited.  
**Removing** the custom `cacheDir` makes it work as expected.

### Config

```ts
const config: UserConfigFnObject = defineConfig(({ mode }) => {
  return {
    plugins: [
      federation({
        name: 'MyRemoteApp',
        manifest: true,
        filename: 'remoteEntry.js',
        exposes: {
          './MyRemote': resolve(__dirname, 'src/federated/MyRemote.tsx'),
        },
        shared: {
          react: { singleton: true },
          'react-dom/client': { singleton: true },
        },
        dts: false,
      }),
    ],
    cacheDir: resolve(__dirname, 'node_modules', '.vite/_myapp_static_'), // <-- triggers the bug
    base: 'myapp/static/',
  };
});
```

### Version

v8.0.3

### Reproduction

-

### Relevant log output

```shell

```

### Validations

Vitalcheffe:
## Problem

When using `@module-federation/vite` with a custom `cacheDir` in `vite.config.ts`, the `module-federation-dev-await-shared-init` plugin skips processing pre-bundled dependency files. This causes a race condition where shared modules (e.g. `react`) are not awaited before bundle execution, resulting in runtime errors like `React is undefined`.

## Root cause

The transform guards on a hardcoded path segment:

```ts
if (!id.includes('.vite/deps/')) return;
```

When `cacheDir` is set to a non-default value (e.g. `node_modules/.vite/_myapp_static_`), Vite places pre-bundled dependencies under a different path, so the check silently skips them.

## Fix

1. Added `depsDir` variable in the `federation()` closure, initialized to the default deps path
2. Updated `configResolved` to resolve the actual deps directory from `config.cacheDir`
3. Changed the transform guard from the hardcoded string to use `depsDir`

This is a minimal fix — 14 lines added, no behavior changes for default configurations.

Fixes #566

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1,4 +1,3 @@
-import { isAbsolute, join, resolve } from 'node:path';
 import type { Plugin } from 'vite';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { getLoadShareImportId } from '../virtualModules/virtualShared_preBuild';
@@ -7,16 +6,6 @@ const { hasPackageDependencyMock, mfWarn } = vi.hoisted(() => ({
   hasPackageDependencyMock: vi.fn(() => false),
   mfWarn: vi.fn(),
 }));
-
-function resolveDepsDir(cacheDir: string | undefined, root: string): string {
-  const normalizePath = (p: string): string => p.replace(/\\/g, '/');
-
-  if (cacheDir) {
-    const resolved = isAbsolute(cacheDir) ? cacheDir : resolve(root, cacheDir);
-    return normalizePath(join(resolved, 'deps')) + '/';
-  }
-  return normalizePath(join(root, 'node_modules', '.vite', 'deps')) + '/';
-}
 
 vi.mock('../utils/packageUtils', async () => {
   const actual =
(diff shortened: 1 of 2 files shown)


gioboa (changes requested):

src/__tests__/cachedir-deps-resolution.test.ts
```diff
@@ -0,0 +1,66 @@
...
+ * module-federation-dev-await-shared-init transform.
+ *
+ * The transform checks `id.includes(depsDir)` to decide whether to
+ * inject top-level awaits for shared module init functions. When a
+ * custom cacheDir is configured, the deps directory moves and the
+ * hardcoded '.vite/deps/' check would miss these files (#566).
+ */
+
+// Simulate the configResolved logic
+function resolveDepsDir(cacheDir: string | undefined, root: string): string {
```
gioboa: The test helper resolveDepsDir does not use normalizePath, but real code does.
please use it in the helper to keep the behaviour aligned.

src/__tests__/cachedir-deps-resolution.test.ts
```diff
@@ -0,0 +1,66 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
```
gioboa: vi and beforeEach are imported but never used.

src/__tests__/cachedir-deps-resolution.test.ts
Vitalcheffe: fixed — helper now uses normalizePath to match the real code behavior on all platforms

src/__tests__/cachedir-deps-resolution.test.ts
Vitalcheffe: removed, thanks

gioboa (approved):
Thanks @Vitalcheffe for your great help 👍

PR #578, opened 2026-03-31
fix: expand the example with the shared library and fixed

Staremang:
This is a continuation of the error https://github.com/module-federation/vite/pull/552/changes

Summary
- Fix getNamedExportsViaRegex to recursively follow export * from './...' re-exports, resolving deeply nested named exports that were previously missing from the generated loadShare module
- Add resolveRelativeModule helper that correctly handles both file extensions and directory index files, preventing existsSync from misidentifying directories as files
- Use getLocalProviderImportPath in remote entry import() generation to prefer the local provider path over the generic shared import source

Test
- Run the vite-vite example with a shared lib using nested export * re-exports (e.g. components/, helpers/search/)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ import { normalizePath, Plugin, UserConfig } from 'vite';
 import addEntry from './plugins/pluginAddEntry';
 import { checkAliasConflicts } from './plugins/pluginCheckAliasConflicts';
 import { PluginDevProxyModuleTopLevelAwait } from './plugins/pluginDevProxyModuleTopLevelAwait';
+import pluginDevRemoteHmr from './plugins/pluginDevRemoteHmr';
 import pluginDts from './plugins/pluginDts';
 import pluginManifest from './plugins/pluginMFManifest';
 import pluginModuleParseEnd from './plugins/pluginModuleParseEnd';
@@ -230,6 +231,7 @@ function federation(mfUserOptions: ModuleFederationOptions): Plugin[] {
     checkAliasConflicts({ shared }),
     normalizeOptimizeDepsPlugin,
     ...pluginDts(options),
+    pluginDevRemoteHmr(options),
     ...addEntry({
       entryName: 'remoteEntry',
       entryPath: remoteEntryId,
--- a/src/plugins/__tests__/pluginDevRemoteHmr.test.ts
+++ b/src/plugins/__tests__/pluginDevRemoteHmr.test.ts
@@ -1,0 +1,226 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import pluginDevRemoteHmr from '../pluginDevRemoteHmr';
+
+const { mfWarn } = vi.hoisted(() => ({
+  mfWarn: vi.fn(),
+}));
+
+vi.mock('../../utils/logger', () => ({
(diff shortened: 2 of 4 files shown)


src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -60,7 +60,7 @@ export function generateLocalSharedImportMap() {
               : useDirectReactImport && pkg === 'react'
                 ? `let pkg = await import("react");
             return pkg;`
-                : `let pkg = await import(${JSON.stringify(getSharedImportSource(pkg, shareItem))});
+                : `let pkg = await import(${JSON.stringify(getLocalProviderImportPath(pkg) || getSharedImportSource(pkg, shareItem))});
```
gioboa: if user explicitly sets shared.foo.import = "/some/custom/entry.js" but foo also resolves to a local workspace package, host will import workspace file instead of configured entry. Let's keep `getSharedImportSource` and then `getLocalProviderImportPath`

src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -186,19 +187,41 @@ function getEsmNamedExports(pkg: string): string[] {
...
 
     return regexNames;
   } catch {
-    return source ? getNamedExportsViaRegex(source) : [];
+    return source ? getNamedExportsViaRegex(source, entryPath) : [];
   }
 }
 
-function getNamedExportsViaRegex(source: string): string[] {
+function resolveRelativeModule(filePath: string, specifier: string): string | undefined {
```
gioboa, line 201: export * with explicit extension still missed

Package entry:
// index.js
`export * from './runtime.js';`

src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -60,7 +60,7 @@ export function generateLocalSharedImportMap() {
               : useDirectReactImport && pkg === 'react'
                 ? `let pkg = await import("react");
             return pkg;`
-                : `let pkg = await import(${JSON.stringify(getSharedImportSource(pkg, shareItem))});
+                : `let pkg = await import(${JSON.stringify(getSharedImportSource(pkg, shareItem) || getLocalProviderImportPath(pkg))});
```
gioboa: 🤔 getSharedImportSource() is always truthy because returns getConcreteSharedImportSource(...) || getPreBuildLibImportId(...). So `|| getLocalProviderImportPath(pkg)` never runs. Am I wrong?

src/virtualModules/virtualRemoteEntry.ts
Staremang: You're right - the original order lets workspace detection override explicit user config. But as you noted in the follow-up comment, simply swapping to `getSharedImportSource() || getLocalProviderImportPath()` doesn't work either, since `getSharedImportSource()` is always truthy.

How about a three-step priority chain instead?                                                                                                                                 
`getConcreteSharedImportSource(pkg, shareItem) || getLocalProviderImportPath(pkg) || getSharedImportSource(pkg, shareItem)` Explicit import config wins first, then workspace path, then prebuild fallback.

gioboa: Yeah it sounds great to me

PR #595, opened 2026-04-02
fix: skip all federation plugins in test env

JammingBen:
When Vitest runs, it sets command="serve" and mode="test". The plugin only branched on command, never checking mode. In serve mode, the hostAutoInit transform generates code using window.origin, which is undefined in test environments like happy-dom or jsdom, causing:

```
Cannot find package 'undefined/remoteEntry.mjs'
```

Since the plugin has no test-mode awareness and MF virtual modules rely on browser globals and runtime initialization unavailable in test environments, the simplest fix is to disable all federation plugins when being in a test environment. Without the plugins, no MF imports are injected and standard Vite module resolution takes over. Users can mock federated remote imports in their tests if needed.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -119,9 +119,9 @@ describe('federation in test environment', () => {
     expect(plugins).toEqual([]);
   });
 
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Image

Issue #606, opened 2026-04-09
[Bug]: Cannot share lit properly


simonihmig:
### Describe the bug

Since upgrading from a much earlier version (`1.8.0`) to more recent ones (`1.12.3`, `1.13.5`, also reproducible on latest `1.14.2`) along side with core MF deps, we are having problems sharing `lit` dependencies properly. 

The problem arises when using [directives](https://lit.dev/docs/templates/directives/). For the better or worse, they are provided by `lit-html` but lit exposes them as re-exports through the main `lit` package, like [here](https://github.com/lit/lit/blob/main/packages/lit/src/directives/class-map.ts). (plus some more "interesting" [shenanigans](https://github.com/lit/lit/blob/main/packages/lit/src/index.ts#L7-L11)). This might play a reason why the build is bundling them instead of sharing, as shown in the following?

I saw that you have a [lit example](https://github.com/gioboa/module-federation-vite-examples/tree/main/lit) already in place for like e2e validation, that's why I built on top of that for my [reproduction](https://github.com/gioboa/module-federation-vite-examples/pull/2). In the [first commit](https://github.com/gioboa/module-federation-vite-examples/pull/2/changes/a8487e2aad3ba2a9760a1d7fac81bdbfb2fc2140), I added some (rather useless) usage of the `classMap` directive, to reproduce the module graph problem as explained above. While `pnpm dev` or `pnpm build && pnpm preview` kinda works (the app renders fine), you can still see in the `dist` output that the code for `classMap` is not shared but bundled AFAICT, as it is in the `/dist/assets/remote-app-Bcr1W7BH.js` chunk.

[...]

PR #607, opened 2026-04-09
fix: support shared package subpaths for Lit in dev

Nsttt:
## What changed

- match bare shared package keys against package subpaths during proxying, so sharing `lit` also covers `lit/*`
- register shared subpaths as their own runtime entries in `localSharedImportMap`
- make serve-mode `__loadShare__` wrappers await the resolved shared module before destructuring named exports
- document subpath share behavior and add regression coverage around the Lit repro

## Why

The Lit repro in module-federation/vite#606 was still failing in dev with `property is not a function` because subpath imports like `lit/decorators.js` and `lit/directives/class-map.js` were bypassing federation or being exported from a pending Promise. That left host and remote with mismatched Lit module instances and broken named exports on cold start.

## Impact

- shared bare packages now cover their package subpaths in dev/build proxying
- Lit host/remote repro renders on exact `pnpm lit:dev` cold start
- named exports from shared ESM packages survive serve-mode wrappers

## Root cause

Two bugs stacked:

1. package subpaths were not consistently proxied/registered under the runtime share map
2. serve-mode shared wrappers destructured `exportModule` before awaiting the async share resolution, so bindings like `LitElement` and `property` could become `undefined`

## Validation

- `pnpm test`
- `pnpm fmt.check`
- exact repro smoke in `module-federation-vite-examples` with `pnpm lit:dev`

### DIFF
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -384,7 +384,7 @@ export const LOAD_SHARE_TAG = '__loadShare__';
 const loadShareCacheMap: Record<string, VirtualModule> = {};
 export function getLoadShareImportId(pkg: string, isRolldown: boolean, command?: string): string {
   if (!loadShareCacheMap[pkg]) {
-    const useESM = isRolldown || command === 'build';
+    const useESM = command === 'serve' || command === 'build' || isRolldown;
     const ext = useESM ? '.mjs' : '.js';
     loadShareCacheMap[pkg] = new VirtualModule(pkg, LOAD_SHARE_TAG, ext);
   }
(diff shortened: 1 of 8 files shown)


src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -402,12 +402,12 @@ export function writeLoadShareModule(
   isRolldown: boolean
 ) {
   if (!loadShareCacheMap[pkg]) {
-    const useESM = isRolldown || command === 'build';
+    const useESM = command === 'serve' || command === 'build' || isRolldown;
```
simonihmig, line 405: Can `command` ever take a different value other than either serve or build? Otherwise this would be a tautology, not?

gioboa: I think you are right, let's refactor the code in a separate PR to keep the changes minimal.

gioboa (commented):
Closed by #608

gioboa:
Anyway, thanks @Nsttt for your help 🙏

PR #608, opened 2026-04-09
fix: support lit shared subpaths

### LINKED ISSUE #606: [Bug]: Cannot share lit properly

gioboa:
Close #606 #607

### DIFF
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -1,7 +1,16 @@
+import { existsSync, readFileSync } from 'fs';
+import path from 'pathe';
 import { Plugin, ResolvedConfig, UserConfig } from 'vite';
+import { mfWarn } from '../utils/logger';
 import { mapCodeToCodeWithSourcemap } from '../utils/mapCodeToCodeWithSourcemap';
 import { NormalizedShared, ShareItem } from '../utils/normalizeModuleFederationOptions';
-import { getIsRolldown, hasPackageDependency, setPackageDetectionCwd } from '../utils/packageUtils';
+import {
+  getIsRolldown,
+  getPackageDetectionCwd,
+  hasPackageDependency,
+  removePathFromNpmPackage,
+  setPackageDetectionCwd,
+} from '../utils/packageUtils';
 import { PromiseStore } from '../utils/PromiseStore';
 import VirtualModule, { assertModuleFound } from '../utils/VirtualModule';
 import {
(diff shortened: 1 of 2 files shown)


src/plugins/pluginProxySharedModule_preBuild.ts
```diff
@@ -22,6 +31,56 @@ function getPrebuildResolutionSource(pkgName: string, shareItem?: ShareItem): st
...
+      } catch {
+        // skip
+      }
+    }
+  }
+  return [];
+}
+
+/**
+ * In dev mode, detects shared packages that are sub-dependencies of other
```
gioboa, line 58: @simonihmig after few different implementations and debugging sessions this solution looks the best one to avoid regressions and apply an automatic fix.

gioboa (commented):

Screenshot 2026-04-10 at 14 13 07

PR #617, opened 2026-04-15
fix: avoid runtime name clashes with internalName

### LINKED ISSUE #612: [Bug]: Module Federation Vite: remote name "scheduler" collides with React's scheduler package
### Describe the bug

I asked an AI assistant to help pinpoint this, and the issue seems to be a name collision rather than a missing expose.

When a remote is named `scheduler`, it conflicts with React’s real `scheduler` package (used by `react-dom/client`). In dev, Module Federation/Vite appears to treat the bare import `scheduler` from React internals as if it were the federated remote, and generates a remote load for:

`loadRemote("scheduler")`

At runtime that gets resolved as expose `"."`, which then fails with:

`[Module Federation] Module . does not exist in container.`

Our remote only exposes `./App`, so the error makes sense once the collision happens.

Why this looks like a bug:
- `import("scheduler/App")` is valid and should resolve to the remote expose
- but unrelated package imports for the real npm package `scheduler` are also being captured
- renaming the remote to something else like `schedulerRemote` or `xxx` makes everything work immediately

So the root issue seems to be that remote name matching is too broad and collides with legitimate package imports from dependencies.

Environment:
- `@module-federation/vite`: `1.14.2`
- `vite`: `8.0.8`
- `react` / `react-dom`: `19.2.5`

Expected:
- `scheduler/App` should resolve to the remote
- bare imports of the real `scheduler` package should stay untouched

Actual:
- the remote named `scheduler` hijacks/import-collides with React’s `scheduler` package and causes `Module . does not exist in container`

Current workaround:
- rename the remote from `scheduler` to a non-conflicting name like `schedulerRemote`

### Version

8.0.8

### Reproduction

https://github.com/metinsut/micro-fe-bug-report

### Relevant log output

```shell
Uncaught Error: [Module Federation] Module . does not exist in container.
    at Module.getExposes (virtual:mf-REMOTE_ENTRY_ID:scheduler__remoteEntry_js:81:44)
    at async Module$1.get (@module-federation_runtime.js?v=b2a02094:1879:39)
    at async RemoteHandler.loadRemote (@module-federation_runtime.js?v=b2a02094:2789:28)
    at async host__loadRemote__scheduler__loadRemote__.mjs:25:26
```

### Validations

gioboa:
Close #612

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -2,7 +2,14 @@ import type { Plugin } from 'vite';
 import { afterAll, beforeEach, describe, expect, it, vi } from 'vitest';
 import { getLoadShareImportId } from '../virtualModules/virtualShared_preBuild';
 
-const { hasPackageDependencyMock, mfWarn } = vi.hoisted(() => ({
+const {
+  getInstalledPackageEntryMock,
+  getInstalledPackageJsonMock,
+  hasPackageDependencyMock,
+  mfWarn,
+} = vi.hoisted(() => ({
+  getInstalledPackageEntryMock: vi.fn(() => undefined),
+  getInstalledPackageJsonMock: vi.fn(() => undefined),
   hasPackageDependencyMock: vi.fn(() => false),
   mfWarn: vi.fn(),
 }));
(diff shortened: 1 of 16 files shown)


gioboa (commented):

Screenshot 2026-04-15 at 12 40 35

gioboa (commented):

Image

PR #618, opened 2026-04-15
fix: handle rolldown outputs more safely

### LINKED ISSUE #616: [Bug]: Silent deadlock when shared dependencies are bundled into the entry chunk via transitive code splitting
### Describe the bug
 
When code splitting causes a shared dependency (e.g., `react`) to be bundled into the entry chunk (`main.js`) — either via `manualChunks` or through transitive imports from other chunked modules — the application silently hangs at load time in production builds.
 
### Root Cause
 
Module Federation transforms shared dependency imports with `top-level await`. The expected execution order is:
 
1. Load synchronous, small `main.js` (entry)
2. Load async `remoteEntry.js` → registers shared modules → creates initial promises
3. Load the real application chunk(s)
4. Shared module `await` resolves because `remoteEntry.js` already ran and the initial promise exists
 
When code splitting pulls a shared dependency like `react` into the entry chunk, the order breaks:
 
1. Load `main.js` — now contains `react` (or a chunk that depends on it)
2. `main.js` hits `await` for the `react` shared module
3. `remoteEntry.js` has **not loaded yet** → no initial promise for `react` exists
4. **Deadlock** — the `await` never resolves, the app silently freezes
 
### Why this is distinct from documented warnings
 
The README warns that `manualChunks` and `codeSplitting: false` are unsupported and auto-ignored. However, this deadlock can also occur **without** explicitly setting either option — for example, when a non-shared chunked package (e.g., an internal library) transitively imports `react`, causing the bundler to hoist `react` into the entry chunk to satisfy the dependency graph.
 
### Additional context
 
- **In `dev` mode this does not reproduce** due to subtle execution timing differences in the dev server, making it especially hard to catch.

```typescript
// vite.config.ts

const config: UserConfigFnObject = defineConfig(({ mode }) => {
  return {
    plugins: [
      reactVite(),
      federation({
        name: 'MyApp',
        hostInitInjectLocation: 'entry',
        manifest: true,
        dts: false,
        filename: 'remoteEntry.js',
        remotes: {
        ...
        },
        exposes: {
          './Wizard': resolve(__dirname, 'src/federated/MyRemote.tsx'),
        },
     shared: {
        'react': { singletone: true },
        'react-dom': { singletone: true }
        'react-dom/client': { singletone: true }
     }
  }),
    ],
    root: resolve(__dirname),
    publicDir: resolve(__dirname, '../../public'),
...
    build: {
      target: 'esnext',
      minify: false,
      chunkSizeWarningLimit: 1000,
      emptyOutDir: true,
      runtimeChunk: 'single',
      rolldownOptions: {
        treeshake: true,
        input: {
          main: resolve(__dirname, 'index.html'),
        },
        output: {
          entryFileNames: '[name].[hash].js',
          chunkFileNames: '[name].[hash].bundle.js',
          assetFileNames: 'assets/[name].[hash].[ext]',
          dir: RELATIVE_CLIENT_DIST_DIR,
          codeSplitting: {
            groups: [
              // !!! --> exclude react from code splitting to avoid bubbling them into main chunk,
              // !!! --> what leads to racing deadlock
              // {
              //   test: /node_modules\/(react|react-dom|react-router)(\/|$)/,
              //   name: 'react',
              // },
            ],
          },
        },
      },
    },
  }
})

export default config
```
 
### Expected Behavior
 
At minimum: a **build-time warning** when a shared dependency ends up in the entry chunk, since this will always produce a deadlock at runtime.
 
Ideally: 
- the plugin should enforce that shared dependencies are **never** co-located with the entry chunk, regardless of how the bundler's chunk splitting resolves the dependency graph.
OR
- initial promise resolved regardless of location and modules order
 
### Steps to Reproduce
 
1. Configure a Module Federation host with `react` in `shared`
2. Add a chunked dependency (e.g., via dynamic import or separate chunk group) that transitively imports `react`
3. Build for production
4. Observe that `react` gets bundled into the entry chunk
5. Load the app → silent hang, no errors in console

### Version

1.14.3

### Reproduction

https://github.com/shevdrakon/vite/tree/repro/issue-616

### Relevant log output

```shell

```

### Validations

gioboa:
Close #616

### DIFF
--- a/examples/vite-vite/vite-remote/vite.config.js
+++ b/examples/vite-vite/vite-remote/vite.config.js
@@ -74,7 +74,7 @@ export default defineConfig({
   ].filter(Boolean),
   build: {
     target: 'chrome89',
-    rollupOptions: {
+    rolldownOptions: {
       output: {
         chunkFileNames: 'static/js/[name]-[hash].js',
         entryFileNames: 'static/js/[name]-[hash].js',
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -153,6 +153,32 @@ describe('module-federation-esm-shims', () => {
     expect(mfWarn).toHaveBeenCalledTimes(1);
   });
 
+  it('removes codeSplitting groups and warns once', () => {
+    const plugin = getEsmShimsPlugin();
+    const config: any = {
+      build: {
+        rolldownOptions: {
+          output: {
+            codeSplitting: {
+              groups: [
+                {
+                  test: /node_modules\/(react|react-dom)(\/|$)/,
+                  name: 'react',
+                },
+              ],
+            },
+          },
+        },
+      },
+    };
+
+    const configHook = typeof plugin.config === 'function' ? plugin.config : plugin.config?.handler;
+    configHook?.call({} as any, config, { command: 'build', mode: 'test' });
+
+    expect(config.build.rolldownOptions.output.codeSplitting).toBeUndefined();
+    expect(mfWarn).toHaveBeenCalledTimes(1);
+  });
+
   it('ignores user manualChunks and warns, keeps federation chunks isolated', () => {
     const plugin = getEsmShimsPlugin();
     const runtimeInitId = virtualRuntimeInitStatus.getImportId();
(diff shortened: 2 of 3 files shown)


gioboa (commented):

Image

PR #620, opened 2026-04-15
fix: fix automatic sharing for export subpaths

### LINKED ISSUE #615: [Bug]: Transitive dependency not shared (lit > lit-html)
### Describe the bug

Since upgrading from a much earlier version (`1.8.0`) to more recent ones (`1.12.3`, `1.13.5`, also reproducible on latest `1.14.2`) along side with core MF deps, we are having problems sharing `lit` dependencies properly. 

The problem arises when using [directives](https://lit.dev/docs/templates/directives/). For the better or worse, they are provided by `lit-html` but lit exposes them as re-exports through the main `lit` package, like [here](https://github.com/lit/lit/blob/main/packages/lit/src/directives/class-map.ts). (plus some more "interesting" [shenanigans](https://github.com/lit/lit/blob/main/packages/lit/src/index.ts#L7-L11)). This might play a reason why the build is bundling them instead of sharing, as shown in the following?

The existing [lit example](https://github.com/gioboa/module-federation-vite-examples/tree/main/lit) serves as the reproduction. While running `pnpm dev` in the `lit` folder (i.e. run both in dev) works in terms of the app being able to run, the `classMap` directive gets bunded into the remote and host, instead of being shared between the two.

You can see this looking at the dist output, or what get's loaded in the browser. Running both in prod (`pnpm build && pnpm preview`), this is what I get when searching for where `classMap` can be found:

Image

You can see that `classMap` is bundled into both the remote and host, not deduped.

Another "fun" way (actually took me hours of debugging) to illustrate the sharing problem is to run either the host with `pnpm dev` and the remote with `preview`, or vice versa. Lit has the interesting fact that it ships separate development and production builds in its [package exports](https://github.com/lit/lit/blob/main/packages/lit/package.json#L20-L23), and the two are not interoperable. In prod it does some [explicit deterministic terser-based mangling](https://github.com/lit/lit/blob/main/rollup-common.js#L91-L260) of method names. If both `lit` and `lit-html` are using the prod output, they still match and everything works. But you cannot mix them. If Lit is in dev mode and lit-html is prod, you will see a `TypeError: currentDirective._$initialize is not a function`, the other way around it'll be `TypeError: i._$AT is not a function`. This is reproducible on that branch, again with mixed `dev` vs. `preview` commands. Not the root cause of anything, just highlighting when sharing is not working. If both `lit` and `lit-html` would be equally shared, they both would be either in dev or prod mode, so no problem.

This is what I get with host served in dev and remote in prod:

Image

The stack trace is showing the point when lit (prod) tries to setup the `classMap` directive (dev). AFAICT, this shouldn't happen if all lit parts were properly shared, as then the mode would be consistent.

### Version

7.3.1

### Reproduction

https://github.com/gioboa/module-federation-vite-examples

### Relevant log output

```shell

```

### Validations

gioboa:
Close #615 

Sharing lit is now sufficient to share the sub-exports as well, as the plugin automatically handles the heavy lifting of sharing them.

This is fixing mixed lit mode too:
- host dev and remote preview
- host preview and remote dev

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -89,6 +89,24 @@ function getEarlyInitPlugin(): Plugin {
   return plugin;
 }
 
+function getEarlyInitPluginWithLitShare(): Plugin {
+  const plugin = federation({
+    name: 'host',
+    filename: 'remoteEntry.js',
+    shared: {
+      lit: {
+        singleton: true,
+      },
+      'lit/directives/class-map.js': {
+        singleton: true,
+      },
+    },
+  }).find((entry) => entry.name === 'vite:module-federation-early-init');
+
+  if (!plugin) throw new Error('vite:module-federation-early-init plugin not found');
+  return plugin;
+}
+
 function getModuleFederationVitePlugin(): Plugin {
   const plugin = federation({
     name: 'host',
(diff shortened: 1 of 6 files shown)


gioboa (commented):

Screenshot 2026-04-16 at 10 06 10

PR #622, opened 2026-04-16
fix: restrict remote named exports plugin to rolldown only

### LINKED ISSUE #619: [Bug]: @module-federation/vite 1.14.3 — Named Imports Broken in Dev Mode
### Describe the bug

> Disclosure: this report has been authored with the help of Claude Code. I've verified the bug myself however, let me know if I can help clarify anything.

When running later versions of @module-federation/vite, (after 1.9.2), dev mode is broken and will not work on Vite versions lower than 8.

Both `pnpm --filter remote build` and `pnpm --filter host build` succeed without errors, confirming this is a dev-only bug.

I've attached a repo with the steps to reproduce. I've verified that updating to Vite 8 does resolve the issue, however there's also a quick patch if you're mid-launch like I am and need something in a pinch:

```ts
{
    name: 'mf-strip-remote-named-exports-dev',
    apply: 'serve' as const,
    configResolved(config) {
      const idx = config.plugins.findIndex(
        (p) => p.name === 'module-federation-remote-named-exports'
      );
      if (idx !== -1) {
        (config.plugins as import('vite').Plugin[]).splice(idx, 1);
      }
    },
  }
```

### Version

6.4.2

### Reproduction

https://github.com/rp4rk/mf-vite-dev-bug-repro

### Relevant log output

```shell
`TypeError: Cannot destructure property 'useCounter' of '__mf_ns_0' as it is undefined`
```

### Validations

gioboa:
Close #619

### DIFF
--- a/src/plugins/__tests__/pluginRemoteNamedExports.test.ts
+++ b/src/plugins/__tests__/pluginRemoteNamedExports.test.ts
@@ -1,8 +1,12 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { parseAst } from 'rollup/parseAst';
 
+const { getIsRolldownMock } = vi.hoisted(() => ({
+  getIsRolldownMock: vi.fn(() => true),
+}));
+
 vi.mock('../../utils/packageUtils', () => ({
-  getIsRolldown: () => true,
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #629, opened 2026-04-17
fix: add backtick support to preload helper regex

### LINKED ISSUE #628: [Bug]: Using Vite 8 with @module-federation/vite breaks cross-origin remote chunk loading
### Describe the bug

This was AI assited.

On Vite 8, Rolldown emits the preload helper’s URL resolver with template literals (e.g. return `````` `/my-app/static/`+e) ``````. The module-federation-fix-preload plugin only matches single/double-quoted base strings, so the rewrite to new URL(e, import.meta.url).href never runs. Chunks stay origin-relative and load from the host origin instead of the remote when the remote is served from another domain. Manifest publicPath: "auto" is unrelated; the bug is in the preload-helper transform. Minimal repro + proposed regex change (include backticks) available.

### Version

v8.0.8

### Reproduction

https://github.com/RalphK66/vite8-mf-bug

### Relevant log output

```shell

```

### Validations

RalphK66:
Fixes #628

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -735,6 +735,36 @@ describe('module-federation-fix-preload', () => {
 
     expect(bundle['preload-helper-abc.js'].code).toContain('new URL(e,import.meta.url).href');
   });
+
+  it('handles backticks in function expression pattern', () => {
+    const plugin = getFixPreloadPlugin();
+    const bundle = {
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-04-18 at 13 13 25

gioboa (approved):
Thanks for your great help @RalphK66 🙏

Issue #368, opened 2025-12-10
[Bug]: CSS assets not included in mf-manifest.json in version >= 1.9.0


jruizu:
### Describe the bug

**Expected behavior:**
CSS files should be listed in the `css` property of both `shared` modules and `exposes` entries in the `mf-manifest.json`, similar to how JavaScript assets are included.

**Actual behavior:**
The `css` arrays in the manifest are empty (`"css":{"async":[],"sync":[]}`), even though CSS files are generated during the build process.

**Affected versions:**
- `@module-federation/vite`: `1.9.3` (and possibly later versions)
- `vite`: `^7.1.9`

**Working version:**
- `@module-federation/vite`: `1.8.1`

1. Create a Vite project with Module Federation
2. Install `@module-federation/vite` version `1.9.3`
3. Configure the plugin with `manifest: true`
4. Import CSS files in your entry point (e.g., `App.tsx`)
5. Build the project: `vite build`
6. Check the generated `dist/mf-manifest.json`

### Configuration Example

```typescript
// vite.config.ts
import { federation } from '@module-federation/vite';

export default defineConfig({
  plugins: [
    federation({
      name: 'myApp',
      filename: 'remoteEntry.js',
      manifest: true,
      exposes: {
        './App': './src/App.tsx',
      },
      shared: {
        react: { singleton: true },
        'react-dom': { singleton: true },
      },
    }),
  ],
  build: {
    cssCodeSplit: false,
    rollupOptions: {
      output: {
        assetFileNames: assetInfo => {
          if (assetInfo.name?.endsWith('.css')) {
            return 'assets/[name]-[hash][extname]';
          }
          return 'assets/[name]-[hash][extname]';
        },
      },
    },
  },
});
```

### Workaround

Downgrade to version `1.8.1`:

```bash
npm install --save-dev @module-federation/vite@1.8.1
# or
pnpm add -D @module-federation/vite@1.8.1
```

### Additional Information

- CSS files are correctly generated in `dist/assets/`
- The issue only affects the manifest generation
- JavaScript assets are correctly included in the manifest
- The same configuration works correctly with version `1.8.1`

### Version

>1.9.0

### Reproduction

https://github.com/geek-corp/module-federation-vite-css-bug-repro

### Relevant log output

```shell

```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
[...]

PR #637, opened 2026-04-21
fix: remove customResolvers

### LINKED ISSUE #634: remove warn: resolve.alias customResolver usage will break in Vite 9
### Describe the bug

Building with Vite 8.0.9 emits the following deprecation warning:

```
`resolve.alias` contains an alias with `customResolver` option. This is deprecated and will be removed in Vite 9. Please use a custom plugin with a resolveId hook and `enforce: 'pre'` instead.
```

The warning is triggered by `@module-federation/vite` itself — the plugin pushes alias entries with a `customResolver` function at three locations in `lib/index.mjs` (in the published 1.14.5):

- `proxyRemotes` plugin, line 2653 (around the `Object.keys(remotes).forEach` block)
- `proxySharedModule` / `proxyPreBuildShared` plugin, line 2772 (the shared-package alias loop)
- `proxyPreBuildShared` plugin, line 2796 (the `async customResolver` for PREBUILD_TAG)

Since `resolve.alias.customResolver` will be removed in Vite 9, users of `@module-federation/vite` will not be able to upgrade once Vite 9 ships unless this is migrated to an `enforce: 'pre'` plugin with a `resolveId` hook.

### Version

@module-federation/vite 1.14.5, Vite 8.0.9

### Reproduction

Any project using `@module-federation/vite` with a shared module configuration (e.g. `react`) will emit the warning during `vite build`.

### Relevant log output

```shell
`resolve.alias` contains an alias with `customResolver` option. This is deprecated and will be removed in Vite 9. Please use a custom plugin with a resolveId hook and `enforce: 'pre'` instead.
```

### Validations

gioboa:
Close #634

### DIFF
--- a/src/plugins/__tests__/pluginProxyRemotes.test.ts
+++ b/src/plugins/__tests__/pluginProxyRemotes.test.ts
@@ -35,6 +35,10 @@ describe('pluginProxyRemotes', () => {
   });
 
   function getSchedulerAlias() {
+    return getSchedulerPluginAndConfig().config.resolve.alias[0];
+  }
+
+  function getSchedulerPluginAndConfig(configOverrides = {}) {
     const plugin = pluginProxyRemotes({
       remotes: {
         scheduler: {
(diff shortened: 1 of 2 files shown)


gioboa (commented):
I'm adding this change in the huge #368 refactoring.

PR #651, opened 2026-04-26
fix: Proxy invariants and SSR bootstrap wrapper for MF remotes

spralle:
## Summary

Two independent bug fixes discovered while working on cross-federation HMR:

### 1. Proxy invariant violation in virtual remote modules

`__mfCreateRemoteProxy` uses a plain empty object `{}` as the Proxy target but copies non-configurable properties (like `default` from `Module` namespace objects) onto it. When the Proxy's `ownKeys` trap returns keys that exist as non-configurable on the target, `getOwnPropertyDescriptor` must return a descriptor consistent with the target - otherwise the engine throws a `TypeError`.

**Fix:** Added `getOwnPropertyDescriptor` trap that checks the target first, and ensured `ownKeys` always includes non-configurable target keys per the Proxy invariant spec.

**Repro:** Any federated remote that re-exports a `default` export triggers this in strict ES module environments.

### 2. SSR bootstrap wrapper incorrectly applied to MF remotes

`pluginAddEntry` wraps entry points with an SSR detection bootstrap (`typeof document === 'undefined'` guard). This wrapper is correct for application entry points but breaks Module Federation remotes - the remote entry runs in a browser context where `document` exists, but the wrapper's conditional logic can interfere with the module's initialization.

**Fix:** Added `_forceClientInjected` option to `addEntry()`. When the federation plugin creates a remote entry, it sets this flag to skip the SSR bootstrap wrapper.

### Testing

- All 270 existing tests pass
- Build succeeds

### DIFF
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -28,11 +28,11 @@ function getFirstHtmlEntryFile(entryFiles: string[]): string | undefined {
 }
 
 const addEntry = ({
-	entryName,
-	entryPath,
-	fileName,
-	inject = 'entry',
-	_forceClientInjected,
+  entryName,
+  entryPath,
+  fileName,
+  inject = 'entry',
+  _forceClientInjected,
 }: AddEntryOptions): Plugin[] => {
   const DEV_HTML_PROXY_PREFIX = 'virtual:mf-html-entry-proxy?';
   const ENTRY_BOOTSTRAP_QUERY = '?mf-entry-bootstrap';
@@ -44,7 +44,7 @@ const addEntry = ({
   let _command: string;
   let emitFileId: string;
   let viteConfig: any;
-	let clientInjected = _forceClientInjected ?? false;
+  let clientInjected = _forceClientInjected ?? false;
   let emittedFileName: string | undefined;
 
   function skipSvelteKitSsrBuild() {
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -19,6 +19,8 @@ interface AddEntryOptions {
   entryPath: string | (() => string);
   fileName?: string;
   inject?: NormalizedModuleFederationOptions['hostInitInjectLocation'];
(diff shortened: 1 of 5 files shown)


src/virtualModules/virtualRemotes.ts
```diff
@@ -108,10 +108,18 @@ export default exportModule?.__esModule ? exportModule.default : exportModule.de
         },
         ownKeys() {
           const mod = getModule();
-          if (!mod) return [];
-          return Reflect.ownKeys(mod);
+          const keys = new Set(mod ? Reflect.ownKeys(mod) : []);
```
gioboa, line 111: Can you add few tests to cover these changes please?

gioboa (changes requested):

src/virtualModules/virtualRemotes.ts
spralle: Added tests - 3 for the Proxy invariants (ownKeys includes non-configurable target keys, getOwnPropertyDescriptor returns target descriptor, get trap delegates to remote module) and 1 for _forceClientInjected (SSR fallback bootstrap is skipped). All 274 tests pass.

src/plugins/pluginAddEntry.ts
```diff
@@ -19,6 +19,8 @@ interface AddEntryOptions {
   entryPath: string | (() => string);
   fileName?: string;
   inject?: NormalizedModuleFederationOptions['hostInitInjectLocation'];
+  /** When true, skip the SSR fallback bootstrap wrapper (used for MF remotes whose HTML is never browser-requested). */
+  _forceClientInjected?: boolean;
```
gioboa: Is there a reason for this _ as prefix?

spralle: No, it's just the styleguide I'm used to, I'll fix it.

spralle: Done, renamed to forceClientInjected in the latest push.

gioboa (approved):
That's awesome 👏 @spralle

Issue #592, opened 2026-04-02
feat: support real HMR rather than live reload


gioboa:
### Describe the bug

Related to #587
PRs are more than welcome.

## HMR

- [ ] Host
- [ ] Remote

PR #653, opened 2026-04-26
fix: shared React Fast Refresh proxy for cross-federation HMR

spralle:
## Summary

Partially addresses #592 — enables React Fast Refresh across Module Federation boundaries by sharing a single `@react-refresh` runtime between host and remote dev servers.

For the full `native` HMR experience (suppressing the full-reload relay so Fast Refresh can work uninterrupted), see the companion PR #654.

### Problem

When a React MF remote runs on a separate dev server (e.g. `localhost:3002`), `@vitejs/plugin-react` injects a `/@react-refresh` preamble that creates an isolated React Fast Refresh registry on the remote's origin. Components registered in the remote's registry are invisible to the host page's `injectIntoGlobalHook` patch, so React Fast Refresh silently does nothing for cross-federation modules.

The existing `pluginDevRemoteHmr` handles cross-federation HMR via a broadcast-and-reload mechanism, but this results in full page reloads rather than fine-grained component updates with state preservation.

### Solution

New plugin `pluginSharedReactRefresh` adds a `configureServer` middleware on MF remote dev servers that intercepts `/@react-refresh` requests and serves a proxy module:

`js
const __rt = await import(window.location.origin + '/@react-refresh');
export const injectIntoGlobalHook = __rt.injectIntoGlobalHook;
export const register = __rt.register;
// ... re-exports all Fast Refresh APIs
`

Since the browser page is always served by the host dev server, `window.location.origin` resolves to the host's `/@react-refresh` - the same module instance loaded by the preamble. This unifies the component registry so `performReactRefresh()` finds and patches all components regardless of which dev server served them.

### Design decisions

- **Middleware over `resolveId`**: `@vitejs/plugin-react`'s `vite:react-refresh` sub-plugin uses `enforce: 'pre'` and typically wins the `resolveId` race. The middleware intercepts the HTTP request before Vite's transform pipeline, making it independent of plugin ordering.
- **`window.location.origin` over config**: No need to pass the host URL through config - the browser page always knows its own origin.
- **Additive, not replacing**: The existing broadcast/reload mechanism in `pluginDevRemoteHmr` continues to work as a fallback. This plugin provides a better experience specifically for React users.
- **Activation conditions**: Only activates when the federation config has `exposes` (is a remote) AND `dev.remoteHmr` is enabled.

### Scope

- New file: `src/plugins/pluginSharedReactRefresh.ts` (58 lines)
- Registration in `src/index.ts` (2 lines)
- New test file: `src/plugins/__tests__/pluginSharedReactRefresh.test.ts` (142 lines, 7 tests)
- No changes to existing plugins or behavior

### Testing

- 7 new tests covering: plugin metadata, activation conditions, middleware interception, query string handling, CORS headers, pass-through behavior
- All 277 tests pass (270 existing + 7 new)
- Build succeeds


src/plugins/pluginSharedReactRefresh.ts
```diff
@@ -0,0 +1,58 @@
...
+): Plugin {
+  const isRemote = Object.keys(options.exposes).length > 0;
+  const hmrEnabled =
+    typeof options.dev === 'object' && options.dev !== null && options.dev.remoteHmr === true;
+
+  return {
+    name: 'module-federation-shared-react-refresh',
+    apply: 'serve',
+
+    configureServer(server) {
```
gioboa: Thanks for your help @spralle 
Is it possible to move this logic in the pluginDevRemoteHMR?

spralle: Done - moved the middleware into pluginDevRemoteHmr's configureServer hook, in the if (isRemote) block right before the existing __mf_hmr metadata endpoint. Deleted the standalone plugin file. Tests moved to pluginDevRemoteHmr.test.ts.

gioboa (approved):
That's awesome 👏 @spralle 
thank you

PR #654, opened 2026-04-26
feat: add remoteHmrStrategy option to control cross-federation HMR

### LINKED ISSUE #592: feat: support real HMR rather than live reload

spralle:
Close #592

## Summary

Addresses #592 — auto-detects React and suppresses the broadcast/relay mechanism so Vite's native HMR (React Fast Refresh) handles cross-federation updates instead of full page reloads.

### Problem

When `dev.remoteHmr` is enabled, `pluginDevRemoteHmr` unconditionally relays remote file changes as `{ type: 'full-reload' }` to the host browser. With React Fast Refresh (and the shared `/@react-refresh` proxy from #653), the component updates fine via native HMR — then the relay immediately nukes it with a full page reload.

### Solution

Auto-detect React from the Vite plugin pipeline. When `vite:react-refresh` or `vite:react-swc:refresh` is present, suppress the broadcast (remote side) and relay (host side) so React Fast Refresh handles it.

Only React is supported for now — it's the only framework with cross-federation HMR plumbing (the shared `/@react-refresh` proxy from #653). Vue, Svelte, Solid have native HMR but no equivalent shared-runtime proxy yet, so they stay on full-reload. Easy to extend once the plumbing exists.

An explicit `dev.remoteHmrStrategy` escape hatch is available:
- `'full-reload'`: Force broadcast/relay even with React
- `'native'`: Force native HMR even without a detected framework
- `undefined` (default): Auto-detect

### Backwards compatibility

- No React plugin → unchanged (full-reload)
- React plugin detected → suppressed (native HMR)
- Explicit config overrides auto-detection

### Testing

- 3 tests: React suppresses broadcast, no-React broadcasts, explicit override
- All 279 tests pass
- Format check passes


src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -467,6 +467,18 @@ interface PluginDevOptions {
...
+   *
+   * - `'full-reload'` (default): The host relays remote change events as full
+   *   page reloads. Works with any framework but loses application state.
+   * - `'native'`: Suppresses the full-reload relay and the remote-side broadcast,
+   *   relying on Vite's native HMR pipeline instead. Use this with frameworks
+   *   whose Vite plugins handle HMR natively (e.g. React with
+   *   `pluginSharedReactRefresh`, Vue, Svelte). Provides fine-grained component
+   *   updates with state preservation.
+   */
+  remoteHmrStrategy?: 'full-reload' | 'native';
```
gioboa: Thanks for the detailed description @spralle 
I think we can be opinionated about this, and infer an automatic behavior based on the detected framework. The idea is to have HMR for the host and for the loaded remote as well. This means that in the host I can see the remote's HMR as well.

spralle: Good call - updated to auto-detect the framework from the Vite plugin pipeline. It now checks for vite:react-refresh, vite:react-swc:refresh, vite:vue, vite-plugin-svelte, and solid in server.config.plugins. When detected, broadcast/relay is suppressed automatically. No config needed - it just works. The explicit remoteHmrStrategy option is kept as an escape hatch for edge cases. 5 tests covering auto-detection, fallback, and override scenarios.

spralle: Scaled it back a bit as only React is auto-detected now.

gioboa: I see, I am highly conservative in adding new settings. Let me test it  👀

spralle: Just let me know what you think is the best direction

gioboa: I prefer to change the type of `remoteHmr` and make it: true/false or 'full-reload'. in the property name we are mentioning Hmr so true is like 'native' option.

spralle:
Good idea, have changed it accordingly.

src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -466,7 +466,19 @@ interface PluginDevOptions {
...
+   * - `false` / `undefined` — HMR disabled (default).
+   * - `true` — HMR enabled with auto-detected strategy: when a React plugin
+   *   is detected (`@vitejs/plugin-react` or `@vitejs/plugin-react-swc`),
+   *   broadcast/relay is suppressed and React Fast Refresh handles updates
+   *   via the shared `/@react-refresh` proxy. Other frameworks fall back to
+   *   full page reloads.
+   * - `'full-reload'` — HMR enabled, always use full page reloads even when
+   *   a framework with native cross-federation HMR is detected.
+   */
+  remoteHmr?: boolean | 'full-reload';
```
gioboa, line 500: 👏 Great, I like it.

gioboa (commented):
Image

I tested this PR with [this repo](https://github.com/gioboa/module-federation-vite-examples/tree/main/react), HMR is not working though. Can you double check it please?

spralle:
showcase_hmr

I seem to have not posted my response. From what I know (not an expert) react fast refresh doesn't like anonymous default exports, it can't track them then you get a state reset like what you see. Also your repo didn't target the right PR? (654?)

gioboa:
I'm testing different versions on that repo. I run my test with the correct one.

gioboa (commented):
Nice call btw. I am wondering if we can patch that plug-in somehow 🤔

gioboa (commented):

Image

gioboa (approved):
Thanks that's great 👏 @spralle

PR #655, opened 2026-04-27
fix(core): resolve version even when import is false

pcfreak30:
- Attempt version resolution even when import: false
- Only suppress mfError on missing version when import: false
- Add test for version resolution with import: false
- Add test for graceful failure when version not found


pcfreak30:
@gioboa

gioboa (approved):
Thanks @pcfreak30 thank you for your help 👏

PR #657, opened 2026-04-27
fix: fix up package resolution

gioboa:


gioboa (commented):

Screenshot 2026-04-27 at 23 16 38

PR #658, opened 2026-04-27
chore: replace deprecated init helper

### LINKED ISSUE #656: chore: MF Init Deprecated
### Describe the bug

It seems that https://github.com/module-federation/core/blob/2dc50ffdba6414e0b7fdcfa453c4bd6ece260807/packages/runtime/src/index.ts#L36 the `init` helper is now deprecated and the plugin should use the new `instance` API.

### Version

v8

### Reproduction

n/a

### Relevant log output

```shell

```

### Validations

gioboa:
Close #656


gioboa (commented):

Screenshot 2026-04-28 at 00 12 21

PR #661, opened 2026-04-28
fix: preserve import-false shared sub-dependencies in dev

### LINKED ISSUE #660: [Bug]: `excludeSharedSubDependencies` incorrectly removes shared packages
### Describe the bug

https://github.com/module-federation/vite/pull/608 has introduced a new method `excludeSharedSubDependencies` that removes duplicated shared dependencies. This breaks setups when a shared package uses a dependency internally but does not re-export or provide it.

The function assumes that if package A depends on package B, sharing A is sufficient and B can be removed. This seems to be correct for packages like `lit` (re-exporting e.g. `lit-html`, `lit-element` etc.), but not for packages that simply consume a dependency internally.

### Details

Given this federation config:

```
 shared: {
   'pinia': { singleton: true, import: false },
   'shared-lib': { singleton: true, import: false },
 }
```

Where `shared-lib` has `pinia` as regular dependency (it uses it, but does not re-export it).

In dev mode, `excludeSharedSubDependencies` removes `pinia` from the shared map. The virtual proxy module for it is never created, causing: `Failed to resolve import "__mf__virtual/__mfe_internal__<remote>__prebuild__pinia__prebuild__.js" from ".__mf__temp/<remote>/localSharedImportMap.js". Does the file exist?`.

Removing this package from the shared config is not an option either. I have a host app that brings all together: it initializes `pinia`, registers it (and `shared-lib`) as shared modules and loads the remote apps. All remote apps as well as `shared-lib` need to use the host's `pinia` instance.

### Suggestion

I'm unsure about the correct way to address this. Possible fixes coming to my mind:

 1. Warn only, don't auto-exclude - let the user decide whether to remove sub-dependencies.
 2. Add an opt-out - e.g., a per-package config flag to prevent exclusion.
 3. Only exclude when `import: false` is not set on the sub-dependency (if the remote explicitly says "don't bundle this, get it from the host", it should be respected).
 4. Declare dependencies such as `pinia` in the above example as peer dependency of `shared-lib`. This would obviously be a solution on my side, not in this library. While I'm not strictly against that, I still feel like this is a situation this library should be able to handle.

I'd be happy get your opinion on this (and to possibly submit a PR with the desired solution).

### Version

1.14.5

### Reproduction

https://github.com/JammingBen/mf-vite-sub-dep-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #660

Preserves `import false` shared sub-dependencies in dev mode without warning.
This respects host-provided shared modules while keeping existing auto-exclusion behavior for other duplicate sub-dependencies.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -724,6 +724,66 @@ describe('pluginProxySharedModule_preBuild', () => {
     readFileSyncMock.mockReset().mockReturnValue('{}');
   });
 
+  it('preserves import false shared sub-dependencies in dev mode without warning', () => {
+    hasPackageDependencyMock.mockReturnValue(false);
+
+    existsSyncMock.mockImplementation(
+      (p: string) => p === '/repo/apps/remote/node_modules/shared-lib/package.json'
+    );
+    readFileSyncMock.mockImplementation((p: string) => {
+      if (p === '/repo/apps/remote/node_modules/shared-lib/package.json') {
+        return JSON.stringify({
+          name: 'shared-lib',
+          dependencies: { pinia: '^3.0.0' },
+        });
+      }
+      return '{}';
+    });
+
+    const shared: NormalizedShared = {
+      'shared-lib': {
+        name: 'shared-lib',
+        from: '',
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-04-28 at 17 44 56

PR #666, opened 2026-04-29
fix: respect shared subpath imports

### LINKED ISSUE #663: [Bug]: Cannot share subpaths
### Describe the bug

After upgrading to 1.14.5, I was seeing that imports with subpaths like `my-pkg/some/path` would not get shared but bundled, even when `my-pkg/` was in `shared`. 

Asking claude code for analysis, this is what I got, not sure though how accurate it is:

```
v1.8.0 — no trailing-slash guard:
  customResolver(source, importer) {
      if (/\.css$/.test(source)) return;
      var loadSharePath = getLoadShareModulePath(source);
      writeLoadShareModule(source, shared[key], command);
      writePreBuildLibPath(source);
      addUsedShares(source);
      writeLocalSharedImportMap();
      return this.resolve(loadSharePath, importer);
  }

  v1.14.5 — added the guard at line 2799:
  customResolver(source, importer) {
      if (/\.css$/.test(source)) return;
      if (useDirectReactImport && source === "react") return;
      if (importer && importer.includes("localSharedImportMap")) return;
      if (key.endsWith("/") && source !== key.slice(0, -1)) return;  // <-- NEW
      // ...
  }

  Corrected Root Cause

  The issue is line 2799 in @module-federation/vite v1.14.5 — a new guard that was added since v1.8.0:

  if (key.endsWith("/") && source !== key.slice(0, -1)) return;

  In v1.8.0, when a trailing-slash shared key like `@my-org/my-pkg/` matched a subpath import like `@my-org/my-pkg/button`, the customResolver would proceed normally —
  intercepting the import and making it shared.

  In v1.14.5, this new guard short-circuits the resolver for any subpath import when the key ends in /. It only allows the bare package name through (e.g., `@my-org/my-pkg`), which
  defeats the entire purpose of the trailing-slash pattern.

  For lit/, this doesn't matter because getLitExportSubpathShares creates dedicated non-trailing-slash entries for each subpath (lit/decorators.js, lit/html.js, etc.), which have their own
  alias entries that bypass this guard. But for every other package, the trailing-slash sharing mechanism is broken.

  This is a regression in @module-federation/vite v1.14.5 that should be reported upstream.
```

I also tried again based on https://github.com/module-federation/vite/pull/657 (using `https://pkg.pr.new/@module-federation/vite@657`), which seems to change a bunch of stuff related to package resolution. Then getting an error like:

```
error during build:
[vite]: Rollup failed to resolve import "/path/to/node_modules/@my-org/my-pkg/index.js" from "virtual:mf-REMOTE_ENTRY_ID:__mfe_internal__xxx__remoteEntry__hash__js".
```
This is what claude came up with:

```
Here's the full chain that causes the build failure:

  1. Shared config has '@my-org/my-pkg/': { singleton: true } (trailing slash)
  2. getOrderedUsedShares() (line 1294-1296) strips the trailing / and adds @my-org/my-pkg to the used shares set — even if no actual import of this package was ever intercepted
  3. generateDirectSharedCacheSeedCode (line 1309) generates await import(importPath) for every used share in the remote entry. For build mode, importPath comes from
  getDirectSharedCacheSeedImportPath
  4. getDirectSharedCacheSeedImportPath calls getProjectResolvedImportPath("@my-org/my-pkg"), which calls getPackageEsmEntryPath, which calls getInstalledPackageEntry
  5. getInstalledPackageEntry (line 273) tries to find the ESM entry point. Since @my-org/my-pkghas no "." export in its exports field (only wildcard patterns like ./*), no module
   field, and no main field, it falls back to "index.js" — a hardcoded default
  6. This produces the absolute path /path/to/node_modules/@my-org/my-pkg/index.js, which is embedded in the remote entry code
  7. Rollup tries to resolve this absolute path, finds the file doesn't exist, and throws the build error
```

To give a bit more context: the package we are trying to share does not have a main index barrel file, only subpaths, exposed via a wildcard package.json export. So the first problem seems to be why it tries to even import from `@my-org/my-pkg` (without subpath). But even if it had a index module, it would be at `node_modules/@my-org/my-pkg/dist/index.js`, not `node_modules/@my-org/my-pkg/index.js`.

### Version

1.14.5

### Reproduction

not available yet, I can try to work on one later this week, wanted to get this report out before a another release is cut. But would love to hear first what state of the plugin this should be based on (last release or main branch), and if there is potentially any work in flight already that could affect this. 

### Relevant log output

```shell

```

### Validations

gioboa:
Close #663

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -418,6 +418,48 @@ describe('pluginProxySharedModule_preBuild', () => {
     expect(preBuildShareItemMap.has('react-dom')).toBe(false);
   });
 
+  it('proxies subpath imports for trailing slash shared packages', async () => {
+    hasPackageDependencyMock.mockReturnValue(false);
+
+    const shared = makeShared();
+    shared['transitive/'] = {
+      ...shared.transitive,
+      name: 'transitive/',
+    };
+    delete shared.transitive;
+
+    const plugins = proxySharedModule({ shared });
+    const proxyPlugin = getProxyPlugin(plugins);
+    const sharedResolvePlugin = getSharedResolvePlugin(plugins);
+    const config: MockUserConfig = {
+      resolve: { alias: [] },
+    };
+
+    callHook(
+      proxyPlugin.config,
+      {
+        meta: createPluginMeta(),
+        resolve: async (id: string) => ({ id: `/resolved/${id}` }),
+      } as unknown as ConfigPluginContext,
+      config,
+      { command: 'serve', mode: 'development' } as ConfigEnv
+    );
+
+    const resolution = await callHook(
+      sharedResolvePlugin.resolveId,
+      {
+        resolve: async (id: string) => ({ id }),
+      } as any,
+      'transitive/button',
+      '/src/main.ts',
+      { isEntry: false }
+    );
+
+    expect((resolution as { id: string }).id).toBeDefined();
+    expect(preBuildShareItemMap.has('transitive/button')).toBe(true);
+    expect(preBuildShareItemMap.has('transitive')).toBe(false);
+  });
+
   it('resolves prebuild aliases to configured share import sources in build mode', async () => {
     hasPackageDependencyMock.mockReturnValue(false);
 
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -237,9 +237,6 @@ export function proxySharedModule(options: {
         }
         if (importer && importer.includes(LOAD_SHARE_TAG)) return;
         if (importer && importer.includes(PREBUILD_TAG)) return;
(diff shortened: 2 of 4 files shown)


gioboa (commented):

Image

simonihmig:
I tried again with this branch, and the previous error is gone, however I now get a bunch of errors like 

```
Could not resolve "/path/to/node_modules/__mf__virtual__development/__mfe_internal__shell__loadShare__lit_mf_1_directives_mf_1_when_mf_3_js__loadShare__.mjs"
```

The files do exist though, at least when I look for them after the build failed, so probably a timing issue?

@gioboa do you think this could be related to the changes here?

gioboa:
If you see them, we should definitely solve them, no matter if it's related to this PR. Can you please create a small reproduction repo?

simonihmig:
I debugged this, and it seems to be an interoperability issue with another plugin adding esbuild plugins. I think I can fix this at the "other" side...

gioboa:
I see, if you need support or changes, I am here 🫡

PR #667, opened 2026-04-29
fix: fix up shared deps Vite optimization interop

### LINKED ISSUE #664: [Bug]: 1.15.0 regression — shared packages bypass `optimizeDeps`, breaking default-style imports of CJS transitive deps in dev
### Describe the bug

After upgrading from `1.14.5` → `1.15.0`, the dev server fails to render as soon as a shared package's source is requested. Default-style imports of CJS-only transitive deps (e.g. `import Emitter from 'tiny-emitter'` inside `@okta/okta-auth-js`) are now served raw from `node_modules`, so the browser hits a `SyntaxError: ... does not provide an export named 'default'`.

The exact same config worked on `1.14.5`. The shared package's transitives used to flow through Vite's dep optimizer (which gives esbuild's CJS→ESM interop). On `1.15.0` they don't.

## Impact

This is a **dev-server-only** regression. Production builds are unaffected:

| Surface | Affected? | Why |
| --- | --- | --- |
| `vite` (dev) | **Yes** — page won't render | MF plugin's resolver intercepts shared packages, bypassing the dep optimizer; raw CJS is served to the browser |
| `vite build` (Rolldown) | No | Rolldown does CJS→ESM interop at bundle-time independent of MF plugin resolution |
| `vite preview` | No | Serves the production bundle |

So this is a DX-only bug, not a deploy blocker — but it makes any local dev of a host that shares packages with CJS transitives unusable until a workaround is applied.

## What the dev server actually serves

Probing the running dev server confirms the resolution path:

```
$ curl -sS http://localhost:5173/node_modules/tiny-emitter/index.js
function E () { ... }
E.prototype = { ... };
module.exports = E;          ← raw CJS, no ESM wrapper
```

```
$ curl -sS http://localhost:5173/node_modules/@okta/okta-auth-js/esm/browser/base/factory.js
import * as __mfLocalShare from "/node_modules/@okta/okta-auth-js/esm/browser/exports/exports/default.js?v=...";
const __mfCacheGlobalKey = "__mf_module_cache__";
...
import Emitter from 'tiny-emitter';   ← default import, no interop available
```

So the MF plugin **does** intercept `@okta/okta-auth-js` (its source is served wrapped with `__mfLocalShare` / `__mfCacheGlobalKey`) but **does not** forward the parent's transitive `tiny-emitter` import to the dep optimizer. On `1.14.5` `tiny-emitter` was discovered by the optimizer normally and served as `node_modules/.vite/deps/tiny-emitter.js` with esbuild's `__commonJSMin` interop wrapping; on `1.15.0` it's served raw from `node_modules/`.

## Bisect

```bash
npm install @module-federation/vite@1.14.5
npm run dev
```

Page renders, no error. Switching back to `1.15.0` re-breaks it.

The same is true regardless of `vite` version (`8.0.8` and `8.0.10` both exhibit the regression on plugin `1.15.0`, and neither does on `1.14.5`).

## Workaround

Force every CJS-only default-imported transitive into the optimizer via `optimizeDeps.include`:

```js
optimizeDeps: {
  include: [
    'tiny-emitter',
    'p-cancelable',
    'cross-fetch',
    '@babel/runtime/regenerator',
    '@babel/runtime/helpers/asyncToGenerator',
    '@babel/runtime/helpers/defineProperty',
    '@babel/runtime/helpers/objectWithoutProperties',
    '@babel/runtime/helpers/slicedToArray',
  ],
}
```

After applying this, `node_modules/.vite/deps/` contains pre-bundled wrappers (esbuild's `__commonJSMin`) for each entry and the page renders. But the list has to be maintained by hand — a new shared package with CJS transitives needs each one enumerated. `1.14.5` did not need any of this for the same input.

## Why this looks like a regression in the resolveId hook

`1.15.0` replaced the deprecated `customResolver` alias entries (#634) with a `resolveId` hook. The pre-`1.15.0` `customResolver` only intercepted the shared module itself, leaving Vite's optimizer to discover and pre-bundle transitives normally. The new hook appears to intercept transitive imports too, routing them to raw `node_modules/...` paths and skipping the optimizer.

Possible directions:

- only intercept the shared package boundary (match `1.14.5` semantics for transitives), or
- explicitly forward transitive imports from shared packages to `optimizeDeps`, or
- mark shared packages with `optimizeDeps.needsInterop` so esbuild's CJS→ESM wrapper is applied even when the resolution is plugin-driven.

## Inline reproduction

(Form field 3 requires a repo URL, but I cannot publish to public GitHub from my work machine. The minimal reproduction is four small files.

```bash
mkdir mf-repro && cd mf-repro
# create the four files below, then:
npm install
npm run dev
# open the URL the dev server prints
# expected: a page with "OktaAuth instance constructed: yes"
# actual: blank page; DevTools console shows the SyntaxError above
```

### `package.json`

```json
{
  "name": "mf-vite-1150-repro",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "vite build"
  },
  "dependencies": {
    "@okta/okta-auth-js": "^7.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
  },
  "devDependencies": {
    "@module-federation/vite": "1.15.0",
    "@vitejs/plugin-react": "^6.0.1",
    "vite": "^8.0.10"
  }
}
```

### `vite.config.mjs`

```js

export default defineConfig({
  plugins: [
    react(),
    federation({
      name: 'host',
      filename: 'remoteEntry.js',
      remotes: {},
      shared: ['react', 'react-dom', '@okta/okta-auth-js'],
      dts: false,
    }),
  ],
  server: { port: 5173 },
});
```

### `index.html`

```html
<!doctype html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>mf-vite 1.15.0 repro</title>
  </head>
  <body>
    <div id="root"></div>
    <script type="module" src="/src/main.jsx"></script>
  </body>
</html>
```

### `src/main.jsx`

```jsx
import React from 'react';
import { createRoot } from 'react-dom/client';
import { OktaAuth } from '@okta/okta-auth-js';

const auth = new OktaAuth({
  issuer: 'https://example.okta.com/oauth2/default',
  clientId: 'irrelevant',
  redirectUri: 'http://localhost:5173/login/callback',
});

createRoot(document.getElementById('root')).render(
  <div>
    <h1>@module-federation/vite 1.15.0 repro</h1>
    <p>OktaAuth instance constructed: {auth ? 'yes' : 'no'}</p>
  </div>
);
```

### Version

vite v8.0.10 (also reproduced on v8.0.8). @module-federation/vite: 1.15.0 (broken) vs 1.14.5 (works).

### Reproduction

Inline in the description above

### Relevant log output

```shell
factory.js?v=...:16 Uncaught (in promise) SyntaxError: The requested module '/node_modules/tiny-emitter/index.js?v=...' does not provide an export named 'default' (at factory.js?v=...:16:8)


(Other CJS leaves throw the equivalent error pointing at `p-cancelable`, `cross-fetch`, `@babel/runtime/regenerator`, etc., depending on which shared-package source path executes first.)
```

### Validations

gioboa:
Close #664

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -325,12 +325,12 @@ export default __mfShared.default ?? __mfShared;`,
             const optimizeDeps = (config.optimizeDeps ??= {});
             optimizeDeps.include ??= [];
             optimizeDeps.exclude ??= [];
-            // Vite 8/Rolldown must keep shared packages outside dependency
-            // optimization, otherwise optimized third-party deps can bypass
-            // the shared loadShare proxy. Vite < 8 still uses esbuild and can
-            // fail when an optimizer entry is also external.
+            // Some packages must stay outside dependency optimization because
+            // their submodules rely on parent initialization order. Other
+            // shared deps should remain optimizable so Vite can apply CJS
+            // interop for transitive dependencies used by prebuild fallbacks.
             const shouldBypassOptimizeDep = isLitShare(key);
-            if (isRolldown || shouldBypassOptimizeDep) {
+            if (shouldBypassOptimizeDep) {
               optimizeDeps.exclude.push(key);
             }
             if (!isRolldown && !shouldBypassOptimizeDep) {
--- a/src/utils/pathNormalization.ts
+++ b/src/utils/pathNormalization.ts
@@ -3,6 +3,7 @@ import { NormalizedModuleFederationOptions } from './normalizeModuleFederationOp
 export const COMMON_SHARED_SUBPATHS: Record<string, string[]> = {
   react: ['react/jsx-runtime', 'react/jsx-dev-runtime'],
   'react-dom': ['react-dom/client', 'react-dom/server', 'react-dom/server.browser'],
+  'solid-js': ['solid-js/web', 'solid-js/store', 'solid-js/html', 'solid-js/h'],
 };
 
 export function removeTrailingSlash(value: string): string {


gioboa (commented):

Screenshot 2026-04-29 at 14 38 25

PR #669, opened 2026-04-29
fix: handle excluded shared keys and scoped remote names in dev bootstrap

Staremang:
## Bug 1: `generateMFManifest` crashes on shared keys removed by `excludeSharedSubDependencies`

The new `excludeSharedSubDependencies` (dev mode only) deletes a shared key from `options.shared` if it is a dependency of another shared package. But the key may already be in `usedShares` (added earlier by `createEarlyVirtualModulesPlugin`, `enforce: "pre"`). When `/mf-manifest.json` is requested, `generateMFManifest` iterates `usedShares`, calls `getNormalizeShareItem(key)` (returns `undefined`), then crashes on `shareItem.version`. Dev server responds 500 and the host fails with `RUNTIME-003: Failed to get manifest`.

`generateLocalSharedImportMap` in the same file already guards against this — `generateMFManifest` was missing the same check.

**Repro:** share two packages where one is listed as a `dependency` of the other (e.g. `@scope/components` deps include `@scope/utils`, both shared).

**Fix** in `src/plugins/pluginMFManifest.ts`: switch `.map()` to `.flatMap()` and skip entries with no `shareItem`.

## Bug 2: Scoped remote names break the auto-preload bootstrap

`pluginAddEntry.ts` builds the host bootstrap with:

```ts
.filter((remote) => remote.includes('/'))
```

intending to keep only sub-path entries (e.g. `remote/App`) and skip bare container names (`remote`). For scoped remotes like `@scope/remote`, the `/` from the scope makes the bare name pass the filter. The generated bootstrap then calls `runtime.loadRemote("@scope/remote")` with no exposed path, which is translated into module `.` and fails with `Module . does not exist in container`. The host never boots.

**Repro:** any host with a scoped remote (`@scope/remote`) where the remote only declares sub-path exposes (no `'.'`).

**Fix** in `src/plugins/pluginAddEntry.ts`: compare against the remote key instead of using the `/` heuristic.

```ts
const remotePreloads = Object.entries(getUsedRemotesMap())
  .flatMap(([remoteKey, remotes]) =>
    Array.from(remotes).filter((remote) => remote !== remoteKey)
  )
  .sort()
  .map((remote) => `runtime.loadRemote(${JSON.stringify(remote)})`)
  .join(',');
```

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -19,6 +19,7 @@ vi.mock('../../utils/packageUtils', () => ({
 
 import addEntry from '../pluginAddEntry';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
+import { addUsedRemote } from '../../virtualModules/virtualRemotes';
 
 type AddEntryPlugin = ReturnType<typeof addEntry>[number];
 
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Image

PR #670, opened 2026-04-29
chore: replace .__mf__temp with virtual module

gu-stav:
We already have a temporary directory, where we store other Vite-related artifacts and would like to configure where the `.__mf__temp` directory is generated, to keep the gitignore files leaner.


gioboa (commented):
I'm going on with this PR because you concern triggered me the idea to verify if that folder is needed or not. apparently we remove that 👀

gioboa (commented):

Image

gioboa:
@gu-stav let me know what you think 🙏

gu-stav:
@gioboa Thanks! I like this approach more! I guess technically this could be seen as a breaking change since some downstream code may depend on the temp folder being there, but since it was clearly marked as temp directory already its probably fine.

While testing this I noticed that `localSharedImportMapPromise` and `exposesMapPromise` cache the rejected promise on transient failures, so the first failure (e.g. Outdated Optimize Dep) permanently breaks shared module loading until a full reload.

I extracted the retry logic into a small `retrySharedInit` helper and reused it for the dynamic imports themselves, and clear the cached promise in a rethrowing .catch so a later call can re-enter the import.

gioboa:
💚 nice, thanks for your help. I'm going to look at this soonish

gioboa (commented):

Image

gioboa (approved):
Thanks @gu-stav LGTM 👏

PR #672, opened 2026-04-30
chore: add `vite-plugin` keyword

gu-stav:
Adding the keyword `vite-plugin` makes it appear in the [Vite Plugin Registry](https://registry.vite.dev).

I think it doesn't need extended metadata - the peer dep ranges are well defined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   },
   "keywords": [
     "vite",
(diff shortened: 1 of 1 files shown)


gioboa (approved):
Nice call 👏 @gu-stav

PR #673, opened 2026-04-30
chore(deps): drop defu

gu-stav:
Even thought `defu` is great, small and self-contained I think the current usage doesn't justify the additional dependency and we can drop it.


gioboa (approved):
That's awesome, thanks for find this out. @gu-stav 🚀

gu-stav:
Ups, I'll take a look at the failures

PR #678, opened 2026-05-01
fix: ensure getInstance works by using runtime init (fixes #675)

### LINKED ISSUE #675: [Bug]: getInstance() returns null in Vite dev although the plugin-created instance exists globally
### Describe the bug

  In Vite dev mode, `@module-federation/vite` creates a host runtime instance during federation bootstrap, but both runtime accessors return `null`:

  - `getInstance()` from `@module-federation/enhanced/runtime`
  - `getInstance()` from `@module-federation/runtime`

  The instance does exist in:

  ```ts
  globalThis.__FEDERATION__.__INSTANCES__

  but it is not returned by either getInstance() API.

### Version

7.3.2

### Reproduction

https://github.com/francesc79/mf-vite-get-instance-repro  -> pnpm i && pnpm dev

### Relevant log output

```shell
{
    "enhancedRuntimeGetInstanceReturnedNull": true,
    "enhancedRuntimeGetInstanceName": null,
    "runtimeGetInstanceReturnedNull": true,
    "runtimeGetInstanceName": null,
    "globalInstanceNames": ["__mfe_internal__host_app"]
  }

  So the plugin-created host instance exists globally, but both public runtime APIs return null.
```

### Validations

francesc79:
Close #675

Replace createInstance-based initialization with runtimeInit in virtualRemoteEntry so the host instance can be correctly retrieved via getInstance.

This removes the internal runtimeInstance state and aligns with the runtime API, preventing duplicated or inaccessible instances in shared/singleton scenarios.

Added regression test to ensure remoteEntry is initialized through runtime init.

### DIFF
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -315,7 +315,7 @@ export function generateRemoteEntry(
   if (typeof __VUE_HMR_RUNTIME__ === 'undefined') {
     globalThis.__VUE_HMR_RUNTIME__ = { createRecord() {}, rerender() {}, reload() {} };
   }
-  import {createInstance, loadRemote} from "@module-federation/runtime";
+  import {init as runtimeInit, loadRemote} from "@module-federation/runtime";
   ${pluginImportNames.map((item) => item[1]).join('\n')}
   ${
     command === 'build'
@@ -326,7 +326,6 @@ export function generateRemoteEntry(
   const initTokens = {}
   const shareScopeName = ${JSON.stringify(options.shareScope)}
   const mfName = ${JSON.stringify(options.internalName)}
-  let runtimeInstance
   let localSharedImportMapPromise
   let exposesMapPromise
   const shouldRetrySharedInitError = ${command !== 'build'} && ((error) => {
@@ -369,19 +368,13 @@ export function generateRemoteEntry(
   async function init(shared = {}, initScope = []) {
     const {usedShared, usedRemotes} = await getLocalSharedImportMap()
     ${generateDirectSharedCacheSeedCode(command)}
-    const runtimeOptions = {
+    const initRes = runtimeInit({
       name: mfName,
       remotes: usedRemotes,
       shared: usedShared,
       plugins: [${pluginImportNames.map((item) => `${item[0]}(${item[2]})`).join(', ')}],
       ${options.shareStrategy ? `shareStrategy: '${options.shareStrategy}'` : ''}
-    };
-    if (!runtimeInstance) {
-      runtimeInstance = createInstance(runtimeOptions);
-    } else {
-      runtimeInstance.initOptions(runtimeOptions);
-    }
-    const initRes = runtimeInstance;
+    });
     // handling circular init calls
     var initToken = initTokens[shareScopeName];
     if (!initToken)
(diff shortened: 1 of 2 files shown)


src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -315,7 +315,7 @@ export function generateRemoteEntry(
   if (typeof __VUE_HMR_RUNTIME__ === 'undefined') {
     globalThis.__VUE_HMR_RUNTIME__ = { createRecord() {}, rerender() {}, reload() {} };
   }
-  import {createInstance, loadRemote} from "@module-federation/runtime";
+  import {init as runtimeInit, loadRemote} from "@module-federation/runtime";
```
gioboa, line 318: Can you look at #656 please?
The actual bug was caused by a PR to solve the #656 issue.

pcfreak30: @francesc79 TLDR `init` is deprecated :upside_down_face: .

francesc79: Tomorrow I ll looking how avoid that!

pcfreak30: I just ran into this same kind of issue with e2e testing b/c only init sets the local var which most of the api still use. this is starting to seem like an upstream bug that needs resolution @gioboa ...

francesc79: Hey @pcfreak30, sorry I missed your work in #656 earlier — I should have checked that first before pushing this PR.

After digging deeper into the runtime behavior, it looks like both approaches hit the same underlying issue:
createInstance does not set the internal FederationInstance, which is still what getInstance() and several runtime APIs rely on.

Because of that, using init (even if deprecated) works, while createInstance alone does not fully initialize the runtime in a compatible way.

At this point, it seems the only clean solution is to fix this upstream in @module-federation/core by updating createInstance to also assign:

`FederationInstance = instance;`

inside the function.

In the meantime, I think it’s reasonable to keep using init as a temporary workaround until the runtime is aligned — if @pcfreak30 and @gioboa agree.

Thanks again for your work on #656 🙏

francesc79: @pcfreak30 @gioboa just a quick update — I’ve opened an upstream issue to address this properly in the runtime:

https://github.com/module-federation/core/issues/4702

Hopefully we can get alignment there so we can rely on createInstance without needing workarounds 👍

gioboa: Great so let's close this one. The solution is to wait the fix on the core project. Thank you all for your commitment 👏

gioboa (commented):
The solution is to keep the new APIs and we are waiting for https://github.com/module-federation/core/issues/4702

pcfreak30:
@gioboa can you confirm though that things work as expected despite the lexical global not getting set?

gioboa:
Based on my tests, the actual implementation is working fine.

simonihmig:
> Based on my tests, the actual implementation is working fine.

Not sure tbh. I noticed that in the original PR, [this example app](https://github.com/module-federation/vite/pull/658/changes#diff-7e7a0449cd9e26fca591bace5b2bb6ad95ebb032307d144224b897de417351bb) was also changed, to not use `loadRemote` directly as imported anymore, most likely due to the fact that this import does not work without the singleton instance being set by `init()`. So having to change this test case would suggest that it was basically hiding the fact that the PR introduced a breaking change. Would you agree?

Beyond that, there is coming some interesting information to light here: https://github.com/module-federation/core/issues/4702#issuecomment-4385635083

mshima:
@gioboa deprecation has been reverted in https://github.com/module-federation/core/pull/4703.
Maybe this PR can be reopened.

gioboa (commented):
Thanks for the heads up @mshima 
I'll revert my PR that replaced the deprecated method. 💪

gioboa:
Thanks @simonihmig and @francesc79 as well 🤗

PR #679, opened 2026-05-01
fix: fix symlinked workspace singleton sharing

### LINKED ISSUE #676: [Bug]: Production singleton shared library using Vaadin is duplicated/evaluated in remote
### Describe the bug

 A shared singleton workspace package is duplicated/evaluated from the remote production bundle when using `@module-federation/vite@1.15.1`.

  The singleton package imports Vaadin at module top level. When the host loads the remote exposed module, the remote loads its own generated shared chunk
  for the singleton package instead of reusing the already-loaded host singleton instance.

  This causes Vaadin top-level side effects to run again and Chrome throws:

  ```text
  InvalidModificationError: Failed to execute 'registerProperty' on 'CSS': The name provided has already been registered.

### Version

7.3.2

### Reproduction

https://github.com/francesc79/mf-vite-vaadin-singleton-repro

### Relevant log output

```shell
pnpm install
  pnpm build
  pnpm serve:remote
  pnpm serve:host

  Open:

  http://localhost:4173

  Remote is served at:

  http://localhost:4174

  ### Expected behavior

  @repro/shared-vaadin-lib is configured as a singleton and should be evaluated/reused once in the page.

  The remote should not load/evaluate its own generated shared chunk for @repro/shared-vaadin-lib.

  ### Actual behavior

  The host first evaluates the shared package:

  [shared-vaadin-lib] evaluated 1

  Then, when the remote exposed module is loaded, Chrome requests the remote-generated shared chunk:

  GET http://localhost:4174/assets/__mfe_internal__remote__loadShare___mf_0_repro_mf_1_shared_mf_2_vaadin_mf_2_lib__loadShare__.mjs-*.js 200

  That chunk evaluates Vaadin again and throws:

  Uncaught InvalidModificationError: Failed to execute 'registerProperty' on 'CSS': The name provided has already been registered.

  Stack points into the remote shared chunk:

  __mfe_internal__remote__loadShare___mf_0_repro_mf_1_shared_mf_2_vaadin_mf_2_lib__loadShare__.mjs-*.js
```

### Validations

francesc79:
Close #676

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -76,6 +76,9 @@ vi.mock('../../utils/packageUtils', () => ({
     if (pkg === 'workspace-shared-lib') {
       return '/repo/packages/workspace-shared-lib/src/index.tsx';
     }
+    if (pkg === 'workspace-esm-only-lib') {
+      return '/repo/apps/remote/node_modules/workspace-esm-only-lib/src/index.ts';
+    }
   }),
   getInstalledPackageJson: vi.fn((pkg: string, opts?: { fromResolvedEntry?: string }) => {
     if (opts?.fromResolvedEntry?.includes('/repo/packages/workspace-shared-lib/')) {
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks @francesc79 i will check it in the next days 👍

gioboa (commented):
build now emits top-level await import(...) and this is breaking the logic. We are running await from TLAs.

gioboa:
@francesc79 I tested this with your repo, it's working great 😊

gioboa (commented):

Image

gioboa (approved):
Thanks @francesc79 for your help

PR #681, opened 2026-05-02
fix(vite): preserve singleton:true packages in excludeSharedSubDependencies

### LINKED ISSUE #680: [Bug]: excludeSharedSubDependencies incorrectly removes singleton: true packages in dev mode
## Describe the bug

In dev mode, `excludeSharedSubDependencies` scans the `dependencies` of each shared package and auto-removes any that also appear in the shared map, to prevent initialization-order issues (e.g. `lit` + `lit-html`).

However, the function has no guard for `singleton: true`, so it also removes packages the user explicitly declared as singletons. For example, if a company SDK or wrapper library lists `react` and `react-dom` in its own `dependencies`, those get silently removed from the shared map — even though they are declared `singleton: true` in the MF config. The result is multiple React instances loading across host and remotes, breaking every hook call.

## Version

Affects `@module-federation/vite` ≤ 1.15.1 (latest at time of writing). Vite 6.x / 7.x.

## Reproduction

https://github.com/ThibautMarechal/mf-vite-singleton-repro

```bash
pnpm install

# Terminal 1
pnpm dev:remote

# Terminal 2
pnpm dev:host
```

Open `http://localhost:5173`. The host terminal shows:

```
[MF warn] "react" is a dependency of shared package "company-sdk" and is also shared separately. [...] Auto-excluding "react" from shared modules for dev mode.
```

The remote `<Button>` (which uses `useState`) fails to render and the browser console shows a hook error.

## Relevant log output

```shell
[MF warn] "react" is a dependency of shared package "company-sdk" and is also shared separately. This may cause initialization order issues in dev mode. Consider sharing only "company-sdk".
  Auto-excluding "react" from shared modules for dev mode.
[MF warn] "react-dom" is a dependency of shared package "company-sdk" and is also shared separately. This may cause initialization order issues in dev mode. Consider sharing only "company-sdk".
  Auto-excluding "react-dom" from shared modules for dev mode.
```

Then at runtime:

```
Uncaught TypeError: Cannot read properties of null (reading 'useState')
```

ThibautMarechal:
## Summary

Closes #680

`excludeSharedSubDependencies` is a dev-mode heuristic that auto-removes shared packages that are sub-dependencies of other shared packages, to prevent initialization-order issues (e.g. `lit` depending on `lit-html`).

However, the function was also silently removing packages explicitly declared with `singleton: true` — such as `react` and `react-dom` — when a company SDK or wrapper library lists them in its `dependencies`. This created multiple React instances across host and remotes, breaking all hooks with `Cannot read properties of null`.

The function already skips packages with `import: false`. This PR adds the same guard for `singleton: true`.

## Changes

- **`src/plugins/pluginProxySharedModule_preBuild.ts`** — one-line fix: add `singleton === true` guard in `excludeSharedSubDependencies`
- **`src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts`**:
  - Update existing `'excludes shared sub-dependencies in dev mode and warns'` test: change lit sub-packages from `singleton: true` → `singleton: false` (they were using `singleton: true` but asserting removal, which would now fail correctly)
  - Add `'preserves singleton: true shared sub-dependencies in dev mode without warning'` — asserts react/react-dom are NOT removed when a wrapper lib lists them as deps
  - Add `'still removes non-singleton sub-dependencies when sibling singletons are preserved'` — asserts the original lit-style behavior is preserved for non-singleton sub-deps
- **`.changeset/lucky-lions-rest.md`** — patch changeset

## Test plan

- [x] All 302 existing tests pass (`pnpm test`)
- [x] New tests cover: singleton sub-dep preserved, non-singleton sub-dep still removed
- [x] TypeScript type-check passes (`pnpm typecheck`)
- [x] Formatting check passes (`pnpm fmt.check`)

### DIFF
--- a/.changeset/lucky-lions-rest.md
+++ b/.changeset/lucky-lions-rest.md
@@ -1,0 +1,14 @@
+---
+"@module-federation/vite": patch
+---
+
+fix: preserve `singleton: true` shared packages in `excludeSharedSubDependencies`
+
+The dev-mode heuristic that auto-removes shared sub-dependencies (to prevent
+initialization-order issues such as `lit`/`lit-html`) was also silently removing
+packages explicitly declared with `singleton: true` — for example `react` and
+`react-dom` when a company SDK or wrapper library lists them in its `dependencies`.
+This caused multiple React instances to load across host and remotes, breaking all hooks.
+
+The fix adds a `singleton: true` guard alongside the existing `import: false` guard
+so that explicitly-declared singletons are never auto-excluded.
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -688,21 +688,21 @@ describe('pluginProxySharedModule_preBuild', () => {
         from: '',
         version: '3.3.2',
         scope: 'default',
-        shareConfig: { singleton: true, requiredVersion: '^3.3.2', strictVersion: false },
+        shareConfig: { singleton: false, requiredVersion: '^3.3.2', strictVersion: false },
       },
       'lit-element': {
         name: 'lit-element',
         from: '',
         version: '4.2.2',
         scope: 'default',
-        shareConfig: { singleton: true, requiredVersion: '^4.2.2', strictVersion: false },
+        shareConfig: { singleton: false, requiredVersion: '^4.2.2', strictVersion: false },
       },
       '@lit/reactive-element': {
         name: '@lit/reactive-element',
         from: '',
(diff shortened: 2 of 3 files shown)


gioboa (commented):
Thanks for your help @ThibautMarechal 
I will look at this soonish

ThibautMarechal:
It fix my issue, but I don't know the impacts of other use cases. If you have another workarround, I can take it :)

gioboa (commented):

Image

gioboa (approved):
Thanks for this fix @ThibautMarechal 👏

PR #685, opened 2026-05-03
fix: fix up SystemJS shared proxy chunks

### LINKED ISSUE #682: [Bug]: runtime.loadShare(...) call needs to wait for the Module Federation runtime to initialize
### Describe the bug

The host__loadShare__react__loadShare__.js has an async execute loading React via runtime.loadShare("react", { requiredVersion: "^19.2.4" }), but the proxy module's setter runs synchronously, leaving React4 = undefined. Calling getAugmentedNamespace(undefined) then throws because Object.prototype.hasOwnProperty.call(null/undefined, ...) throws a TypeError.

Problem is that loadShare.js uses an async execute because the underlying runtime.loadShare(...) call needs to wait for the Module Federation runtime to initialize and coordinate module sharing.

react and react-dom are CommonJS packages. When @module-federation/vite wraps them as shared modules, Rollup generates a getAugmentedNamespace CJS→ESM proxy that tries to read .arguments on a strict-mode function — which throws in ES module context.

### Version

1.14.5 and 1.15.0

### Reproduction

Vite rollup output config as "system" and react/react-dom shared across using module federation

### Relevant log output

```shell
2__mfe_internal__host__loadShare__react__loadShare__.js_commonjs-proxy-CmUQzkpA.js:17 Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at hasOwnProperty (<anonymous>)
    at getAugmentedNamespace (__mfe_internal__host__loadShare__react__loadShare__.js_commonjs-proxy-CmUQzkpA.js:17:45)
    at Object.execute (__mfe_internal__host__loadShare__react__loadShare__.js_commonjs-proxy-CmUQzkpA.js:45:55)
    at r (systemjs.min@6.14.1.js:4:3638)
    at v (systemjs.min@6.14.1.js:4:4074)
    at systemjs.min@6.14.1.js:4:3979
    at Array.forEach (<anonymous>)
    at v (systemjs.min@6.14.1.js:4:3948)
    at systemjs.min@6.14.1.js:4:3551
```

### Validations

gioboa:
Close #682

### DIFF
--- a/integration/host-build.test.ts
+++ b/integration/host-build.test.ts
@@ -58,10 +58,11 @@ describe('host build', () => {
     const bootstrapAsset = output.output.find(
       (item) => item.type === 'asset' && item.fileName.includes('mf-entry-bootstrap')
     );
-    expect(bootstrapAsset?.source).toContain('const { initHost } = await import(');
+    expect(bootstrapAsset?.source).toContain('const { initHost } = await __mfImport(');
     expect(bootstrapAsset?.source).toContain('const runtime = await initHost();');
     expect(bootstrapAsset?.source).toContain('runtime.loadRemote("remote1/Module")');
-    expect(bootstrapAsset?.source).toContain('})().then(() => import(');
+    expect(bootstrapAsset?.source).toContain('})().then(() => __mfImport(');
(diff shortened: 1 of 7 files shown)


gioboa (commented):

Image

PR #686, opened 2026-05-04
chore: bump vite dep

gioboa:


gioboa (commented):

Image

PR #687, opened 2026-05-04
fix: use protocol relative fallback origin

simonihmig:
In our setup, we use `ignoreOrigin` and also have vite configured to run on TLS. That used to work, prior to #638, but since then it had a hardcoded `http://` fallback origin, so the `remoteEntry.js` request using http would fail to get served by vite listening on https. This change is making use of a protocol relative URL ( `//host`), as it seems to have been before.


src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
```diff
@@ -5,14 +5,14 @@ import type {
...
 import { getDefaultMockOptions } from '../../utils/__tests__/helpers';
-import { normalizeModuleFederationOptions } from '../../utils/normalizeModuleFederationOptions';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
+import { normalizeModuleFederationOptions } from '../../utils/normalizeModuleFederationOptions';
+import { getHostAutoInitPath } from '../../virtualModules';
+import pluginProxyRemoteEntry from '../pluginProxyRemoteEntry';
 
 describe('pluginProxyRemoteEntry', () => {
-  it('uses an absolute fallback origin for dev host init in SSR/module-runner contexts', async () => {
+  it('uses an inlined data-URL origin for dev host init in SSR/module-runner contexts, protocol relative fallback origin otherwise', async () => {
```
simonihmig, line 15: Not entirely sure about the intention of this test. AFAICT, in SSR mode (`typeof window === 'undefined') it wouldn't use the fallback origin at all (unlike what the title said), but an inlined data-URL. 

I hope the changes in this test make sense, but please double-check! :)

gioboa: Thanks, It was misleading 🙏

src/plugins/pluginProxyRemoteEntry.ts
```diff
@@ -118,7 +118,7 @@ export default function ({
             const publicPath = JSON.stringify(
               resolvePublicPath(options, viteConfig.base) + options.filename
             );
-            const fallbackOrigin = `http://${host}:${viteConfig.server?.port}`;
+            const fallbackOrigin = `//${host}:${viteConfig.server?.port}`;
```
gioboa, line 121: Nice call 👏

gioboa (commented):

Image

gioboa (approved):
Thanks @simonihmig for your amazing help 👏

PR #688, opened 2026-05-04
chore: replace magic-string with helper

gioboa:


gioboa (commented):

Image

PR #695, opened 2026-05-06
revert: replace deprecated init helper

### LINKED ISSUE #675: [Bug]: getInstance() returns null in Vite dev although the plugin-created instance exists globally

gioboa:
Close #675

Reverts module-federation/vite#658

Deprecation has been reverted in https://github.com/module-federation/core/pull/4703 so It's safer to undo this change as well.

### DIFF
--- a/examples/vite-webpack-rspack/host/src/App.tsx
+++ b/examples/vite-webpack-rspack/host/src/App.tsx
@@ -1,3 +1,4 @@
+import { registerRemotes } from '@module-federation/runtime';
 import React, { lazy, StrictMode, Suspense, useEffect } from 'react';
 import { createRoot } from 'react-dom/client';
 import { Footer } from './components/Footer';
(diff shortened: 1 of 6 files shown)


gioboa (commented):

Image

PR #696, opened 2026-05-06
fix: skip host cache seeding for `import:false` modules during build

JammingBen:
In build mode, `generateHostAutoInitSharedCacheSeedCode` was emitting static `import()` calls for shared modules with `import: false`. This caused the bundler to resolve those modules and their transitive dependencies, which fails under strict package managers like pnpm where transitive deps are not hoisted.

The fix skips cache seeding in build mode for `import:false` modules. The `runtime.loadShare()` loop that follows will handle acquiring these modules at runtime instead. Serve mode is unaffected since Vite's dev server resolves on-demand without requiring all transitive deps to be available at build time.

See here for a simple reproducer: https://github.com/JammingBen/mf-vite-seed-repro


gioboa (commented):

Image

gioboa (approved):
Thanks for your help @JammingBen

PR #698, opened 2026-05-07
fix: restore async loadShare fallback for import:false shared modules

JammingBen:
This is a follow-up for https://github.com/module-federation/vite/issues/660. After updating to `v1.15.2`, I found some more issues in our setup related to the refactor in d7d1b4d.

It replaced the async `loadShare` approach for `import:false` shared modules with a synchronous check on the global module cache. This works when the host uses `@module-federation/vite` (which generates `hostAutoInit` to pre-populate the cache), but breaks when the host uses `@module-federation/runtime` directly.

Restore top-level await with `initPromise` as a fallback: first check the cache (fast path), and if empty, await the runtime and call `loadShare()` to resolve the module from the host's registered shared.

I'm not sure if this is the best solution, but it works for our setup. I created a minimal reproduction repo in case you want to test it yourself: https://github.com/JammingBen/mf-vite-import-repro

### DIFF
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -96,7 +96,8 @@ describe('shared dependencies', () => {
     expect(remoteEntry).toBeDefined();
     expect(loadShare).toBeDefined();
     expect(remoteEntry!.code).toContain('const versions =');
-    expect(remoteEntry!.code).toContain('shared[pkg]');
+    expect(remoteEntry!.code).toContain('[pkg]');
+    expect(remoteEntry!.code).toContain('const provider = versions && versions[Object.keys(versions)[0]]');
     expect(remoteEntry!.code).toContain('__mfModuleCache.share[pkg]');
     expect(remoteEntry!.code).not.toContain('initRes.loadShare(pkg');
     expect(loadShare!.code).toContain('initPromise.then');
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -80,6 +80,33 @@ describe('shared dependencies', () => {
     expect(localSharedImportMap!.code).toContain('import: false');
   });
 
+  it('supports import:false named imports with runtime-registered host shares', async () => {
+    const output = await buildFixture({
+      fixture: 'shared-remote',
+      mfOptions: {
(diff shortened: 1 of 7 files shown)


gioboa (commented):

Image

gioboa (commented):
I'm proposing these changes to avoid using top-level await. Please feel free to change, add, or remove anything.

gioboa:
Screenshot 2026-05-07 at 12 48 20

JammingBen:
> I'm proposing these changes to avoid using top-level await. Please feel free to change, add, or remove anything.

It doesn't work for me unfortunately. Trying it in my setup and the minimal reproduction repo in https://github.com/JammingBen/mf-vite-import-repro, console prints: `Error: [Module Federation] Shared module 'vue' must be provided by host`. It looks like the cache still doesn't get populated (correctly).

gioboa:
Let me try to craft a fix with your repo.

gioboa:

Screenshot 2026-05-07 at 21 35 02

gioboa (approved):
I tested your repo and it's working fine 💪

gioboa:
Screenshot 2026-05-07 at 21 38 12

JammingBen:
Thanks for your help on this! I still can't get it to work properly though, sorry 🙈 I updated the reproduction repo to be a bit more like the real world setup by having the host app setup Vue and the remote providing a component that the host app then uses. Vue is a shared dependency, since the remote app should just use the host app's Vue instance.

The console now prints: `index.ts:9 Uncaught (in promise) TypeError: defineComponent is not a function`

Looks like the cache now gets populated, but there is still something weird with how the shared dependencies get resolved. The latest working version is `1.14.2` (before https://github.com/module-federation/vite/pull/608 got merged).

On a sidenote, the reproduction repo is now at least somewhat close to our use case and setup. I guess that might be the case for others as well. Do you think we could have an integration test for this? I could look into it myself if you agree, as soon as I find a little spare time.

gioboa:
Yes please, contact me on discord or LinkedIn, I would like to chat about one thing. Thank you 🙏

540792740:
That worked perfectly!

PR #702, opened 2026-05-08
fix: fix import-false shared export race

gioboa:


gioboa (commented):

Screenshot 2026-05-09 at 00 18 55

PR #718, opened 2026-05-18
fix: handle side-effect only modules in normalizeRuntimeShare

simonihmig:
This is fixing an issue if your shared deps cover a module without exports (only side-effects). Ran into this when sharing `lit/` which covers [`lit/polyfill-support`](https://github.com/lit/lit/blob/main/packages/lit-element/src/polyfill-support.ts). The browser seems to expose this with an (empty) special module namespace object as `default`. The `__mfNormalizeRuntimeShare` function would then return this one instead of the original module object. In the `const exportModule = normalizedModule === mod ? {...mod} : normalizedModule;` statement, `normalizedModule` would not be equal to `mod`, so it would try to `defineProperty` on that namespace object, which fails as that is a exotic esm object that you cannot mutate. 

This can be reproduced when adding `lit/` in the lit example app [here](https://github.com/simonihmig/module-federation-vite-examples/blob/main/lit/host/vite.config.ts#L24).


gioboa (commented):

Screenshot 2026-05-18 at 12 04 19

gioboa (commented):
Thanks for your amazing help @simonihmig

PR #720, opened 2026-05-18
feat: add createModuleFederationConfig helper

tonoizer:
This PR adds `createModuleFederationConfig` helper for Vite config DX to further follow more the module federation recommendations to use the `module-federation.config.ts`file structure (https://module-federation.io/guide/build-plugins/plugins-rspack.html#create)

```ts
createModuleFederationConfig(options) => options
```

The goal is DX consistency with the official Module Federation integrations for Rspack, Rsbuild, and Rspress, where users are encouraged to define their federation config in a dedicated `module-federation.config.ts` file and pass that object into the bundler plugin.

# Why

- Aligns Vite with the existing Module Federation config-file convention.
- Gives users a simple, typed entry point for `module-federation.config.ts`.
- Improves discoverability and keeps docs/examples consistent across bundlers.
- Keeps Vite-specific typing by using `@module-federation/vite`'s own `ModuleFederationOptions` type instead of the generic SDK plugin type.

# Example

```ts
// module-federation.config.ts
import { createModuleFederationConfig } from '@module-federation/vite';

export default createModuleFederationConfig({
  name: 'remote_a',
  exposes: {
    './Feature': './src/Feature.tsx',
  },
});
```

```ts
// vite.config.ts
import moduleFederationConfig from './module-federation.config';

export default defineConfig({
  plugins: [federation(moduleFederationConfig)],
});
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1101,4 +1101,13 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
   ];
 }
 
-export { federation, type ModuleFederationOptions, type PluginManifestOptions };
+function createModuleFederationConfig(options: ModuleFederationOptions): ModuleFederationOptions {
+  return options;
+}
+
+export {
+  createModuleFederationConfig,
+  federation,
+  type ModuleFederationOptions,
+  type PluginManifestOptions,
+};
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,8 +1,8 @@
 import { readFileSync, writeFileSync } from 'fs';
 import { createRequire } from 'module';
 import path from 'pathe';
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks for your help @tonoizer

src/index.ts
```diff
@@ -1101,4 +1101,13 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
   ];
 }
 
-export { federation, type ModuleFederationOptions, type PluginManifestOptions };
+function createModuleFederationConfig<T extends ModuleFederationOptions>(options: T): T {
```
gioboa, line 1104: Generic preserves the caller’s exact config type.

PR #721, opened 2026-05-18
feat: Vue HMR + HMR adapter for other frameworks

Staremang:
## Summary

Adds cross-federation HMR support for Vue and reshapes the dev-remote HMR plugin around a clear adapter pattern.

- **Vue support**: SFCs in remote modules now hot-reload in the host page. Two collisions are fixed:
  - `__VUE_HMR_RUNTIME__` guard injected into the host's HTML head pins the first-loaded Vue runtime, so the remote's Vue copy can't silently overwrite it.
  - `_sfc_main.__hmrId` literals in remote code are rewritten with a federation-name prefix, so host/remote SFCs sharing a path don't collide on the shared HMR registry.
- **Adapter pattern**: framework-specific HMR fixes live behind an `HmrAdapter` interface (`src/plugins/hmr/`). React (`/@react-refresh` proxy) and Vue ship as built-in adapters. Hooks are grouped under `remote` / `host` namespaces so the side they run on is part of the data shape, not implicit gating in the plugin.
- **Full-reload extracted**: the metadata endpoint, remote broadcast, and host Node-to-Node WS relay live in `src/plugins/hmr/fullReload.ts`. The main plugin is now a thin orchestrator over adapters + three setup functions.
- **Strategy resolution**: `'native'` when a matched adapter handles HMR, `'full-reload'` otherwise (or when explicitly requested via `dev.remoteHmr: 'full-reload'`). Resolved once in `configResolved`, reused across hooks.


gioboa (approved):
Thanks @Staremang that's a great piece of code 👏

gioboa:

Screenshot 2026-05-19 at 22 26 48

PR #725, opened 2026-05-19
fix: preserve expose exports when rollup input matches expose entry

### LINKED ISSUE #724: [Bug]: 1.15.x loses named exports on exposed module when rollup input matches the expose entry
### Summary

`@module-federation/vite@1.15.x` loses the named exports on the exposed module when `rollupOptions.input` resolves to the same absolute path as one of the `exposes` entries. `loadRemote("remote/Expose")` resolves to an empty `{}` and any named import the host expected is `undefined`.

This is the pattern any setup that generates its remote-entry file at build time hits — the generated file ends up being both the rollup input of the federation build and the expose target. In our case it's a `sanity dev`-style flow; the result is the host's render call throwing `did not expose a render function`.

### Repro

<https://github.com/gu-stav/mf-vite-1.15-expose-bug>

```sh
pnpm install
pnpm build:remote && pnpm preview:remote &
pnpm dev:host
```

Open <http://localhost:5173>:

```
loadRemote("remote/Expose") returned:
  keys: []
  render is function: false
  → BUG: render is missing
```

Pinning `remote/package.json` to `1.14.5` makes it pass.

### Root cause

In the built `dist/assets/virtualExposes-*.js`, the expose factory imports a bootstrap shim instead of the real chunk:

```js
// 1.14.5 — works
"./Expose": async () => {
  const e = await u(() =>
    f(() => import("./expose-<hash>.js").then((s) => s.Z), [...]),
  );
  Object.assign(t, e);
  // ...
};

// 1.15.5 — broken (no `.then((s) => s.Z)`, and the imported chunk is the bootstrap shim, not the real module)
"./Expose": async () => {
  const e = await u(() => c(() => import("./expose-<hash>.js"), []));
  Object.assign(t, e);
  // ...
};
```

The shim body is `(async () => { await initHost(); })().then(() => import(realEntry))`. The chain's resolved value is discarded, so the dynamic import in `virtualExposes` resolves to an empty namespace and `Object.assign(t, e)` copies nothing.

### Bisect

```
# bad: 6cb1bd9 1.15.0
# good: a5182ef 1.14.5
# first bad commit: d7d1b4d refactor: remove TLAs and one customResolver (#638)
```

PR #638 is also referenced as the cause of #674 (which fixed a different `__moduleExports` symptom from the same change). I suspect there's a remaining case in `pluginAddEntry`'s build-time transform path where the entry chunk gets wrapped with `getBootstrapSource` but `virtualExposes`' resolution still points at the wrapped chunk instead of the real one.

gu-stav:
Fixes #724.

`pluginAddEntry`'s build-time transform wraps every `rollupOptions.input` entry with a host-init bootstrap IIFE that loads the original module only for its side-effects and discards the namespace. When the same file is also an `exposes` target, `virtualExposes`' dynamic `import(exposePath)` resolves to that wrapped chunk and `loadRemote("remote/Expose")` hands back an empty namespace.

The plugin already has a signal for "this build is a remote whose output is never browser-requested as a page entry": `forceClientInjected`, set whenever `Object.keys(options.exposes).length > 0`. It was being used to gate the SSR fallback but not the `rollupOptions.input` clause — so for a build with exposes, the rollup-input wrap fired anyway and clobbered the exposed module's exports.

Init for these builds is guaranteed via `loadRemote → virtualExposes → remoteEntry.init`, so the wrap is both redundant and lossy for them. Extending the existing `forceClientInjected` gate to cover the rollup-input clause restores 1.14.5 behavior without re-introducing the TLA / static-import patterns #638 removed.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -194,6 +194,40 @@ describe('pluginAddEntry', () => {
     expect(result?.code).not.toContain('globalThis.System.import(src)');
   });
 
+  it('skips host-init bootstrap on rollupOptions.input entries when the build has exposes (#724)', async () => {
+    const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mf-add-entry-expose-input-'));
+    const exposePath = path.join(tempDir, 'src/expose.ts');
+
+    const plugins = addEntry({
+      entryName: 'hostInit',
+      entryPath: '/virtual/hostInit.js',
+      inject: 'entry',
+      forceClientInjected: true,
+    });
+    const buildPlugin = plugins[1];
+
+    runConfig(
+      buildPlugin,
+      {} as ConfigPluginContext,
+      { build: { rollupOptions: { input: exposePath } } },
+      { command: 'build', mode: 'production' }
+    );
+    runConfigResolved(buildPlugin, {
+      root: tempDir,
+      base: '/',
+      command: 'build',
+      build: { rollupOptions: { input: exposePath } },
+    } as unknown as ResolvedConfig);
+
+    const originalCode = 'export function render(root) { root.textContent = "hi" }';
+    const result = await runTransform(buildPlugin, originalCode, exposePath);
+
+    // The transform must skip — wrapping with the bootstrap IIFE would
+    // discard the named exports and virtualExposes' dynamic import of this
+    // same path would resolve to an empty namespace, breaking loadRemote.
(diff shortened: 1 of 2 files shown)


gioboa (commented):
This PR fixes the example repro because forceClientInjected is true whenever exposes exist, but the condition is too broad: it skips normal hostInit injection for every remote that has exposes, instead of only skipping exposed files.
I created #726 with a more complete solution so I'm closing this one.
Thanks for you help @gu-stav  🙏

PR #734, opened 2026-05-21
fix: fix HTML input federation bootstrap build

### LINKED ISSUE #729: [Bug]: Version 1.15.5 completely broken on fresh projects
## Bug 1: Broken build — remoteEntry references non-existent files

**Package:** `@module-federation/vite@1.15.5`  
**Affects:** host app build output

After `vite build` in `vite-host`, the generated `remoteEntry.js` references chunk files that do not exist in the output directory. Loading the remote in a host results in a network 404 for every exposed module chunk.

In dev mode the problem is different but related: instead of real module imports from the remote, the host receives empty proxy objects — exposed components and utilities are effectively `undefined` at runtime.

**Steps to reproduce:**

```bash
pnpm build
pnpm preview
# inspect console in browser devtools
```

**Expected:** `remoteEntry.js` references actual emitted chunks.  
**Actual:** references point to files absent from `dist/`.

## Bug 2: Build crashes when `input` is an HTML file

**Package:** `@module-federation/vite@1.15.5`  
**Affects:** host app build

The host uses a named HTML entry in `rolldownOptions.input`:

```ts
// vite-host/vite.config.ts
build: {
  rolldownOptions: {
    input: {
      main: "indexProd.html",
    },
  },
},
```

Running `vite build` in `vite-host` throws an error caused by the MF plugin. Removing the `federation()` plugin from the config makes the build succeed — confirming the plugin is the root cause.

**Steps to reproduce:**

```bash
pnpm build
```

**Expected:** build completes successfully.  
**Actual:** build crashes with an error from the MF plugin when processing the HTML input entry.

## Bug 3: `filename` with hash pattern — remoteEntry file not emitted

**Package:** `@module-federation/vite@1.15.5`  
**Affects:** remote app build output

When `filename` in the `federation()` config contains a hash pattern, the file is never written to `dist/`:

```ts
// vite-remote/vite.config.ts
federation({
  name: "@test/remote",
  filename: "remoteEntry-[hash].js",
  // ...
})
```

Two failure modes depending on whether `varFileName` is set:

- **`varFileName` specified** — build completes but `remoteEntry-[hash].js` is absent; the manifest and internal references point to the `varFileName` value instead.
- **`varFileName` not specified** — build completes with no `remoteEntry` file emitted at all; the remote is completely unreachable.

Removing the hash placeholder (`filename: "remoteEntry.js"`) restores normal behavior.

**Steps to reproduce:**

```bash
pnpm build
# no remoteEntry-*.js file in dist/
```

**Expected:** `dist/remoteEntry-[hash].js` emitted and referenced correctly.  
**Actual:** file absent from `dist/`; references broken or missing entirely.

### Version

1.15.5

### Reproduction

https://github.com/alternativniy/mf-vite-8-named-exports

### Relevant log output

```shell

```

### Validations

### My system

CachyOS Linux
NodeJS v26.2.0

gioboa:
Close #729

### DIFF
--- a/integration/build.test.ts
+++ b/integration/build.test.ts
@@ -14,10 +14,11 @@ describe('build', () => {
     it('remoteEntry contains federation runtime init with correct name', async () => {
       const output = await buildFixture({ mfOptions: BASIC_REMOTE_MF_OPTIONS });
       const remoteEntry = findChunk(output, 'remoteEntry');
+      const allCode = getAllChunkCode(output);
       expect(remoteEntry).toBeDefined();
-      expect(remoteEntry!.code).toContain('basicRemote');
-      expect(remoteEntry!.code).toContain('localSharedImportMapPromise');
-      expect(remoteEntry!.code).toContain('getExposesMap');
+      expect(allCode).toContain('basicRemote');
+      expect(allCode).toContain('localSharedImportMapPromise');
+      expect(allCode).toContain('getExposesMap');
     });
 
     it('exposed module content is included in output', async () => {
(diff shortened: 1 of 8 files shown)


gioboa (commented):

Screenshot 2026-05-21 at 22 39 06

PR #735, opened 2026-05-21
fix: fix up loaded-first support

### LINKED ISSUE #732: [Bug]: shareStrategy loaded-first loading remotes when the host is first initialised
### Describe the bug

I have `shareStrategy` set to `loaded-first` which loads up the remotes when you require the component to be rendered rather than when the host is initialised.

When i upgraded the package from 14.5.0 to 1.15.5 the remotes are loaded when the runtime is initialised causing the app to error if the remote isn't available.

I have reproduced the issue in the react example of the MF Vite examples

### Version

7.3.3

### Reproduction

https://github.com/JamesMHenderson/module-federation-vite-examples/tree/main/react

### Relevant log output

```shell
Uncaught (in promise) Error: [ Federation Runtime ]: Failed to get manifest. #RUNTIME-003
args: {"manifestUrl":"http://localhost:3000/conductor/remotes/error/mf-manifest.json","moduleName":"remote_error","hostName":"__mfe_internal__Conductor"}
View the docs to see how to solve: https://module-federation.io/guide/troubleshooting/runtime#runtime-003
Original Error Message:
 SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data
```

### Validations

gioboa:
Close #732

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1,12 +1,20 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-const { hasPackageDependencyMock, normalizedSharedMock, usedRemotesMapMock, writeSyncSpy } =
-  vi.hoisted(() => ({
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Image

PR #736, opened 2026-05-21
fix: solve circular remote import deadlock

### LINKED ISSUE #733: [Bug]: Two remotes that statically import each other's exposed modules hang forever (host never resolves loadRemote)
### Describe the bug

Two remotes that statically import each other's exposed modules deadlock when the host loads both. Page hangs, no error, no log.

Setup (full repro in the link, plain TS, no frameworks):

- `alpha` (:3001) exposes `./Service`, statically imports `beta/Service`
- `beta` (:3002) exposes `./Service`, statically imports `alpha/Service`
- `host` (:5173) does `Promise.all([import('alpha/Service'), import('beta/Service')])`

In dev both `loading…` slots never fill. I added a 10s setTimeout that paints the page red so you can spot it without opening DevTools.

Prod build is cleaner evidence. The auto-generated `mf-entry-bootstrap-0.js` does:

```js
const __mfRemotePreloads = [
  runtime.loadRemote("alpha/Service"),
  runtime.loadRemote("beta/Service"),
];
await Promise.all(__mfRemotePreloads);
```

before importing the user entry. That Promise.all never settles, so `main.ts` never runs and the page sits empty with zero console output. The hang is in the federation runtime itself.

Same shape as #180 (closed via #191). 1.1.7 fixed something there and now also logs `The remote "alpha" is already registered`, but the bidirectional case still hangs on both 1.1.7 and 1.15.5.

Tried on 1.15.5, all still hang:

- `entry: '.../remoteEntry.js'` vs `entry: '.../mf-manifest.json'`
- `hostInitInjectLocation: 'entry'` (matches the upstream `examples/vite-vite/vite-remote` config)
- with and without `shared: {}`

Workarounds that sidestep the cycle:

- one side uses `await import('peer/Module')` inside an event handler, resolves lazily
- move the shared module into a third remote that nothing consumes back
- use `shared:` (the maintainer's suggestion on #180). That dedupes a common npm/workspace package across remotes. It's not the same as one remote consuming another remote's exposed module, which is what this repro needs.

Env: Node 24.12.0, pnpm 10.26.2, Windows 11, Chromium.

Happy to PR a fix if someone points me at the right direction.

### Version

v1.15.5 with vite 8.0.14 (also reproduced on 1.1.7 + vite 6.4.2)

### Reproduction

https://github.com/kamranasad7/mf-vite-bidirectional-repro

### Relevant log output

```shell
Dev: no output. Host script runs, dynamic imports never resolve.

Prod: no output. main.ts never runs because the bootstrap awaits
Promise.all([loadRemote('alpha/Service'), loadRemote('beta/Service')])
first and that promise never settles.
```

### Validations

gioboa:
Close #733

### DIFF
--- a/integration/host-build.test.ts
+++ b/integration/host-build.test.ts
@@ -60,7 +60,8 @@ describe('host build', () => {
     );
     expect(bootstrapAsset?.source).toContain('const { initHost } = await __mfImport(');
     expect(bootstrapAsset?.source).toContain('const runtime = await initHost();');
-    expect(bootstrapAsset?.source).toContain('runtime.loadRemote("remote1/Module")');
+    expect(bootstrapAsset?.source).toContain('__mfPreloadRemote("remote1/Module")');
+    expect(bootstrapAsset?.source).toContain('runtime.loadRemote(remote)');
     expect(bootstrapAsset?.source).toContain('})().then(() => __mfImport(');
     expect(bootstrapAsset?.source).toContain('globalThis.System.import(src)');
     expect(bootstrapAsset?.source).toContain('hostInit');
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -390,8 +390,9 @@ describe('pluginAddEntry', () => {
       | { code: string }
       | undefined;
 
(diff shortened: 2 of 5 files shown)


gioboa (commented):

Image

PR #737, opened 2026-05-22
fix: host bootstrap respects shareStrategy and tolerates offline remotes

Staremang:
## Problem

Run only the host in dev (without its remotes) and the host renders a **blank page** - even screens that use no remote. The console shows `RUNTIME-003: Failed to get manifest`.

## Why

The plugin's generated bootstrap eagerly loads every remote before importing the app entry:

```js
await Promise.all(remotes.map(r => runtime.loadRemote(r))); // fails here
import("/src/main.ts"); // never runs
```

Two bugs:

1. It preloads remotes even with `shareStrategy: 'loaded-first'`, which is meant to load remotes only on demand.
2. `Promise.all` makes one offline remote fatal – the bootstrap rejects and the app entry never loads.

## The fix

In `getBootstrapSource` (`src/plugins/pluginAddEntry.ts`):

1. With `shareStrategy: 'loaded-first'`, skip the preload – the host boots, each remote loads only when its screen opens.
2. With `version-first`, use `Promise.allSettled` so an offline remote can't abort startup.

This matches the official runtime: it preloads remotes only for `version-first` and skips failed ones instead of crashing.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -17,6 +17,26 @@ vi.mock('../../utils/packageUtils', () => ({
   hasPackageDependency: () => true,
 }));
 
+const mockMfOptions = vi.hoisted(() => ({
+  shareStrategy: 'version-first' as 'version-first' | 'loaded-first',
+}));
+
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #739, opened 2026-05-22
fix: resolve scoped remote names

Staremang:
## Problem
If remote name `@scope/remote` contains a slash. The code took only the first path segment => `@scope`, which is not a configured remote 

## Fix
This is a quick fix. It can be optimized, but it will require major interface changes.


gioboa (commented):

Screenshot 2026-05-22 at 17 13 24

PR #741, opened 2026-05-22
fix: fix up loaded-first config

### LINKED ISSUE #740: [Bug]: Lazy-loading problem in loaded-first
### Describe the bug

`shareStrategy: 'loaded-first'` uses a lazy proxy for remote modules. While the module is unloaded, every access on the proxy does `throw <Promise>`.

This is designed for **React Suspense**, which catches thrown Promises. **Vue does not** – a thrown Promise is just an ordinary error to it.

Image

https://github.com/Staremang/module-federation-vite-examples/commit/b688f42e1f1e8077842f657ae136d7780eb7ce51

### Version

b1701518069656dbbd325edc4c802dd219e8a946 commit (next version)

### Reproduction

https://github.com/Staremang/module-federation-vite-examples/tree/example/loaded-first-strategy/vue

### Relevant log output

```shell
Uncaught (in promise) Error: [object Promise]
```

### Validations

gioboa:
Close #740

### DIFF
--- a/src/virtualModules/__tests__/virtualRemotes.test.ts
+++ b/src/virtualModules/__tests__/virtualRemotes.test.ts
@@ -62,14 +62,14 @@ describe('generateRemotes', () => {
     expect(code).toContain('runtime.loadRemote("remote/Button")');
   });
 
-  it('defers remote loading until proxy use for loaded-first', () => {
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-05-22 at 22 55 26

PR #742, opened 2026-05-23
fix: framework-agnostic remote exports, deferred loaded-first, and unified SSR gates

tonoizer:
- Replace the React-specific remote `Proxy` with real module resolution (`__mfUnwrapRemoteDefault` + top-level `await`) for **build**, **`version-first` dev**, and **SSR**.
- Restore **deferred** client loading for `shareStrategy: 'loaded-first'` via a minimal framework-agnostic proxy (`__mfCreateDeferredRemoteProxy`); SSR still awaits the real remote module.
- Centralize SSR feature flags in `getSsrCapabilities()` so `ssrEntryLoader` injection and dev SSR bootstrap stay aligned (Vite 8+ dev; build/preview on all supported Vite majors).
- Serve `*.exposes.js` over HTTP in dev for SSR fetch chains (builds on #723).

## Motivation

The previous React-only proxy fixed `React.lazy` / hydration quirks but was wrong for SSR (a proxy is not a real module) and blocked other frameworks. A single always-`await` path also regressed `loaded-first`, which must not call `loadRemote` until a remote export is actually read on the client.

This PR keeps one await-based path for SSR and `version-first`, and only defers on the **browser** when `shareStrategy: 'loaded-first'`.

## Examples branch (manual verification)

Validated against the SSR examples on: [**https://github.com/tonoizer/module-federation-vite-examples/tree/feat/ssr-vue-react-demos**
](https://github.com/gioboa/module-federation-vite-examples/pull/6)

(`vue-ssr`, `react-ssr`, `tanstack`, `nuxt` — build, preview, dev, Playwright; default `version-first`, no `shareStrategy` required in demos)


gioboa (changes requested):
Thanks for your help @tonoizer

src/virtualModules/virtualRemotes.ts
```diff
@@ -42,10 +51,26 @@ export function getRemoteFromId(id: string, remotes: Record<string, RemoteObject
   return remoteName ? remotes[remoteName] : undefined;
 }
 
-export function generateRemotes(id: string, command: string, enableSsrInit = false) {
+type RemoteExportStrategy = 'await-real' | 'deferred-client';
+
+function resolveRemoteExportStrategy(command: string, shareStrategy: string): RemoteExportStrategy {
+  if (command === 'build') return 'await-real';
```
gioboa: This breaks loaded-first in build. resolveRemoteExportStrategy() forces build to await-real, then lines 320-321 + 244-249 start and await loadRemote() at module evaluation. A production host with shareStrategy: 'loaded-first' can block/fail during module evaluation even when the remote is not used.

src/virtualModules/virtualRemotes.ts
```diff
@@ -14,11 +15,19 @@ const cacheRemoteMap: {
...
-  const cacheKey = `${remote}__${command}__${enableSsrInit ? 'ssr' : 'no-ssr'}`;
+
+export function getRemoteVirtualModule(
+  remote: string,
+  command: string,
+  enableSsrInit = false,
+  consumer: RemoteConsumer = 'unified'
+) {
+  const { shareStrategy } = getNormalizeModuleFederationOptions();
+  const consumerKey = command === 'build' ? 'build' : consumer;
```
gioboa: The build cache key collapses all consumers to `build`.
If client/SSR wrappers need different behavior, whichever resolves first wins.

tonoizer: The concerns should be addressed in the latest commit: 7e6e994

src/virtualModules/virtualRemotes.ts
tonoizer: The concerns should be addressed in the latest commit: 7e6e994

tonoizer:
@gioboa Do I need to change additional things or is this PR alright?

gioboa (changes requested):
Unfortunately this PR is adding top level awaits and this is problematic with Nuxt and other frameworks. Can you replace TLAs with a chain of promises? We are already doing that in the codebase.

gioboa (commented):
Thanks for your help @tonoizer 🙏 
This is a great improvement

gioboa:

Screenshot 2026-06-10 at 01 47 31

PR #743, opened 2026-05-25
feat: added named chunks for some splitting reasons

ASnow:


README.md
```diff
@@ -158,6 +158,15 @@ export default defineConfig({
       // moduleParseTimeout for large codebases where total build time may
       // exceed the fixed timeout value.
       moduleParseIdleTimeout: 10,
+      // Assign regular imported modules to chunks managed by this plugin.
+      // Federation runtime chunks keep priority and cannot be overridden.
+      // Shared dependencies are also protected from chunkMap to preserve
+      // Module Federation bootstrap order.
+      // Keys can be package names, resolved ids, or relative file suffixes.
+      chunkMap: {
```
gioboa, line 166: Thanks for your help @ASnow 
Let's use the existing Rollup/Rolldown chunk options instead of introducing a new chunkMap API.
Rolldown should use output.codeSplitting.groups; manualChunks only stays for Rollup compatibility.
The plugin should always keep federation runtime and shared modules in their required chunks, but for regular app modules it should respect the chunk names/groups already configured in the bundler.
This needs careful testing because of chunk ordering, it can break MF bootstrap. Biggest edge cases are reserved chunk names (runtimeInit, __loadShare__) and user groups accidentally catching shared deps.

gioboa (changes requested):

gioboa (commented):
Is it still valid?

gioboa (changes requested):
@ASnow thanks for opening this PR.
I'm closing this for now because it's stale and with conflicts, feel free to create a new PR with your improvements. :muscle:

PR #745, opened 2026-05-26
chore(examples): use @module-federation/enhanced package

smeng9:
Make the examples consistent with https://module-federation.io/guide/runtime/index.html#installation

> Most projects should install @module-federation/enhanced


gioboa (approved):
That’s great @smeng9 
Thank you 🙏

PR #746, opened 2026-05-27
fix: lazy-load DTS plugin when enabled

### LINKED ISSUE #744: [Bug]: Implicit dependency on `typescript`
### Describe the bug

Hi!

When importing `@module-federation/vite` you end up importing `typescript` no matter your options:

`@module-federation/vite/src/index.ts`
https://github.com/module-federation/vite/blob/dfeb4538c40b12e94ef0b1f5cf7892584ad658b8/src/index.ts#L9

`@module-federation/vite/src/plugins/pluginDts.ts`
https://github.com/module-federation/vite/blob/dfeb4538c40b12e94ef0b1f5cf7892584ad658b8/src/plugins/pluginDts.ts#L12

`@module-federation/core/packages/dts-plugin/src/core/index.ts`
https://github.com/module-federation/core/blob/91b3a02b0c15e3d61d8ac1e3fa3d2e7801ee9b5c/packages/dts-plugin/src/core/index.ts#L1

`@module-federation/core/packages/dts-plugin/src/core/configurations/remotePlugin.ts`
https://github.com/module-federation/core/blob/91b3a02b0c15e3d61d8ac1e3fa3d2e7801ee9b5c/packages/dts-plugin/src/core/configurations/remotePlugin.ts#L4

`typescript` is properly declared as a peer dependency of `@module-federation/dts-plugin` but not provided by `@module-federation/vite`:

```bash
# yarn explain peer-requirements
p5c17cd → ✘ @module-federation/vite@npm:1.15.5 [3304a] doesn't provide typescript to @module-federation/dts-plugin@npm:2.4.0 [89bce]
```

Because I don't have `typescript` in my project, the build fails with this error, despite setting `dts: false`:

```
error during build:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'typescript' imported from /Users/.../node_modules/@module-federation/dts-plugin/dist/esm/expose-rpc-B-zm7DmJ.mjs
```

---

Plugins could be imported lazily; when `dts` is set to `false`, the plugin would not even be loaded instead of featuring an early return (https://github.com/module-federation/vite/blob/dfeb4538c40b12e94ef0b1f5cf7892584ad658b8/src/plugins/pluginDts.ts#L296-L298)

### Version

@module-federation/vite@npm:1.15.5

### Reproduction

can be added if needed

### Relevant log output

```shell

```

### Validations

gioboa:
Close #744

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,6 @@ import { version as viteVersion } from 'vite';
 import addEntry from './plugins/pluginAddEntry';
 import { checkAliasConflicts } from './plugins/pluginCheckAliasConflicts';
 import pluginDevRemoteHmr, { shouldIgnoreFile } from './plugins/pluginDevRemoteHmr';
-import pluginDts from './plugins/pluginDts';
 import pluginManifest from './plugins/pluginMFManifest';
 import pluginModuleParseEnd from './plugins/pluginModuleParseEnd';
 import pluginProxyRemoteEntry from './plugins/pluginProxyRemoteEntry';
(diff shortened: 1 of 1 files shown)


gioboa (commented):

Image

PR #747, opened 2026-05-27
fix: avoid resolving aliases as packages

gioboa:


gioboa (commented):

Screenshot 2026-05-27 at 22 35 53

PR #751, opened 2026-05-28
fix: content-hash bootstrap entry asset

### LINKED ISSUE #750: [Bug]:  Bootstrap entry asset is not content-hashed
### Describe the bug

The bootstrap entry emitted by the plugin lands as `mf-entry-bootstrap-0.js` no content hash. Since this file is referenced from `index.html` as the app entry point, stale browser/CDN caches serve the old bootstrap after deploy, breaking app load.

### Reproduction

Minimal Vite app with:

```ts
federation({ name: 'host', remotes: {}, shared: { react: { singleton: true } } })
```

### Version

1.15.5

### Relevant log output

```shell
dist/index.html                  → <script src="/mf-entry-bootstrap-0.js">
dist/mf-entry-bootstrap-0.js     ← no hash
```

### Validations

cschroeter:
## Summary

- Append a short sha256 hash of the bootstrap source to the emitted bootstrap filename so it participates in the standard cache-busting flow.
- Without the hash, `mf-entry-bootstrap-0.js` is referenced from `index.html` but never invalidates in browser/CDN caches after a deploy, breaking app load until caches expire.
- Update test for the new `mf-entry-bootstrap-<index>-<hash>.js` shape.

Fixes #750

## Before

```
dist/index.html               → <script src="/mf-entry-bootstrap-0.js">
dist/mf-entry-bootstrap-0.js  ← no hash, stale on deploy
```

## After

```
dist/index.html                        → <script src="/mf-entry-bootstrap-0-3f1c9a2b.js">
dist/mf-entry-bootstrap-0-3f1c9a2b.js  ← content-hashed, invalidates on deploy
```

## Test plan

- [x] `pnpm test` (417 passing)
- [x] `pnpm typecheck`
- [x] Existing `pluginAddEntry` build-mode test updated to assert the new filename shape and that the HTML references the hashed bootstrap

### DIFF
--- a/.changeset/hash-bootstrap-entry.md
+++ b/.changeset/hash-bootstrap-entry.md
@@ -1,0 +1,14 @@
+---
+"@module-federation/vite": patch
+---
+
+fix: content-hash the bootstrap entry asset
+
+The bootstrap entry emitted by the plugin was named `mf-entry-bootstrap-<index>.js`
+with no content hash. Because `index.html` references this file as the app entry,
+browsers and CDNs kept serving the stale bootstrap after a deploy, breaking app load
+until caches expired.
+
+The fix appends a short sha256 hash of the bootstrap source to the filename
+(`mf-entry-bootstrap-<index>-<hash>.js`) so the emitted asset participates in the
+normal cache-busting flow alongside every other built asset.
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Image

gioboa (approved):
Thanks @cschroeter 
That's simply amazing 🚀

PR #757, opened 2026-05-29
fix: fix Vue sharing and assets

gioboa:


gioboa (commented):

Image

PR #758, opened 2026-05-29
fix: fix shared cache version keys

### LINKED ISSUE #755: [Bug]: singleton: false does not enable side-by-side loading of multiple versions
### Summary

  When two remotes declare the same `shared` dependency at incompatible major versions with `singleton: false`, the official [`shared` documentation](https://module-federation.io/configure/shared) states:

  > "If singleton mode is not enabled, and the shared dependencies between the remote application and the host application have different versions, each will load their own dependencies."

  In practice with `@module-federation/vite@1.15.2`, both remotes still resolve to a **single shared instance** — whichever remote evaluates first wins the slot, and the second remote silently runs against a version it wasn't built for. This is observably identical to `singleton: true` behavior.

  ### Reproduction

  - Host with no shared declaration for `pkg`.
  - Remote A `package.json`: `"zustand": "^4.5.5"` → declares `shared: { zustand: { singleton: false } }`.
  - Remote B `package.json`: `"zustand": "^5.0.2"` → declares `shared: { zustand: { singleton: false } }`.

  In a single page session, load A then B (or B then A). The second remote does not bundle its own copy and does not get its own instance — it reads from the same share slot the first remote wrote.

  Concretely, with the order B → A, A's `import create from 'zustand'` (default import, exists in v4, removed in v5) resolves to v5's namespace object instead of a function. Calling `create(...)` at module top-level throws `TypeError: ... is not a function`. (Reversing the order, the silent fallback works only because v4 still has `create` as a named export — so the `singleton: false` symptom is asymmetric.)

  ### Expected behavior

  Per the docs, each remote should load and use its own version when versions are incompatible. The share scope should be able to hold both `zustand@4.5.7` and `zustand@5.0.13` simultaneously, with each remote resolving to its matching version.

  ### Actual behavior

  The chunk wrapper generated by the plugin keys the share scope by **package name only**, with no version dimension. From `node_modules/@module-federation/vite/lib/index.cjs` around L1276-1289 (`writeLoadShareModule`):

  ```js
  loadShareCacheMap[pkg].writeSync(`
    ${prebuildImportLine}
    ${devDynamicImportLine}
    ${importLine}
    let exportModule = __mfModuleCache.share[${pkg}]
      exportModule = __mfLocalShare;
      __mfModuleCache.share[${pkg}] = exportModule;
    }
    ${exportLine}
  `, true);
  ```

  `__mfModuleCache.share[pkg]` is a single slot per package name. There is no `[version]` (or `[requiredVersion]`) sub-key, so two versions cannot coexist regardless of `singleton` / `requiredVersion` / `strictVersion`. This diverges from the webpack `ModuleFederationPlugin`, whose share scope is
  `__webpack_require__.S[scope][pkg][version]` — a three-level structure that does support side-by-side versions.

  I confirmed at runtime by probing the share scope in a production build after both remotes loaded:

  ```js
  globalThis.__mf_module_cache__.share.zustand
  // → single object holding only one version's exports
  ```

  ### Environment

  - `@module-federation/vite`: 1.15.2
  - `@module-federation/runtime`: 2.4.0
  - Vite: 8.0.11
  - Build mode: production (`vite build`)
  - Browser: Chromium 142

  ### Questions

  1. Is this an intentional simplification of the share-scope semantics for the Vite plugin, or a known gap from webpack parity?
  2. Is there an issue or PR I missed tracking version-keyed share scopes for the Vite plugin?
  3. If multi-version isn't on the roadmap, would you accept a Pull Request to fix it?

### Version

1.15.2

### Reproduction

Example above

### Relevant log output

```shell

```

### Validations

gioboa:
Close #755

### DIFF
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -98,7 +98,7 @@ describe('shared dependencies', () => {
     expect(allCode).toContain('const versions =');
     expect(allCode).toContain('[pkg]');
     expect(allCode).toContain('const provider = versions && versions[Object.keys(versions)[0]]');
-    expect(allCode).toContain('__mfModuleCache.share[pkg]');
+    expect(allCode).toContain('__mfModuleCache.share[cacheKey]');
     expect(allCode).not.toContain('initRes.loadShare(pkg');
     expect(loadShare!.code).toContain('initPromise.then');
     expect(loadShare!.code).toContain('__mfModuleCache.share["pathe"]');
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -24,6 +24,11 @@ vi.mock('fs', async (importOriginal) => {
 });
 
 vi.mock('../../utils/packageUtils', () => ({
+  getSharedCacheKey: (
+    pkg: string,
+    shareItem: { version?: string; shareConfig: { singleton?: boolean } }
+  ) =>
+    shareItem.shareConfig.singleton || !shareItem.version ? pkg : `${pkg}@${shareItem.version}`,
   hasPackageDependency: hasPackageDependencyMock,
   getInstalledPackageEntry: vi.fn(() => undefined),
   getInstalledPackageJson: vi.fn((pkg: string) => {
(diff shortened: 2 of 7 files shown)


gioboa (commented):

Image

PR #759, opened 2026-05-29
fix: fix duplicate named remote helper

### LINKED ISSUE #756: [Bug]: Multiple named federated imports breaks Vue HMR setup
### Describe the bug

When setting:
```
dev: {
    remoteHmr: true,
},
```

v1.16.1’s named-remote-import transform emits `__mfCreateNamedRemoteProxy` functions more than once in the same module when multiple named federated imports exist, like:
```
import { useAuth } from 'common/auth';
import { useShellStore } from 'shell/shellStore';
```

 which causes:

`SyntaxError: Identifier '__mfCreateNamedRemoteProxy' has already been declared`

### Version

8.0.14

### Reproduction

Example above

### Relevant log output

```shell

```

### Validations

gioboa:
Close #756


gioboa (commented):

Image

PR #761, opened 2026-05-29
chore: update @vitejs/plugin-react in examples and migrate to react compiler

smeng9:
Hi @gioboa 

In latest version of React, it encourages the adoption of React Compiler https://react.dev/learn/react-compiler/introduction .

I have updated the example to support react compiler.

However I have one question regarding the shared dependency cache key, why is it different from the @module-federation/enhanced ones? This took almost an hour to figure out why the code is failing without the CACHE_KEY patch. For normal users they shouldn't need this setup for it to work.

### DIFF
--- a/examples/vite-runtime-register/vite-remote/package.json
+++ b/examples/vite-runtime-register/vite-remote/package.json
@@ -14,7 +14,9 @@
     "react-dom": "19.2.4"
   },
   "devDependencies": {
-    "@vitejs/plugin-react": "^5.1.2",
+    "@rolldown/plugin-babel": "^0.2.3",
+    "@vitejs/plugin-react": "^6.0.2",
+    "babel-plugin-react-compiler": "^1.0.0",
     "vite": "^8.0.10"
   }
 }
--- a/examples/vite-runtime-register/vite-host/package.json
+++ b/examples/vite-runtime-register/vite-host/package.json
@@ -14,7 +14,9 @@
     "react-dom": "19.2.4"
   },
   "devDependencies": {
-    "@vitejs/plugin-react": "^5.1.2",
+    "@rolldown/plugin-babel": "^0.2.3",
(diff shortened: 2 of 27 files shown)


gioboa (commented):
Thanks for your help @smeng9 

> why is it different from the @module-federation/enhanced ones?

Can you elaborate more about this please?

smeng9:
> Can you elaborate more about this please?

If I don't perform the following then the host app will crash with an exception `called useMemoCache is not defined` when rendering the remote app. I am wondering why this step is needed if we already declare all the sharing in both host and remote.

```
if (globalThis[CACHE_KEY].share["react"] === undefined) {
  globalThis[CACHE_KEY].share["react"] = React;
}
```

examples/vite-runtime-register/vite-remote/package.json
smeng9, line 12: Hi @gioboa

I also noticed I need to set this value to 1.16.2. If I keep using the workspace:* in this merge request, the example will fail even after I build the workspace. Any idea why?

smeng9:
> Thanks for your help @smeng9
> 
> > why is it different from the @module-federation/enhanced ones?
> 
> Can you elaborate more about this please?

The previous message regarding previous commit is outdated. I have updated the @module-federation/vite package so it no longer needs custom injection of react instance into __mf_module_cache__.

Can you retry the playwright test? I am not sure why it is failing on CI.

gioboa (commented):
multi-example script is failing I guess

smeng9:
Hi @gioboa 

I have applied a fix in the tests, so it should now support both react 18 without react-compiler and react 19 with react-compiler. Can you re-trigger the ci jobs again?

gioboa (commented):

Image

gioboa (approved):
Thanks @smeng9 
it looks great to me 👏

Issue #762, opened 2026-06-01
[Bug]: shared singleton importing another shared singleton deadlocks host bootstrap (regressed in 1.15.0)


falldowngoboone:
### Describe the bug

When one shared `singleton` module statically imports **another** shared `singleton` module, the host hangs forever during bootstrap. The host loads `mf-entry-bootstrap` → `hostInit` → both `loadShare` virtual modules, then deadlocks: `loadRemote` is never reached, `remoteEntry.js` is never requested, and nothing is logged. Silent hang.

Regressed in **1.15.0**; last good version is **1.14.5**.

### Reproduction

Minimal standalone repo (~19 files, no frameworks/services):
**https://github.com/falldowngoboone/mf-vite-shared-singleton-deadlock**

```bash
npm install
npx playwright install chromium webkit
npm run build
npm test
```

The setup:
- `@repro/i18n` — a shared `singleton` (trivial synchronous module).
- `@repro/api` — a shared `singleton` that does `import { i18n } from '@repro/i18n'`.
- host declares both `singleton`, consumes `@repro/api` at bootstrap, then `loadRemote('remote/Widget')`.

The built host's network trace stops dead after the shared-init chunks:

```
200  /mf-entry-bootstrap-0-*.js
200  /assets/hostInit-*.js
200  /assets/…host__loadShare__…repro_api__loadShare__.mjs
200  /assets/…host__loadShare__…repro_i18n__loadShare__.mjs
     ⟶ (hang; no remoteEntry.js request, no error)
```

### Minimal trigger (bisected by removing one trait at a time)

| api imports i18n | i18n is shared `singleton` | result |
|---|---|---|
| ✓ | ✓ | ❌ deadlock |
| ✗ | ✓ | ✅ passes |
| ✓ | ✗ (regular dep) | ✅ passes |

The deadlock requires **both** the import relationship *and* both modules being shared singletons. Top-level async init in the imported module is **not** required — a trivial synchronous module deadlocks identically.

### Version bisect

| `@module-federation/vite` | `vite` | result |
|---|---|---|
| 1.16.2 | 8.0.14 | ❌ deadlock |
| 1.14.5 | 8.0.8 | ✅ passes |

(1.14.x requires vite ≤ 8.0.8 — 8.0.9+ ships Rolldown ≥ rc.16, which rejects 1.14.x's `loadShare` TLA with `[REQUIRE_TLA]`. 1.15.0 removed that TLA, which is the rewrite that introduced this deadlock.)

[...]

PR #765, opened 2026-06-01
perf: materialize each shared source's loadShare artifacts at most once

falldowngoboone:
## Summary

`proxyPreBuildShared:resolve-shared-loadShare` is a `resolveId` hook (`enforce: 'pre'`), so it fires **once per importing module** for the whole build. On every call that matches a shared key it regenerates all of:

- `writeLoadShareModule` (the `__loadShare__` virtual module),
- `writePreBuildLibPath` (the `__prebuild__` fallback),
- `writeLocalSharedImportMap` (the **entire** local-shared import map), and
- `refreshHostAutoInit`,

and then redirects via `this.resolve`. These artifacts are a **pure function of the shared source**, so a singleton imported by N modules rewrites all of it N times — an `O(modules × shared)` cost that dominates builds for shared-heavy hosts.

This PR memoizes the heavy writes per shared source with a `Set<string>` on the plugin closure: the writes run **at most once per source**, while the import redirect (`getLoadShareModulePath` + `this.resolve`) still runs on **every** resolution, so resolution behavior is unchanged. Net change is 15 lines in one file.

## Why this is more than a micro-optimization

The redundant work isn't cheap. Each materialization calls `getPackageNamedExports()` (in `virtualShared_preBuild.ts`), which is **uncached** and `require()`s the package to introspect its exports — and it's called **twice** per materialization (once via `writeLoadShareModule`, once via `writePreBuildLibPath`). On top of that, `writeLocalSharedImportMap` regenerates by sorting and stringifying **every** used share on each call. Multiplying that by every importing module is where the time goes.

This is a regression introduced by the 1.15 rewrite. Dropping `customResolver` was the right call (and Vite 9 requires it anyway); this just closes the one hot path that replacement opened.

## Measured impact

Real host with ~26 shared singletons, cold build, 3-run average:

| Config | Build |
|---|---|
| pre-1.15 (`customResolver`) | ~4.5s |
| current `main` | ~8.0s |
| this PR | ~2.0s |

`PLUGIN_TIMINGS` on `main` attributed ~30% to `resolve-shared-loadShare`, ~28% to `module-federation-virtual-modules`, ~15% to `resolve-prebuild`. With this change the "spent significant time in plugins" warning disappears entirely. (The wall-clock numbers are from a private host, but the structural argument above is reproducible from the source alone — the cost scales with singleton count × importing modules.)

## Why it's safe

- **Output is byte-identical.** Same chunk count, same manifest — verified against `main`. The change only removes redundant regeneration; it never changes *what* gets written.
- **The skipped writes are idempotent.** `writeLoadShareModule`/`writePreBuildLibPath` are keyed by the same source and produce identical content on repeat; `addUsedShares` is a `Set.add`; `writeLocalSharedImportMap`/`refreshHostAutoInit` regenerate purely from the accumulated `usedShares` set, so re-running them when nothing new was added yields identical output. (`writeLocalSharedImportMap` already had a content-diff guard, so it never re-invalidated redundantly — this just also skips paying the regeneration cost.)
- **The memo key maps to a stable payload.** The `shareSource` is a pure function of `(source, key)`, and the share key is itself derived deterministically from the source, so a given `shareSource` always pairs with the same `ShareItem` config — there's no path where one source memoizes another's config.

## Anticipated questions

**Why a `Set` on the closure instead of memoizing inside `writeLoadShareModule`/`writePreBuildLibPath`?**
Those two could memoize their own content by key, but `writeLocalSharedImportMap` and `refreshHostAutoInit` are *aggregate* regenerators (they rebuild from all used shares), so they can't be memoized by a single source key at that level. Gating the whole block once per source is the change that actually removes the aggregate-regeneration cost, and it keeps the fix to one file. Pushing per-call memoization down into the writers as well would be a fine follow-up but isn't required for the win.

**`configResolved` already materializes every top-level shared key — could the hook skip those entirely (1→0 instead of N→1)?**
Yes, in principle: seeding the set with the keys `configResolved` writes would let the hook do zero work for plain top-level singletons and only materialize genuinely-new subpath sources. I left that out deliberately — it couples two currently-independent hooks and has to mirror `configResolved`'s skip predicate exactly (the `key.endsWith('/')` and direct-React cases). Happy to add it if you'd prefer, but the minimal change already captures the bulk of the regression.

**Does this affect dev/HMR?**
No. Shared config isn't hot-editable, and the existing `VirtualModule` object caches already assume that stability, so memoizing per source is consistent with current behavior. The set lives on the plugin closure, so it's per-build and cleared on each new build.

**Relationship to the deadlock work in #762?**
Orthogonal. This was found while investigating #762 and lives in the same area (`virtualShared_preBuild` resolve path), but it touches a different file and a different phase (resolution vs. chunk output) and produces identical output — there's no ordering dependency between them.

## Testing

- `pnpm typecheck` clean
- `pnpm test` — 425 passing
- `pnpm test:integration` — 34 passing

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -8,10 +8,24 @@ import type {
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
 
-const { hasPackageDependencyMock, existsSyncMock, readFileSyncMock } = vi.hoisted(() => ({
+const {
+  hasPackageDependencyMock,
+  existsSyncMock,
+  readFileSyncMock,
+  addUsedSharesMock,
+  refreshHostAutoInitMock,
+  writeLoadShareModuleMock,
+  writeLocalSharedImportMapMock,
+  writePreBuildLibPathMock,
+} = vi.hoisted(() => ({
   hasPackageDependencyMock: vi.fn<(pkg: string) => boolean>(),
   existsSyncMock: vi.fn<(path: string) => boolean>(() => false),
   readFileSyncMock: vi.fn<(path: string) => string>(() => '{}'),
+  addUsedSharesMock: vi.fn<(pkg: string) => void>(),
+  refreshHostAutoInitMock: vi.fn<() => void>(),
+  writeLoadShareModuleMock: vi.fn(),
+  writeLocalSharedImportMapMock: vi.fn<() => void>(),
+  writePreBuildLibPathMock: vi.fn(),
 }));
 
 vi.mock('fs', async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


gioboa (commented):
That looks great 👍 @falldowngoboone 
I will look at this soon

gioboa (commented):

Image

gioboa (approved):
Thanks @falldowngoboone for your amazing help 🥳

PR #766, opened 2026-06-01
fix: isolate __vite_preload helper into its own chunk to break shared-singleton loadShare deadlock

### LINKED ISSUE #762: [Bug]: shared singleton importing another shared singleton deadlocks host bootstrap (regressed in 1.15.0)

falldowngoboone:
Fixes #762.

## The bug

When one shared `singleton` module statically imports **another** shared `singleton`, the host hangs forever during bootstrap: it loads `mf-entry-bootstrap` → `hostInit` → both `loadShare` virtual modules, then deadlocks — `loadRemote` is never reached, `remoteEntry.js` is never requested, nothing is logged. A silent hang.

Regressed in **1.15.0** (last good: 1.14.5). The 1.15.0 `loadShare` rewrite (#638 / #645) removed the `customResolver` and introduced the top-level-`await` form of the local fallback, which is what closes the cycle described below.

## Root cause

Rolldown injects the `__vite_preload` helper as a special runtime module. Left to automatic chunking, it **hoists the helper into whichever shared-singleton `loadShare` chunk first uses it**. Each workspace-source singleton's `loadShare` wrapper does a top-level `await import(localSource)` (the `usesLazyLocalFallback` path), and that singleton's source statically imports the other singleton. The hoisted helper adds a static import from one `loadShare` chunk to the sibling, closing a ring of three top-level-await modules:

```
i18n loadShare chunk ──(static import, only to get the __vite_preload helper)──▶ api loadShare chunk
        ▲                                                                              │
        │                                                          (TLA: await import(api's own source))
        └──────────────(static import)──────── api source ◀────────────────────────────┘
                                          (api's source statically imports @byte/i18n)
```

Three top-level-await modules in a static-import cycle → none of their evaluation promises can resolve → deadlock. The lethal ingredient is **TLA inside the cycle** (plain cyclic imports are fine); the rewrite added the TLA and the hoisted-helper static edge closed the ring.

**Precondition (why it's rarely reported):** both modules must be shared `singleton` **and** consumed as workspace source — `exports` pointing at TS source (`"exports": { "import": "./src/index.ts" }`), which routes them through the `usesLazyLocalFallback` TLA path. Projects that share built npm packages take the non-TLA path and never hit this. This is a legitimate, common monorepo setup, not a misconfiguration.

## The fix

Isolate the `__vite_preload` helper into its **own** dependency-free, TLA-free chunk so no `loadShare` chunk statically imports it from a sibling — the ring never closes. Implemented by migrating the plugin's chunking from `manualChunks` to `codeSplitting` (all in `src/index.ts`):

- A high-priority `test`-based group pulls the preload helper into its own `vite-preload-helper` chunk.
- A dynamic `name(id)` group reproduces the existing `runtimeInit` / `loadShare` isolation.
- `ensureCodeSplitting` gains a guard so it strips user-supplied `codeSplitting.groups` (as before) but never the plugin's own groups.

### Why not `manualChunks`?

`manualChunks` **cannot** relocate the preload helper — Rolldown treats it as a special runtime module whose placement `manualChunks` does not govern (verified on rolldown 1.0.2: returning a name for the helper from `manualChunks` produced byte-identical output). Only `codeSplitting` / `advancedChunks` can move it. Within `codeSplitting`, the helper is captured by a `test` group but is **not** passed to the dynamic `name()` function, so the helper must be its own `test` group.

## Validation

- `pnpm typecheck` clean; `pnpm test` **425 passed** (29 files); `pnpm test:integration` **34 passed** (9 files).
- **Minimal repro** (https://github.com/falldowngoboone/mf-vite-shared-singleton-deadlock): deadlocks deterministically on the pre-fix plugin and passes on the fixed plugin, on both Chromium and WebKit. Same repo, same machine, only the plugin version swapped:

  | Plugin | isolated `vite-preload-helper` chunk? | Playwright (Chromium + WebKit) |
  |---|---|---|
  | pre-fix | no | 2 failed (deadlock, 30s timeout both browsers) |
  | this fix | yes | 2 passed |

- **Real ~26-singleton host:** built on this fix and inspected the actual chunks — `vite-preload-helper` chunk created (dependency-free, no TLA), all cross-`loadShare` cycle edges removed, all 37 `loadShare` chunks preserved; e2e (chromium + webkit) green.

## Related

- #369 — prior precedent: isolating the preload helper to prevent a shared-module deadlock (Vite-7 era). This is the Rolldown-era re-introduction of that class of bug.
- #638 / #645 — the 1.15.0 `loadShare` rewrite that introduced the regression.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -348,8 +348,9 @@ describe('module-federation-esm-shims', () => {
       : undefined;
   };
 
-  it('removes codeSplitting false, warns once, and installs federation groups', () => {
+  it('removes codeSplitting false, warns once, and installs bundler-appropriate isolation', () => {
     const plugin = getEsmShimsPlugin();
+    const runtimeInitId = virtualRuntimeInitStatus.getImportId();
     const config: any = {
       build: {
         rollupOptions: { output: { codeSplitting: false } },
@@ -359,9 +360,16 @@ describe('module-federation-esm-shims', () => {
 
     runConfig(plugin, {} as ConfigPluginContext, config, { command: 'build', mode: 'test' });
 
-    // `codeSplitting: false` is replaced with the federation groups.
-    expect(Array.isArray(config.build.rollupOptions.output.codeSplitting.groups)).toBe(true);
+    // Rollup (Vite 5–7) gets `manualChunks` and never `codeSplitting` (which Rollup
+    // rejects as an unknown output option).
+    expect(config.build.rollupOptions.output.codeSplitting).toBeUndefined();
+    expect(typeof config.build.rollupOptions.output.manualChunks).toBe('function');
+    expect(config.build.rollupOptions.output.manualChunks(`/virtual/${runtimeInitId}`)).toBe(
+      'runtimeInit'
+    );
+    // Rolldown (Vite 8+) gets the `codeSplitting` groups and no `manualChunks`.
     expect(Array.isArray(config.build.rolldownOptions.output.codeSplitting.groups)).toBe(true);
+    expect(config.build.rolldownOptions.output.manualChunks).toBeUndefined();
     expect(mfWarn).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks @falldowngoboone 
I will look at this soon

falldowngoboone:
I'll look into those failing integration tests.

gioboa (commented):

Image

gioboa (approved):
Thanks @falldowngoboone 
it looks great to me 🚀

PR #771, opened 2026-06-02
fix: walk static-import graph when collecting expose async assets

### LINKED ISSUE #770: [Bug]: manifest exposes[*].assets.js.async is empty when lazy imports live in transitively-reachable chunks
### Describe the bug

`pluginMFManifest` emits an empty `exposes[*].assets.js.async` array for any exposed module whose lazy `import()` calls live in a **sibling chunk** reached via static import from the matched expose entry, rather than directly in the matched chunk itself.

The cause is in [`src/utils/cssModuleHelpers.ts`](https://github.com/module-federation/vite/blob/main/src/utils/cssModuleHelpers.ts#L188-L196): `processModuleAssets` iterates the bundle, matches a chunk by checking whether one of its `modules` resolves to `mfOptions.exposes[].import`, then inspects **only that chunk's own** `dynamicImports`. When Rollup splits the expose entry into multiple sibling chunks reachable via static imports, only the chunk whose module path matches the expose's `import` value is visited — sibling chunks (carrying the dynamic-import edges) are never visited under that expose's `matchKey`, so their entries never reach the manifest.

This blocks `@module-federation/runtime`'s `preloadRemote({ resourceCategory: 'all' })`. The runtime relies on the manifest's `async` array to know what to preload; when it's empty, lazy chunks fetch on demand at navigation time instead of in parallel with the sync burst.

**Expected `mf-manifest.json`:**

```json
[{
  "id": "demo:router",
  "name": "router",
  "assets": {
    "js": {
      "sync": ["assets/router-<hash>.js"],
      "async": [
        "assets/admin-<hash>.js",
        "assets/hostnames-<hash>.js",
        "..."
      ]
    },
    "css": { "sync": [], "async": [] }
  },
  "path": "./router"
}]
```

**Actual:**

```json
[{
  "id": "demo:router",
  "name": "router",
  "assets": {
    "js": {
      "sync": ["assets/router-<hash>.js"],
      "async": []
    },
    "css": { "sync": [], "async": [] }
  },
  "path": "./router"
}]
```

…despite six lazy page chunks existing on disk in `dist/assets/`.

**Chunk graph:**

```
assets/router-<hash>.js   ← matched as ./router expose (re-exports `routes`)
  ↓ chunk.imports (STATIC)
assets/routes-<hash>.js   ← contains the lazy(() => import(...)) calls
  ↓ chunk.dynamicImports (DYNAMIC)
assets/admin-<hash>.js, hostnames-<hash>.js, ...
```

`processModuleAssets` matches the chunk that contains `src/router.tsx` (the expose's `import` value), then iterates THAT chunk's `dynamicImports` — which is `[]` because the dynamic edges live in the sibling chunk. The sibling chunk's `dynamicImports` are never associated with the `./router` expose because the matcher doesn't walk the static-import graph.

This shape (small expose entry that re-exports from a heavier sibling routes module) is the natural pattern for federated remotes that support both standalone and federated modes — the routes module has multiple consumers, so Rollup keeps it as its own chunk.

**Fix sketch:** walk `chunk.imports` transitively from the matched chunk and accumulate each visited chunk's `dynamicImports` as async assets under the expose's `matchKey`. ~20-line BFS in `cssModuleHelpers.ts`. Strictly additive — only adds entries to `async` arrays, never changes sync arrays. PR with the fix + a regression test is up at #771.

### Version

1.14.5

### Reproduction

https://github.com/DanBeckDev/mfevite-async-chunk-repro

### Relevant log output

```shell
# Built dist/mf-manifest.json (relevant slice):
{
  "exposes": [{
    "id": "demo:router",
    "name": "router",
    "assets": {
      "js": { "sync": ["assets/router-<hash>.js"], "async": [] },
      "css": { "sync": [], "async": [] }
    },
    "path": "./router"
  }]
}

# But six lazy page chunks exist in dist/assets/:
$ ls dist/assets/{admin,hostnames,prefixes,api-keys,settings,profile}-*.js
dist/assets/admin-B34pVfmU.js
dist/assets/api-keys-5NwVNRb2.js
dist/assets/hostnames-614vIVm6.js
dist/assets/prefixes-DPTotWX1.js
dist/assets/profile-4FGXLcwa.js
dist/assets/settings-C7cMGFaP.js

# And the sibling routes chunk DOES have the dynamic-import edges
# (which never make it into the manifest):
$ grep -oE "(admin|hostnames|prefixes|api-keys|settings|profile)-[A-Za-z0-9_-]{8}" dist/assets/routes-*.js | sort -u
admin-B34pVfmU
api-keys-5NwVNRb2
hostnames-614vIVm6
prefixes-DPTotWX1
profile-4FGXLcwa
settings-C7cMGFaP
```

### Validations

DanBeckDev:
Fixes #770.

## Problem

`processModuleAssets` (`src/utils/cssModuleHelpers.ts`) was only inspecting the **matched chunk's direct** `dynamicImports`. When Rollup splits the expose entry into multiple sibling chunks reachable via static imports — a common shape when the expose entry re-exports from a larger routes/config module that owns the `lazy()` declarations — sibling chunks carry the dynamic-import edges, but the matcher never visits them under the expose's `matchKey`. Result: `mf-manifest.json` emits empty `exposes[*].assets.js.async` arrays despite Vite producing real async chunks in `dist/assets/`.

Downstream: `@module-federation/runtime`'s `preloadRemote({ resourceCategory: 'all' })` has nothing to preload — lazy chunks fetch on demand at navigation time instead of in parallel with the sync burst.

Full repro: https://github.com/DanBeckDev/mfevite-async-chunk-repro (linked from the issue).

## Change

Switch the one-level `dynamicImports` scan to a BFS from the matched chunk through `chunk.imports` (static-import graph), accumulating each visited chunk's `dynamicImports` as async assets under the expose's `matchKey`. `visited: Set<string>` guards against cycles.

- Strictly additive — only adds entries to `async` arrays
- Sync arrays unchanged
- No behavior change for the single-chunk case (the existing `'processes module assets'` test still passes verbatim)

```diff
-      // Handle dynamic imports
-      if (fileData.dynamicImports) {
-        for (const dynamicImport of fileData.dynamicImports) {
-          const importData = bundle[dynamicImport];
-          if (!importData) continue;
+      // Handle dynamic imports reachable through the static-import graph
+      const visited = new Set<string>();
+      const queue: string[] = [fileName];
+      while (queue.length) {
+        const cur = queue.shift()!;
+        if (visited.has(cur)) continue;
+        visited.add(cur);
+        const chunk = bundle[cur];
+        if (!chunk || chunk.type !== 'chunk') continue;

-          const isCss = isCSSFile(dynamicImport);
-          trackAsset(filesMap, matchKey, dynamicImport, true, isCss ? 'css' : 'js');
+        if (chunk.dynamicImports) {
+          for (const dynamicImport of chunk.dynamicImports) {
+            if (!bundle[dynamicImport]) continue;
+            const isCss = isCSSFile(dynamicImport);
+            trackAsset(filesMap, matchKey, dynamicImport, true, isCss ? 'css' : 'js');
+          }
+        }
+        if (chunk.imports) {
+          for (const imp of chunk.imports) queue.push(imp);
         }
       }
```

## Test

New regression test in `src/utils/__tests__/cssModuleHelpers.test.ts` (`'tracks dynamic imports reachable through the static-import graph'`) covers the transitive case: matched chunk with `imports: ['sibling.js']`, sibling chunk with `dynamicImports: ['async.js']`. Without the patch the test fails with `async: []`; with the patch it passes with `async: ['async.js']`.

Verified locally:

- Full suite: 426/426 pass (425 existing + 1 new)
- Same suite with patch reverted (test alone, not source): the new test fails as expected, all 425 existing pass — confirms no regression in existing coverage and the new test catches the bug
- Empirical end-to-end: compiled `.mjs` patched in the repro's `node_modules`; manifest's `async[]` goes from `[]` to listing all 6 lazy page chunks; chunk hashes unchanged

### DIFF
--- a/src/utils/cssModuleHelpers.ts
+++ b/src/utils/cssModuleHelpers.ts
@@ -184,14 +184,25 @@ export const processModuleAssets = (
         }
       }
 
-      // Handle dynamic imports
-      if (fileData.dynamicImports) {
-        for (const dynamicImport of fileData.dynamicImports) {
-          const importData = bundle[dynamicImport];
-          if (!importData) continue;
+      // Handle dynamic imports reachable through the static-import graph
+      const visited = new Set<string>();
+      const queue: string[] = [fileName];
+      while (queue.length) {
+        const cur = queue.shift()!;
+        if (visited.has(cur)) continue;
+        visited.add(cur);
+        const chunk = bundle[cur];
+        if (!chunk || chunk.type !== 'chunk') continue;
 
-          const isCss = isCSSFile(dynamicImport);
-          trackAsset(filesMap, matchKey, dynamicImport, true, isCss ? 'css' : 'js');
+        if (chunk.dynamicImports) {
+          for (const dynamicImport of chunk.dynamicImports) {
+            if (!bundle[dynamicImport]) continue;
+            const isCss = isCSSFile(dynamicImport);
+            trackAsset(filesMap, matchKey, dynamicImport, true, isCss ? 'css' : 'js');
+          }
+        }
+        if (chunk.imports) {
+          for (const imp of chunk.imports) queue.push(imp);
         }
       }
     }
--- a/src/utils/cssModuleHelpers.ts
+++ b/src/utils/cssModuleHelpers.ts
@@ -184,11 +184,11 @@ export const processModuleAssets = (
         }
       }
 
-      // Handle dynamic imports reachable through the static-import graph
+      // Handle dynamic imports from static chunks reachable from the matched expose chunk
       const visited = new Set<string>();
       const queue: string[] = [fileName];
-      while (queue.length) {
-        const cur = queue.shift()!;
+      for (let queueIndex = 0; queueIndex < queue.length; queueIndex++) {
+        const cur = queue[queueIndex];
         if (visited.has(cur)) continue;
         visited.add(cur);
         const chunk = bundle[cur];
--- a/src/utils/__tests__/cssModuleHelpers.test.ts
+++ b/src/utils/__tests__/cssModuleHelpers.test.ts
@@ -184,9 +184,16 @@ describe('cssModuleHelpers', () => {
         },
         'sibling.js': {
           ...createChunk('sibling.js'),
+          imports: ['nested.js'],
           dynamicImports: ['async.js'],
         },
+        'nested.js': {
+          ...createChunk('nested.js'),
+          imports: ['entry.js'],
+          dynamicImports: ['nested-async.js'],
+        },
         'async.js': createChunk('async.js'),
+        'nested-async.js': createChunk('nested-async.js'),
       } satisfies Record<string, OutputBundleItem>;
 
       const filesMap = {};
@@ -196,7 +203,7 @@ describe('cssModuleHelpers', () => {
 
       expect(filesMap).toEqual({
         module1: {
-          js: { sync: ['entry.js'], async: ['async.js'] },
+          js: { sync: ['entry.js'], async: ['async.js', 'nested-async.js'] },
           css: { sync: [], async: [] },
         },
       });
--- a/src/utils/__tests__/cssModuleHelpers.test.ts
+++ b/src/utils/__tests__/cssModuleHelpers.test.ts
@@ -173,6 +173,35 @@ describe('cssModuleHelpers', () => {
       });
     });
 
+    it('tracks dynamic imports reachable through the static-import graph', () => {
+      const bundle = {
+        'entry.js': {
+          ...createChunk('entry.js'),
+          modules: {
+            module1: createRenderedModule(),
+          },
+          imports: ['sibling.js'],
+        },
+        'sibling.js': {
+          ...createChunk('sibling.js'),
+          dynamicImports: ['async.js'],
+        },
+        'async.js': createChunk('async.js'),
+      } satisfies Record<string, OutputBundleItem>;
+
+      const filesMap = {};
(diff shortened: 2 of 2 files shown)


src/utils/cssModuleHelpers.ts
```diff
@@ -184,14 +184,25 @@ export const processModuleAssets = (
...
 
-      // Handle dynamic imports
-      if (fileData.dynamicImports) {
-        for (const dynamicImport of fileData.dynamicImports) {
-          const importData = bundle[dynamicImport];
-          if (!importData) continue;
+      // Handle dynamic imports from static chunks reachable from the matched expose chunk
+      const visited = new Set<string>();
+      const queue: string[] = [fileName];
+      for (let queueIndex = 0; queueIndex < queue.length; queueIndex++) {
```
gioboa, line 190: Thanks @DanBeckDev I changed this with a for loop.
```
Local Node microbench, 100k items:
  - shift(): ~744 ms
  - index loop: ~0.048 ms
  - index loop faster by ~99.99%
  - ~15,590x speedup

Reason: Array.shift() reindexes array each pop, effectively O(n²). Index loop is O(n).
For small bundles, real-world impact tiny, but index loop is objectively better.
```

gioboa (commented):

Screenshot 2026-06-03 at 15 03 19

gioboa (approved):
Thanks for your help @DanBeckDev 
Nice work 👏

PR #772, opened 2026-06-02
fix: detect shared named exports from configured import source

majashryer9:
## Summary

Fix named export detection for shared modules when `shareConfig.import` is explicitly configured.

Shared wrapper generation already uses `shareConfig.import` as the concrete source when provided, but named export detection previously still resolved exports from the shared package key. For explicit import paths that are not standalone package entries, this could cause named exports to be missed and the generated wrapper to fall back to a less precise export shape.

This change resolves and parses the configured import source first, then falls back to the existing package-based detection for backward compatibility.

## Problem

Given a shared config like:

```ts
shared: {
  '@repo/custom-shared-source': {
    import: '/repo/packages/custom-shared-source',
    singleton: true,
  },
}
```
the generated shared wrapper imports from the configured source:
```ts
'/repo/packages/custom-shared-source'
```
However, named export detection previously inspected only the shared key:
```ts
'@repo/custom-shared-source'
```
If that key is not a resolvable package entry, named export detection returns no exports even though the configured import source has valid named exports.

## Changes

- Added resolution for string `shareConfig.import` values during named export detection.
- Added file-like resolution for configured import paths, including extensionless paths and directory `index.*` entries.
- Reused the existing ESM export parsing logic for resolved configured import files.
- Preserved existing package-based named export detection as the fallback path.
- Left `import: false` behavior unchanged, since that mode has no local implementation source to inspect.
- Added unit coverage for:
  - named exports from explicit shareConfig.import sources
  - prebuild wrapper generation
  - loadShare wrapper generation
  - fallback to package-based detection
  - existing package-based behavior

## Why This Is Backward Compatible

- When `shareConfig.import` is not provided, behavior is unchanged.
- When `shareConfig.import` is provided but cannot be resolved or parsed, the implementation falls back to the existing package-based named export detection.
- `import: false` continues to use the existing package-based dev dependency detection behavior.

### DIFF
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -123,16 +123,19 @@ function resolveConfiguredImportPath(importSource: string): string | undefined {
     return resolveFileLikeModule(path.resolve(projectRoot, importSource));
   }
 
+  const esmEntry = getInstalledPackageEntry(importSource, {
+    conditions: ['browser', 'import', 'module', 'default'],
+    resolveSubpathWithRequire: false,
+  });
+  if (esmEntry) return esmEntry;
+
   try {
     const projectRequire = createRequire(
       new URL(`file://${path.join(projectRoot, 'package.json')}`)
     );
     return projectRequire.resolve(importSource);
   } catch {
-    return getInstalledPackageEntry(importSource, {
-      conditions: ['browser', 'import', 'module', 'default'],
-      resolveSubpathWithRequire: false,
-    });
+    return undefined;
   }
 }
 
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -83,11 +83,9 @@ function getPackageEsmEntryPath(pkg: string): string | undefined {
   );
 }
 
-function getEsmNamedExports(pkg: string): string[] {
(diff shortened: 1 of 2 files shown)


src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -111,6 +109,53 @@ function getEsmNamedExports(pkg: string): string[] {
...
+  if (path.isAbsolute(importSource)) {
+    return resolveFileLikeModule(importSource);
+  }
+
+  const projectRoot = getPackageDetectionCwd();
+  if (importSource.startsWith('.')) {
+    return resolveFileLikeModule(path.resolve(projectRoot, importSource));
+  }
+
+  const esmEntry = getInstalledPackageEntry(importSource, {
```
gioboa, line 126: I changed your code @majashryer9 and I moved few lines before. For bare `shareConfig.import` values, named export detection now checks the package’s ESM/browser entry first, matching the module path Vite is expected to bundle.
This avoids reading the Node require conditional entry first, which can expose different named exports and cause generated shared wrappers to export names that do not exist in the actual imported module.

majashryer9: Makes sense, thanks @gioboa!

gioboa (commented):

Image

gioboa (commented):
Thanks @majashryer9 
it looks great to me 💪

gioboa (approved):

PR #773, opened 2026-06-03
fix: use public federation runtime names

### LINKED ISSUE #764: [Bug]: __mfe_internal__ prefix is forced for all microfrontends
### Describe the bug

Overview:
Since @module-federation/vite 1.14.5 when building an application the `__mfe_internal__ ` prefix is added to the name of the microfrontend, which is then used as the module federation instance name.

PR with changes that is causing the issue:
https://github.com/module-federation/vite/pull/617/changes

Why this is an issue:
1. Module federation instance finding
When building with 1.14.4 and lower, I can safely assume that the instance name of the module federation will be equal to the name of the module that is set in the config. However, in higher versions the prefix will be added, essentially breaking all functionality relying on the instance name - https://github.com/module-federation/core/blob/b439039c5497e327e40653e2f32fb958cfc1939a/packages/runtime/src/index.ts#L111.
The prefix is also referenced only in the vite plugin - https://github.com/search?q=org%3Amodule-federation+__mfe_internal__&type=code.

2. Package sharing strategy implications
When shareStrategy is 'version-first' and there are 2 MFEs with the same version of the package, I believe the resolution of the chosen instance is determined by comparison of the MFE names. Since newer versions of the plugin set the prefix, a change of behavior can be observed once the application is upgrading the package version.
The code that is impacted with the change is the following - https://github.com/module-federation/core/blob/b439039c5497e327e40653e2f32fb958cfc1939a/packages/runtime-core/src/shared/index.ts#L446.

Stackblitz reproduction:
1. App with plugin version 1.14.4 - https://stackblitz.com/edit/vitejs-vite-hbsvfnbk
2. App with plugin version ^1.16.2 - https://stackblitz.com/edit/vitejs-vite-gpntfeek

In both cases:
- run the application
- investigate __FEDERATION__ object and the instance for the app

Before the change, the name is `test_mfe` and after is `__mfe_internal__test_mfe`.

### Version

^7.0.14

### Reproduction

https://stackblitz.com/edit/vitejs-vite-gpntfeek

### Relevant log output

```shell

```

### Validations

gioboa:
Close #764

### DIFF
--- a/integration/remote-entry-isolation.test.ts
+++ b/integration/remote-entry-isolation.test.ts
@@ -2,10 +2,11 @@ import { resolve } from 'path';
 import { describe, expect, it } from 'vitest';
 import type { ModuleFederationOptions } from '../src/utils/normalizeModuleFederationOptions';
 import { buildFixture, FIXTURES } from './helpers/build';
+import { isRollupChunk } from './helpers/assertions';
 import { findChunk, getAllChunkCode, getChunkNames } from './helpers/matchers';
 
 const ISOLATION_MF_OPTIONS = {
-  name: 'isolatedRemote',
+  name: 'scheduler',
   filename: 'remoteEntry.js',
   exposes: {
     './shared': resolve(FIXTURES, 'shared-remote', 'exposed-module.js'),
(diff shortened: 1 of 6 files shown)


gioboa (commented):

Screenshot 2026-06-03 at 18 19 23

PR #774, opened 2026-06-03
fix: use explicit shared subpath fallback

### LINKED ISSUE #767: [Bug]: Shared package subpath loadShare wrapper resolves to package root in 1.16.2
### Describe the bug

  • `@module-federation/vite@1.16.2` appears to regress shared package subpath handling during production build.

  When both a package root and one of its subpaths are declared as shared singletons, the generated `loadShare` wrapper for the subpath can use the root
  package `loadShare` wrapper as its local fallback instead of the subpath's local fallback chunk.

  This breaks when the subpath exports symbols that the package root does not export.

  ## Minimal reproduction

  Repository:

  https://github.com/francesc79/mf-vite-shared-subpath-repro

  Install and run:

  ```bash
  pnpm install
  pnpm repro

  The reproduction has a local package with these exports:

  {
    "exports": {
      ".": {
        "import": "./src/index.ts"
      },
      "./media": {
        "import": "./src/media.ts"
      }
    }
  }

  The app imports both:

  import { rootValue } from '@repro/hooks';
  import { useAvatar } from '@repro/hooks/media';

  And configures both as shared singletons:

  shared: {
    '@repro/hooks': { singleton: true },
    '@repro/hooks/media': { singleton: true },
  }

  ## Expected behavior

  The generated loadShare wrapper for @repro/hooks/media should use the ./media fallback, for example:

  import * as __mfLocalShare from './media-<hash>.js';

  ## Actual behavior

  With @module-federation/vite@1.16.2, the generated loadShare wrapper for @repro/hooks/media uses the root package wrapper as its fallback instead:

  await import('./_virtual_mf___repro_mf_1_hooks__loadShare__...js');

  The check script reports:

  FAIL: the @repro/hooks/media loadShare wrapper uses the root @repro/hooks loadShare wrapper as its fallback.

  At runtime, this shape causes named imports from the subpath to be missing or non-functions if the root package does not re-export them.

  ## Regression range

  The same reproduction passes with @module-federation/vite@1.16.1.

  Only changing the package version back to 1.16.1 makes the generated wrapper import the correct media-<hash>.js chunk.

  ## Suspected source

  The regression looks related to the 1.16.2 changes in src/plugins/pluginProxySharedModule_preBuild.ts, especially the new tagged importer proxy logic
  from:

  fix: fix Vue sharing and assets (#757)

  The new logic allows some cross-wrapper shared imports to be proxied instead of always skipping LOAD_SHARE_TAG / PREBUILD_TAG importers. That appears to
  be correct for the Vue case handled by the PR, but for package subpaths it can collapse the subpath fallback to the package root.

  ## Environment

  - @module-federation/vite: 1.16.2
  - Vite: 8.0.16
  - Package manager: pnpm
  - Build mode: production

### Version

8.0.16

### Reproduction

https://github.com/francesc79/mf-vite-shared-subpath-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #767

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -17,6 +17,7 @@ const {
   writeLoadShareModuleMock,
   writeLocalSharedImportMapMock,
   writePreBuildLibPathMock,
+  getInstalledPackageEntryMock,
 } = vi.hoisted(() => ({
   hasPackageDependencyMock: vi.fn<(pkg: string) => boolean>(),
   existsSyncMock: vi.fn<(path: string) => boolean>(() => false),
@@ -26,6 +27,7 @@ const {
   writeLoadShareModuleMock: vi.fn(),
   writeLocalSharedImportMapMock: vi.fn<() => void>(),
   writePreBuildLibPathMock: vi.fn(),
+  getInstalledPackageEntryMock: vi.fn<(pkg: string) => string | undefined>(() => undefined),
 }));
 
 vi.mock('fs', async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #781, opened 2026-06-05
fix: detect enum and destructuring exports when scanning shared named…

### LINKED ISSUE #780: [Bug]: `export enum` re-exported via `export *` is dropped from the shared module `loadShare` proxy
## Describe the bug

The static export scanner that builds the Module Federation `loadShare` proxy
for a shared package misses two common export forms when they are reached
through `export * from './...'`:

1. **`export enum` declarations** (TypeScript enums — runtime values).
2. **destructuring exports**, e.g. `export const { a: renamedA } = obj;` — the
   exact shape Redux Toolkit's `createSlice` produces:
   `export const { addItem: createActionAddItem } = slice.actions;`

Neither is re-exported by the generated proxy, so a consumer importing those
names across the share boundary fails at runtime with:

```
SyntaxError: The requested module
'/@id/__x00__virtual:mf:...loadShare...shared_mf_2_lib...mjs'
does not provide an export named 'Color'
```

(and likewise `does not provide an export named 'createActionAddItem'`).

Plain `export const NAME` / `export function NAME` re-exported the same way
(`export *`) are present in the proxy. Names listed explicitly in an
`export { NAME }` block are also fine; only the `export *` path combined with
these two declaration forms is affected.

## Reproduction

Minimal repro (pnpm workspace, host + a shared workspace package):

```
packages/shared-lib/src/values.ts
  export enum Color { Red = 'red', Blue = 'blue' }      // dropped
  export const GREETING = 'hello'                       // ok
  export function shout(text: string) { return text.toUpperCase() }  // ok
  const bag = { rawA: () => 'a', rawB: () => 'b' }
  export const { rawA: renamedA, rawB: renamedB } = bag // dropped (destructuring)

packages/shared-lib/src/index.ts
  export * from './values'        // star re-export -> the trigger

packages/host/vite.config.ts
  federation({
    name: 'host',
    runtime: 'enhanced',
    shared: { 'shared-lib': { singleton: true } },
  })

packages/host/src/main.ts
  import { Color, GREETING, shout, renamedA } from 'shared-lib'  // Color/renamedA throw
```

Steps:

```bash
pnpm install
pnpm verify   # headless check, exits 1 when the bug is reproduced
# or: pnpm dev, then open the page and look at the console
```

Output of the headless check (before fix):

```
loadShare proxy module: /@id/__x00__virtual:mf:...loadShare__shared_mf_2_lib__loadShare__.mjs
  re-exports GREETING (const)           : true
  re-exports shout (function)           : true
  re-exports Color (enum)               : false
  re-exports renamedA (destructured)    : false
```

A complete runnable repository is available here:
https://github.com/francesc79/mf-vite-enum-export-bug

## Expected behavior

The `loadShare` proxy for `shared-lib` re-exports `Color` (the enum) alongside
`GREETING` and `shout`, so consumers can import the enum across the share
boundary, exactly as they can without Module Federation.

## Root cause

The plugin determines a shared package's named exports statically. For
`export * from './values'`, the names are collected by recursively following the
star re-export in `getNamedExportsViaRegex` (`lib/index.mjs`). The declaration
regex only recognizes `function | const | let | var | class`, and there is no
handling for destructuring patterns:

```js
const declRegex = new RegExp(
  `export\\s+(?:async\\s+)?(?:function(?:\\*\\s*|\\s+\\*?\\s*)|const\\s+|let\\s+|var\\s+|class\\s+)(${JS_IDENTIFIER_PATTERN})`,
  'gu',
);
```

- `enum` (and `namespace`) are not in the alternation, so enums reached through
  a star re-export are never added.
- `export const { ... } = ...` / `export const [ ... ] = ...` are not matched by
  `declRegex` (the next token is `{`/`[`, not an identifier) and are also not
  matched by the `export { ... }` list regex (the leading `const` breaks it), so
  destructured bindings are dropped.

(Names in an explicit `export { ... }` block survive because they go through a
separate regex.)

A fix is to add `enum\s+` / `namespace\s+` to the alternation and add a pass for
destructuring exports:

```js
const declRegex = new RegExp(
  `export\\s+(?:async\\s+)?(?:function(?:\\*\\s*|\\s+\\*?\\s*)|const\\s+|let\\s+|var\\s+|class\\s+|enum\\s+|namespace\\s+)(${JS_IDENTIFIER_PATTERN})`,
  'gu',
);

// destructuring exports: export const { a, b: alias } = ... / export const [x] = ...
const destructureRegex = /export\s+(?:const|let|var)\s+(\{[^}]*\}|\[[^\]]*\])\s*=/g;
const bindingNameRegex = new RegExp(`^(${JS_IDENTIFIER_PATTERN})`, 'u');
while ((match = destructureRegex.exec(source)) !== null) {
  const inner = match[1].slice(1, -1);
  for (const part of inner.split(',')) {
    let token = part.split('=')[0].trim();
    if (!token || token.startsWith('...')) continue;
    if (token.includes(':')) token = token.slice(token.indexOf(':') + 1).trim();
    const b = token.match(bindingNameRegex);
    if (b && isValidEsmExportName(b[1])) names.add(b[1]);
  }
}
```

A ready-to-apply patch against 1.16.3 is included in the repo (for reference,
not applied) at `patches/@module-federation__vite@1.16.3.patch`.

## Environment

- `@module-federation/vite`: 1.16.3 (also reproduced on 1.16.1 — the `declRegex`
  is identical in both)
- `vite`: 8.0.16
- `runtime`: `enhanced`
- Node: v24
- OS: Linux

## Additional context

The enum is a runtime value, so this is not merely a typings issue: the missing
named export breaks the ES module contract and throws at load time in the
consumer.

francesc79:
… exports

`getNamedExportsViaRegex` only matched function/const/let/var/class declarations, so `export enum`/`export namespace` and destructuring exports (e.g. `export const { addItem: createActionAddItem } = slice.actions`) reached via `export *` were dropped from the shared `loadShare` proxy. Importing those names across the share boundary failed at runtime with `does not provide an export named '<name>'`.

Add `enum`/`namespace` to the declaration regex and a dedicated pass for object/array destructuring exports (handling renames, defaults and rest).

Closes #780.

### DIFF
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -196,9 +196,9 @@ function getNamedExportsViaRegex(
     if (isValidEsmExportName(name)) names.add(name);
   }
 
-  // Destructuring exports, e.g. `export const { a, b: alias } = obj;` or
-  // `export const [first] = arr;` — the shape Redux Toolkit's `createSlice`
-  // produces (`export const { addItem: createActionAddItem } = slice.actions`).
+  // Destructuring exports, e.g. `export const { a, b: alias, ...rest } = obj;`
+  // or `export const [first, ...others] = arr;` — the shape Redux Toolkit's
+  // `createSlice` produces (`export const { addItem: createActionAddItem } = slice.actions`).
   // These are matched by neither `declRegex` (the next token is `{`/`[`) nor the
   // `export { ... }` list regex below (the leading `const` breaks it).
   const destructureRegex = /export\s+(?:const|let|var)\s+(\{[^}]*\}|\[[^\]]*\])\s*=/g;
@@ -206,8 +206,11 @@ function getNamedExportsViaRegex(
   while ((match = destructureRegex.exec(source)) !== null) {
     const inner = match[1].slice(1, -1);
     for (const part of inner.split(',')) {
+      // strip a default value (`= ...`); rest elements (`...x`) never have one
       let token = part.split('=')[0].trim();
-      if (!token || token.startsWith('...')) continue;
+      // rest element `...rest` -> the bound name is `rest`
+      if (token.startsWith('...')) token = token.slice(3).trim();
+      if (!token) continue;
       // object rename `key: alias` -> the bound name is the alias
       if (token.includes(':')) token = token.slice(token.indexOf(':') + 1).trim();
       const bindingMatch = token.match(bindingNameRegex);
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -187,13 +187,32 @@ function getNamedExportsViaRegex(
   const declRegex = new RegExp(
     `export\\s+(?:async\\s+)?(?:` +
       `function(?:\\*\\s*|\\s+\\*?\\s*)` +
-      `|const\\s+|let\\s+|var\\s+|class\\s+)(${JS_IDENTIFIER_PATTERN})`,
+      `|const\\s+|let\\s+|var\\s+|class\\s+|enum\\s+|namespace\\s+)(${JS_IDENTIFIER_PATTERN})`,
     'gu'
   );
   let match: RegExpExecArray | null;
   while ((match = declRegex.exec(source)) !== null) {
     const name = match[1];
     if (isValidEsmExportName(name)) names.add(name);
+  }
+
+  // Destructuring exports, e.g. `export const { a, b: alias } = obj;` or
+  // `export const [first] = arr;` — the shape Redux Toolkit's `createSlice`
+  // produces (`export const { addItem: createActionAddItem } = slice.actions`).
+  // These are matched by neither `declRegex` (the next token is `{`/`[`) nor the
+  // `export { ... }` list regex below (the leading `const` breaks it).
+  const destructureRegex = /export\s+(?:const|let|var)\s+(\{[^}]*\}|\[[^\]]*\])\s*=/g;
+  const bindingNameRegex = new RegExp(`^(${JS_IDENTIFIER_PATTERN})`, 'u');
+  while ((match = destructureRegex.exec(source)) !== null) {
+    const inner = match[1].slice(1, -1);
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Image

src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -196,6 +196,25 @@ function getNamedExportsViaRegex(
...
+  // produces (`export const { addItem: createActionAddItem } = slice.actions`).
+  // These are matched by neither `declRegex` (the next token is `{`/`[`) nor the
+  // `export { ... }` list regex below (the leading `const` breaks it).
+  const destructureRegex = /export\s+(?:const|let|var)\s+(\{[^}]*\}|\[[^\]]*\])\s*=/g;
+  const bindingNameRegex = new RegExp(`^(${JS_IDENTIFIER_PATTERN})`, 'u');
+  while ((match = destructureRegex.exec(source)) !== null) {
+    const inner = match[1].slice(1, -1);
+    for (const part of inner.split(',')) {
+      let token = part.split('=')[0].trim();
+      if (!token || token.startsWith('...')) continue;
```
gioboa: Here we are skipping `...` so the PR description `handling renames, defaults and rest` is not 100% accurate.
Would you like to cover this too?

gioboa (changes requested):
Thanks @francesc79 for your help.

src/virtualModules/virtualShared_preBuild.ts
francesc79: Good catch — you're right. Rather than relax the description, I updated the scan to actually cover rest elements: ...rest (object) and [first, ...others] (array) bind real
  runtime values, so they're now captured (strip the leading ..., keep the bound name). Added test coverage for both. Thanks for the review!

gioboa: Thank you 💪

gioboa (approved):
Thanks @francesc79 you rock 🚀

francesc79:
Thanks @gioboa, glad to help 🚀😉

PR #782, opened 2026-06-05
fix: fix up Nuxt dev bootstrap loop

gioboa:


gioboa (commented):

Screenshot 2026-06-05 at 13 44 17

PR #786, opened 2026-06-06
fix: respect entryFileNames directory for bootstrap file output

pcfreak30:
The MF bootstrap file was emitted with a hardcoded fileName (mf-entry-bootstrap-N.js) at the build root, bypassing Rollup's entryFileNames pattern. When users configure entryFileNames to place JS in a subdirectory (e.g. static/js/), the bootstrap file still landed at the root while all other entries went to static/js/.

Derive the directory prefix from the emitted hostInit file path at build time (since Vite/Rolldown normalizes entryFileNames before plugins can read it). Rebase the import paths inside the bootstrap to be relative to its new location instead of the HTML root, preventing doubled paths like static/js/static/js/.

Falls back to root-level output when no subdirectory is detected for backward compatibility.

---

The new bootstrap concept broke how I use MF, and so this adds back the needed flexibility for my use case.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -967,7 +967,7 @@ describe('pluginAddEntry', () => {
     expect(bootstrapFile).toBeDefined();
     expect(bootstrapFile!.fileName).toMatch(/^static\/js\/mf-entry-bootstrap-0-[a-f0-9]{8}\.js$/);
     expect(bootstrapFile!.source as string).toContain('__mfImport("./hostInit-abc.js")');
-    expect(bootstrapFile!.source as string).toContain('__mfImport("../src/main.tsx")');
+    expect(bootstrapFile!.source as string).toContain('__mfImport("../../src/main.tsx")');
   });
 
   it('emits bootstrap file at root when entryFileNames is not set', () => {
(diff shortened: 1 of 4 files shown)


pcfreak30:
@gioboa

pcfreak30:
@gioboa please re-approve workflow run

gioboa (commented):
Thanks for your help @pcfreak30 
I have two small comments:
- With base: "/app/", the bootstrap file is emitted under static/js/, but the generated host init import is rebased to ../app/static/js/hostInit-abc.js. From static/js/, this resolves incorrectly; it should import the sibling file as ./hostInit-abc.js.
- Absolute URLs should not be rebased. When renderBuiltUrl or similar returns https://..., rebaseImport() currently treats it like a local bare path and rewrites it to ../https://..., producing an invalid import URL.

gioboa (changes requested):

pcfreak30:
@gioboa done

pcfreak30:
@gioboa if you still see issues, let me know how to re-prod as an example so I can troubleshoot.

gioboa (commented):

Screenshot 2026-06-07 at 22 09 53

gioboa (approved):
That's amazing thanks @pcfreak30 👏

PR #789, opened 2026-06-07
fix: fix Vue loaded-first static imports

### LINKED ISSUE #787: [Bug]: Static Vue component import fails with loaded-first strategy
### Describe the bug

Hello,

Since version 1.16, statically importing a remote Vue component no longer works when using the **loaded-first** strategy.

### Steps to reproduce

1. Clone https://github.com/clemvnt/vite-mf-import-vue-comp-issue
2. Run `npm install` in `host-app` and `remote-app` directories.
3. Run `npm run dev` in both directories.
4. Open the `host-app` application.

### Expected behavior

The remote counter component should be rendered correctly.

### Actual behavior

The remote counter component is not rendered, and the browser console reports the following error:
```
(index):1 Uncaught (in promise) Promise {<fulfilled>: undefined}[[Prototype]]: Promise[[PromiseState]]: "fulfilled"[[PromiseResult]]: undefined
```

### Additional information

* Without **loaded-first** strategy, it works.
* With version 1.14.5, it works.
* When the remote component is imported dynamically using `import()` + `defineAsyncComponent`, it works. 
* In my real-world setup, the remote component is statically imported inside a host component that is itself lazy-loaded via a dynamic `import()` + `defineAsyncComponent`. The issue still occurs in that scenario. In the repro, the commented LazyComp code represents this case.

### Versions

Vite 8.0.16
@module-federation/vite 1.16.4

### Reproduction

https://github.com/clemvnt/vite-mf-import-vue-comp-issue

### Validations

gioboa:
Close #787

### DIFF
--- a/src/virtualModules/__tests__/virtualRemotes.test.ts
+++ b/src/virtualModules/__tests__/virtualRemotes.test.ts
@@ -4,6 +4,7 @@ import { getRemoteVirtualModule, generateRemotes } from '../virtualRemotes';
 const mockOptions = vi.hoisted(() => ({
   shareStrategy: 'version-first' as 'version-first' | 'loaded-first',
 }));
+const hasPackageDependencyMock = vi.hoisted(() => vi.fn((_pkg: string) => false));
 
 vi.mock('../../utils/packageUtils', async () => {
   const actual = await vi.importActual<typeof import('../../utils/packageUtils')>(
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-06-07 at 21 50 29

PR #790, opened 2026-06-07
fix: await nested remote static imports

### LINKED ISSUE #785: React hook exposed from Vite remote fails with static import, but works with dynamic import
### Describe the bug

In a React + TypeScript + Vite setup using `@module-federation/vite`, a host exposes two React hooks:

```ts
// host/src/useEmployeeAccess.ts
export { useEmployeeAccess } from './RemoteProviders';
export { useEmployeeAccess as default } from './RemoteProviders';
```

```ts
// host/src/useHeaderConfig.ts
export { useHeaderConfig } from './RemoteProviders';
export { useHeaderConfig as default } from './RemoteProviders';
```

The host renders the remote inside the matching providers:

```tsx
// host/src/App.tsx
<RemoteProviders>
  <RemoteApp />
</RemoteProviders>
```

The remote then consumes the exposed hooks.

When the remote uses a static named import, the app can fail at runtime. In this repro the failure is intermittent, so you may need to refresh the page multiple times before the console error appears:

```tsx
import { useEmployeeAccess } from 'viteViteHost/useEmployeeAccess';
import { useHeaderConfig } from 'viteViteHost/useHeaderConfig';

export default function RemoteApp() {
  const { selectedEmployee } = useEmployeeAccess();
  const { title } = useHeaderConfig();

  return <pre>{JSON.stringify({ selectedEmployee, title }, null, 2)}</pre>;
}
```

However, if the remote loads the same exposed module with dynamic import, the app loads:

```ts
const { useHeaderConfig } = await import('viteViteHost/useHeaderConfig');
```

In this repo, `vite-remote/src/App.tsx` already has the dynamic import workaround commented out. To test the workaround, comment out the static import and uncomment the dynamic import.

## Expected / unclear behavior

I am not fully sure what the intended behavior is for React hook/function modules exposed through `@module-federation/vite`.

The behavior I am trying to understand is whether a statically imported exposed hook module is expected to be usable the same way as the same module loaded with dynamic import:

```tsx
import { useHeaderConfig } from 'viteViteHost/useHeaderConfig';

const { title } = useHeaderConfig();
```

If static imports of exposed hooks/functions are not expected to work in React projects, it would be helpful to know that and document the recommended pattern.

## Actual behavior

The static named import path can fail at runtime when the remote calls the exposed hook during render.

The app errors when destructuring the hook result:

```tsx
const { title } = useHeaderConfig();
```

The same exposed module works when loaded with dynamic import.

The failure is timing-dependent in this minimal reproduction. On some loads the page may render successfully; after several refreshes, the remote can fail during render and the console shows the destructuring error.

## Question / clarification

Is this usage expected to work with `@module-federation/vite`?

I am not assuming this is definitely a bug. I am trying to understand whether exposing host-owned React hooks/functions is supported in React projects using this Vite plugin, especially with static named imports.

In this repro, both exposed modules are hook functions. The host providers are mounted, React is shared as a singleton, and the same exposed module works when loaded dynamically. I am not sure whether static imports of exposed hook/function modules are expected to behave the same way as dynamic imports in React projects.

Moving these hooks into a shared package is not a viable workaround for my use case. The hooks read host-owned provider/context state, and the remote needs to consume that live state from the shell rather than import an independent copy of the hook from a shared library.

## Reproduction

Repository: https://github.com/bashiq/vite-mfe

Steps:

1. Install dependencies in both apps.
2. Start the host on port `5175`.
3. Start the remote on port `5176`.
4. Open the host app.
5. Refresh the host page several times. If needed, hard refresh with DevTools open and cache disabled.
6. With the static import, the remote can fail during render and show the console error.
7. Change `useHeaderConfig` in the remote to dynamic import and the app loads consistently.

The dynamic import workaround is already present as commented lines in `vite-remote/src/App.tsx`.

## Environment

- `@module-federation/vite`: `1.16.4`
- `vite`: `8.0.10`
- `react`: `19.2.4`
- `react-dom`: `19.2.4`
- TypeScript: `5.9.3`
- Browser: Chrome
- OS: macOS

## Question

Is exposing and consuming host-owned hook modules an expected use case for `@module-federation/vite` in React projects? If yes, should static named imports and dynamic imports behave the same for those modules? If not, what import/architecture pattern is recommended?

### Version

8.0.10

### Reproduction

https://github.com/bashiq/vite-mfe

### Relevant log output

```shell
App-BdaS3I_E.js:1 Uncaught TypeError: Cannot destructure property 'selectedEmployee' of 'C(...)' as it is undefined.
    at D (App-BdaS3I_E.js:1:4550)
    at Mo (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:47545)
    at Cc (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:69967)
    at Bc (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:80222)
    at Iu (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:115515)
    at Nu (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:114598)
    at Mu (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:114437)
    at bu (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:111306)
    at fd (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:8:122802)
    at MessagePort.ae (_virtual_mf___mfe_internal__viteViteHost__loadShare__react_mf_2_dom_mf_1_client__loadShare__.mjs-ClfXvo60.js:1:1858)
```

### Validations

gioboa:
Close #785

Fixes a race where a remote module could render before its static federated imports finished resolving. Static named/namespace remote import rewrites now expose a dependency promise, and remote wrappers wait for it before caching/rendering the loaded module. This avoids unresolved named import proxies returning undefined, no use of top-level await.

### DIFF
--- a/src/plugins/__tests__/pluginRemoteNamedExports.test.ts
+++ b/src/plugins/__tests__/pluginRemoteNamedExports.test.ts
@@ -98,6 +98,11 @@ describe('pluginRemoteNamedExports', () => {
     it('rewrites named import', async () => {
       const result = await transform('import { foo } from "remoteApp/utils";');
       expect(result).toContain('import { __moduleExports as');
+      expect(result).toContain('__mf_remote_pending as __mf_ns_0_pending');
+      expect(result).toContain(
+        'export const __mf_remote_dependency_pending = Promise.all([__mf_ns_0_pending]);'
+      );
+      expect(result).not.toContain('await ');
       expect(result).toContain('__mfCreateNamedRemoteProxy');
       expect(result).toContain('__mfCreateNamedRemoteProxy(__mf_ns_0, "foo")');
       expect(result).not.toContain('import { foo }');
@@ -188,7 +193,13 @@ describe('pluginRemoteNamedExports', () => {
   describe('namespace imports', () => {
     it('rewrites namespace import', async () => {
       const result = await transform('import * as utils from "remoteApp/utils";');
-      expect(result).toContain('import { __moduleExports as utils }');
+      expect(result).toContain(
+        'import { __moduleExports as utils, __mf_remote_pending as utils__mf_pending }'
+      );
+      expect(result).toContain(
+        'export const __mf_remote_dependency_pending = Promise.all([utils__mf_pending]);'
+      );
+      expect(result).not.toContain('await ');
       expect(result).not.toContain('import *');
     });
   });
@@ -235,6 +246,11 @@ describe('pluginRemoteNamedExports', () => {
     it('rewrites named re-export', async () => {
       const result = await transform('export { foo } from "remoteApp/utils";');
       expect(result).toContain('import { __moduleExports as');
+      expect(result).toContain('__mf_remote_pending as __mf_ns_0_pending');
+      expect(result).toContain(
+        'export const __mf_remote_dependency_pending = Promise.all([__mf_ns_0_pending]);'
+      );
+      expect(result).not.toContain('await ');
       expect(result).toContain('__mf_re_');
       expect(result).toContain('as foo');
     });
@@ -274,7 +290,13 @@ describe('pluginRemoteNamedExports', () => {
         '/src/app.tsx',
         true
       );
-      expect(result).toContain('import { __moduleExports as utils }');
+      expect(result).toContain(
+        'import { __moduleExports as utils, __mf_remote_pending as utils__mf_pending }'
+      );
+      expect(result).toContain(
+        'export const __mf_remote_dependency_pending = Promise.all([utils__mf_pending]);'
+      );
+      expect(result).not.toContain('await ');
     });
 
     it('rewrites default + named import via fallback', async () => {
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Screenshot 2026-06-07 at 22 58 31

PR #791, opened 2026-06-07
fix: add missing manifest additionalData

gioboa:


gioboa (commented):

Screenshot 2026-06-07 at 23 46 22

PR #793, opened 2026-06-09
chore: drop CJS build preserving compatibility

gioboa:
This is `not` a breaking change because:
 - It drops the CommonJS build and publishes the package as ESM-only, while keeping CommonJS consumers working on Node versions that support require(esm): Node ^20.19.0 or >=22.12.0.
 - It aligns package exports and build output to .js/.d.ts, removes stale CJS output paths, and centralizes ESM-native module resolution through import.meta.resolve.
- It also keeps user project compatibility for .cjs/.mjs/.mts inputs, meaning consumers can still use CJS config/files, ESM .mjs/.mts sources, and CJS/ESM remote/shared dependencies; only this package’s published build output drops CJS.

### DIFF
--- a/.github/publish-ci/node-esm/package.json
+++ b/.github/publish-ci/node-esm/package.json
@@ -9,8 +9,5 @@
   },
   "dependencies": {
     "@module-federation/vite": "workspace:*"
-  },
-  "devDependencies": {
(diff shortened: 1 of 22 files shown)


gioboa (commented):

Screenshot 2026-06-09 at 14 28 34

PR #795, opened 2026-06-09
chore: replace pathe with node:path

gioboa:


gioboa (commented):

Image

PR #796, opened 2026-06-09
chore: replace estree-walker with local walker

gioboa:


gioboa (commented):

Screenshot 2026-06-09 at 18 21 18

PR #797, opened 2026-06-09
chore: remove es-module-lexer dependency safely

gioboa:


gioboa (commented):

Image

Issue #792, opened 2026-06-08
[Bug]: Remote component using RTK crashes with `resolveDispatcher() is null`


alternativniy:
### Describe the bug

**Package:** `@module-federation/vite@1.16.4`  
**Affects:** dev mode, any remote component that calls `useSelector` or any other Redux hook

When the host renders a component exposed by the remote that calls `useSelector`, React throws:

```
Uncaught TypeError: can't access property "useRef", resolveDispatcher() is null
```

Full stack:

```
useRef                         React
useSyncExternalStoreWithSelector  react-redux (…al-store-with-selector.development.js:35)
useSelector2                   Redux
AppSub                         AppSub.tsx:6
```

The `resolveDispatcher() is null` error means `useSelector` → `useSyncExternalStoreWithSelector` → `useRef` is called against a **different React instance** than the one rendering the component tree. Despite `react` being declared `singleton: true` in the MF shared config, `@module-federation/vite@1.16.4` fails to deduplicate React — the remote loads its own copy, causing hooks to operate on a detached dispatcher.

**Steps to reproduce:**

1. Clone [repository](https://github.com/alternativniy/mf-vite-8-named-exports)
2. `pnpm install:deps`
3. `pnpm dev`
4. Open `http://localhost:5001`
5. Host renders the remote `AppSub` component — crash on mount

**Expected:** `react` singleton shared correctly; `useSelector` works inside remote components.  
**Actual:** Remote loads its own React instance; any hook call from `react-redux` (or similar libraries using `useSyncExternalStore`) throws `resolveDispatcher() is null` and crashes the component tree.

### Version

1.16.4

### Reproduction

https://github.com/alternativniy/mf-vite-8-named-exports

### Relevant log output

```shell
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://react.dev/link/invalid-hook-call for tips about how to debug and fix this problem. <anonymous code>:1:147461
    overrideMethod <anonymous code>:1
    React 2
    useSyncExternalStoreWithSelector use-sync-external-store-with-selector.development.js:35
    useSelector2 Redux
    AppSub AppSub.tsx:6
    React 10
    performWorkUntilDeadline scheduler.development.js:45
    (Асинхронный: EventHandlerNonNull)
[...]

PR #798, opened 2026-06-10
feat(ssr): support mf-manifest.json as remote entry URL

tonoizer:
## Summary

SSR hosts can use `entry: "…/mf-manifest.json"` interchangeably with `remoteEntry.js`. On the server, `ssrEntryLoader` reads `metaData.ssrRemoteEntry` from the manifest and loads the Node SSR entry. This unlocks [Module Federation Chrome DevTools](https://module-federation.io/guide/debug/chrome-devtool) manifest-based discovery.

**Scope (intentionally narrow):**
- `ssrEntryLoader` — manifest-first SSR resolution for `.json` entries; HEAD-check `ssrRemoteEntry`; fallback to `__mf_server__` / URL convention; existing `remoteEntry.js` path unchanged
- `index.ts` — `configEnvironment` sets `ENV_TARGET: "node"` on server/ssr Vite environments only (client stays `"web"`)

**Out of scope:** no changes to virtual remote wrappers, shared prebuild codegen, or client-side manifest handling (handled by `@module-federation/enhanced` runtime).

## Usage

Remote (`manifest: true` required):

```ts
federation({ name: "remote", filename: "remoteEntry.js", manifest: true, exposes: { … } });
```

Host (either entry style):

```ts
remotes: {
  remote: {
    type: "module",
    name: "remote",
    entry: "http://localhost:4174/mf-manifest.json", // or …/remoteEntry.js
  },
},
```

## Test plan

- [x] `pnpm test` (ssrEntryLoader manifest-as-entry + ENV_TARGET configEnvironment tests)
- [x] `pnpm playwright test --project=multi-example` (10/10)


tonoizer:
I can also adjust the vite examples PR (https://github.com/gioboa/module-federation-vite-examples/pull/6) to use the manifest instead of the remoteEntry.js as a demo.

MANIFEST_SSR.md
```diff
@@ -0,0 +1,34 @@
+# feat(ssr): support `mf-manifest.json` as remote entry URL
```
gioboa: do we need this file?

gioboa (commented):
Thanks for your help

MANIFEST_SSR.md
tonoizer: no I will remove it

src/index.ts
```diff
@@ -392,8 +391,7 @@ export default __mfShared.default ?? __mfShared;`,
             // their submodules rely on parent initialization order. Other
             // shared deps should remain optimizable so Vite can apply CJS
             // interop for transitive dependencies used by prebuild fallbacks.
-            const shouldBypassOptimizeDep =
-              isLitShare(key) || (key === 'react' && hasPackageDependency('react-redux', root));
```
gioboa, line 397: I recently added this specific check in this [PR](https://github.com/module-federation/vite/pull/800/changes) to solve #792.
I need to test that scenario again with this new code.

gioboa (commented):

Image

src/index.ts
tonoizer: Reverted the change hope that's okey now thanks for the good feedback

gioboa (commented):
I did a check with LLM and there are few things to fix.
Findings:

  1. src/index.ts:627 removes LOAD_SHARE_TAG / PREBUILD_TAG from parse-end exclusion. That reverts the recent “reduce latency” behavior and
  makes generated shared modules gate parsePromise again. Risk: large builds wait/timeout on MF-generated virtual modules unrelated to app
  parse completion.
  2. src/utils/ssrEntryLoader.ts:550 now falls through from /__mf_ssr__/ ModuleRunner failure/unavailable into fetchEsmToTempFile. For dev SSR
  this path is known broken: served SSR entry imports virtual:mf-exposes-ssr:*, and temp-file import can’t resolve Vite virtual ids. This
  silently masks runner failures and does wasted/brittle fetch/import work.
  3. src/index.ts:1105 configEnvironment overwrites existing env-level define.ENV_TARGET. Root config() deliberately preserves user-defined
  ENV_TARGET; this hook should likely do same or warn.

gioboa (changes requested):

src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -51,10 +51,13 @@ export function setLocalSharedImportMapInvalidator(invalidator: (() => void) | u
   invalidateLocalSharedImportMap = invalidator;
 }
 
+let prevLocalSharedImportMapContent: string | undefined;
 export function writeLocalSharedImportMap() {
-  // The map is generated by the loader. Avoid rebuilding a large string every
-  // time another shared module is discovered during resolution.
-  invalidateLocalSharedImportMap?.();
+  const nextContent = generateLocalSharedImportMap();
```
gioboa: I did this change in [#804](https://github.com/module-federation/vite/pull/804/changes)
is there a specific reason to revert the change?

src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -481,20 +480,12 @@ export function toViteOptimizedDepVirtualId(id: string): string {
 }
 
 export function getCachedLoadSharePkg(id: string): string | undefined {
-  // Most resolved ids are not loadShare virtual ids. Fast reject before
-  // normalization/decoding work on the resolveId hot path.
-  if (!id.includes(LOAD_SHARE_TAG)) return;
```
gioboa, line 549: I did this change in #804
is there a specific reason to revert the change?

src/virtualModules/virtualRemoteEntry.ts
tonoizer: Good catch. No, there wasn’t a specific reason to revert this. It was an accidental regression from the merge/rebase while working on the SSR manifest changes. I restored the #804 behavior so we keep the existing hot-path optimization intact.

src/virtualModules/virtualShared_preBuild.ts
tonoizer: Good catch. No, there wasn’t a specific reason to revert this. It was an accidental regression from the merge/rebase while working on the SSR manifest changes. I restored the #804 behavior so we keep the existing hot-path optimization intact.

src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -634,19 +634,30 @@ function generateLazyWorkspaceSingletonExports(
...
       __mfApplyLazyShareExports(exportModule);`
-          : `initPromise.then(() =>
-        import(${escapeGeneratedStringLiteral(importSource)}).then((mod) => {
-          exportModule = __mfNormalizeShareModule(mod);
-          __mfModuleCache.share[${escapeGeneratedStringLiteral(cacheKey)}] = exportModule;
-          __mfApplyLazyShareExports(exportModule);
-        })
-      );`
+          : `if (import.meta.env.SSR) {
+        const mod = await import(${escapeGeneratedStringLiteral(importSource)});
```
gioboa: I apologise if I am adding repeated comments.
Here we are adding a TLA and this is breaking the possibility to consume ESM from CJS.
This is because require() cannot be used on an ESM graph with top-level await.
we should fix it.

`The module is fully synchronous (contains no top-level await);` from [Node official docs](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require)

tonoizer: No worries happy to contribute and help the mf Vite plugin to grow, always glad for feedback 🙏

src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -634,23 +634,31 @@ function generateLazyWorkspaceSingletonExports(
...
+          })
+        );
+      }`
       }
     } else {
       __mfApplyLazyShareExports(exportModule);
     }
     export { __mf_default as default };${namedExportLine}`;
 
-  return eagerLocalFallback
```
gioboa, line 659: The CI is failing because we need to keep supporting this scenario

gioboa (approved):
Thanks @tonoizer 
I manually tested it with the Nuxt example and SSR manifest is working fine.

gioboa (commented):

Image

PR #799, opened 2026-06-10
fix: fix up React HMR

gioboa:


gioboa (commented):

Screenshot 2026-06-10 at 13 19 06

PR #804, opened 2026-06-10
chore: reduce latency

gioboa:
Performance improvements for large shared graphs:

  - Cache virtual module lookup by id.
  - Cache shared-key matching instead of scanning all shared keys per resolve.
  - Fast-path loadShare id parsing.
  - Exclude generated loadShare/prebuild modules from parse tracking.
  - Use idle parse timeout by default.
  - Resolve parse wait on buildEnd.
  - Defer local shared import-map generation to loader.


gioboa (commented):

Screenshot 2026-06-10 at 23 44 58

PR #807, opened 2026-06-11
fix: order shared dependency preload

### LINKED ISSUE #805: [Bug]: Workspace package shared as singleton: named exports are `undefined` in production builds (regression in 1.16.6) — "X is not a function"
## Summary

In a **production build**, a **workspace package shared as a `singleton`** that has **named exports** generates a `loadShare` wrapper whose exports are assigned **asynchronously** (inside `initPromise.then(() => import(...).then(...))`).

Any consumer that uses one of those named exports **synchronously, at module-evaluation time** (e.g. a Redux store calling `configureStore(...)` at the top level) reads `undefined` and throws:

```
Uncaught (in promise) TypeError: X is not a function
```

This is a **regression in 1.16.6** — the same setup works in **1.16.5**.

## Versions

| package | version |
|---|---|
| `@module-federation/vite` | **1.16.6** (broken) / `1.16.5` (works) |
| `vite` | 8.0.16 |
| node | 24.x |
| package manager | pnpm workspaces |

## Reproduction

Minimal repro: https://github.com/francesc79/mf-vite-workspace-singleton-repro

```bash
git clone https://github.com/francesc79/mf-vite-workspace-singleton-repro
cd mf-vite-workspace-singleton-repro
pnpm install
pnpm --filter @repro/host build
pnpm --filter @repro/host preview   # http://localhost:4173
```

pnpm workspace with three packages:

```
packages/
  shared-lib/   @repro/shared-lib  — exports `makeStore` (named function)
  core/         @repro/core        — imports { makeStore }, calls makeStore(41) at top level
  host/         @repro/host        — MF host, shares BOTH as singleton, consumes @repro/core
```

`host/vite.config.js` shares both workspace packages as singletons, listing `@repro/core` **before** `@repro/shared-lib` so the MF init loop resolves `@repro/core` first — at which point `@repro/shared-lib`'s share cache is not yet populated.

### Expected (and actual on 1.16.5)

```
OK — core store.value = 41
```

### Actual on 1.16.6

Page stays on `loading…`, console shows:

```
Uncaught (in promise) TypeError: e is not a function
    at assets/src-XXXX.js
```

(`e` is the minified `makeStore`.)

## Root cause

`src/virtualModules/virtualShared_preBuild.ts`, in `writeLoadShareModule`:

```ts
const usesLazyLocalFallback =
  isWorkspacePackage && shareItem.shareConfig.singleton === true;

if (usesLazyLocalFallback) {
  exportLine = generateLazyWorkspaceSingletonExports(
    namedExports,
    lazyLocalFallbackSource,
    cacheKey,
    command !== 'build',   // eagerLocalFallback = false for production builds
  );
}
```

For `command === 'build'`, `eagerLocalFallback` is `false`, so the generated wrapper assigns the named exports **inside an async continuation**:

### Generated wrapper — 1.16.6 (broken)

```js
let __mf_0;                 // makeStore
let __mf_default;
const __mfApplyLazyShareExports = (mod) => {
  __mf_0 = mod["makeStore"];
  __mf_default = mod.default ?? mod;
};
let exportModule = __mfModuleCache.share["@repro/shared-lib"];
  initPromise.then(() =>
    import("./src-XXXX.js").then((mod) => {     // <-- async
      exportModule = __mfNormalizeShareModule(mod);
      __mfModuleCache.share["@repro/shared-lib"] = exportModule;
      __mfApplyLazyShareExports(exportModule);  // <-- exports assigned here, later
    })
  );
} else {
  __mfApplyLazyShareExports(exportModule);
}
export { __mf_default as default };
export { __mf_0 as makeStore };                 // undefined until the .then runs
```

When `exportModule === undefined` (share cache not yet populated — the common case when this wrapper is pulled into another module's synchronous evaluation), `makeStore` stays `undefined` for the entire synchronous phase. A consumer doing `const store = makeStore(41)` at module top level throws `makeStore is not a function`.

### Generated wrapper — 1.16.5 (works)

1.16.5 used **top-level `await import(...)`**, so the export bindings are populated before the module finishes evaluating:

```js
let i = __mfModuleCache.share["@repro/shared-lib"];
if (i === undefined) {
  i = __mfNormalizeShareModule(await import("./src-XXXX.js"));  // <-- top-level await
  __mfModuleCache.share["@repro/shared-lib"] = i;
}
const { LABEL: a, makeStore: o } = i;   // synchronous, after await
export { o as makeStore, a as LABEL };  // real function at export time
```

## Bisect

The only commit between `1.16.5` and `1.16.6` that touches the build path:

- `b07ce14` — *fix: framework-agnostic remote exports, deferred loaded-first, and unified SSR gates (#742)* — rewrote `virtualShared_preBuild.ts` / `virtualRemotes.ts`.

(`#800 "avoid duplicate react with redux"` is `serve`-only and unrelated to the production-build failure.)

## Suggested fix

Either:

- keep the **top-level `await import()`** form (1.16.5 behaviour) for the production-build path of `usesLazyLocalFallback`, so named/default exports are live before evaluation completes; or
- if the deferred form must stay, guarantee the share cache is populated before any consumer wrapper can be synchronously evaluated.

## Real-world impact

Any monorepo that shares workspace packages as singletons and builds a Redux (or similar) store at module top level inside one of those packages hits this. The deferred-assignment wrapper only works for consumers that touch the exports **asynchronously** (e.g. React components at render time); it breaks any **synchronous** top-level use of a named export.

### Version

8.0.16

### Reproduction

https://github.com/francesc79/mf-vite-workspace-singleton-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #805 

Fixes build-mode shared singleton initialization when one shared workspace package depends on another. Shared preload order now respects package dependencies, preventing consumers from evaluating before their shared dependencies are cached, without eager local fallback imports that double-initialize singletons.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -48,6 +48,33 @@ vi.mock('../../utils/packageUtils', () => {
       shareItem.shareConfig.singleton || !shareItem.version ? pkg : `${pkg}@${shareItem.version}`,
     hasPackageDependency: hasPackageDependencyMock,
     packageNameEncode: (name: string) => name.replace(/[^a-zA-Z0-9_-]/g, '_'),
+    getPackageName: (packageString: string) => {
+      const match = packageString.match(/^(?:@[^/]+\/)?[^/]+/);
+      return match ? match[0] : packageString;
+    },
+    getInstalledPackageJson: (pkg: string) => {
+      if (pkg === '@repro/core') {
+        return {
+          path: '/repo/packages/core/package.json',
+          dir: '/repo/packages/core',
+          packageJson: {
+            name: '@repro/core',
+            dependencies: {
+              '@repro/shared-lib': 'workspace:*',
+            },
+          },
+        };
+      }
+      if (pkg === '@repro/shared-lib') {
+        return {
+          path: '/repo/packages/shared-lib/package.json',
+          dir: '/repo/packages/shared-lib',
+          packageJson: {
+            name: '@repro/shared-lib',
+          },
+        };
+      }
+    },
   };
 });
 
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-06-11 at 15 31 44

PR #808, opened 2026-06-11
fix: fix dev remote proxy exports

gioboa:


gioboa (commented):

Screenshot 2026-06-11 at 12 05 49

PR #809, opened 2026-06-11
fix(manifest): advertise types archive url in mf-manifest

giorgiPapava:
## Problem

A vite **producer** generates and serves its type archive in dev (`/@mf-types.zip`, `/@mf-types.d.ts`) and emits it on build, but `mf-manifest.json` hardcodes:

```jsonc
"types": { "path": "", "name": "" }
```

in **both** the dev-server middleware and the build manifest. There is no `zip`/`api`.

A **host** consuming that remote via `@module-federation/dts-plugin` reads `metaData.types.zip` to locate the archive (`requestRemoteManifest`), and throws when it's missing:

```
[ERROR] requestRemoteManifest - fetch manifest failed,
Error: Can not get <remote>'s types archive url!, <remote> will be ignored
```

So remote type hints are silently dropped for any host consuming a vite remote — even though the archive is right there and reachable (HTTP 200). Runtime is unaffected; only cross-remote types break.

The webpack/rspack producer (`@module-federation/enhanced`) populates `metaData.types.zip`/`.api` correctly — the vite plugin just never wired it through.

## Fix

Populate `metaData.types.zip` / `.api` with the same relative paths the dts plugin already serves (`<typesFolder>.zip` / `<typesFolder>.d.ts`), resolved against `publicPath` by the consumer. Applied at both manifest sites via a shared `resolveTypesMeta(dts)` helper that:

- returns `{ path: '', name: '' }` (unchanged) when `dts === false` or `generateTypes === false`
- otherwise advertises `zip`/`api`, honoring a custom `generateTypes.typesFolder`

`DEFAULT_PUBLIC_TYPES_FOLDER` is exported from `pluginDts.ts` so both modules share one source of truth.

## Tests

Added 4 cases to `pluginMFManifest.test.ts` (enabled → advertised, custom `typesFolder`, `dts: false` → empty, `generateTypes: false` → empty). Full suite green (505 passed), typecheck + `oxfmt` clean.

## Open question for maintainers

The dts plugin only generates/serves the archive for **TypeScript projects** (`isTSProject`). This PR gates on `dts`/`generateTypes` only, matching the dev middleware's default-on behavior. If you'd prefer to also gate on `isTSProject` to avoid advertising a 404 for non-TS remotes, happy to wire that through.

## Repro

A vite remote exposing any module + any host consuming it through `@module-federation/dts-plugin`'s dynamic remote type hints. Before: host logs the error above and gets no remote types. After: host downloads `@mf-types.zip` and resolves the remote's `.d.ts`.

### DIFF
--- a/src/plugins/__tests__/pluginMFManifest.test.ts
+++ b/src/plugins/__tests__/pluginMFManifest.test.ts
@@ -126,6 +126,7 @@ async function runGenerateBundleWithManifest(
         }
       | undefined
     >;
+    dts?: unknown;
   } = {},
   command: 'serve' | 'build' = 'build',
   base = '/'
@@ -135,6 +136,7 @@ async function runGenerateBundleWithManifest(
     filename: 'remoteEntry.js',
     getPublicPath: undefined,
     varFilename: undefined,
+    dts: runtime.dts,
     manifest: manifestOptions,
     exposes: runtime.exposePaths || {},
     remotes: {},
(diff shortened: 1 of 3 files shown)


gioboa (approved):
LGTM 👏  Thanks @giorgiPapava

gioboa:

Screenshot 2026-06-11 at 14 43 17

PR #810, opened 2026-06-12
fix: reduce host bootstrap waterfall with modulepreload

jgodi:
The MF bootstrap loads hostInit → remoteEntry → _virtual_mf* → index via runtime dynamic imports the browser's preload scanner never sees, causing a serial waterfall on cold cache. The existing fix-preload plugin is gated on `exposes` so pure hosts were unoptimized.

Adds `pluginHostInitPreload` which injects `<link rel="modulepreload">` for all four host-init chunks via `transformIndexHtml` (post, build-only), with no `exposes` gate so all hosts benefit automatically.


src/plugins/pluginHostInitPreload.ts
```diff
@@ -0,0 +1,76 @@
...
+ * the entry. `modulepreload` only fetches and compiles (never executes), so the MF init
+ * order is unchanged — only the network stalls are eliminated.
+ */
+import type { HtmlTagDescriptor, Plugin, Rollup } from 'vite';
+
+// Chunks emitted by the MF bootstrap that the browser's preload scanner never sees,
+// because they're loaded via runtime dynamic import() calls. Matching by Rollup chunk
+// name (not the hashed filename) survives content-hash changes. `_virtual_mf*` is
+// matched by prefix since it carries a generated app-specific suffix.
+const HOST_INIT_CHUNKS: ReadonlyArray<{ label: string; match: (name: string) => boolean }> = [
```
gioboa: Thanks for your help 🙏 
I'm moving this logic into an already present plugin, hold on a sec.

jgodi: Sounds good, jumping into a few meetings and I can push up changes and fix the CI checks too.

gioboa: Moving the logic to `src/plugins/pluginAddEntry.ts` is helpful. 

  - avoids extra plugin registration
  - reuses existing base / renderBuiltUrl handling
  - avoids noisy missing-chunk warnings
  - keeps behavior gated to hostInit

gioboa:
It's working fine:

Screenshot 2026-06-12 at 22 18 01

gioboa:

Screenshot 2026-06-12 at 22 18 39

gioboa (approved):
That's a nice catch 💯 
Thank you @jgodi

PR #814, opened 2026-06-13
fix: improve hostInit browsers support

### LINKED ISSUE #813: [Bug]: host app crashes in Safari with "initHost is not a function"
### Describe the bug

**Summary:** A Module Federation **host crashes on startup in Safari** (macOS + iOS) with `TypeError: initHost is not a function` — Chrome and Firefox load the same build fine. It happens whenever top-level await is lowered to an emulation: i.e. `vite-plugin-top-level-await` is in the plugin list, or `build.target` is below `es2022`. In that case the host bootstrap calls `initHost()` before the hostInit module has finished initialising it, and Safari loses the resulting timing race.

The rest of this issue is the detailed mechanism.

---

The host entry bootstrap emitted by `getBootstrapSource` ([`src/plugins/pluginAddEntry.ts`](https://github.com/module-federation/vite/blob/main/src/plugins/pluginAddEntry.ts)) destructures `initHost` straight off the dynamic `import()` of the hostInit chunk and calls it, **without awaiting that module's exported `__tla` promise**:

```js
// dist/assets/mf-entry-bootstrap-0-<hash>.js
(async () => {
  const { initHost } = await import("./hostInit-<hash>.js");
  await initHost();
})().then(() => import("./index-<hash>.js"));
```

This is correct only under **native** top-level await, where `await import()` does not resolve until the imported module's TLA settles.

When the hostInit chunk's TLA is **lowered to an emulation** — by `vite-plugin-top-level-await`, which is required for any `build.target` below `es2022` — `initHost` is assigned **asynchronously** inside the emulated `__tla` chain:

```js
// dist/assets/hostInit-<hash>.js (emulated TLA)
let a, l, c = Promise.all([f]).then(async () => { /* … */ l = async function(){ /* … */ }; a = l(); });
export { c as __tla, a as hostInitPromise, l as initHost };
```

`l` (`initHost`) is `undefined` until `c` (`__tla`) settles. The bootstrap awaits the **import** (which resolves after the module's *synchronous* evaluation) but **not** `__tla`, so it reads `initHost` while it is still `undefined`.

**Why it is browser-specific:** the single shipped artifact is identical across browsers — the difference is microtask ordering. **V8 (Chrome)** runs the `__tla` `.then` callback before the bootstrap's `await import()` continuation, so `initHost` is assigned in time. **JavaScriptCore (Safari)** settles the `import()` continuation first, so `initHost` is read while still `undefined`:

```
Unhandled Promise Rejection: TypeError: initHost is not a function. (In 'initHost()', 'initHost' is undefined)
```

**Note:** the remoteEntry loader inside `hostInit` already handles this correctly — `import("../remoteEntry.js").then(async r => (await r.__tla, r))`. The host bootstrap is the one place that does not await `__tla`.

**Fix sketch:** await the hostInit module's `__tla` before reading `initHost` — a no-op under native TLA (`__tla` is `undefined`). Mirrors the existing remoteEntry loader. PR with the fix + a regression test to follow.

```diff
-  const { initHost } = await import("./hostInit-<hash>.js");
+  const __mfHostInit = await import("./hostInit-<hash>.js");
+  await __mfHostInit.__tla;
+  const { initHost } = __mfHostInit;
```

### Version

1.16.6

### Reproduction

https://github.com/DanBeckDev/mfevite-safari-inithost-repro

### Relevant log output

```shell
# dist/assets/mf-entry-bootstrap-0-<hash>.js — reads initHost WITHOUT awaiting __tla:
(async () => {
  const { initHost } = await __mfImport("./hostInit-<hash>.js");
  await initHost();
})().then(() => __mfImport("./index-<hash>.js"));

# dist/assets/hostInit-<hash>.js — initHost (`l`) is assigned asynchronously inside
# the __tla chain and exported as a live binding, so it is undefined when the
# bootstrap reads it:
let a,l,c=Promise.all([(()=>{try{return f}catch{}})()]).then(async()=>{ /* … */ l=async function(){ /* … */ }; a=l() });
export{c as __tla,a as hostInitPromise,l as initHost};
```

### Validations

DanBeckDev:
Fixes #813.

## Problem

`getBootstrapSource` (`src/plugins/pluginAddEntry.ts`) emits a host bootstrap that destructures `initHost` from the hostInit chunk and calls it without awaiting that module's `__tla` promise:

```js
const { initHost } = await import("./hostInit-<hash>.js");
await initHost();
```

This assumes **native** top-level await. When the hostInit chunk's TLA is lowered to an emulation (`vite-plugin-top-level-await`, or any `build.target` below `es2022`), `initHost` is assigned asynchronously inside the `__tla` chain and is `undefined` until it settles. The `await import()` resolves after the module's *synchronous* evaluation, so reading `initHost` is a microtask race — V8 wins it, **JavaScriptCore (Safari) loses** → `TypeError: initHost is not a function` on host bootstrap.

Full details + minimal repro in #813.

## Change

Await the hostInit module's exported `__tla` promise before reading `initHost`:

```diff
     const importCode = `
 (async () => {
-  const { initHost } = await ${importExpression(initSrc)};
+  const __mfHostInit = await ${importExpression(initSrc)};
+  await __mfHostInit.__tla;
+  const { initHost } = __mfHostInit;
   ${preloadBlock}
 })().then(() => ${importExpression(entrySrc)});
 `;
```

- No-op under native TLA (`__tla` is `undefined` → `await undefined`)
- Mirrors the existing remoteEntry loader, which already does `await r.__tla`
- Covers both the standard and `useSystemImportFallback` bootstraps (the latter delegates to the same template)

## Test

Updated the three bootstrap assertions in `src/plugins/__tests__/pluginAddEntry.test.ts` to the new shape, and added a regression assertion that `__tla` is awaited **before** `initHost()` is called (a plain `toContain` would still pass if the ordering regressed).

Verified locally:

- Full suite: 507/507 pass
- `pnpm build` (tsdown) clean

## Not included

The SvelteKit inline-start (`rewriteSvelteKitInlineStart`) and Nuxt dev-mount injections use the same `…then(({ initHost }) => initHost())` shape and share this defect, but I can't repro those framework paths — happy to fold them in here or as a follow-up if you'd prefer.

### DIFF
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -271,9 +271,10 @@ const addEntry = ({
           .join(',')
       : '';
 
+    const hostInitCall = `${importExpression(initSrc)}.then(({ initHost }) => initHost())`;
     const preloadBlock = remotePreloads
       ? `
-  const runtime = await initHost();
+  const runtime = await ${hostInitCall};
   const __mfPreloadRemote = (remote) => {
     const pendingKey = "__mf_pending__" + remote;
     if (!__mfModuleCache.remote[pendingKey]) {
@@ -292,22 +293,10 @@ const addEntry = ({
   };
   const __mfRemotePreloads = [${remotePreloads}];
   await Promise.allSettled(__mfRemotePreloads);`
-      : `await initHost();`;
+      : `await ${hostInitCall};`;
 
-    // The hostInit chunk's top-level await may be lowered to an emulated
-    // `__tla` promise (e.g. by vite-plugin-top-level-await, or any build target
-    // below es2022). Under that lowering the dynamic import resolves after the
-    // module's *synchronous* evaluation — before its async init assigns
-    // `initHost` — so destructuring `initHost` immediately races the init and
-    // reads `undefined` in engines that settle the import microtask first
-    // (notably Safari/JavaScriptCore; V8 happens to win the race). Await the
-    // module's exported `__tla` promise before reading `initHost`. No-op under
-    // native TLA, where `__tla` is undefined.
     const importCode = `
 (async () => {
-  const __mfHostInit = await ${importExpression(initSrc)};
-  await __mfHostInit.__tla;
-  const { initHost } = __mfHostInit;
   ${preloadBlock}
 })().then(() => ${importExpression(entrySrc)});
 `;
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -294,9 +294,20 @@ const addEntry = ({
   await Promise.allSettled(__mfRemotePreloads);`
       : `await initHost();`;
 
+    // The hostInit chunk's top-level await may be lowered to an emulated
+    // `__tla` promise (e.g. by vite-plugin-top-level-await, or any build target
+    // below es2022). Under that lowering the dynamic import resolves after the
+    // module's *synchronous* evaluation — before its async init assigns
+    // `initHost` — so destructuring `initHost` immediately races the init and
+    // reads `undefined` in engines that settle the import microtask first
+    // (notably Safari/JavaScriptCore; V8 happens to win the race). Await the
+    // module's exported `__tla` promise before reading `initHost`. No-op under
+    // native TLA, where `__tla` is undefined.
     const importCode = `
 (async () => {
-  const { initHost } = await ${importExpression(initSrc)};
+  const __mfHostInit = await ${importExpression(initSrc)};
+  await __mfHostInit.__tla;
+  const { initHost } = __mfHostInit;
   ${preloadBlock}
 })().then(() => ${importExpression(entrySrc)});
 `;
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -271,10 +271,9 @@ const addEntry = ({
           .join(',')
       : '';
 
-    const hostInitCall = `${importExpression(initSrc)}.then(({ initHost }) => initHost())`;
     const preloadBlock = remotePreloads
       ? `
-  const runtime = await ${hostInitCall};
+  const runtime = await initHost();
   const __mfPreloadRemote = (remote) => {
     const pendingKey = "__mf_pending__" + remote;
     if (!__mfModuleCache.remote[pendingKey]) {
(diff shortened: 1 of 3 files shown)


PercussiveRepair (approved):

gioboa (commented):
Thanks @DanBeckDev
I would like automatically support all browsers without any additional plugins, no matter what the build output is.

DanBeckDev:
I can see the integration tests are failing, let me look into it.

DanBeckDev:
Thanks @gioboa 

I have fixed the failing CI.

On "support all browsers without any additional plugins, no matter the build output" — agreed that's the right bar. I'd split it into two pieces:

1. Bootstrap correctness (this PR). Even with vite-plugin-top-level-await added — today's way to ship sub-es2022 targets — the host bootstrap reads initHost before the hostInit module's __tla settles, so it crashes in Safari (Chrome just happens to win the microtask race). Awaiting __tla is correct for any output: a no-op under native TLA, the right gate under lowered TLA.
2. No plugin required (the broader goal). Sub-es2022 targets still need users to add vite-plugin-top-level-await, because MF's own hostInit / loadShare chunks emit top-level await and esbuild won't lower it below es2022. For MF to "just work" with no extra plugin on any target, MF would need to own that lowering for its generated chunks — and a bootstrap that awaits __tla is the precondition for that.

So this PR is (1), a standalone Safari fix. 

(2) would require a bit more thought in my opinion, either MF applying the TLA lowering to its own generated chunks, or restructuring hostInit to export a synchronously-available init promise the bootstrap awaits.

src/plugins/pluginAddEntry.ts
```diff
@@ -271,9 +271,10 @@ const addEntry = ({
           .join(',')
       : '';
 
+    const hostInitCall = `${importExpression(initSrc)}.then(({ initHost }) => initHost())`;
```
gioboa: This implementation is not speculating on `__tla`
Do you like it @DanBeckDev ? is it working in your scenario?

gioboa (commented):

Screenshot 2026-06-13 at 14 12 51

src/plugins/pluginAddEntry.ts
DanBeckDev: I am just testing this fix on my fork of the repo and the application with the problem, I should have an answer in ~15 ish minutes.

DanBeckDev: I tested this commit in WebKit (Playwright, which reproduces the real staging crash), it still throws the same TypeError: initHost is not a function. Same result as the stock build; the await __tla version was the only one that booted.

`.then(({ initHost }) => initHost())` reads initHost the moment the import resolves, which under lowered TLA is before the __tla chain assigns it, same race as the original, it just loses in JSC/Safari and wins in V8.

The two ways forward I see (but I am happy to hear your thoughts):
1. Await __tla before reading initHost (the original code)
2. Avoid __tla entirely,  but that needs hostInit to export an init signal created outside the lowered TLA chain, since once TLA is lowered every export (initHost, hostInitPromise) is assigned async too. Which is obviously a bigger change.

gioboa: I'll revert my commit and it's good to go 👍

gioboa: Thanks for your check 🙏

gioboa:
Thanks @DanBeckDev for you amazing support 👏

PR #815, opened 2026-06-13
fix: solve optimizeDeps clash

### LINKED ISSUE #812: [Bug]: react ends up in both optimizeDeps.include and exclude → "The entry point \"react\" cannot be marked as external" (regression in 1.16.6)
### Describe the bug

After updating from `1.14.5` to `1.16.7`, the host dev server (`vite`) fails during dep optimization:

```
✘ [ERROR] The entry point "react" cannot be marked as external
```

The host uses `@vitejs/plugin-react`, has `react-redux` as a dependency, and shares `react` as a singleton.

Root cause — `react` lands in **both** `optimizeDeps` lists:
- `@vitejs/plugin-react` adds `react` to `optimizeDeps.include`.
- Since #800 (1.16.6, *"avoid duplicate react with redux in dev mode"*), the plugin adds `react` to `optimizeDeps.exclude` when `react-redux` is present.

An included dep is an esbuild entry point; an excluded one is externalized → esbuild throws. On `1.14.5` (before #800) `react` was only in `include`, so it worked. Confirmed by logging the resolved config: `include` contains `react`, `exclude` is `["react", "react/compiler-runtime"]`.

**Workaround** — a host plugin reconciling the lists (include wins):

```ts
{
  name: 'mf-reconcile-optimizedeps',
  configResolved(config) {
    const od = config.optimizeDeps;
    if (od?.include && od?.exclude) {
      const included = new Set(od.include);
      od.exclude = od.exclude.filter((dep) => !included.has(dep));
    }
  }
}
```

Note: the opposite (exclude wins) lets dev start but then crashes at runtime with `Uncaught ReferenceError: module is not defined` (white screen) — `react` gets externalized while `react-dom` stays optimized, breaking CJS interop. Keeping `react` in `include` avoids both.

### Version

`@module-federation/vite` 1.16.7, Vite 6.4.1

### Reproduction

https://github.com/Sultanyaron/federation-redux-example

A minimal `host` + `remote`, both with `@vitejs/plugin-react`, `react-redux` as a dependency, and `react` shared as a singleton:

```bash
cd host
npm install
npm run dev
```

The host dev server fails on startup with the error above.

### Relevant log output

```shell
✘ [ERROR] The entry point "react" cannot be marked as external
error: Build failed with 1 error:
error: The entry point "react" cannot be marked as external
    at failureErrorWithLog (.../esbuild/lib/main.js:1467:15)
```

### Validations

- [x] Read the docs
- [x] Read the common issues list
- [x] Checked there isn't already a duplicate issue
- [x] The provided reproduction is a minimal reproducible example of the bug

gioboa:
Close #812
#792 is still fixed.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1017,6 +1017,28 @@ describe('vite:module-federation-early-init', () => {
     expect(config.optimizeDeps.include).not.toContain('react');
   });
 
+  it('keeps included shared react out of dev optimizeDeps exclude when react-redux is installed', () => {
+    hasPackageDependencyMock.mockImplementation(
+      (dependency: string): boolean => dependency === 'react-redux'
+    );
+    const plugin = getEarlyInitPluginWithReactShared();
+    const config: any = {
+      root: process.cwd(),
+      optimizeDeps: {
+        include: ['react'],
+        exclude: [],
+      },
+    };
+
+    runConfig(plugin, {} as ConfigPluginContext, config, {
+      command: 'serve',
+      mode: 'test',
+    });
+
+    expect(config.optimizeDeps.include).toContain('react');
+    expect(config.optimizeDeps.exclude).not.toContain('react');
+  });
+
   it('leaves ENV_TARGET undefined for Astro mixed builds', () => {
     hasPackageDependencyMock.mockImplementation(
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-06-13 at 15 31 08

PR #819, opened 2026-06-15
fix: fix cyclic workspace singleton sharing

### LINKED ISSUE #818: [Bug]: Cyclic shared singletons get `undefined` bindings in production builds → `Cannot read properties of undefined` at startup (1.16.8)
### Describe the bug

In a **production build**, when several workspace packages are shared as
**singletons** and they form a **dependency cycle**, the generated eager
share-init loop can load a **consumer before the producer it depends on**. The
producer's `loadShare` wrapper assigns its exported bindings inside a deferred
`initPromise.then(() => import(...).then(...))` continuation, so a consumer that
uses one of those bindings **synchronously at module-evaluation time** sees
`undefined` and the app crashes on startup:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getLogger')
```

Because the crashing module aborts, its own exports never initialise, producing
the follow-on errors in downstream chunks:

```
Uncaught (in promise) TypeError: <fn> is not a function
```

> The simpler **non-cyclic** case (a single singleton consumed synchronously)
> was a regression in **1.16.6** and is **fixed in 1.16.8**. The **cyclic**
> variant described here still reproduces on **1.16.8**.

## Reproduction

Minimal repo: https://github.com/francesc79/mf-1168-shared-init-order-repro

```bash
pnpm install
pnpm --filter @spike/host build
pnpm --filter @spike/host preview     # http://localhost:4173
```

Open the page.

It is a pnpm workspace with four packages forming a cycle
`gamma → beta → alpha → gamma`, all shared as singletons by the host:

```
packages/
  alpha/   @spike/alpha  — logging service; default export is an object with getLogger()
                           (imports @spike/gamma -> closes the cycle)
  beta/    @spike/beta   — imports alpha's default export and calls
                           logging.getLogger('beta') AT MODULE TOP LEVEL
  gamma/   @spike/gamma  — imports beta's getValue(); the host entry renders through it
  host/    @spike/host   — MF host; shares alpha/beta/gamma as singletons
```

### Expected

```
OK — beta getValue() = 42
```

### Actual (1.16.8)

The page stays on `loading…`, console shows:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getLogger')
    at assets/src-XXXX.js          (the `beta` chunk)
```

## Environment

| | |
|---|---|
| `@module-federation/vite` | **1.16.8** |
| `@module-federation/enhanced` | 2.5.1 (`runtime: 'enhanced'`) |
| `vite` | 8.x (Rolldown) |
| `shareStrategy` | `version-first` |
| node | 24.x |
| package manager | pnpm workspaces |

### 1. The cycle defeats dependency-first ordering

With a cycle there is no valid topological order, so the generated eager init
loop in `hostInit-….js` keeps the registration order and loads the **consumer
before the producer**:

```js
// hostInit-….js  (sequential, awaited)
const shared = {
  "@spike/beta":  { shareConfig: { singleton: true, requiredVersion: "^0.0.1" } },
  "@spike/gamma": { shareConfig: { singleton: true, requiredVersion: "^0.0.1" } },
  "@spike/alpha": { shareConfig: { singleton: true, requiredVersion: "^0.0.1" } }, // <- LAST
};
for (const [name, cfg] of Object.entries(shared)) {
  cache.share[key] === undefined &&
    await runtime.loadShare(name, …).then(/* cache.share[key] = … */);
}
```

`loadShare('@spike/beta')` evaluates the `beta` chunk, whose top level needs
`@spike/alpha` — but `@spike/alpha` is loaded **last**, so its share cache entry
is still empty at that moment.

### 2. The producer wrapper hands out a deferred (undefined) binding

Generated wrapper for `@spike/alpha` (`_virtual_mf…alpha__loadShare__.js`):

```js
let s;                                  // the default-export binding (undefined)
const c = (mod) => { /* … */ s = mod.default ?? mod };
const l = cache.share["@spike/alpha"];
l === undefined
  ? initPromise.then(() =>
      import("./src-alpha.js").then((mod) => {  // <-- async continuation
        cache.share["@spike/alpha"] = normalize(mod);
        c(cache.share["@spike/alpha"]);          // <-- s assigned HERE, later
      }))
  : c(l);
export { s as t };                      // <-- exported binding, undefined until the .then runs
```

Because the share cache is empty when the wrapper is first evaluated, it takes
the **deferred** branch: `s` is only assigned inside the `initPromise.then(...)`
continuation. Note the decoupling — `loadShare` later populates
`cache.share["@spike/alpha"]`, but that does **not** retroactively assign the
wrapper's local `s`.

### 3. The consumer reads the binding synchronously → throws

Generated `beta` chunk (`src-XXXX.js`):

```js
import { t as logging } from "./_virtual_mf…alpha__loadShare__.js";
logging.getLogger("beta").info("beta module evaluated");  // logging === undefined -> THROW
export function getValue() { return 42; }
export const BETA_LABEL = "beta";
```

`logging` is the still-`undefined` `s` binding → `logging.getLogger` throws.
Because `beta`'s evaluation aborts, its `getValue` / `BETA_LABEL` exports never
initialise, so any later consumer of `beta` throws `<fn> is not a function`.

## Why it matters

Real monorepos that share their workspace packages as singletons almost always
have cycles between those packages (e.g. a `utility ⇄ core ⇄ framework`
triangle, where the lowest-level lib obtains a logger at module top level). On
1.16.8 the deferred-assignment wrapper makes that arrangement crash at startup
in production builds.

## Suggested fix

Either:

- For the production-build path, restore the **top-level `await import(...)`**
  form so the wrapper's exported bindings are live before the module finishes
  evaluating (this is what fixed the simple non-cyclic case); or
- Guarantee the share cache for every shared singleton is populated **before**
  any consumer wrapper can be evaluated, even when the shared graph is cyclic
  (don't rely on dependency-first ordering, which a cycle makes impossible).

## Notes

- No framework (React/etc.) is needed to reproduce; the failure is purely in the
  generated MF shared-init code.
- The deterministic trigger is the **cycle** + a **top-level synchronous use** of
  a shared singleton's binding. Break the cycle, or make the use lazy/async, and
  the page renders `OK`.

### Version

8.0.16

### Reproduction

https://github.com/francesc79/mf-1168-shared-init-order-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #818

Fixes workspace singleton sharing when shared packages form a dependency cycle.
In that case, lazy loadShare wrappers can expose uninitialized copied bindings during module evaluation, causing startup crashes. The fix detects cyclic workspace singleton graphs and uses live ESM re-exports only for those packages, avoiding stale or undefined wrapper locals. 

`Non-cyclic workspace singletons keep the existing lazy/cache-first path, preserving normal Module Federation singleton negotiation and preventing duplicate initialization`


gioboa (commented):
@francesc79 Can you check this pre-release please?

gioboa (commented):

Image

francesc79:
@gioboa  Thanks for the fix here. I verified that the original minimal repro is fixed by this PR.

  I also updated the repro repository with a new remaining issue, so it now demonstrates a root package + shared subpath ordering problem directly.

  Repro repo:
  `/home/zazos/Development/flex4/spikes/mf-1168-shared-init-order-repro`

  With this PR version:

  ```js
  "@module-federation/vite": "https://pkg.pr.new/@module-federation/vite@819"

  the host shares:

  shared: {
    '@spike/core': { singleton: true },
    '@spike/hooks': { singleton: true },
    '@spike/hooks/media': { singleton: true },
  }

  @spike/core depends on the root package @spike/hooks and synchronously uses an export from it at module evaluation time:

  import { BaseEvent } from '@spike/hooks';

  export class CurrentRowChangedEvent extends BaseEvent {}

  The generated hostInit-*.js preload order is currently:

  @spike/hooks/media
  @spike/core
  @spike/hooks

  That makes @spike/core evaluate before the root @spike/hooks share cache is populated, so the generated @spike/hooks loadShare wrapper still exports BaseEvent as undefined.

  Runtime error:

  Uncaught (in promise) TypeError: Class extends value undefined is not a constructor or null

  I think the issue is in orderSharedDependenciesFirst: getPackageName('@spike/hooks') and getPackageName('@spike/hooks/media') both resolve to @spike/hooks, so the subpath shared key
  overwrites the root shared key in sharedKeyByPackageName. Then @spike/core's dependency on @spike/hooks is ordered against @spike/hooks/media instead of the root package.

  I verified locally that preferring the explicit root shared key fixes the repro:

  const sharedKeyByPackageName = new Map();

  sharedPackages.forEach((pkg) => {
    const packageName = getPackageName(pkg);
    const existing = sharedKeyByPackageName.get(packageName);

    if (!existing || pkg === packageName) {
      sharedKeyByPackageName.set(packageName, pkg);
    }
  });

  With that change, the generated order becomes:

  @spike/hooks
  @spike/core
  @spike/hooks/media

  and the page renders successfully.

  Do you want me to open a separate issue for this remaining case, or would you prefer to track it in this PR?

gioboa:
let me continue with this PR

gioboa:
@francesc79 I added the use case in the official tests. can you verify this new fix?

francesc79:
Thanks for the fix. I can confirm that the second issue also seems to be resolved.

However, during additional testing I still found a remaining problem. Based on a first analysis, it looks like there is still a race condition in the generated shared wrapper for workspace singletons.

When the application is loaded directly (page reload), it can remain stuck on the splash screen even though the MF assets are loaded successfully. The failure appears to happen because some shared named exports are still undefined when they are consumed by top-level code.

A deeper investigation points to the shared wrapper generated by @module-federation/vite, where live exports are initialized only after initPromise.then(...) completes. This seems to allow consumers to access the exports before they are populated.

I'll perform a more detailed analysis tomorrow and provide a more precise root-cause investigation with additional findings.

gioboa:
> I'll perform a more detailed analysis tomorrow and provide a more precise root-cause investigation with additional findings.

if you can recreate the issue with the vite-vite example in this repo, it will be amazing. Thanks for your help btw.

PR #820, opened 2026-06-15
fix: fix up React federation order edge case

gioboa:
Fixes a React federation plugin-order edge case where react could end up in both optimizeDeps.include and exclude, causing esbuild to externalize an entry point and crash. Federation now skips optimizer entry-point proxying, and normalizeOptimizeDeps removes include/exclude conflicts after all plugins mutate config.


gioboa (commented):

Image

PR #822, opened 2026-06-15
fix: use share scope in loadShare cache key to prevent cross-scope collisions

### LINKED ISSUE #821: [Bug]: loadShare cache key ignores share scope — breaks multi-scope setups
## Summary

The `loadShare` virtual module uses `__mf_module_cache__.share` with keys like `"react"` or `"react@18.3.1"`. These keys don't include the share scope, so when multiple scopes exist on the same page (e.g., `react-18` and `react-19`), whichever scope populates the cache first wins. All subsequent lookups — regardless of scope — return the wrong version.

## Reproduction

Three Vite apps in a monorepo:

**vite-host** (React 19.2.7):
```js
// vite.config.js
federation({
  name: 'Host',
  remotes: {
    Remote19: 'http://localhost:3001/mf-manifest.json',
    Remote18: 'http://localhost:3002/mf-manifest.json',
  },
  shared: {
    react: { singleton: true, shareScope: 'react-19' },
    'react-dom': { singleton: true, shareScope: 'react-19' },
  },
})
```

**vite-remote-19** (React 19.2.7, port 3001):
```js
federation({
  name: 'Remote19',
  exposes: { './Widget': './src/Widget.tsx' },
  shared: {
    react: { singleton: true, shareScope: 'react-19' },
    'react-dom': { singleton: true, shareScope: 'react-19' },
  },
})
```

**vite-remote-18** (React 18.3.1, port 3002) — uses `@module-federation/bridge-react` for cross-major isolation:
```js
federation({
  name: 'Remote18',
  exposes: { './Widget': './src/BridgeWidget.tsx' },
  shared: {
    react: { singleton: true, shareScope: 'react-18' },
    'react-dom': { singleton: true, shareScope: 'react-18' },
  },
})
```

**Steps:**
1. Start all three dev servers
2. Open the host in a browser

**Expected:** Remote18 loads React 18.3.1 from the `react-18` scope. Remote19 loads React 19.2.7 from the `react-19` scope. Both work.

**Actual:** Remote18 gets React 19.2.7 from the cache (keyed as just `"react"`), crashes with `TypeError: Cannot read properties of undefined (reading 'ReactCurrentDispatcher')`.

## Root Cause

`getSharedCacheKey()` in `src/utils/packageUtils.ts`:

```ts
export function getSharedCacheKey(pkg: string, shareItem: ShareItem) {
  return shareItem.shareConfig.singleton || !shareItem.version
    ? pkg                           // "react" — no scope
    : `${pkg}@${shareItem.version}`;
}
```

The generated loadShare module does:
```js
let exportModule = __mfModuleCache.share["react"]  // flat key, no scope
```

Additionally, three inline cache-key constructions in `virtualRemoteEntry.ts` (the `__FEDERATION__.__SHARE__` bridge loop, the post-init `import:false` loop, and the `hostAutoInit` preload path) build keys the same way, bypassing `getSharedCacheKey()`.

The MF runtime's own `shareScopeMap` IS scope-aware. The Vite plugin's generated code just doesn't use the scope.

## Suggested Fix

**`src/utils/packageUtils.ts`** — prefix with `shareItem.scope` (already normalized to a string, defaults to `'default'`):

```ts
export function getSharedCacheKey(pkg: string, shareItem: ShareItem) {
  const prefix = `${shareItem.scope}:`;
  return shareItem.shareConfig.singleton || !shareItem.version
    ? `${prefix}${pkg}`
    : `${prefix}${pkg}@${shareItem.version}`;
}
```

**`src/virtualModules/virtualRemoteEntry.ts`** — emit a runtime helper for the three inline sites that can't call the build-time function:

```js
const __mfGetSharedCacheKey = (pkg, singleton, version, scope) => {
  const prefix = scope ? (scope + ":") : "";
  return singleton || !version ? prefix + pkg : prefix + pkg + "@" + version;
};
```

Also add `version` and `scope` to the `usedShared` preload config so the runtime helper has access:

```ts
scope: ${JSON.stringify(shareItem.scope)},
```

### Files changed

| File | What |
|------|------|
| `src/utils/packageUtils.ts` | `getSharedCacheKey()`: add scope prefix |
| `src/virtualModules/virtualRemoteEntry.ts` | Emit `__mfGetSharedCacheKey` helper; update 3 inline cache key sites; add scope/version to preload config |

Single-scope setups get a `"default:"` prefix — different string, same dedup semantics, no behavioral change.

## Environment

- `@module-federation/vite` 1.16.8
- Vite 8.x
- React 18 + React 19 on the same page via bridge components and separate share scopes

### Version

Vite 8.x

### Reproduction

https://github.com/RalphK66/repro-mf-vite-scope-unaware-cache-key

### Relevant log output

```shell

```

### Validations

RalphK66:
Closes #821

### DIFF
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -94,7 +94,7 @@ describe('shared dependencies', () => {
     const remoteEntry = findChunk(output, 'remoteEntry');
     const loadShare = output.output
       .filter(isRollupChunk)
-      .find((chunk) => chunk.code.includes('__mfModuleCache.share["mock-shared-dep"]'));
+      .find((chunk) => chunk.code.includes('__mfModuleCache.share["default:mock-shared-dep"]'));
 
     expect(remoteEntry).toBeDefined();
     expect(loadShare).toBeDefined();
@@ -104,7 +104,7 @@ describe('shared dependencies', () => {
     expect(allCode).toContain('__mfModuleCache.share[cacheKey]');
     expect(allCode).not.toContain('initRes.loadShare(pkg');
     expect(loadShare!.code).toContain('initPromise.then');
-    expect(loadShare!.code).toContain('__mfModuleCache.share["mock-shared-dep"]');
+    expect(loadShare!.code).toContain('__mfModuleCache.share["default:mock-shared-dep"]');
     expect(loadShare!.code).toContain('init');
     expect(loadShare!.code).not.toContain('await initPromise');
     expect(allCode).not.toContain('from"mock-shared-dep"');
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -42,9 +42,13 @@ vi.mock('fs', async (importOriginal) => {
 vi.mock('../../utils/packageUtils', () => ({
   getSharedCacheKey: (
     pkg: string,
-    shareItem: { version?: string; shareConfig: { singleton?: boolean } }
-  ) =>
-    shareItem.shareConfig.singleton || !shareItem.version ? pkg : `${pkg}@${shareItem.version}`,
+    shareItem: { version?: string; scope?: string; shareConfig: { singleton?: boolean } }
+  ) => {
+    const prefix = `${shareItem.scope || 'default'}:`;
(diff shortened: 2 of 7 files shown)


gioboa (commented):

Image

gioboa (approved):
Thanks @RalphK66 that's a nice fix  🎉

PR #825, opened 2026-06-16
fix: bind acyclic workspace shared singletons eagerly when consumed by a peer (#823)

### LINKED ISSUE #823: [Bug]: Acyclic shared-singleton graph crashes at startup when a package is shared together with one of its subpath exports (producer registered after its consumers)
### Describe the bug

With runtime: 'enhanced' + shareStrategy: 'version-first', registering a workspace package and one of its subpath exports as shared singletons (e.g. shared-toolkit and shared-toolkit/extra) defeats the dependency-first share ordering. The root package (the one that actually exports the value, e.g. a log object) gets registered after its consumers in the generated hostAutoInit loop. A consumer that reads the shared export at module-evaluation time then reads undefined:

Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getLogger')

The shared dependency graph is fully acyclic — no cycle is involved. This looks like the root-vs-subpath ordering case from #818 still being reachable.

Minimal reproduction: https://github.com/francesc79/mf-shared-init-order-repro

Environment

- @module-federation/vite: 1.16.8 (the build that includes the #818/#819 fixes — installed via https://pkg.pr.new/@module-federation/vite@819)
- @module-federation/enhanced: 2.5.1
- vite: 8.0.16 (Rolldown)
- host: runtime: 'enhanced', shareStrategy: 'version-first', exposes an entry, consumes a remote
- Node 24, production build served with vite preview (not the dev server)

Reproduction steps

git clone https://github.com/francesc79/mf-shared-init-order-repro
cd mf-shared-init-order-repro
pnpm install
pnpm build      # builds remote-app then host-app
pnpm preview    # host on :3000, remote on :3001

Open http://localhost:3000 with DevTools → Network → Disable cache, then reload.

Package graph (acyclic)

shared-toolkit   (leaf; exports `log.getLogger`; ALSO exports subpath `shared-toolkit/extra`)
shared-hooks     (leaf)
shared-domain    -> shared-toolkit, shared-hooks      (reads log.getLogger at MODULE TOP LEVEL)
shared-widgets   -> shared-toolkit, shared-toolkit/extra, shared-hooks
shared-shell     -> shared-domain, shared-widgets, shared-toolkit, shared-hooks
host-app         -> shared-shell, shared-domain, shared-toolkit   (exposes ./HostEntry, consumes remote-app)
remote-app       -> shared-shell, shared-domain, shared-toolkit   (exposes ./RemotePanel)

Host shared config (all singletons):

shared: {
  'shared-toolkit':       { singleton: true },
  'shared-toolkit/extra': { singleton: true },   // <-- subpath of the same package
  'shared-hooks':         { singleton: true },
  'shared-domain':        { singleton: true },
  'shared-widgets':       { singleton: true },
  'shared-shell':         { singleton: true },
}

Expected

The host renders; every shared singleton's exported bindings are assigned before any consumer module evaluates.

Actual

The page stays at loading… and throws Cannot read properties of undefined (reading 'getLogger') (deterministically in this minimal setup). A consumer that reads a shared function binding at top level instead throws <fn> is not a function.

Root cause (from the generated bundle)

1. The producer is registered last. The generated hostAutoInit chunk iterates shares in this order:

"shared-toolkit/extra", "shared-hooks", "shared-domain", "shared-widgets", "shared-shell", "shared-toolkit"

1. The root shared-toolkit (which exports log) is ordered after shared-domain/shared-widgets/shared-shell, its consumers. Without the subpath singleton, the order is correctly producer-first (shared-toolkit, shared-hooks, shared-domain, …) and there is no crash. This points at getPackageName('shared-toolkit/extra') === 'shared-toolkit' colliding with the root entry in orderSharedDependenciesFirst.

1. Quick check:
grep -oE '"shared-[a-z/]+":\{shareConfig' packages/host-app/dist/assets/*hostAutoInit*.js
2. The loadShare wrapper assigns its exported bindings asynchronously. The generated shared-toolkit wrapper only assigns log inside the deferred branch:

share['shared-toolkit'] === void 0
  ? initPromise.then(() => import('./src-*.js').then(e => { share['shared-toolkit'] = …; /* assign log */ }))
  : /* assign synchronously */
3. The consumer reads at module-evaluation time. shared-domain runs const logger = log.getLogger('shared-domain') at module top level. Because (1) registers the producer after the consumer, the wrapper takes the deferred branch and log is still undefined when shared-domain evaluates → the TypeError.

So the crash is purely a share-registration ordering problem, reachable on an acyclic graph via a root + subpath shared singleton.

Why this matters

Sharing a package together with a subpath (pkg + pkg/sub) as singletons is a common, valid configuration. A real (much larger, acyclic) app that does this fails intermittently at startup with the same error — the minimal repro just pins the same ordering defect deterministically.

Expected fix (either is fine)

1. In orderSharedDependenciesFirst, ensure the root package entry is ordered before its consumers even when subpath entries of the same package are also shared (don't let a pkg/sub entry displace pkg in the package-name → key mapping); or
2. Detect at build time that a shared singleton's exported bindings cannot be guaranteed assigned before a consumer evaluates and emit a clear, actionable error instead of shipping output that throws at runtime.

Notes

- Reproduces against the production build (vite preview), not the dev server.
- No dependency cycle is involved (the graph is a DAG).
- runtime: 'enhanced' / shareStrategy: 'version-first' and the exposed entry / remote are kept for fidelity; the dominant trigger is the root + subpath singleton ordering.

### Version

8.0.16

### Reproduction

https://github.com/francesc79/mf-shared-init-order-repro

### Relevant log output

```shell

```

### Validations

MohammadYusif:
## Summary

- Fixes #823: a Module Federation host crashes at startup with `Cannot read properties of undefined (reading 'getLogger')` on an **acyclic** shared-singleton graph where a workspace package is shared together with one of its subpath exports (`pkg` + `pkg/sub`) and is consumed by a peer shared singleton.
- A workspace singleton's `loadShare` wrapper binds its exports either eagerly (synchronously, safe at module-evaluation time) or lazily (deferred via `initPromise.then`). The eager path was only taken when `pkg` had a *cyclic* workspace-singleton dependency, so acyclic graphs where a consumer reads the producer's bindings at evaluation time still raced and read `undefined`.

## Changes

- `src/virtualModules/virtualShared_preBuild.ts`: generalize the gating predicate `hasCyclicWorkspaceSingletonDependency` → `isWorkspaceSingletonConsumedByPeer`. A workspace singleton now binds eagerly whenever *any peer shared singleton* transitively depends on it. This subsumes the existing cyclic behaviour (each node in a cycle is reached from its peer) and additionally covers the acyclic root + subpath scenario. Leaf / top-level-only singletons remain lazy.
- `src/virtualModules/__tests__/virtualShared_preBuild.test.ts`: add a regression test (`emits eager local re-exports for acyclic workspace singletons consumed by a peer`) reproducing #823, plus mock wiring for the `workspace-producer` / `workspace-consumer` packages. Verified the test fails without the source change and passes with it.

Fixes #823

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -85,6 +85,12 @@ vi.mock('../../utils/packageUtils', () => ({
     if (pkg === 'workspace-cycle-b') {
       return '/repo/packages/workspace-cycle-b/src/index.ts';
     }
+    if (pkg === 'workspace-producer' || pkg.startsWith('workspace-producer/')) {
+      return '/repo/packages/workspace-producer/src/index.ts';
+    }
+    if (pkg === 'workspace-consumer') {
+      return '/repo/packages/workspace-consumer/src/index.ts';
+    }
   }),
   getInstalledPackageJson: vi.fn((pkg: string, opts?: { fromResolvedEntry?: string }) => {
     if (opts?.fromResolvedEntry?.includes('/repo/packages/workspace-shared-lib/')) {
@@ -120,6 +126,23 @@ vi.mock('../../utils/packageUtils', () => ({
         packageJson: {
           name: 'workspace-cycle-b',
           dependencies: { 'workspace-cycle-a': 'workspace:*' },
+        },
+      };
+    }
+    if (opts?.fromResolvedEntry?.includes('/repo/packages/workspace-producer/')) {
+      return {
+        path: '/repo/packages/workspace-producer/package.json',
+        dir: '/repo/packages/workspace-producer',
+        packageJson: { name: 'workspace-producer' },
+      };
+    }
+    if (opts?.fromResolvedEntry?.includes('/repo/packages/workspace-consumer/')) {
+      return {
+        path: '/repo/packages/workspace-consumer/package.json',
+        dir: '/repo/packages/workspace-consumer',
+        packageJson: {
+          name: 'workspace-consumer',
+          dependencies: { 'workspace-producer': 'workspace:*' },
         },
       };
     }
@@ -192,6 +215,8 @@ vi.mock('fs', () => ({
       filePath.endsWith('/repo/packages/workspace-name-mismatch/package.json') ||
       filePath.endsWith('/repo/packages/workspace-cycle-a/package.json') ||
       filePath.endsWith('/repo/packages/workspace-cycle-b/package.json') ||
+      filePath.endsWith('/repo/packages/workspace-producer/package.json') ||
+      filePath.endsWith('/repo/packages/workspace-consumer/package.json') ||
       filePath.endsWith('/repo/packages/custom-shared-source/index.ts')
   ),
   readFileSync: vi.fn((filePath: string) => {
@@ -531,6 +556,12 @@ vi.mock('module', async (importOriginal) => {
           }
           if (pkg === 'workspace-cycle-b') {
             return '/repo/packages/workspace-cycle-b/src/index.ts';
+          }
+          if (pkg === 'workspace-producer' || pkg.startsWith('workspace-producer/')) {
+            return '/repo/packages/workspace-producer/src/index.ts';
+          }
+          if (pkg === 'workspace-consumer') {
+            return '/repo/packages/workspace-consumer/src/index.ts';
           }
           if (
             pkg === 'mock-package-reexport-type' ||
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks @MohammadYusif for your help 👏 
Cloud you test this pre-release @francesc79 ?

gioboa (commented):

Image

francesc79:
Confirmed fixed!
Thank you @gioboa @MohammadYusif for the quick turnaround!

gioboa (approved):
Thanks @MohammadYusif LGTM 💪 
Thank you @francesc79 for your fast check.

PR #827, opened 2026-06-16
fix: rethrow failed dev remote loads

### LINKED ISSUE #824: [Bug]: loaded-first stragegy does not error component when it fails to load
### Describe the bug

If I have `loaded-first` share strategy set and the remote fails to load it doesn't error and the Suspense fallback remains displayed. This prevents displaying the error route fallback in react router

### Version

7.3.5

### Reproduction

https://github.com/JamesMHenderson/mf-vite

### Relevant log output

```shell

```

### Validations

gioboa:
Close #824

Fixes loaded-first dev remote failures being swallowed, which left Suspense stuck on its fallback instead of letting React Router show the error route.
  - Clear pending remote cache on loadRemote() failure
  - Rethrow the error for React lazy/Suspense consumers
  - Prevent internal sync listener unhandled rejection noise

### DIFF
--- a/src/virtualModules/__tests__/virtualRemotes.test.ts
+++ b/src/virtualModules/__tests__/virtualRemotes.test.ts
@@ -190,6 +190,17 @@ describe('generateRemotes', () => {
     expect(exports.__moduleExports).toBe(remoteModule);
   });
 
+  it('rejects remote pending when dev loadRemote fails', async () => {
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

JamesMHenderson:
@gioboa  Thanks for looking into these issues that I have raised and fixing them so quickly!

PR #828, opened 2026-06-16
fix: reuse singleton share cache aliases

### LINKED ISSUE #826: [Bug]: react fails to share when share is specified as an empty object
### Describe the bug

When I have a host that has the config

```
shared: {
          react: {
            singleton: true,
            requiredVersion: dependencies.react,
            version: dependencies.react,
          },
          'react-dom': {
            singleton: true,
            requiredVersion: dependencies['react-dom'],
            version: dependencies['react-dom']
          },
        },
```

and the remote has the config

```
shared: {
        react: {
        },
        'react-dom': {
        },
      },
```

When I access a useState hook in the remote I get the error ` Uncaught TypeError: can't access property "useState", dispatcher is null`

This doesn't happen if the remote has specific versions for it's shared dependencies

This happens when I use the package version 1.16.8, but doesn't using 1.16.1

### Version

7.3.5

### Reproduction

https://github.com/JamesMHenderson/mf-vite/tree/feat/shared-empty

### Relevant log output

```shell
Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
```

### Validations

gioboa:
Close #826

Fixes remote shared modules with empty config loading local React instead of host singleton.

  - Alias host singleton cache entries to the remote’s expected versioned key.
  - Prevents duplicate React instances and invalid hook calls.
  - Covers both newly bridged runtime providers and already cached singleton shares.
  - Adds regression test for versioned remote share lookup.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -686,4 +686,90 @@ describe('virtualRemoteEntry', () => {
       'const cacheKey = __mfGetSharedCacheKey(pkg, share.shareConfig?.singleton, share.version, share.scope);'
     );
   });
+
+  it('aliases external singleton providers to the remote share cache key', async () => {
+    normalizedSharedMock.mockReturnValue({
+      react: {
+        name: 'react',
+        from: '',
+        version: '18.3.1',
+        scope: 'default',
+        shareConfig: {
+          singleton: false,
+          requiredVersion: '^18.3.1',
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #830, opened 2026-06-16
fix: prefer installed shared version over requiredVersion (#829)

### LINKED ISSUE #829: [Bug]: shared requiredVersion is inferred as provider version before installed package version
### Describe the bug

`normalizeShareItem()` in `src/utils/normalizeModuleFederationOptions.ts` appears to give `inferVersionFromRequiredVersion(requiredVersion)` higher precedence than the installed package version.

Current logic in `src/utils/normalizeModuleFederationOptions.ts`:

```ts
const explicitVersion =
  typeof shareItem === 'object'
    ? shareItem.version || inferVersionFromRequiredVersion(shareItem.requiredVersion)
    : undefined;

const version = explicitVersion || searchPackageVersion(key);
```

This makes `requiredVersion` act as the provider `version` whenever `version` is not explicitly configured and `requiredVersion` contains a concrete semver value.

I understand `inferVersionFromRequiredVersion()` may be intended as a fallback, especially for cases where version resolution fails or `import: false` is used. The issue is the fallback order: the installed package version should be preferred before inferring a provider version from a consumer constraint.

Minimal example:

```json
{
  "dependencies": {
    "react": "19.2.7"
  }
}
```

```ts
federation({
  name: "requiredVersionRemote",
  filename: "remoteEntry.js",
  manifest: true,
  dts: false,
  exposes: {
    "./App": "./src/App.js",
  },
  shared: {
    react: {
      singleton: true,
      requiredVersion: "^19.1.1",
    },
  },
});
```

Expected generated shared metadata:

```json
{
  "name": "react",
  "version": "19.2.7",
  "requiredVersion": "^19.1.1"
}
```

Actual generated shared metadata:

```json
{
  "name": "react",
  "version": "19.1.1",
  "requiredVersion": "^19.1.1"
}
```

`requiredVersion` is a consumer constraint, while `version` should describe the actual package version provided by the build. If the inferred `requiredVersion` value is advertised as the provider version, the runtime receives provider metadata that does not match the module this build can actually provide.

Depending on `strictVersion`, share strategy, and fallback behavior, this can surface as false positives or false negatives in semver matching:

- If the installed package is `19.0.0` and `requiredVersion` is `^19.1.1`, the provider may be advertised as `19.1.1`, so the metadata can satisfy a range even though the local fallback/provider would not.
- If the installed package is `19.2.7` and the provider is advertised as `19.1.1`, a host requiring `^19.2.0` may fail semver matching even though the installed fallback/provider would satisfy it.

I think a safer precedence would be:

```ts
const explicitVersion =
  typeof shareItem === 'object' ? shareItem.version : undefined;

const inferredVersion =
  typeof shareItem === 'object'
    ? inferVersionFromRequiredVersion(shareItem.requiredVersion)
    : undefined;

const version = explicitVersion || searchPackageVersion(key) || inferredVersion;
```

This keeps `inferVersionFromRequiredVersion()` as a last-resort fallback, but avoids overriding the actual installed provider version.

I can submit a small PR with a regression test if this behavior is not intentional.

The reproduction repository README includes the exact setup and verification command:

```bash
pnpm install
pnpm repro
```

The reproduction builds a Vite remote, reads `dist/mf-manifest.json`, and compares the generated React provider `version` against `node_modules/react/package.json`.

### Version

1.16.8

### Reproduction

https://github.com/dmchoi77/mf-vite-required-version-precedence-repro

### Relevant log output

```shell
> mf-vite-required-version-precedence-repro@1.0.0 repro
> vite build && node scripts/inspect-shared-version.mjs

vite v8.0.16 building client environment for production...
✓ built

Reproduced: generated shared metadata advertises the wrong provider version.

Installed react version: 19.2.7
Manifest react version:  19.1.1
Manifest requiredVersion: ^19.1.1
Expected provider version: 19.2.7

The provider version should come from react/package.json. requiredVersion is a consumer constraint.
```

### Validations

dmchoi77:
## Summary

Fixes #829.

This changes shared version normalization so `inferVersionFromRequiredVersion(requiredVersion)` is only used as a last-resort fallback.

Provider `version` now resolves in this order:

1. Explicit shared `version`
2. Installed package version from `package.json`
3. Version inferred from `requiredVersion`

## Problem

Previously, `normalizeShareItem()` treated the inferred `requiredVersion` value as part of the explicit version path:

```ts
const explicitVersion =
  typeof shareItem === 'object'
    ? shareItem.version || inferVersionFromRequiredVersion(shareItem.requiredVersion)
    : undefined;

const version = explicitVersion || searchPackageVersion(key);
```

With `react@19.2.7` installed and `requiredVersion: '^19.1.1'`, the generated shared metadata advertised `react` as provider version `19.1.1`.

`requiredVersion` is a consumer constraint, while `version` should describe the actual package version this build can provide.

## Changes

- Split explicit shared `version` from the inferred `requiredVersion` fallback.
- Prefer `searchPackageVersion(key)` before using the inferred fallback.
- Added a regression test for installed `react@19.2.7` with `requiredVersion: '^19.1.1'`.

## Validation

```bash
pnpm exec vitest run src/utils/__tests__/normalizeModuleFederationOption.test.ts
pnpm test
pnpm run typecheck
pnpm run fmt.check
pnpm run build
```

Results:

- `src/utils/__tests__/normalizeModuleFederationOption.test.ts`: 30 passed
- `pnpm test`: 34 files, 520 tests passed
- `typecheck`: passed
- `fmt.check`: passed
- `build`: passed


gioboa:
Fixes shared version precedence so the installed package version is preferred over a version inferred from requiredVersion.

Example: if react@19.2.7 is installed and shared config has requiredVersion: '^19.1.1', the manifest now advertises provider version: '19.2.7' instead of incorrectly inferring version: '19.1.1'.

requiredVersion remains the consumer constraint; version now reflects what the build actually provides.

gioboa (commented):

Image

gioboa (approved):
Thanks @dmchoi77 
Nice catch 👏 that's a great find.

dmchoi77:
@gioboa Thanks! Glad it helped 😄

PR #834, opened 2026-06-17
fix: fix dev manifest hashed remote entry

### LINKED ISSUE #833: [Bug]: mf-manifest.json contains incorrect file names or missing them completely in dev mode
### Describe the bug

When I run a remote app, which is in a module federation setup, in dev mode, I noticed that there are incorrect or missing file names in the mf-manifest.json file which is making the remote app inaccessible from the host. I am attaching 2 sample mf-manifest.json files, one is generated from vite and one from rsbuild. As you can see, main issues are: 

1. remoteEntry's name in vite config is "remoteEntry-[hash]" and there's no such file generated
2. file names for shared modules are missing completely

When running it in preview mode, it looks good.

 
[vite.config.json](https://github.com/user-attachments/files/29038482/vite.config.json)

[rsbuild.config.json](https://github.com/user-attachments/files/29038483/rsbuild.config.json)

### Version

v8.0.13

### Reproduction

-

### Relevant log output

```shell

```

### Validations

gioboa:
Close #833

- Fix dev mf-manifest.json remote entry when filename contains [hash].
- Serve stable dev remoteEntry.js while preserving existing virtual entry handling.
- Add dev shared asset fallback to remoteEntry.js.
- Fix prod SSR fallback to derive from emitted hashed remote entry.
- Add regression tests for dev and build manifest filenames.

### DIFF
--- a/src/plugins/__tests__/pluginMFManifest.test.ts
+++ b/src/plugins/__tests__/pluginMFManifest.test.ts
@@ -127,13 +127,15 @@ async function runGenerateBundleWithManifest(
       | undefined
     >;
     dts?: unknown;
+    filename?: string;
+    bundle?: OutputBundle;
   } = {},
   command: 'serve' | 'build' = 'build',
   base = '/'
 ): Promise<Record<string, string>> {
   getNormalizeModuleFederationOptions.mockReturnValue({
     name: 'basicRemote',
-    filename: 'remoteEntry.js',
+    filename: runtime.filename || 'remoteEntry.js',
     getPublicPath: undefined,
     varFilename: undefined,
     dts: runtime.dts,
@@ -199,7 +201,7 @@ async function runGenerateBundleWithManifest(
       }) as ResolvedId,
   };
 
-  await runGenerateBundle(buildPlugin, ctx, makeBundle());
+  await runGenerateBundle(buildPlugin, ctx, runtime.bundle || makeBundle());
   return emitted;
 }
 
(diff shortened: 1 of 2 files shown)


gioboa (commented):
@akiryazov
Can I ask you to test this pre-release please?

akiryazov:
Yes, of course!

gioboa (commented):

Image

akiryazov:
I can confirm that file names are looking better now, remoteEntry's name is remoteEntry.js and I see it also present for the shared modules. The only issue now is with what the MF exposes, there are no linked files for the exposed paths. All the arrays are empty there

gioboa:
> The only issue now is with what the MF exposes, there are no linked files for the exposed paths. All the arrays are empty there

I pushed the fix, try this new release please.
`npm i https://pkg.pr.new/@module-federation/vite@2690494`

akiryazov:
Sorry for the delay, I was investigating and testing until now and will share my observations: 

- Now I see remoteEntry.js included everywhere
- Since you mentioned yesterday that publicPath '/' resolves to 'auto' internally, I removed it from the configuration. However, this seems to not work for me, not sure if it's specifically for my case or not. For example when I try to access the remote app from the host, it tries to look for https://$hostURL/remoteEntry.js instead of https://$remoteURL/remoteEntry.js . Adding "publicPath": "auto" explicitly fixes it.  It is a big ecosystem and I don't have access to the host app to check its configuration. However, I think I might change that if needed further investigation.
- I managed to connect to the remote only by explicitly adding filename: "remoteEntry.js". Otherwise, for me it looks like it fallbacks to some code when file name is not specified, which names the entry point "remoteEntry-hash.js". At least this is what I saw in @id folder in the source files. 
- After successfully connecting to the remote on dev, I am able to load the remote files, but getting "$RefreshSig$ is not defined" error. Also not sure if it's related to the package or the project. 
- Preview mode still works fine

gioboa:
I see, thanks for you check. I'll merge this PR and let's continue the discussion on the other points.

akiryazov:
Thank you too!

PR #835, opened 2026-06-17
fix: await host init in dev for bridge SSR hydration entries

tonoizer:
## Summary

Fixes a dev-only race when `hostInitInjectLocation: "entry"` is used with SSR hosts that have no root `index.html` (Nitro, TanStack Start, etc.).

Production builds already wrap hydration entries with `getBootstrapSource()` (`await initHost()` before the client entry runs). In Vite dev (`serve`), hydration entry fallbacks only received a side-effect `import` of the host-init module, so `hydrateRoot` could run before the Module Federation runtime finished initializing.

This shows up most clearly with **`@module-federation/bridge-react`** (`createLazyComponent` / `loadRemote`): bridge resolves the MF runtime synchronously on first render via `getInstance()`. If `initHost()` has not completed, remotes fail on first paint with: "Module Federation runtime is not initialized" 

Production preview for the same apps works — the gap was dev-only client entry injection.

## Affected setup

- `hostInitInjectLocation: "entry"`
- No `index.html` at dev time (or missing from project root)
- Client entry contains `hydrateRoot`, `createRoot`, or `ReactDOM.render`
- Vite `command: "serve"`

Typical frameworks: Nitro + React (`entry-client.tsx`), TanStack Start (`client.tsx` + `StartClient`).

`React.lazy(() => import('remote/…'))` with a route loader that pre-imports remotes may mask the same race; bridge should not require an app-level `await hostInitPromise` shim.

## Root cause

In `pluginAddEntry.ts`, `isHydrationEntryFallback` matches client hydration modules when rollup has no registered input entry. For `waitsForInit` injections, `usesDevEntryFallback` forced dev serve to prepend only `import hostInit` instead of the full bootstrap wrapper used in `build`.

The old comment about SSR modules needing synchronous exports does not apply here — server entries do not match `isHydrationEntryFallback` (no mount/hydrate calls).

Nuxt dev is unchanged: `entry.async.js` is still skipped (`skipNuxtDevEntryAsyncInject`); host init is injected at mount time in `entry.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,9 @@ export default defineConfig({
       // Optional parameter that controls where the host initialization script is injected.
       // By default, it is injected into the index.html file.
       // You can set this to "entry" to inject it into the entry script instead.
-      // Useful if your application does not load from index.html.
+      // Recommended for SSR hosts without index.html (Nitro, TanStack Start) so
+      // initHost() completes before hydrateRoot and @module-federation/bridge-react
+      // remotes render on first paint.
       hostInitInjectLocation: "html", // or "entry"
       // Controls whether all CSS assets from the bundle should be added to every exposed module.
       // When false (default), the plugin will not process any CSS assets.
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Screenshot 2026-06-17 at 15 44 50

gioboa (approved):
Thanks @tonoizer LGTM 💪

PR #839, opened 2026-06-18
fix(ssr): initialize host before Vue and remote entries

tonoizer:
## Summary

Fixes dev SSR entry initialization for apps that do not use an `index.html` client entry.

- Wrap Vue client entries that mount with `app.mount(...)` so `hostInit` runs before the first render
- Keep exposes-only remotes eligible for the hydration-entry fallback when `hostInitInjectLocation` is `entry`
- Avoid consuming that fallback on Module Federation internal virtual modules or one-off unrelated hydration modules
- Register common Zustand subpaths (`zustand/vanilla` and `zustand/react`) as loadShare modules during early init

## Why

React SSR entries already got a host-init fallback, but custom Vue SSR apps and producer remotes can also render before the Module Federation runtime is initialized. When that happens, bridge remotes can fail on first render because `getInstance()` has no initialized runtime yet. These changes make the fallback apply to the real app entry, keep internal virtual modules out of the way, and add the shared Zustand subpaths that need the same early loadShare setup.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1132,6 +1132,43 @@ describe('vite:module-federation-early-init', () => {
     expect(virtualModule?.code).toContain('jsx');
   });
 
+  it('registers zustand shared subpath loadShare modules during early init', () => {
+    const plugin = (
+      federation({
+        name: 'host',
+        filename: 'remoteEntry.js',
+        shared: {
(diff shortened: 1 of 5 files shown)


gioboa (commented):

image

gioboa (approved):
Thanks @tonoizer LGTM 👏

PR #840, opened 2026-06-20
fix: surface SSR entry HTTP errors

dmchoi77:
## Summary

This surfaces HTTP fetch failures when loading SSR remote entries and their transitive ESM imports.

Previously, non-ok HTTP responses could still be treated as module source and written to a temporary file. This made SSR entry failures harder to diagnose, especially when the actual response was a `404`, `500`, or an error page.

SSR module fetches now throw a dedicated error with the failed URL and HTTP status details.

## Problem

`fetchEsmToTempFile()` read the response body before checking whether the HTTP response was successful:

```ts
const res = await fetch(url);
let code = await res.text();
```

Because `res.ok` was not checked, a failed response could continue through the normal SSR module transformation path.

For example, if `remoteEntry.ssr.js` returned `404 Not Found`, the loader could attempt to transform and import that response instead of reporting the actual fetch failure.

This also affected transitive relative imports, where a missing nested chunk did not clearly report which URL failed.

## Changes

- Added `SsrEntryHttpError` for SSR module HTTP failures.
- Included the failed module URL, status code, status text, and a short body preview in the error.
- Updated `fetchEsmToTempFile()` to throw when `fetch()` returns a non-ok response.
- Preserved the existing fallback behavior for non-HTTP SSR loading errors.
- Skipped Vite preload helper specifiers while collecting relative transitive imports.
- Added regression tests for a non-ok SSR entry response and a non-ok transitive HTTP module response.

## Validation

```bash
pnpm exec vitest run src/utils/__tests__/ssrEntryLoader.test.ts
```

Results:

- `src/utils/__tests__/ssrEntryLoader.test.ts`: added coverage for SSR entry HTTP failures and nested import HTTP failures.

### DIFF
--- a/src/utils/__tests__/ssrEntryLoader.test.ts
+++ b/src/utils/__tests__/ssrEntryLoader.test.ts
@@ -13,6 +13,8 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 type FetchEntry = {
   ok: boolean;
+  status?: number;
+  statusText?: string;
   text?: string;
   json?: unknown;
   headers?: Record<string, string>;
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks @dmchoi77 it looks great to me 🎉

PR #841, opened 2026-06-20
fix: select import:false shared providers by strategy

dmchoi77:
## Summary

This fixes provider selection for `import: false` shared modules when the shared scope contains multiple providers for the same package.

Previously, the remote entry bridge picked the first provider from the shared scope object. That made the selected shared module depend on object key order rather than the configured share strategy, provider load state, version ordering, and strict version behavior.

When `import: false` shared modules are configured, the generated remote entry imports the public runtime share helper and emits a small bridge for selecting providers before seeding the `__mf_module_cache__` entry.

This aligns the `import: false` bridge with the documented [`shareStrategy`](https://module-federation.io/configure/sharestrategy#sharestrategy) behavior for `version-first` and `loaded-first`.

## Problem

The `import: false` bridge selected a provider like this:

```ts
const provider = versions && versions[Object.keys(versions)[0]];
```

When multiple providers existed for the same shared package, this could select whichever version appeared first in the object.

Example failure before this change:

```ts
const versions = {
  '18.3.1': react18Provider,
  '19.2.0': react19Provider,
};
```

If `18.3.1` was inserted first, `Object.keys(versions)[0]` selected the React 18 provider. A remote configured with `react` as `singleton`, `import: false`, and `requiredVersion: '^19.0.0'` could then seed `__mf_module_cache__.share['default:react']` with React 18 before the exposed module imported React. Since singleton cache entries are reused, later shared imports could keep reading the wrong provider instead of the React 19 provider expected by `version-first`.

The generated share config also did not include `strictVersion`, so strict version mismatch handling was not preserved when the bridge delegated selection to the runtime helper.

## Changes

- Added `strictVersion` to generated local and used shared configs in [`virtualRemoteEntry.ts`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L142) and [`generateUsedSharedPreloadConfig`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L187).
- Added a small provider selection bridge emitted into `remoteEntry` only when `import: false` shared modules are configured: [`sharedProviderSelectionHelperCode`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L296) and [`needsSharedProviderSelectionHelper`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L407).
- Imported `share` from `@module-federation/runtime/helpers` for that bridge so strategy, range, singleton, and strict-version decisions are delegated to the runtime `getRegisteredShare` path: [`runtimeShare` import](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L432).
- Changed the base `@module-federation/runtime` alias to an exact match and added an `import: false`-only exact alias for `@module-federation/runtime/helpers` derived from the same runtime implementation: [`getRuntimeHelpersImplementation`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/index.ts#L180) and [`config` alias setup](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/index.ts#L1073). This keeps custom `implementation` consumers and strict pnpm installs on one runtime package.
- For `version-first`, the bridge uses a selection-only provider map with `loaded`, `loading`, and `lib` neutralized before delegating to the runtime helper: [`__mfCreateProviderSelectionVersions`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L299). This prevents an already materialized lower version from winning over a higher compatible provider during `import: false` cache seeding.
- Updated the `import: false` bridge to use the helper instead of the first object key for both [`global share pre-seeding`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L501) and [`shared init argument bridging`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/virtualRemoteEntry.ts#L579).
- Added focused tests for provider selection, first-key removal, conditional helper imports, helper aliasing, custom runtime implementations, and global share scope pre-seeding in [`virtualRemoteEntry.test.ts`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/virtualModules/__tests__/virtualRemoteEntry.test.ts#L799) and [`index.test.ts`](https://github.com/dmchoi77/vite/blob/d314f63f1ece83466d2cdca0206d8829d2874593/src/__tests__/index.test.ts#L1434).

### DIFF
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -100,7 +100,8 @@ describe('shared dependencies', () => {
     expect(loadShare).toBeDefined();
     expect(allCode).toContain('const versions =');
     expect(allCode).toContain('[pkg]');
-    expect(allCode).toContain('const provider = versions && versions[Object.keys(versions)[0]]');
+    expect(allCode).toContain('__mfSelectSharedProvider');
+    expect(allCode).not.toContain('versions[Object.keys(versions)[0]]');
     expect(allCode).toContain('__mfModuleCache.share[cacheKey]');
     expect(allCode).not.toContain('initRes.loadShare(pkg');
     expect(loadShare!.code).toContain('initPromise.then');
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -187,6 +187,25 @@ function getModuleFederationVitePlugin(): FederationPlugin {
   return plugin;
 }
 
+function getModuleFederationVitePluginWithImportFalse(implementation?: string): FederationPlugin {
+  const plugin = (
+    federation({
+      name: 'host',
+      filename: 'remoteEntry.js',
+      implementation,
+      shared: {
+        react: {
+          singleton: true,
+          import: false,
+        },
+      },
+    }) as Plugin[]
+  ).find((entry) => entry.name === 'module-federation-vite');
+
+  if (!plugin) throw new Error('module-federation-vite plugin not found');
+  return plugin;
+}
+
 function resolvesQuickly(promise: Promise<unknown>) {
   return Promise.race([
     promise.then(() => true),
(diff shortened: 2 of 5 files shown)


dmchoi77:
@gioboa  CI was failing because an integration test still asserted the old first-key provider selection logic.
I updated the assertion to expect the new `__mfSelectSharedProvider` path and pushed the fix. Could you please review it when you get a chance?🙏

gioboa (commented):
Thanks for this PR, I'll check this out in few hours 👍

gioboa (commented):

Screenshot 2026-06-22 at 11 13 10

gioboa (approved):
Thanks @dmchoi77 
this looks awesome to me 👍

PR #843, opened 2026-06-22
fix: manage default share key compatibility

### LINKED ISSUE #842: [Bug]: 1.16.9 (#822) scope-prefixed loadShare cache keys break new-remote-on-old-host shared singletons (React → null)
## Describe the bug

PR #822 ("use share scope in loadShare cache key to prevent cross-scope collisions", released in **1.16.9**) changed the generated `loadShare` runtime code to prefix every shared-module cache key with the scope name (`default:`). This is **not backward compatible** across plugin versions: a remote built with 1.16.9 reads the shared scope under `default:<name>`, but a host built with an older plugin (≤1.16.8) populated it under the un-prefixed `<name>`.

When such a 1.16.9 remote is consumed by an older host, every shared singleton resolves to `undefined`. For React this surfaces as a hard crash:

```
TypeError: Cannot read properties of null (reading 'useEffect')
    at e.useEffect (..._loadShare___...feature_flag_client_public__loadShare__.js)
```

The remote loads its `loadShare` chunk, looks up `share["default:react"]`, finds nothing (the host only set `share["react"]`), and ends up with a `null` React instance — invalid hook call.

## Evidence (build-artifact diff, 1.16.6 → 1.16.9)

Same project, same source, only `@module-federation/vite` changed. The emitted `mf-manifest.json` `shared` section is **byte-for-byte identical** between the two versions (versions, `requiredVersion`, ids all unchanged) — so this is **not** a version-negotiation change (#830 is not involved). The only difference is the cache-key string in the generated `loadShare` chunks:

```diff
- M = ve.share.react;
- M === void 0 && ((M = ye(c)), (ve.share.react = M));
+ M = ve.share[`default:react`];
+ M === void 0 && ((M = ye(c)), (ve.share[`default:react`] = M));

- it = nt.share[`react/jsx-runtime`];
+ it = nt.share[`default:react/jsx-runtime`];

- Z = Y.share[`@ad-lucy/feature-flag-client-public`];
+ Z = Y.share[`default:@ad-lucy/feature-flag-client-public`];
```

Because the shared scope is a single runtime object read/written by both host and remote, the prefixed reader (`default:react`) never finds what the un-prefixed writer (`react`) stored.

## Reproduction

1. Host application built with `@module-federation/vite` ≤ 1.16.8, sharing `react`/`react-dom` as singletons.
2. Remote built with `@module-federation/vite` 1.16.9, consuming the host's shared React (`shared: { react: { singleton: true }, 'react-dom': { singleton: true }, ... }`).
3. Load the remote through the host → `Cannot read properties of null (reading 'useEffect')`.

Bisected across releases against the integrated/host load: **1.16.6, 1.16.7, 1.16.8 all work; 1.16.9 breaks.** Reverting the remote to ≤1.16.8 fixes it. Setting `requiredVersion: false` on the shares does **not** help (confirms it's the cache key, not version matching).

## Expected behavior

A remote built with a newer plugin should still resolve shared singletons populated by a host built with an older plugin. The scope-prefixed cache key should either:

- fall back to reading the legacy un-prefixed key when the prefixed entry is absent, or
- be introduced in a way that doesn't require host and remote to be on matching plugin versions (since hosts and remotes are deployed and upgraded independently — that independence is the whole point of Module Federation).

## Actual behavior

Shared singleton lookups miss across the version boundary, returning `undefined`/`null` and crashing the remote (invalid React hook call).

## Versions

- `@module-federation/vite`: remote on **1.16.9** (broken), **1.16.8 / 1.16.7 / 1.16.6** all OK; host on an older release that writes un-prefixed share keys
- Vite: 8.x (Rolldown)
- React: 18.3.1 (shared as singleton)
- Build mode: `vite build` (production), remote consumed via `mf-manifest.json`

## Suggested fix

Make the prefixed-key reader fall back to the un-prefixed key (e.g. `share["default:react"] ?? share["react"]`), or otherwise preserve cross-version compatibility of the shared scope, so #822's collision fix doesn't break independently-versioned host/remote pairs.

### Version

1.16.9

### Reproduction

https://github.com/alfredosalzillo/mf-vite-1169-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #842 

- Adds default-scope cache aliases for backward compatibility between old flat keys and new scoped keys.
  - Examples:
      - react also available as default:react
      - default:react also available as react
      - react@1.2.3 also available as default:react@1.2.3
      - default:react@1.2.3 also available as react@1.2.3
- Does not alias custom scopes like react-18:react. (Those are in new versions)
- Existing keys are never overwritten.

### DIFF
--- a/src/virtualModules/__tests__/virtualRuntimeInitStatus.test.ts
+++ b/src/virtualModules/__tests__/virtualRuntimeInitStatus.test.ts
@@ -46,6 +46,40 @@ describe('virtualRuntimeInitStatus', () => {
     expect(cache.remote).toBe(remote);
   });
 
+  it('aliases default-scoped and legacy share cache keys both ways', async () => {
+    const { getRuntimeModuleCacheBootstrapCode } = await import('../virtualRuntimeInitStatus');
+    const legacyReact = { source: 'legacy-react' };
+    const legacyVersioned = { source: 'legacy-versioned' };
+    const defaultVue = { source: 'default-vue' };
+    const defaultVersioned = { source: 'default-versioned' };
+    const customScoped = { source: 'custom-scope' };
+    const existingDefaultReact = { source: 'existing-default-react' };
+    (globalThis as any).__mf_module_cache__ = {
+      share: {
+        react: legacyReact,
+        'default:react': existingDefaultReact,
+        'react@1.2.3': legacyVersioned,
+        'default:vue': defaultVue,
+        'default:vue@2.0.0': defaultVersioned,
+        'react-18:react': customScoped,
+      },
+      remote: {},
+    };
+
+    const share = runCode<Record<string, unknown>>(
+      getRuntimeModuleCacheBootstrapCode(),
+      'return __mfModuleCache.share;'
+    );
+
+    expect(share['default:react']).toBe(existingDefaultReact);
+    expect(share.react).toBe(legacyReact);
+    expect(share['default:react@1.2.3']).toBe(legacyVersioned);
+    expect(share.vue).toBe(defaultVue);
+    expect(share['vue@2.0.0']).toBe(defaultVersioned);
+    expect(share['react-18:react']).toBe(customScoped);
+    expect(share['default:react-18:react']).toBeUndefined();
+  });
+
   it('promise bootstrap with enableSsrInit resolves to SSR no-op runtime', async () => {
     const { getRuntimeInitPromiseBootstrapCode } = await import('../virtualRuntimeInitStatus');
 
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-06-22 at 16 25 41

gioboa:
@alfredosalzillo Can you test this pre-release please?
`npm i https://pkg.pr.new/@module-federation/vite@843`

gioboa:
Screenshot 2026-06-22 at 17 31 34

PR #854, opened 2026-06-24
fix: resolve manifest remoteEntry to the container, not a same-named expose

### LINKED ISSUE #853: [Bug]: manifest `remoteEntry` points at an expose chunk when chunk names collide
### Describe the bug

A remote won't load when an exposed module's output chunk shares a base name with the container `filename`. The host's federation runtime throws:

```
[ Federation Runtime ]: The remote entry interface does not contain "init" #RUNTIME-002
```

The build itself is correct — the container is emitted to `remoteEntry.js` and exports `init`/`get`. The bug is the manifest pointer: `mf-manifest.json` advertises `metaData.remoteEntry` as the hashed expose chunk (which only exports the exposed module), not the container.

**Why:** `findRemoteEntryFile` resolves the container by stripping the extension off `filename` (`remoteEntry.js` → `remoteEntry`) and returning the **first** bundle chunk with that name:

```ts
if (
  filename.replace(/[\[\]]/g, '_').replace(/\.[^/.]+$/, '') === fileData.name ||
  fileData.name === 'remoteEntry'
) {
  return fileData.fileName;
}
```

If an exposed module is sourced from a file named like `filename`, its chunk is **also** named `remoteEntry`. On Vite 8 / rolldown it's ordered before the container, so it wins the lookup. (Vite 7 ordered them the other way, which is why this stayed hidden.)

This is independent of the naming style: a kebab `filename` like `remote-entry.js` collides via the stripped-name match, and the camelCase default `remoteEntry.js` also via the hardcoded `fileData.name === 'remoteEntry'` branch.

**Expected:** `metaData.remoteEntry` → `remoteEntry.js` (the container).
**Actual:** `metaData.remoteEntry` → `assets/remoteEntry-<hash>.js` (an expose — no `init`).

Fix in #854.

### Version

vite 8.1.0 (`@module-federation/vite` 1.16.10)

### Reproduction

https://github.com/gu-stav/1379-module-federation-vite-reproduction

```sh
pnpm install
pnpm repro   # vite build && node verify.mjs
```

The only trick is the collision: container `filename: "remoteEntry.js"` plus an exposed module sourced from `src/remoteEntry.js`.

### Relevant log output

```shell
manifest advertises remoteEntry: assets/remoteEntry-eKVaqVFW.js

  remoteEntry.js  (the real container) exports: export { getExposes as get, init }
  assets/remoteEntry-eKVaqVFW.js  (advertised) exports: export { widget }

BUG: the manifest points at "assets/remoteEntry-eKVaqVFW.js", not the "remoteEntry.js" container.
That chunk has no `init`, so a host loading this manifest throws #RUNTIME-002.
```

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a minimal reproducible example of the bug.

gu-stav:
Fixes #853.

### Problem

When an exposed module's output chunk shares a base name with the container `filename`, `mf-manifest.json` advertises `metaData.remoteEntry` as the **expose chunk** instead of the container. A host loading that manifest throws `#RUNTIME-002` ("The remote entry interface does not contain `init`") and never mounts.

`findRemoteEntryFile` resolves the container by stripping the extension off `filename` (`remoteEntry.js` → `remoteEntry`) and returning the first bundle chunk with that name. An exposed module sourced from a file named like `filename` ends up with a chunk of the same name — and on Vite 8 / rolldown it's ordered first, so it wins the lookup. (Vite 7 ordered them the other way, which hid it.)

### Fix

Prefer the chunk whose `fileName` is exactly the configured `filename`. The container is emitted with that exact name (`addEntry({ fileName })`), so it's unambiguous. The existing name match stays as a fallback for the hashed/dev filenames it can't equal.

### DIFF
--- a/src/plugins/__tests__/pluginMFManifest.test.ts
+++ b/src/plugins/__tests__/pluginMFManifest.test.ts
@@ -422,6 +422,31 @@ describe('pluginMFManifest', () => {
     expect(manifest.metaData.ssrRemoteEntry.name).toBe('remoteEntry-a1b2c3d4.ssr.js');
   });
 
+  // An expose basenamed like the container gets the same chunk name; even emitted first,
+  // the manifest must point at the container, not the expose. Covers a camelCase
+  // `remoteEntry.js` and a kebab `remote-entry.js` filename.
+  it.each(['remote-entry', 'remoteEntry'])(
+    'resolves the build remoteEntry to the %s container, not a same-named expose',
+    async (base) => {
+      const filename = `${base}.js`;
+      const exposeFile = `assets/${base}-abc123.js`;
+      const container = {
+        ...(makeBundle()['remoteEntry.js'] as OutputChunk),
+        fileName: filename,
+        name: base,
+      };
+      const bundle: OutputBundle = {
+        [exposeFile]: { ...container, fileName: exposeFile, facadeModuleId: '/src/expose.js' },
+        [filename]: container,
+      };
+
+      const emitted = await runGenerateBundleWithManifest(true, { bundle, filename });
+      const manifest = JSON.parse(emitted['mf-manifest.json']);
+
+      expect(manifest.metaData.remoteEntry.name).toBe(filename);
+    }
+  );
+
   it('emits companion stats file using manifest fileName suffix', async () => {
     const emitted = await runGenerateBundleWithManifest({
       fileName: 'path/custom-manifest.json',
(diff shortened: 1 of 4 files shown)


gioboa (commented):

image

gioboa (approved):
Thanks @gu-stav for your amazing help 🚀

PR #856, opened 2026-06-25
fix: ignore federated specifiers inside comments and strings

### LINKED ISSUE #855: [Bug]: import('remote/foo') inside comments and strings is treated as a real import
### Describe the bug

The plugin matches `import('remoteApp/x')` with a regex against the raw source, so it also matches that text when it shows up inside a `//` comment, `/* */` block, double-quoted string, or template literal. It then replaces the match with the ~20-line dynamic-import wrapper, which spills out of the comment or string and breaks the file.

I hit this in a SvelteKit + MF host. A `+page.ts` had a comment that mentioned `import('quizzes/App')` as docs. The plugin rewrote the comment, hover-preload broke silently, no console error. Found it by diffing against a sibling MFE.

The repro has four files, each with the specifier in a different place: line comment, block comment, double-quoted string, template literal. All four reproduce on `1.16.10`. For `trap_block.js`, `trap_string.js`, and `trap_template.js`, the bug fires on the initial `GET /src/<file>`. For `trap.js` the initial request returns the entry-bootstrap shim (HTTP 200) and the actual transform runs on the re-import at `/src/trap.js?mf-entry-bootstrap`, where it 500s the same way.

The code is in [`src/plugins/pluginRemoteNamedExports.ts`](https://github.com/module-federation/vite/blob/main/src/plugins/pluginRemoteNamedExports.ts). Two collectors: `collectFromAST` (walks the AST, fine) and `collectFromRegex` (regex over raw source). The dispatcher hits the regex one whenever AST returns nothing:

```ts
if ((!imports || imports.length === 0) && REGEX_FALLBACK_EXTENSIONS_RE.test(id)) {
  imports = collectFromRegex(code, matchesRemoteImport);
}
```

A file with `import('remote/x')` only inside a comment is exactly when AST returns nothing. So it goes to the regex, which doesn't care about comments.

I have a PR ready, will link it after this is filed. It feeds `stripLiteral(code)` to the regex (same library Vite uses for the same reason since vitejs/vite#8054), and changes the dispatcher so the regex only runs when acorn actually threw.

### Version

8.0.12

### Reproduction

https://github.com/kamranasad7/mf-vite-comment-trap

### Relevant log output

```shell
[vite] Internal server error: Failed to parse source for import analysis because the content contains invalid JS syntax. If you are using JSX, make sure to name the file with the .jsx or .tsx extension.
  Plugin: vite:import-analysis
  File: src/trap_string.js:2:0
  1  |  // String-literal variant: the specifier appears only inside a string value.
  2  |  const docs = "Federated imports look like: import('remoteApp/App').then(function(__mf_m__) {
     |                                                                                              ^
  3  |    var __mf_pending__ = __mf_m__ && __mf_m__.__mf_remote_pending;
  4  |    var __mf_ready__ = __mf_pending__ && typeof __mf_pending__.then === "function"
      at TransformPluginContext._formatLog (vite/dist/node/chunks/node.js:30408:39)
      at TransformPluginContext.error (vite/dist/node/chunks/node.js:30405:14)
      at TransformPluginContext.transform (vite/dist/node/chunks/node.js:27626:10)
      at async EnvironmentPluginContainer.transform
      at async loadAndTransform

Note: trap_block.js and trap_template.js produce no log, they return HTTP 200 with the __mf_* expansion embedded inside the comment literal. That's a silent failure mode.
```

### Validations

kamranasad7:
Closes #855 .

The regex fallback in `pluginRemoteNamedExports` was scanning raw source text, so it matched `import('remoteApp/x')` text inside comments, strings, and template literals and rewrote it with the multi-line dynamic-import wrapper. The wrapper spilled out of the comment or string and either broke `vite:import-analysis` or silently corrupted the surrounding text. Repro: https://github.com/kamranasad7/mf-vite-comment-trap (all four traps reproduce on `1.16.10`; for `trap.js` the bug fires on the `?mf-entry-bootstrap` re-import).

Two changes, both in `src/plugins/pluginRemoteNamedExports.ts`:

1. `collectFromRegex` now scans `stripLiteral(code)` instead of `code`. [strip-literal](https://github.com/antfu/strip-literal) blanks comment and string contents while keeping offsets and quote delimiters intact, which Vite has used since [vitejs/vite#8054](https://github.com/vitejs/vite/pull/8054) for the same problem. In-comment and in-string matches disappear, real imports still match. Captures are read from the original `code` via the `d` flag's `match.indices`, since stripLiteral also blanks the module-name string. A small `readMatch` helper resolves the keyword-aligned bounds and source string.

2. Dispatcher: only fall through to the regex tier when acorn threw, not when it returned an empty list.

```diff
 let imports: ImportInfo[] | undefined;
+let parsedSuccessfully = false;

 try {
   const ast = this.parse(code);
   imports = await collectFromAST(ast, code, matchesRemoteImport);
+  parsedSuccessfully = true;
 } catch {
   if ((id.includes('.vue') || id.includes('.svelte')) && /^\s*</.test(code)) return;
-  imports = collectFromRegex(code, matchesRemoteImport);
 }

-if ((!imports || imports.length === 0) && REGEX_FALLBACK_EXTENSIONS_RE.test(id)) {
+if (!parsedSuccessfully) {
   imports = collectFromRegex(code, matchesRemoteImport);
 }
```

strip-literal alone is enough to fix all four traps. The dispatcher tightening is cleanup: regex tier skips every healthy file, two duplicate `collectFromRegex` call sites consolidate, and the now-unused `REGEX_FALLBACK_EXTENSIONS_RE` constant goes.

## Tests

13 new tests in `src/plugins/__tests__/pluginRemoteNamedExports.test.ts` under `describe('comment / string trap')`. Each of the four trap positions (line comment, block comment, single/double-quoted string, template literal) × {AST path, regex fallback path}, plus mixed cases where a real import sits next to an in-commented one, plus a JSX regex-fallback case with an in-comment specifier.

Verified phase by phase on this branch:

| | Target file tests |
|---|---|
| baseline, no fix | 8 trap-test failures |
| + strip-literal only | 65 / 65 |
| + dispatcher tightening | 65 / 65 |

Full suite: 9 pre-existing failures on `upstream/main` (in `pluginAddEntry`, `pluginMFManifest`, `virtualShared_preBuild`, `index`, `packageUtils`, `normalizeModuleFederationOption`); count unchanged through all three phases.

End-to-end against the linked repro: all four trap files return 200 with their source unchanged (no `__mf_*` expansion in the body).

## Dependency

`strip-literal@^3.1.0` (~3 KB, zero runtime deps, by antfu).

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,8 +72,7 @@
   "dependencies": {
     "@module-federation/dts-plugin": "2.6.0",
     "@module-federation/runtime": "2.6.0",
-    "@module-federation/sdk": "2.6.0",
-    "strip-literal": "^3.1.0"
+    "@module-federation/sdk": "2.6.0"
   },
   "devDependencies": {
     "@playwright/test": "1.58.2",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
   "dependencies": {
     "@module-federation/dts-plugin": "2.6.0",
     "@module-federation/runtime": "2.6.0",
-    "@module-federation/sdk": "2.6.0"
+    "@module-federation/sdk": "2.6.0",
+    "strip-literal": "^3.1.0"
   },
   "devDependencies": {
     "@playwright/test": "1.58.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,9 +17,6 @@ importers:
       '@module-federation/sdk':
         specifier: 2.6.0
         version: 2.6.0
-      strip-literal:
-        specifier: ^3.1.0
-        version: 3.1.0
     devDependencies:
       '@playwright/test':
         specifier: 1.58.2
@@ -4969,9 +4966,6 @@ packages:
   js-tokens@4.0.0:
     resolution: {integrity: sha512-RdJUflcE3cUzKiMqQgsCu06FPu9UdIJO0beYbPhHN4k6apgJtifcoCtT9bcxOpYBtpD2kCM6Sbzg4CausW/PKQ==}
 
-  js-tokens@9.0.1:
-    resolution: {integrity: sha512-mxa9E9ITFOt0ban3j6L5MpjwegGz6lBQmM1IJkWeBZGcMxto50+eWdjC/52xDbS2vy0k7vIMK0Fe2wfL9OQSpQ==}
-
   js-yaml@4.1.0:
     resolution: {integrity: sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==}
     hasBin: true
@@ -6354,9 +6348,6 @@ packages:
   strip-final-newline@2.0.0:
     resolution: {integrity: sha512-BrpvfNAE3dcvq7ll3xVumzjKjZQ5tI1sEUIKr3Uoks0XUl45St3FlatVqef9prk4jRDzhW6WZg+3bk93y6pLjA==}
     engines: {node: '>=6'}
-
-  strip-literal@3.1.0:
-    resolution: {integrity: sha512-8r3mkIM/2+PpjHoOtiAW8Rg3jJLHaV7xPwG+YRGrv6FP0wwk/toTpATxWYOW0BKdWwl82VT2tFYi5DlROa0Mxg==}
 
   style-loader@3.3.4:
     resolution: {integrity: sha512-0WqXzrsMTyb8yjZJHDqwmnwRJvhALK9LfRtRc6B4UTWe8AijYLZYZ9thuJTZc2VfQWINADW/j+LiJnfy2RoC1w==}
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -72,7 +72,8 @@
   "dependencies": {
     "@module-federation/dts-plugin": "2.6.0",
     "@module-federation/runtime": "2.6.0",
-    "@module-federation/sdk": "2.6.0"
+    "@module-federation/sdk": "2.6.0",
+    "strip-literal": "^3.1.0"
```
gioboa: is there a way to avoid this dependency?

gioboa (commented):
Thanks for your help @kamranasad7

package.json
gioboa: I did this.

gioboa (commented):

image

gioboa (approved):
Thanks @kamranasad7 for your help 👏

kamranasad7:
@gioboa I'm glad I was helpful. Thanks for approving!

Issue #858, opened 2026-06-26
React duplicate instance in dev mode: regression in 1.16.5, 1.16.9, 1.16.10 — 1.16.6 is the only working version


abhayzmp:
## Description

In a Vite + React 19 Module Federation setup, the remote MFE fails to load inside the host with a **duplicate React instance** error in dev mode. The issue is version-specific:

| Version | Result |
|---------|--------|
| 1.16.4 | ✅ Works |
| 1.16.5 | ❌ Broken — `Invalid hook call` / `Cannot read properties of null (reading 'useSyncExternalStore')` |
| **1.16.6** | **✅ Works** |
| 1.16.7 | Not tested |
| 1.16.8 | Not tested |
| 1.16.9 | ❌ Broken — `TypeError: Cannot read properties of undefined (reading '0')` at RemoteApp (hooks dispatcher index) |
| 1.16.10 | ❌ Broken — same error as 1.16.5 |

1.16.6 included `fix: avoid duplicate react with redux in dev mode` (#800) and `fix: fix up React HMR` (#799), which fixed the issue. However, this fix regressed in 1.16.9 and 1.16.10.

## Environment

- `@module-federation/vite`: see table above
- `@module-federation/runtime`: 2.6.0
- React: 19.2.7
- Vite: ^7.3.1
- pnpm workspace (monorepo) — catalog-pinned versions, pnpm v11.8.0
- Architecture: host (webConsole, port 5173) loads remote (maps, port 5001) via `loadRemote`

## Error in 1.16.5 / 1.16.10

```
installHook.js:1 Error: Invalid hook call. Hooks can only be called inside of the body of a function component.
TypeError: Cannot read properties of null (reading 'useSyncExternalStore')
```

## Error in 1.16.9

```
TypeError: Cannot read properties of undefined (reading '0')
    at RemoteApp (RemoteApp.tsx:25)  // line with useRef()
```

Both point to React being loaded as two separate instances instead of sharing the singleton from the host's scope.

## Steps to reproduce

1. Host app using `@module-federation/vite` as a Vite plugin, configured with `shared` React packages as singletons
2. Remote app similarly configured
3. Start both dev servers
4. Load the remote inside the host via `loadRemote`
5. Remote renders a component that uses React hooks → error

## pnpm workspace catalog (`pnpm-workspace.yaml`)

```yaml
catalogs:
  vite:
    '@module-federation/runtime': 2.6.0
[...]

PR #837, opened 2026-06-18
fix: normalize trailing slash shares

gioboa:
Close #836

Align Vite shared normalization with Module Federation docs by treating known trailing-slash shares like react/ and react-dom/ as their package roots while preserving common subpath sharing.
This prevents Vite/Rolldown from resolving invalid package imports such as react-dom/.

PR #800, opened 2026-06-10
fix: avoid duplicate react with redux in dev mode

gioboa:
Close #792

In dev only, if react-redux is present, we skip Vite pre-bundling for shared react.
This keeps React resolved through Module Federation sharing.

Result: 
host, remote, and react-redux all use the same React instance.

PR #861, opened 2026-06-26
fix: exclude shared singleton React from dev optimizeDeps

dmchoi77:
Partially addresses #858

## Summary

This prevents shared singleton React from being pre-bundled by Vite's dev `optimizeDeps` pipeline.

Previously, `react` was excluded from dev dependency optimization only when `react-redux` was installed. React-only or React-without-Redux Module Federation setups could still have `react` pushed into `optimizeDeps.include`, allowing Vite to pre-bundle a remote-local React copy under `/node_modules/.vite/deps`.

The fix changes the bypass condition to use the shared configuration itself: if `react` is configured as a shared singleton, it stays out of dev pre-bundling.

## Problem

React singleton sharing depends on bare React imports going through the federation shared/loadShare path.

In dev mode, Vite can rewrite optimized dependencies to files such as:

```txt
/node_modules/.vite/deps/react.js
```

Once that happens, the remote can evaluate against its own optimized React instance while the host's `react-dom` renders with the host React dispatcher. That produces duplicate React symptoms such as:

```txt
Invalid hook call
Cannot read properties of null (reading 'useSyncExternalStore')
```

The previous guard was too narrow:

```ts
key === 'react' && hasPackageDependency('react-redux', root)
```

That only protected projects with `react-redux` installed. Shared singleton React should not depend on whether Redux is present.

## Changes

- Changed the dev optimizeDeps bypass condition for React to use `shareItem.shareConfig.singleton === true`.
- Kept the existing Lit bypass behavior unchanged.
- Added a regression test that verifies shared singleton React is excluded from dev optimizeDeps even when `react-redux` is not installed.

## Tests

- Added coverage in `src/__tests__/index.test.ts` for:
  - `react-redux` absent,
  - `react` configured as a shared singleton,
  - dev mode optimizeDeps excluding `react`,
  - dev mode optimizeDeps not including `react`.

## Verification

```sh
pnpm exec vitest run src/__tests__/index.test.ts -t "excludes shared singleton react from dev optimizeDeps without react-redux"
pnpm exec vitest run src/__tests__/index.test.ts
pnpm exec vitest run --dir src
git diff --check -- src/index.ts src/__tests__/index.test.ts
```

Results:

- Regression test failed before the implementation change and passed after it.
- `src/__tests__/index.test.ts` passed: 56 tests.
- TypeScript check passed.
- Source tests passed: 34 files / 563 tests.

`pnpm exec vitest run --dir src` still prints the existing `MaxListenersExceededWarning`, but all tests pass.

## Compatibility Notes

This change only affects dev-mode optimizeDeps handling for shared singleton `react`.

Non-singleton React remains on the existing optimizeDeps include path. User-provided `optimizeDeps.include` still takes precedence through the existing include check, preserving the current explicit override behavior.

Projects that need React singleton sharing should avoid explicitly adding `react` to `optimizeDeps.include`, because that intentionally opts back into Vite dev pre-bundling.


gioboa (commented):
I see, that check was introduced for #792. 
Can you check if this change is working in that case too?

dmchoi77:
I checked the #792 case as well.

I ran the existing #800 regression test for the `react-redux` path:

```sh
pnpm exec vitest run src/__tests__/index.test.ts -t "excludes shared react from dev optimizeDeps when react-redux is installed"
```

It passes with this change, so the #792 behavior is preserved.

abhayzmp:
## Tested #861 against the repro — it's necessary but **not sufficient**

Thanks for the quick PR @dmchoi77. I tested it directly against the minimal repro (https://github.com/abhayzmp/mf-vite-repro) and unfortunately the duplicate-React error still reproduces. I believe the bug has **two independent causes** and #861 only addresses the first.

### How I tested

I applied #861's exact source change to the installed plugin in `node_modules`:

```js
// src/index.ts (built: lib/index.js)
- const shouldBypassOptimizeDep = isLitShare(key) || key === "react" && hasPackageDependency("react-redux", root);
+ const shouldBypassOptimizeDep = isLitShare(key) || key === "react" && shareItem.shareConfig?.singleton === true;
```

Important detail: the host and remote resolve to **two separate installed copies** of the plugin, because they're on different Vite majors:

- host → `@module-federation/vite@1.16.10` resolved against **vite@7.3.6**
- remote → `@module-federation/vite@1.16.10` resolved against **vite@8.1.0**

I patched **both** copies (verified the `shouldBypassOptimizeDep` line in each now uses the `singleton` gate), then restarted with forced re-optimization:

```bash
pnpm dev --force   # vite --force on both apps → "Forced re-optimization of dependencies"
```

### Result

The remote still fails with the same error:

```
TypeError: Cannot read properties of null (reading 'useSyncExternalStore')
Invalid hook call. ... You might have more than one copy of React in the same app
```

The stack trace shows `react` + `react-i18next` resolving from the **remote** origin (`http://localhost:5001`) while `react-dom` resolves from the **host** (`http://localhost:5173`) — i.e. the remote is still ending up with its own React instance even though React is now correctly excluded from its dev pre-bundle.

### Why I think there are two causes

**Cause 1 — the `react-redux` gate (what #861 fixes).** #800 only excluded React from dev `optimizeDeps` when `react-redux` was installed. #861 re-gates that on `singleton`, which is correct. ✅

**Cause 2 — the 1.16.9 share-scope cache-key / singleton-aliasing rewrite (not addressed).** After the `1.16.9` cluster, even with React correctly excluded from pre-bundling, the remote's `loadShare` no longer resolves onto the host's React singleton. The relevant PRs:

- #822 — *use share scope in loadShare cache key to prevent cross-scope collisions*
- #828 — *reuse singleton share cache aliases* (its own description: *"Fixes remote shared modules with empty config loading local React instead of host singleton… Prevents duplicate React instances and invalid hook calls"*)
- #837 — *normalize trailing slash shares* (`react/`, `react-dom/`)
- #830 — *prefer installed shared version over requiredVersion*
- #841 / #843 (1.16.10) — *select import:false shared providers by strategy* / *manage default share key compatibility* (`default:react` ⇄ `react` aliasing)

This matches the bisection: **1.16.6 is the last working version** precisely because it predates this rewrite. The dev `optimizeDeps` exclude (whether automatic or set manually) stopped being sufficient from 1.16.9 onward.

### Bisection (no `react-redux` in the stack)

| `@module-federation/vite` | Result |
|---|---|
| 1.16.4 | ✅ Works |
| 1.16.5 | ❌ `Invalid hook call` |
| **1.16.6** | **✅ Works** |
| 1.16.9 | ❌ `Cannot read properties of undefined (reading '0')` |
| 1.16.10 | ❌ `Cannot read properties of null (reading 'useSyncExternalStore')` |
| 1.16.10 + #861 | ❌ Still `useSyncExternalStore` (tested as above) |

### Suggestion

#861 looks correct and worth merging on its own (the `react-redux` gate shouldn't have been there). But a complete fix for #858 also needs the share-scope **singleton-aliasing** path restored so a remote's `loadShare` resolves onto the host's React singleton in dev — likely a regression introduced by #828 / #837 / #843.

Happy to re-test any follow-up patch against the same repro.

abhayzmp:
Now that the repo is public and accessible, please feel free to play around, clone and test your changes on it. thanks

gioboa (commented):

Image

dmchoi77:
I checked this further, and I agree: #861 is necessary but not sufficient, so I’ll keep this PR scoped as “Partially addresses #858”.

I’ve prepared a follow-up patch for the second path you identified. It adds canonical shared cache keys with default-scope compatibility aliases (`default:react` ⇄ `react`) and applies the shared cache read/write helpers across `loadShare`, `remoteEntry` provider bridging, singleton aliasing, host auto-init preload, and `react/compiler-runtime`.

I’ll open it as a separate PR instead of expanding #861, so this one can stay focused on the optimizeDeps gating fix. Once it’s up, could you re-test it against the same repro?

gioboa (approved):
Thanks @dmchoi77 LGTM 👏

gioboa:
Thanks @abhayzmp for your check.

PR #864, opened 2026-06-26
fix: fix workspace singleton named exports

### LINKED ISSUE #863: [Bug]: eager workspace singleton fallback breaks monorepo remote sharing with host singletons
### Describe the bug

A regression introduced by commit `5624ab7fc9443d8ac01d4ace51962cdf35556e86` appears to break a valid monorepo federation setup where a host app and a remote app both depend on shared workspace packages.

I've reproduced the issue as minimally as I can in this repository https://github.com/rfsmart-anelson/module-fed-repro-5624ab7

- shared workspace packages:
  - `p1`
  - `p2` (depends on `p1`)
- federated apps:
  - `host`
  - `remote` (loaded by `host` at runtime)

`p1` and `p2` are shared singletons. `host` and `remote` are in the same monorepo/workspace.

After `5624ab7`, when `remote` is built inside the workspace, the generated `__loadShare__` shim for `p1` changes from the normal prebuild-backed path or lazy workspace singleton to the eager workspace singleton.

This appears to cause `remote` to fall back to its own local `p1` instance instead using the host's instance. The result is duplicate module instances at runtime. In my case, upstream code executes, but singleton state (for example Zustand stores) does not match the data in stores mutated by the host.

This only happens when the remote is built as part of the same workspace. If I remove `remote` from the workspace and install shared packages from a tarball, the generated shared code changes back to the non-workspace prebuild path and the problem goes away.

Relevant code path:
- `src/virtualModules/virtualShared_preBuild.ts`
- `isWorkspaceSingletonConsumedByPeer(pkg)`
- `writeLoadShareModule()`
- `generateEagerWorkspaceSingletonExports()`

The behavior change seems to be:
- before: workspace singleton peer case used lazy/prebuild behavior
- after `5624ab7`: the remote emits eager workspace singleton fallback for `p1`

Observed emitted shapes:

Broken in-workspace remote build (`p1` eager workspace path):
```js
var exportModule = __mfModuleCache.share["default:p1"];
if (exportModule === void 0) {
  Promise.resolve().then(() => {
    if (__mfModuleCache.share["default:p1"] === void 0)
      __mfModuleCache.share["default:p1"] =
        __mfNormalizeShareModule(main_exports);
  });
  exportModule = main_exports;
}
```

## Repro steps
Repro can be found in the repository readme https://github.com/rfsmart-anelson/module-fed-repro-5624ab7

### Version

v1.16.9

### Reproduction

https://github.com/rfsmart-anelson/module-fed-repro-5624ab7

### Relevant log output

```shell
Navigated to http://localhost:3002/
App.tsx:10 host store state prior to remote load {p1: 'p1ModifiedByHost', p2: 'p2ModifiedByHost', setP1: ƒ, setP2: ƒ}
App.tsx:4 remote store state {p1: 'p1Init', p2: 'p2Init', setP1: ƒ, setP2: ƒ}
react-dom-client.production.js:4329 Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at areHookInputsEqual (react-dom-client.production.js:4329:32)
    at Object.updateCallback [as useCallback] (react-dom-client.production.js:5157:24)
    at exports.useCallback (react.production.js:486:33)
    at useStore (react.mjs:8:11)
    at useBoundStore (react.mjs:16:39)
    at App (App.tsx:7:14)
    at basicStateReducer (react-dom-client.production.js:4551:41)
    at updateReducerImpl (react-dom-client.production.js:4628:13)
    at updateReducer (react-dom-client.production.js:4555:10)
    at Object.useState (react-dom-client.production.js:5703:14)
areHookInputsEqual @ react-dom-client.production.js:4329
updateCallback @ react-dom-client.production.js:5157
(anonymous) @ react.production.js:486
useStore @ react.mjs:8
(anonymous) @ react.mjs:16
App @ App.tsx:7
basicStateReducer @ react-dom-client.production.js:4551
updateReducerImpl @ react-dom-client.production.js:4628
updateReducer @ react-dom-client.production.js:4555
useState @ react-dom-client.production.js:5703
(anonymous) @ react.production.js:526
App @ App.tsx:13
renderWithHooks @ react-dom-client.production.js:4351
updateFunctionComponent @ react-dom-client.production.js:6497
beginWork @ react-dom-client.production.js:7473
performUnitOfWork @ react-dom-client.production.js:11305
workLoopSync @ react-dom-client.production.js:11182
renderRootSync @ react-dom-client.production.js:11163
performWorkOnRoot @ react-dom-client.production.js:10728
performWorkOnRootViaSchedulerTask @ react-dom-client.production.js:12078
performWorkUntilDeadline @ scheduler.production.js:151
postMessage
(anonymous) @ scheduler.production.js:202
(anonymous) @ scheduler.production.js:325
scheduleTaskForRootDuringMicrotask @ react-dom-client.production.js:12055
processRootScheduleInMicrotask @ react-dom-client.production.js:11978
(anonymous) @ react-dom-client.production.js:12095
scheduler.production.js:120 [Violation] 'message' handler took 3502ms
```

### Validations

gioboa:
Close #863

Fixes eager workspace singleton named exports to read from the resolved shared module instance instead of re-exporting the local fallback, preventing duplicate shared state between host and remote.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1,3 +1,4 @@
+import path from 'node:path';
 import type {
   ConfigEnv,
   ConfigPluginContext,
@@ -44,6 +45,8 @@ import { federation } from '../index';
 import VirtualModule from '../utils/VirtualModule';
 import { getPreBuildLibImportId, LOAD_SHARE_TAG, PREBUILD_TAG } from '../virtualModules';
 import { virtualRuntimeInitStatus } from '../virtualModules/virtualRuntimeInitStatus';
+
+const REACT_EXAMPLE_ROOT = path.join(process.cwd(), 'examples/vite-vite/vite-host');
 
 type FederationPlugin = Plugin;
 function createChunk(fileName: string, code: string): Rollup.OutputBundle[string] {
@@ -1193,7 +1196,7 @@ describe('vite:module-federation-early-init', () => {
   it('registers common shared subpath loadShare modules during early init', () => {
     const plugin = getEarlyInitPluginWithReactShared();
     const config: any = {
-      root: process.cwd(),
+      root: REACT_EXAMPLE_ROOT,
       optimizeDeps: {
         include: [],
         exclude: [],
--- a/src/plugins/__tests__/pluginDevRemoteHmr.test.ts
+++ b/src/plugins/__tests__/pluginDevRemoteHmr.test.ts
@@ -1,9 +1,12 @@
+import path from 'node:path';
 import type { IncomingMessage, ServerResponse } from 'http';
 import type { HtmlTagDescriptor, MinimalPluginContextWithoutEnvironment } from 'vite';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import pluginDevRemoteHmr, { shouldIgnoreFile } from '../pluginDevRemoteHmr';
 import { normalizeModuleFederationOptions } from '../../utils/normalizeModuleFederationOptions';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
+
+const REACT_EXAMPLE_ROOT = path.join(process.cwd(), 'examples/vite-vite/vite-remote');
 const { mfWarn } = vi.hoisted(() => ({
   mfWarn: vi.fn(),
 }));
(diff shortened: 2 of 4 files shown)


gioboa (commented):

Image

PR #867, opened 2026-06-28
fix(virtualExposes): broaden CSS dedup selector to match any existing stylesheet

pcfreak30:
The dedup check in injectCssAssets only matched links with the data-mf-href attribute, meaning stylesheets injected by Vite's own CSS module system or MF runtime's createLink were invisible to the check. This caused duplicate <link> tags for the same CSS URL.

Replace the selector from link[rel="stylesheet"][data-mf-href="..."] to link[rel="stylesheet"][href="..."] so it detects any existing stylesheet with the same URL regardless of who created it. Remove the data-mf-href attribute from newly created links since it is no longer needed for dedup.

---

I had a case of both code paths triggering causing duplicate css, while trying to do webapp perf optimizations. This has resolved that.

### DIFF
--- a/src/virtualModules/__tests__/virtualExposes.test.ts
+++ b/src/virtualModules/__tests__/virtualExposes.test.ts
@@ -94,7 +94,7 @@ describe('virtualExposes', () => {
         }),
       },
       querySelector: vi.fn((selector: string) => {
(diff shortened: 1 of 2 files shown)


pcfreak30:
@gioboa

gioboa (approved):
Thanks @pcfreak30 for this nice improvement 👏

PR #868, opened 2026-06-30
fix: stop build load hook from reading arbitrary filesystem paths

tonoizer:
## Summary

Stops the build-time federation `load` hook from falling back to `readFileSync(id)` when a `loadShare` or `loadRemote` virtual module cannot be found in the in-memory virtual module registry.

Instead, the hook now returns `null` and lets Vite/Rollup continue normal resolution.

## Why

The hook currently checks for federation marker tags in the module id and then reads the id from disk if `VirtualModule.findById(id)` does not return code.

That fallback is too broad: a crafted id containing a federation marker can cause the build hook to attempt reading an arbitrary filesystem path. Missing virtual federation modules should not be resolved by reading from disk.

## Changes

- Return `null` when a matching `loadShare`/`loadRemote` id is not a registered virtual module.
- Keep the existing transformation behavior for valid registered virtual modules.
- Add a regression test that verifies a crafted path-like id is not read and returns `null`.

## Validation

- `./node_modules/.bin/vitest run src/__tests__/index.test.ts`
- `./node_modules/.bin/tsc -p tsconfig.json --noEmit`

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -394,6 +394,17 @@ describe('module-federation-esm-shims', () => {
     );
   });
 
+  it('returns null when build load hook cannot resolve a virtual module', () => {
+    const plugin = getEsmShimsPlugin();
+    runConfig(plugin, {} as ConfigPluginContext, { build: {} }, { command: 'build', mode: 'test' });
+    const result = callHook(
+      plugin.load,
+      {} as any,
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

gioboa (approved):
Thanks @tonoizer 
Nice catch 👏

PR #869, opened 2026-06-30
fix: harden dev runner endpoint validation

tonoizer:
## Summary

Hardens the Vite 8 dev `__mf_runner__` endpoint used by SSR remotes.

The endpoint now validates runner requests before forwarding them into Vite's `fetchModule` path:

- allows only the supported `fetchModule` and `getBuiltins` invoke names
- rejects unsafe module ids and importers before calling `fetchModule`
- requires non-virtual module ids to resolve inside the remote project's root
- removes the broad allowance for any path containing `node_modules`

## Why

The dev runner endpoint is intentionally exposed over HTTP so a host can fetch transformed SSR modules from a remote dev server. Because that endpoint forwards request-controlled module ids into Vite's module runner, it should be strict about request shape, payload size, and which module ids are allowed.

Previously, unsupported invoke names were rejected only after parsing an unbounded body, and module ids/importers were not validated before `fetchModule`. A broad `node_modules` allowance also accepted absolute paths outside the current project root.


gioboa (changes requested):
it looks good, I added tests to solve few small things in the current implementation.
Thank you 🚀

gioboa (commented):

Image

gioboa (approved):
Thanks @tonoizer LGTM 🚀

PR #872, opened 2026-07-03
feat(ssrEntryLoader): manifest-version-keyed caching, revalidate(), and opt-in vm share-scope strategy

Nsttt:
## Summary

Closes two gaps in the SSR entry loader compared to the webpack/rspack node runtime (`@module-federation/node`): **stale remotes on long-running servers** and **no share-scope semantics on the server**. Both are additive and default-off; the default temp-file behavior is unchanged.

### 1. Revalidation — remote redeploys without host restarts

Today the loader's caches (`ssrEntryCache`, `manifestFetchCache`, `tempFileCache`) plus Node's ESM module cache live until process exit, so a remote redeployed at the same URL is served stale until the host restarts.

- Every SSR entry resolution now carries a **version key** derived from the remote's manifest (`metaData.buildInfo.buildVersion` + FNV content hash). The key flows into temp-file names, all cache keys, and the entry import itself (`?v=<versionKey>` query), so a new version can never be served from Node's module cache for a reused path.
- New **`maxAgeMs`** plugin option: when a cached resolution is older than this, the loader re-fetches the manifest; if the version changed, per-remote caches are dropped and the new build is loaded. Omit for today's cache-forever behavior.
- New **`revalidate(entryUrl?)`** export (analogue of `@module-federation/node/utils`' revalidate): clears loader caches (scoped by remote origin or globally) and best-effort clears global federation instances' module caches.

### 2. Opt-in `strategy: 'vm'` — real share scope on the server

The temp-file strategy pins bare shared imports to host-resolved `file://` paths (`resolvedShared`) — singletons, but no version negotiation. The new opt-in strategy evaluates the remote's ESM graph with `vm.SourceTextModule` and links bare imports through a linker:

1. the host's federation share scope (`instance.loadShare` on `globalThis.__FEDERATION__` instances that declare the package as shared) — real share-scope semantics,
2. the build-time `resolvedShared` file map (same source as the temp-file strategy),
3. plain host `import(specifier)`.

Requires `--experimental-vm-modules`; degrades to the temp-file strategy with a one-time warning when the API is unavailable. The browser-preload neutralization is extracted as `neutralizeBrowserPreloadHelpers()` and shared by both strategies. The vm strategy is a lazily imported chunk, so nothing changes for consumers who don't opt in.

```ts
federation({
  runtimePlugins: [
    ['@module-federation/vite/ssrEntryLoader', {
      resolvedShared,
      maxAgeMs: 30_000,     // optional: revalidate remotes every 30s
      strategy: 'vm',       // optional: share-scope linking on the server
    }],
  ],
})
```

## Tests

- ~12 new tests: revalidation semantics (stale-manifest re-fetch, version-change reload, scoped/global `revalidate()`), and real `vm.SourceTextModule` evaluation — relative-import graphs, share-scope linking, share-declaration-based instance selection, `resolvedShared` fallback, preload-helper neutralization, HTTP error propagation, and version-keyed cache separation.
- vitest now passes `--experimental-vm-modules` to workers (top-level `execArgv`) so the vm tests exercise the real API; they skip themselves where the flag is absent.
- Full suite: **610 passed, 1 environment-skipped**; no new `tsc` errors; build emits the new lazy `ssrVmStrategy` chunk.

While adding these tests I also fixed a latent order-dependence in the suite: mocked-crypto temp-file names collide across tests, and a failed import of that path poisons Node's resolution cache for later tests — the `?v=` import query resolves this structurally.

## Notes for reviewers

- `revalidate()` clears the loader's caches, but hosts holding direct references to previously loaded remote modules keep those references — documented on the function.
- Convention-resolved entries (no manifest) get a stable `unversioned` key and cannot be auto-revalidated; only manifest-resolved entries can.
- Pre-existing on `main` and untouched here: `tsc` fails on `pluginDts.ts` (duplicate `@module-federation/dts-plugin` type identity via lockfile). Happy to fix in a follow-up.

Context: this incubates "node runtime parity" pieces for the Vite loader discussed while building `@module-federation/nuxt` SSR support — revalidation and server share-scope semantics being the two biggest gaps versus `@module-federation/node`.

### DIFF
--- a/src/utils/__tests__/ssrEntryLoader.test.ts
+++ b/src/utils/__tests__/ssrEntryLoader.test.ts
@@ -34,7 +34,7 @@ function makeFetchMock(responses: Record<string, FetchEntry>) {
   });
 }
 
-async function freshLoader() {
+async function freshLoaderModule() {
   vi.resetModules();
   vi.mock('path', () => ({
     default: {
@@ -58,7 +58,11 @@ async function freshLoader() {
     default: { createRequire: vi.fn() },
     createRequire: vi.fn(),
   }));
-  const { default: factory } = await import('../ssrEntryLoader');
+  return await import('../ssrEntryLoader');
+}
+
+async function freshLoader() {
+  const { default: factory } = await freshLoaderModule();
   return factory;
 }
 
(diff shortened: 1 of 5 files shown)


gioboa (approved):
Thanks @Nsttt 
LGTM 👏

PR #866, opened 2026-06-28
fix: await pending share loads before entry import in bootstrap

pcfreak30:
The bootstrap code generated by pluginAddEntry calls initHost() then immediately imports the entry module. However, some loadShare modules (workspace singletons with lazy fallback, and import:false modules) do not populate their exports synchronously during init().

Instead they schedule an initPromise.then() callback that dynamically imports the real module and assigns exports asynchronously. The bootstrap never waited for these callbacks to complete.

This means the entry module (and any code it transitively runs) could access shared exports (e.g. react, react-dom) before the async initPromise.then() + import() chain resolved, reading undefined.

The fix tracks these deferred share load promises in __mfModuleCache.pendingShareLoads. Two code paths are wrapped:

1. generateLazyWorkspaceSingletonExports (virtualShared_preBuild.ts): the client-side initPromise.then(() => import(...)) that assigns workspace singleton exports after the ESM import resolves.

2. generateDeferredHostProvidedExports (virtualShared_preBuild.ts): the initPromise.then() that waits for the host runtime to provide a shared module the remote does not have locally (import: false).

The bootstrap in pluginAddEntry.ts now awaits
Promise.all(__mfModuleCache.pendingShareLoads) after initHost() and before importing the entry.

PR #874, opened 2026-07-03
fix: resolve remote init after import:false shared cache seeding

### LINKED ISSUE #873: [Bug]: Race condition with import: false shared deps in production builds — loadShare proxy uses .then() instead of top-level await
### Describe the bug

When using shared dependencies with import: false (e.g. react), the generated loadShare proxy chunk uses a non-awaited .then() to resolve the shared module. This causes an intermittent race condition in production builds where the expose chunk is evaluated before the shared module is available.

When using shared dependencies with `import: false` (e.g. `react`), the
generated `loadShare` proxy chunk uses a non-awaited `.then()` to resolve the
shared module. This causes an intermittent race condition in production builds
where the expose chunk is evaluated before the shared module is available.

## Reproduction

Configure a remote with shared `react` using `import: false`:

```ts
federation({
  name: 'mfm',
  shared: {
    react: { singleton: true, import: false },
    'react-dom': { singleton: true, import: false },
  },
  exposes: {
    './contact-form': './src/form/contact-form.tsx',
  },
})
```

The `contact-form` module uses Formik, which calls `React.createContext()` at
module level.

After `vite build`, the generated `loadShare__react` chunk contains:

```js
var Z = o.share[`default:react`]
Z === void 0
  ? i.then(() => {
      // ← NOT awaited!
      Z = o.share[`default:react`]
      if (Z === void 0) throw Error('...')
      X(Z) // populates createContext, useState, etc.
    })
  : X(Z)
```

When `Z` is `undefined` (host hasn't populated the share cache yet), `.then()`
is fired but **not awaited**. The module continues evaluating with all export
variables still `undefined`. The `contact-form` chunk then crashes with:

```
TypeError: x is not a function  (x = createContext)
```

This is **intermittent** — depends on whether the host's `init()` has populated
`o.share['default:react']` before the proxy chunk is evaluated by the browser.

## Root Cause

In `lib/index.js`, `generateDeferredHostProvidedExports` (~line 1362) generates:

```js
initPromise.then(() => {
  exportModule = __mfReadSharedCache(...);
  __mfApplyHostProvidedExports(exportModule);
});
```

This is fire-and-forget. Export variables remain `undefined` until the promise
resolves asynchronously, but dependent chunks are already being evaluated.

## Expected Behavior

Since the build output is ESM (`import` statements), the proxy should use
**top-level await**:

```js
  await initPromise;
  exportModule = __mfReadSharedCache(...);
  if (exportModule === undefined) throw new Error('...');
  __mfApplyHostProvidedExports(exportModule);
} else {
  __mfApplyHostProvidedExports(exportModule);
}
```

This blocks chunk evaluation until the host has registered the shared module.

## Notes

- Only affects **production builds**. In dev mode, `eagerLocalFallback` is set
  to `true` (line 1434: `command !== "build"`), which uses a synchronous
  fallback path.
- `experiments.asyncStartup` from the SDK types is not implemented in the Vite
  plugin.
- Top-level await is valid since the output is ESM.
- Any library calling `React.createContext()`, `useRef`, or similar at module
  level (Formik, React Router, etc.) is affected.
- Workaround: Post-build patching the generated chunk to replace `.then()` with
  `await`.

## Environment
- `@module-federation/vite`: 1.16.12
- `@module-federation/runtime`: 2.6.0
- `vite`: 8.1.0
- Bundler: Rolldown (via Vite 8)

### Version

v1.16.12

### Reproduction

https://github.com/gylxan/mf-vite-race-condition-repo

### Relevant log output

```shell

```

### Validations

dmchoi77:
# fix: resolve remote init after import:false shared cache seeding

fixes #873
Follow-up to #866

## Summary

This fixes the remaining `import: false` shared initialization race on top of the latest host bootstrap changes.

#866 added `pendingShareLoads` so host bootstraps can wait for deferred shared export binding before importing the application entry. However, the generated remote entry still resolved its federation init promise before it finished `initializeSharing` and before it seeded host-provided `import: false` shared modules into `__mf_module_cache__`.

That meant a generated `loadShare` wrapper waiting on `initPromise` could resume while the remote entry was still preparing the host-provided shared cache.

This change moves `initResolve(initRes)` after the `import: false` shared cache seeding loop, so consumers that wait for federation init only resume once the remote entry has finished populating those cache entries.

## Problem

Before this change, generated remote entry init was ordered like this:

```ts
initRes.initShareScopeMap("default", shared);
initResolve(initRes);

await initRes.initializeSharing("default", ...);

for (const [pkg, share] of Object.entries(usedShared)) {
  if (share.shareConfig?.import !== false) continue;
  __mfWriteSharedCache(__mfModuleCache.share, cacheDescriptor, resolved);
}
```

For `import: false`, the remote has no local fallback. It must wait for the host-provided shared provider and seed the module cache before exposed modules read generated shared wrappers.

Resolving `initPromise` before that cache seeding created a timing window:

- generated `loadShare` wrappers could observe that federation init was complete;
- the `default:react` / default-scope compatible cache entry could still be missing;
- exposed modules could then evaluate against unready shared exports.

In React-based remotes, that can surface as named exports such as `createContext` being read before the host-provided React module is available.

## Changes

- Moved `initResolve(initRes)` after `initializeSharing` and after the `import: false` shared cache seeding loop in generated remote entries: [`virtualRemoteEntry.ts`](https://github.com/dmchoi77/vite/blob/3d9a79ec929d4acf02e869c2702a6c4d761d7c69/src/virtualModules/virtualRemoteEntry.ts#L562).
- Preserved the latest shared cache compatibility helpers from `main`, including `__mfReadSharedCache`, `__mfWriteSharedCache`, and default-scope aliases: [`virtualRemoteEntry.ts`](https://github.com/dmchoi77/vite/blob/3d9a79ec929d4acf02e869c2702a6c4d761d7c69/src/virtualModules/virtualRemoteEntry.ts#L574).
- Added a generated remote entry ordering assertion to keep `import: false` shared cache handling before `initResolve(initRes)`: [`virtualRemoteEntry.test.ts`](https://github.com/dmchoi77/vite/blob/3d9a79ec929d4acf02e869c2702a6c4d761d7c69/src/virtualModules/__tests__/virtualRemoteEntry.test.ts#L493).
- Tightened integration coverage to assert generated `loadShare` chunks use the latest `pendingShareLoads` path while the remote entry continues to seed the shared cache directly: [`shared-deps.test.ts`](https://github.com/dmchoi77/vite/blob/3d9a79ec929d4acf02e869c2702a6c4d761d7c69/integration/shared-deps.test.ts#L109).

## Tests

- Updated generated-code coverage for remote entry init ordering.
- Updated shared dependency integration coverage for the current `pendingShareLoads` + shared cache helper behavior.

## Verification

```sh
pnpm test
pnpm test:integration
pnpm run typecheck
pnpm run build
```

Results:

- Source tests passed: 37 files / 622 passed, 1 skipped.
- Integration tests passed: 10 files / 41 tests.
- TypeScript check passed.
- Package build passed.

`pnpm test` still prints the existing `MaxListenersExceededWarning`, but all tests pass.

I also re-verified the fix against a local Vite React Router Module Federation PoC with the local package linked in and the remote configured with `react` and `react-dom` as `import: false`.

PoC result:

- Remote entry served: pass.
- Remote manifest served: pass.
- Host index served: pass.
- Remote route SSR marker: pass.
- Remote data SSR marker: pass.
- Browser hydration and CSR click check: pass.
- Browser console errors: empty.
- Page errors: empty.

### DIFF
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -110,6 +110,7 @@ describe('shared dependencies', () => {
     expect(allCode).toContain('__mfWriteSharedCache(__mfModuleCache.share');
     expect(allCode).not.toContain('initRes.loadShare(pkg');
     expect(loadShare!.code).toContain('initPromise.then');
+    expect(loadShare!.code).toContain('pendingShareLoads');
     expect(loadShare!.code).toContain('default:mock-shared-dep');
     expect(loadShare!.code).toContain('mock-shared-dep');
     expect(loadShare!.code).toContain('init');
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -492,6 +492,9 @@ describe('virtualRemoteEntry', () => {
     expect(code).toContain('const mfName = "host"');
     expect(code).toContain('share.shareConfig?.import !== false');
     expect(code).toContain('const versions = shared?.[pkg]');
+    expect(code.indexOf('share.shareConfig?.import !== false')).toBeLessThan(
+      code.indexOf('initResolve(initRes)')
+    );
     expect(code).not.toContain('initRes.loadShare(pkg');
     expect(code).not.toContain('import exposesMap from');
     expect(code).not.toContain('import {usedShared, usedRemotes} from');
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -560,7 +560,6 @@ export function generateRemoteEntry(
     if (initScope.indexOf(initToken) >= 0) return;
     initScope.push(initToken);
     initRes.initShareScopeMap('${options.shareScope}', shared);
-    initResolve(initRes)
     try {
       await retrySharedInit(async () => {
         await Promise.all(await initRes.initializeSharing('${options.shareScope}', {
@@ -584,6 +583,7 @@ export function generateRemoteEntry(
       const resolved = await Promise.resolve(mod);
       __mfWriteSharedCache(__mfModuleCache.share, cacheDescriptor, __mfNormalizeRuntimeShare(resolved));
     }
+    initResolve(initRes)
     return initRes
   }
 


gioboa (commented):

Image

gioboa (approved):
Thanks for this great fix @dmchoi77 👏

PR #875, opened 2026-07-03
fix: await remote pending before named remote bindings

### LINKED ISSUE #871: [Bug?]: I don't understand `throw ensurePending()`
### Describe the bug

Hi!

I'm doing very unorthodox things with `@module-federation/vite` as I'm trying to modernize a legacy (and somewhat custom) webpack module federation step by step without breaking changes. I cannot create a minimal reproduction at this time because of the number of moving parts implied in this bug report, but I'll try to be as clear as possible. Maybe you'll (especially @gioboa who introduced `ensurePending` in b170151) figure all out faster than me

### Part 1: Registration

I'm building a frontend with federated modules. All modules register themselves in a global record: `appShell.remotes`. 

```ts
// Something like this
appShell.remotes['my-module'] = {
  init() {},
  get() {
    // All modules are expected to return a default function when `get` is called with `./init`
  }
}
```

### Part 2: Initialization

This `appShell.remotes` record is iterated to call `.init(__webpack_share_scopes__.default)` on all remotes

```ts
Object.values(window.appShell.remotes).map(container => container.init(__webpack_share_scopes__.default))
```

https://github.com/Jahia/app-shell/blob/9887cae416772b158581e841c2db07bd59fa0d3e/src/javascript/jahia.js#L7

And then `.get('./init')` is called on all remotes

```ts
Object.values(window.appShell.remotes).map(async (container) => {
  const factory = await container.get('./init');
  // ...
})
```

https://github.com/Jahia/app-shell/blob/9887cae416772b158581e841c2db07bd59fa0d3e/src/javascript/appShell.js#L62

https://github.com/Jahia/app-shell/blob/9887cae416772b158581e841c2db07bd59fa0d3e/src/javascript/appShell.js#L10-L25

### Part 3: Exposing CKEditor 5

Remotes can also choose to expose other things than `./init`, usually `.`. That's the case for [Jahia/richtext-ckeditor5](https://github.com/Jahia/richtext-ckeditor5) that uses `.` to expose CKEditor 5:

```ts
// webpack.config.js
new ModuleFederationPlugin({
    exposes: {
        '.': './src/javascript/shared',
    },
});

// src/javascript/shared.js
export * from 'ckeditor5';
```

https://github.com/Jahia/richtext-ckeditor5/blob/2f9b7ce9a202d23d246e870d9e1a51bee8f067af/webpack.config.js#L138-L140

https://github.com/Jahia/richtext-ckeditor5/blob/2f9b7ce9a202d23d246e870d9e1a51bee8f067af/src/javascript/shared/index.js#L1

### Part 4: Retrieving CKEditor 5

Because of the legacy nature of my federation, I created a custom loader that enables a new protocol: `window:` (e.g. `remotes: { ckeditor5: "window:appShell.remotes.richtextCkeditor5" }`)

```ts
export default function FederationWindowPlugin(): ModuleFederationRuntimePlugin {
  return {
    name: "federation-window-plugin",
    async loadEntry({ remoteInfo }) {
      if (remoteInfo.type !== "var") return;
      const { protocol, pathname } = new URL(remoteInfo.entry);
      if (protocol !== "window:") return;
      let mod: any = window;
      for (const segment of pathname.split(".")) mod = mod[segment];
      return mod;
    },
  };
}
```

https://github.com/Jahia/javascript-components/blob/68005126a36c0627e12e1db334b5e2d81155f90f/packages/vite-federation-plugin/src/federation-window-plugin.ts#L8-L20

Consuming code is compiled from:

```ts
import { View } from 'ckeditor5';

class Foo extends View {};

// Init function exposed as  `./init`
export default function init() {
  new Foo();
}
```

Into (simplified):

```ts
var exportModule = __mfCreateDeferredRemoteProxy();
var View = __mfCreateNamedRemoteProxy(exportModule, "View");

class Foo extends View {}; // 💣 Will throw here when reading `View.prototype`

export default function init() {
  new Foo();
}
```

At runtime this code produces something weird: it throws a promise (this should be an error object with a human-readable message at least)

Image

`const mod = getModule();` returns `undefined` because `__mfModuleCache` contains `__mf_pending__ckeditor5`, not yet `ckeditor5`.

Thanks to this investigation I can think of some workarounds, but what can I do for my code to work without any change?

Since we're in an ESM module, `@module-federation/vite` could insert a top-level await between "imports" and userland-code to ensure that all dependencies are loaded. Or maybe something else entirely, my understanding of `@module-federation/vite` is too shallow for now.

If hoping on a quick call is easier for you, [you can book a slot in my calendar](https://calendar.app.google/1RsH6WNBt5XD9MDc8)

Thank you in advance

### Version

8.1.3

### Reproduction

x

### Relevant log output

```shell

```

### Validations

dmchoi77:
# fix: await remote pending before named remote bindings

Fixes #871

## Summary

This fixes a consumer-side ordering gap in the Rolldown named remote export transform.

`pluginRemoteNamedExports` rewrites static named, namespace, and explicit re-export imports from federated remotes so consumers can use named exports even when the bundler does not provide synthetic named exports.

Before this change, the rewritten module imported both `__moduleExports` and `__mf_remote_pending`, and it also exported `__mf_remote_dependency_pending` for parent remote loaders to await. However, the transformed module itself still evaluated the generated named bindings immediately after the import statement.

That left a timing window where code like this could evaluate a named binding before the remote proxy module had finished resolving:

```ts
import { View } from "remoteApp";

class Foo extends View {}
```

This change adds an `await __mf_remote_pending` step before evaluating rewritten named bindings, namespace imports, and explicit named re-exports.

## Issue #871

#871 reports a legacy federation setup that consumes CKEditor 5 through a custom `window:` remote loader:

```ts
import { View } from "ckeditor5";

class Foo extends View {}
```

The generated code created a named remote proxy for `View` and then evaluated `class Foo extends View {}` before the remote module had resolved. Class heritage evaluation reads from the base value immediately, so the proxy attempted to access a still-pending remote export and surfaced the `ensurePending()` promise throw.

The issue suggests that, because the consumer is an ESM module, the plugin could insert a top-level await between the generated import section and userland code. This PR implements that ordering for the named remote export rewrite path.

## Problem

The previous generated output for a named remote import was ordered like this:

```ts
import {
  __moduleExports as __mf_ns_0,
  __mf_remote_pending as __mf_ns_0_pending
} from "remoteApp";

function __mfCreateNamedRemoteProxy(...) { ... }
const __mf_is_proxy_1 = __mf_ns_0 && __mf_ns_0.__mf_is_remote_proxy;
const { View: __mf_named_2 } = __mf_is_proxy_1 ? {} : __mf_ns_0;
const View = __mf_is_proxy_1
  ? __mfCreateNamedRemoteProxy(__mf_ns_0, "View")
  : __mf_named_2;

export const __mf_remote_dependency_pending = Promise.all([__mf_ns_0_pending]);
```

The dependency promise was available to parent generated remote loaders, but the current module did not wait before evaluating its own generated binding declarations.

That matters for runtime-sensitive named imports, especially values used immediately during module evaluation, such as class bases, framework components, hooks, or stores.

## Changes

- Await `__mf_remote_pending` before evaluating rewritten static named remote imports.
- Await `__mf_remote_pending` before exposing rewritten namespace remote imports.
- Await `__mf_remote_pending` before evaluating explicit named re-exports from remotes.
- Kept `__mf_remote_dependency_pending = Promise.all([...])` so parent remote loaders can still wait for nested remote dependencies.
- Added a regression test that verifies the pending await is emitted before generated named binding evaluation for `class Foo extends View {}`.
- Updated existing transform expectations to assert the generated awaits for AST and fallback paths.

## Tests

- Updated `src/plugins/__tests__/pluginRemoteNamedExports.test.ts` for named imports, namespace imports, explicit named re-exports, and fallback rewrites.
- Added coverage for immediate named binding use during module evaluation.

## Verification

```sh
pnpm exec vitest run src/plugins/__tests__/pluginRemoteNamedExports.test.ts
git diff --check -- src/plugins/pluginRemoteNamedExports.ts src/plugins/__tests__/pluginRemoteNamedExports.test.ts
```

Results:

- `pluginRemoteNamedExports` test file passed: 55 tests.
- TypeScript check passed.

## Compatibility Notes

This only affects transformed modules that statically import or explicitly re-export named bindings from configured remotes.

Default-only remote imports remain unchanged. Dynamic imports continue to use the existing promise wrapper. `export * from "remote"` remains unsupported and continues to warn.

### DIFF
--- a/src/plugins/pluginRemoteNamedExports.ts
+++ b/src/plugins/pluginRemoteNamedExports.ts
@@ -213,7 +213,7 @@ function applyRewrites(
           ms.overwrite(
             imp.start,
             imp.end,
-            `import { __moduleExports as ${imp.namespaceLocal}, __mf_remote_pending as ${pendingId} } from ${src};`
+            `import { __moduleExports as ${imp.namespaceLocal}, __mf_remote_pending as ${pendingId} } from ${src};\nawait ${pendingId};`
           );
         } else {
           const nsId = `__mf_ns_${counter++}`;
@@ -225,7 +225,7 @@ function applyRewrites(
           importParts.push(`__moduleExports as ${nsId}`);
           importParts.push(`__mf_remote_pending as ${pendingId}`);
 
-          let rewrite = `import { ${importParts.join(', ')} } from ${src};`;
+          let rewrite = `import { ${importParts.join(', ')} } from ${src};\nawait ${pendingId};`;
           if (imp.named.length > 0) {
             const isProxyId = `__mf_is_proxy_${counter++}`;
             const tempNames = imp.named.map((_s) => `__mf_named_${counter++}`);
@@ -270,7 +270,11 @@ function applyRewrites(
         const syncLine = `${pendingId}.then(() => {\n${assignLines}\n});`;
         const exportLine = `export { ${vars.map((v) => `${v.tmp} as ${v.exported}`).join(', ')} };`;
 
-        ms.overwrite(imp.start, imp.end, `${importLine}\n${varLines}\n${syncLine}\n${exportLine}`);
+        ms.overwrite(
+          imp.start,
+          imp.end,
+          `${importLine}\nawait ${pendingId};\n${varLines}\n${syncLine}\n${exportLine}`
+        );
         changed = true;
         break;
       }
--- a/src/plugins/pluginRemoteNamedExports.ts
+++ b/src/plugins/pluginRemoteNamedExports.ts
@@ -181,22 +181,55 @@ function applyRewrites(
   // Per-file counter — deterministic regardless of file processing order.
   let counter = 0;
   let namedProxyHelperDeclared = false;
+  let namespaceProxyHelperDeclared = false;
   const dependencyPendingIds: string[] = [];
   const namedProxyHelper = `function __mfCreateNamedRemoteProxy(ns, key) {
+  const readValue = () => {
+    try {
+      return ns[key];
+    } catch (error) {
+      if (error && typeof error.then === "function") return undefined;
+      throw error;
+    }
+  };
+  const syncPrototype = (value, target) => {
+    if (value && value.prototype && Object.getPrototypeOf(target.prototype) !== value.prototype) {
+      Object.setPrototypeOf(target.prototype, value.prototype);
+    }
+  };
   const target = function (...args) {
-    const value = ns[key];
+    const value = readValue();
+    syncPrototype(value, target);
     return typeof value === "function" ? value.apply(this, args) : value;
   };
   return new Proxy(target, {
     get(_target, prop) {
       if (prop === "then") return undefined;
-      const value = ns[key];
+      const value = readValue();
+      syncPrototype(value, target);
+      if (prop === "prototype" && (value == null || value.prototype == null)) return target.prototype;
       if (prop === Symbol.toPrimitive) return () => value;
       const item = value == null ? undefined : value[prop];
       return typeof item === "function" ? item.bind(value) : item;
     },
     apply(target, thisArg, args) {
       return target.apply(thisArg, args);
+    },
+    construct(target, args, newTarget) {
+      const value = readValue();
+      syncPrototype(value, target);
+      if (typeof value === "function") return Reflect.construct(value, args, newTarget);
+      return Reflect.construct(target, args, newTarget);
+    }
+  });
+}`;
+  const namespaceProxyHelper = `function __mfCreateNamespaceRemoteProxy(ns) {
+  if (!ns || !ns.__mf_is_remote_proxy) return ns;
+  return new Proxy(ns, {
+    get(target, prop) {
+      if (prop === "then") return undefined;
+      if (prop in target) return target[prop];
+      return __mfCreateNamedRemoteProxy(target, prop);
     }
   });
 }`;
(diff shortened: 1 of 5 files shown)


src/plugins/pluginRemoteNamedExports.ts
```diff
@@ -225,7 +225,7 @@ function applyRewrites(
           importParts.push(`__moduleExports as ${nsId}`);
           importParts.push(`__mf_remote_pending as ${pendingId}`);
 
-          let rewrite = `import { ${importParts.join(', ')} } from ${src};`;
+          let rewrite = `import { ${importParts.join(', ')} } from ${src};\nawait ${pendingId};`;
```
gioboa: This fix introduces top-level await. We should instead reuse the existing dependency-pending exports and await them in bootstrap/remote-loader paths before evaluating consumers.

gioboa (changes requested):

src/plugins/pluginRemoteNamedExports.ts
dmchoi77: @gioboa I understand the concern about introducing top-level await in transformed consumer modules.

One clarification: the existing remote-loader wait works for modules returned from `runtime.loadRemote()`, but the reported failure can also happen in a host entry/static consumer module. In that path, bootstrap starts `import(entrySrc)`, and that import starts evaluating code like `class Foo extends View`. So bootstrap cannot read the module's `__mf_remote_dependency_pending` export before evaluation has already started.

Is the intended direction to add a pre-entry dependency preflight/virtual module, or another mechanism, so bootstrap can await those remote pending promises before importing the actual entry?

gioboa: I proposed one alternative solution. can you look at it? feel free to change it if you like.

GauBen:
Thank you so much for working on this! I confirm that 6e3ca87 fixed my original bug.

The issue I have now is that since we're manipulating proxies and that ckeditor5 uses `Map` objects to register plugins, plugins conflict:

```ts
// In host code:
import { ContextualBalloon } from 'ckeditor5';

plugins.set(ContextualBalloon, ...); // Real `ContextualBalloon` class

// In remote code:
import { ContextualBalloon } from 'ckeditor5';

plugins.set(ContextualBalloon, ...); // Proxy `ContextualBalloon` class, !== real `ContextualBalloon`
```

At runtime this code crashes with `plugincollection-plugin-name-conflict`

Would it be possible to drop proxies altogether?

gioboa:
> Would it be possible to drop proxies altogether?

I did a new implementation from scratch on #879 
@GauBen can you check it please?

src/plugins/pluginRemoteNamedExports.ts
dmchoi77: @gioboa I appreciate the alternative implementation. I see how #879 addresses the concern I raised here.

It moves the wait before exposed module evaluation by preloading direct remote dependencies, and avoids relying on `__mf_remote_dependency_pending` after the consumer module has already started evaluating. It also avoids named export proxies, which should help with the CKEditor class identity issue GauBen reported.

This direction makes sense to me.

dmchoi77:
Superseded by #879, which addresses the same race by preloading exposed remote dependencies before evaluation and avoids the named export proxy identity issue reported with CKEditor.

Closing this PR in favor of #879.

PR #876, opened 2026-07-03
fix: resolve DTS Runtime Plugin Path

gioboa:
Fixes pnpm cold-start resolution for DTS dynamic remote type hints. The internally injected DTS runtime plugin is now resolved to an actual file path before being added to Vite’s optimizeDeps.include, while user runtime plugins remain unchanged. Adds a focused regression test covering the DTS-only behavior.


gioboa (commented):

Image

PR #879, opened 2026-07-06
fix: preload exposed remote dependencies first

### LINKED ISSUE #871: [Bug?]: I don't understand `throw ensurePending()`

gioboa:
Close #871 #875

Preloads direct remote dependencies before exposed module evaluation, removing named proxy identity mismatches while avoiding top-level await.

### DIFF
--- a/integration/fixtures/nested-remote-class/entry.js
+++ b/integration/fixtures/nested-remote-class/entry.js
@@ -1,0 +1,1 @@
+export const app = true;
--- a/integration/fixtures/nested-remote-class/exposed-init.js
+++ b/integration/fixtures/nested-remote-class/exposed-init.js
@@ -1,0 +1,7 @@
+import { View } from 'ckeditor5';
+
+class Foo extends View {}
+
+export default function init() {
+  return Foo;
+}
--- a/integration/fixtures/nested-remote-class/index.html
+++ b/integration/fixtures/nested-remote-class/index.html
@@ -1,0 +1,6 @@
+<!doctype html>
+<html>
+  <body>
+    <script type="module" src="/entry.js"></script>
+  </body>
+</html>
--- a/integration/remote-dependency-pending.test.ts
+++ b/integration/remote-dependency-pending.test.ts
@@ -1,0 +1,44 @@
+import { resolve } from 'path';
+import { describe, expect, it } from 'vitest';
+import type { ModuleFederationOptions } from '../src/utils/normalizeModuleFederationOptions';
+import { buildFixture, FIXTURES } from './helpers/build';
+import { getAllChunkCode } from './helpers/matchers';
+
+const REMOTE_DEPENDENCY_MF_OPTIONS = {
+  name: 'classRemote',
+  filename: 'remoteEntry.js',
+  exposes: {
+    './init': resolve(FIXTURES, 'nested-remote-class', 'exposed-init.js'),
+  },
+  remotes: {
+    ckeditor5: {
+      name: 'ckeditor5',
+      entry: 'http://localhost:3002/remoteEntry.js',
+      type: 'module',
+    },
+  },
+  dts: false,
+} satisfies Partial<ModuleFederationOptions>;
+
+describe('remote dependency pending', () => {
+  it('waits at expose loading for nested remote named imports without TLA', async () => {
+    const output = await buildFixture({
+      fixture: 'nested-remote-class',
+      mfOptions: REMOTE_DEPENDENCY_MF_OPTIONS,
+    });
+
+    const allCode = getAllChunkCode(output);
+
+    expect(allCode).toMatch(/\b(?:const|var)\s+\{\s*View\s*\}\s*=\s*exportModule/);
+    expect(allCode).not.toContain('const pendingPrototype = {};');
+    expect(allCode).toContain('__loadRemote__ckeditor5__loadRemote__');
+    expect(allCode).toContain('const dependencyPending = importModule && importModule.__mf_remote_dependency_pending;');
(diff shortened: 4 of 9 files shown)


gioboa (commented):

Image

GauBen:
Yep it works!

I'm now fighting with duplicate react versions but that's unrelated

dmchoi77:
@gioboa  Thanks, this direction makes sense to me. I also verified that it covers the reported direct-import case.

One question about dependency discovery coverage: `collectRemoteDependencies` appears to scan only the exposed source file for direct `import ... from` / `import("...")` remote references.

Do we also need to handle cases where the remote dependency is reached indirectly, for example:

```js
// exposed init module
import "./plugin";

// plugin.js
import { View } from "ckeditor5";
class Foo extends View {}
```

or via re-exports from the exposed module:
```
export { View } from "ckeditor5";
export * from "ckeditor5";
```
In those cases, the expose wrapper may not preload ckeditor5 before importing the exposed module, so evaluation can still reach the transformed remote binding before __mf_remote_pending is awaited.

Maybe these are out of scope for this PR, but I wanted to check whether the preload scan is intended to cover only direct remote imports or the exposed module dependency graph as well.

GauBen:
Thank you very much @gioboa and @dmchoi77 for your reactivity on this issue!

PR #880, opened 2026-07-06
fix: fix dev host auto init hash

gioboa:
Ensures dev host auto init resolves hash-pattern remote entry filenames to stable URLs, preventing literal placeholder imports during serve.


gioboa (commented):

Image

PR #883, opened 2026-07-06
fix: reuse shared cache before loading

### LINKED ISSUE #881: [Bug]: Shared dependencies are downloaded again, even when the share scope already contains them
### Describe the bug

When several remotes share the same dependency in the same share scope, navigating from one remote to another downloads the shared dependency again — even though it is already present in the share cache in a matching version.

## Setup

Typical shell architecture: a thin shell that shares nothing loads three remotes via `@module-federation/bridge-react`.

| App   | Shared deps       | Version | Share scope |
| ----- | ----------------- | ------- | ----------- |
| shell | –                 | –       | –           |
| app-1 | react, react-dom  | 19.2.7  | `react@19`  |
| app-2 | react, react-dom  | 19.2.7  | `react@19`  |
| app-3 | react, react-dom  | 18.3.1  | `react@18`  |

## Steps

1. `npm run install:all && npm run build && npm run preview`
2. Open http://localhost:5000, DevTools → Network, filter `loadShare`, "Disable cache" on.
3. The shell loads app-1; `Object.keys(globalThis.__mf_module_cache__.share)` shows the scope seeded exactly once: `react@19:react`, `react@19:react-dom`, `react@19:react-dom/client`, `react@19:react/jsx-runtime`.
4. Navigate to **App 2**.

## Expected

app-2's shared modules are served from the already-seeded `react@19` scope — no further downloads for react/react-dom.

## Actual

react and react-dom are downloaded again (~200 kB raw, see log output below). This happens on every first visit of a remote.

### Version

vite 8.1.3, @module-federation/vite 1.16.12 (@module-federation/runtime 2.6.0)

### Reproduction

https://github.com/sodehl/mf-shared-scope-issue

### Relevant log output

```shell
# Network requests when navigating app-1 → app-2 (share scope react@19 already seeded):
_virtual_mf___mfe_internal__app2__loadShare__react__loadShare__.js-C5HeWLQI.js                        12.1 kB
_virtual_mf___mfe_internal__app2__loadShare__react_mf_2_dom__loadShare__.js-CL4WwmwQ.js                6.1 kB
_virtual_mf___mfe_internal__app2__loadShare__react_mf_1_jsx_mf_2_runtime__loadShare__.js-DINkSGTN.js   1.6 kB
_virtual_mf___mfe_internal__app2__loadShare__react_mf_2_dom_mf_1_client__loadShare__.js-DMJE1atM.js  180.3 kB
```

### Validations

gioboa:
Close #881 

Make singleton shared fallback lazy during build and reuse cached shared modules before importing local providers, preventing duplicate downloads.

### DIFF
--- a/e2e/vite-webpack-rspack/tests/index.spec.ts
+++ b/e2e/vite-webpack-rspack/tests/index.spec.ts
@@ -1,5 +1,15 @@
 import { expect, test } from '@playwright/test';
 
+function isDynamicRemoteSharedPayload(url: string) {
+  if (!url.startsWith('http://localhost:4002/')) return false;
+  if (url.includes('__loadShare__')) return false;
+  return (
+    url.includes('__prebuild__react') ||
+    url.includes('__prebuild__lodash') ||
+    /\/node_modules\/\.vite\/deps\/(?:react|react-dom|lodash)(?:[._-]|$)/.test(url)
+  );
+}
+
 test.describe('Vite Host Tests', () => {
   test.beforeEach(async ({ page, baseURL }) => {
     await page.goto(baseURL!);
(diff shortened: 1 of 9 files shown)


dmchoi77:
Hi! @gioboa One thing I’m wondering about from the CI results:

Vite 5/6/7 integration passes, but Vite 8 fails because it cannot find the `localSharedImportMap` chunk.

The Playwright trace also shows the browser trying to fetch `virtual:mf-REMOTE_ENTRY_ID:...` directly, which then fails with CORS / `net::ERR_FAILED`.

Could this be caused by a behavior difference in how Vite 8/Rolldown handles `import(/* @vite-ignore */ "virtual:...")` compared to Vite 5/6/7?

gioboa (commented):
@dmchoi77 Thanks for your check 
This is not a simple fix, the pipeline with all the frameworks is failing as well 🥲

gioboa (commented):

Image

PR #890, opened 2026-07-09
fix: resolve wildcard subpath patterns in package exports

jonasfrisell:
### Problem

Sharing a workspace package whose `package.json` uses **wildcard `exports` subpaths** (the layout shadcn/ui and Base UI generate) fails in dev. Given:

```json
{
  "exports": {
    "./components/*": "./components/*.tsx"
  }
}
```

the generated `loadShare` module throws:

```
The requested module '...' does not provide an export named 'Button'
```

(same for `Select`, `CommandDialog`, etc.). The shared module exposes only a `default` export.

### Root Cause

`getPackageExportsTarget` (`src/utils/packageUtils.ts`) resolves a subpath against the `exports` map by **exact-key lookup only** (`record[subpath]`). For `@scope/ui/components/button`, `subpath` is `./components/button`, but the map only contains the wildcard key `./components/*`, so the lookup returns `undefined`, resolution falls back to `index.js`, and named-export detection yields `[]`.

The chain that surfaces it:

```
getSharedNamedExports
  -> getPackageNamedExports        // require() fails on a .tsx source
  -> getEsmNamedExports
  -> getPackageEsmEntryPath
  -> getInstalledPackageEntry({ resolveSubpathWithRequire: false })
  -> getPackageExportsTarget       // returns undefined for the wildcard subpath
```

### Fix

Replace the exact-key lookup in `getPackageExportsTarget` with Node's `exports` subpath pattern matching (`PACKAGE_IMPORTS_EXPORTS_RESOLVE`): the exact key wins first, otherwise the best pattern by `PATTERN_KEY_COMPARE` (longest literal `patternBase`, then longest key), and the captured `patternMatch` is substituted into the resolved target. Two small helpers do this (`matchExportsSubpath` and `substituteExportsWildcard`), covering string, conditional-object, and fallback-array targets.

### Tests

Adds 5 cases to `src/utils/__tests__/packageUtils.test.ts`: wildcard resolution into string, conditional-object, and fallback-array targets, plus longest-prefix and equal-base tiebreak selection.

- Verified the 5 new tests **fail on `main`** (exact-key lookup) and **pass with the fix**; the 13 pre-existing tests pass unchanged.
- Full suite: **636 passing**. `pnpm fmt.check` and `pnpm typecheck` clean.

### Impact / Scope

- **Purely additive.** The exact-key path is byte-identical, so every package that resolved before resolves identically. Only previously unresolved wildcard subpaths change, from broken to correct. The `.` (root) path is untouched.
- **Faithful to Node's resolution.** A shared package's subpaths resolve to the same target Node itself would, so the shared copy matches what the host app imports directly.
- This is the entry-resolver counterpart to the recent workspace named-export fixes (#864, #860, #870), which addressed `virtualShared_preBuild.ts` but not the `exports`-map resolution in `packageUtils.ts` where this gap lived.

### Reproduction

Any workspace monorepo where a shared package's `package.json` `exports` uses a wildcard subpath pointing at source (e.g. `"./components/*": "./components/*.tsx"`) and lists the package in `shared`. Importing any component subpath from the federated module then throws in dev.

### DIFF
--- a/src/utils/__tests__/packageUtils.test.ts
+++ b/src/utils/__tests__/packageUtils.test.ts
@@ -87,6 +87,118 @@ describe('getInstalledPackageJson', () => {
 
     expect(entry).toBe(path.join(packageDir, 'dist/browser.js'));
   });
+
+  it('resolves wildcard subpath exports (e.g. "./components/*")', () => {
+    const packageName = 'mf-test-wildcard-ui';
+    const root = mkdtempSync(path.join(tmpdir(), 'mf-vite-wildcard-'));
+    tempDirs.push(root);
+
+    const hostDir = path.join(root, 'apps/host');
+    const packageDir = path.join(hostDir, 'node_modules', packageName);
+    mkdirSync(path.join(packageDir, 'components'), { recursive: true });
+    writeFileSync(path.join(hostDir, 'package.json'), JSON.stringify({ name: 'host' }));
+    writeFileSync(
+      path.join(packageDir, 'package.json'),
+      JSON.stringify({
+        name: packageName,
+        exports: { './components/*': './components/*.tsx' },
+      })
+    );
+    writeFileSync(
+      path.join(packageDir, 'components/button.tsx'),
+      'export const Button = () => null;'
+    );
+
+    const entry = getInstalledPackageEntry(`${packageName}/components/button`, {
+      cwd: hostDir,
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

gioboa (approved):
Thanks @jonasfrisell for your great improvement 🚀

PR #892, opened 2026-07-09
fix: seed subpath shares before their package root

YurySolovyov:
### Problem

Regression in **1.16.14** (works in 1.16.13), dev serve only. When a symlinked (`file:`/workspace) package is shared as a singleton **together with one of its exact-key subpath exports**, and the package's own source imports that subpath through a self-referencing bare specifier:

```js
// packages/widgets/index.js
import { builder } from 'widgets/api';

export const root = builder.post({ path: '/search' }); // ← TypeError
```

```js
// vite.config.js
shared: {
  widgets: { singleton: true },
  'widgets/api': { singleton: true },
}
```

the subpath share resolves with **undefined named exports** at module-evaluation time:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'post')
```

Sharing only the subpath (without the root) works; `vite build` is unaffected. This is the dev-serve reincarnation of the failure family previously fixed for production builds in #767 (fixed by #774) and #805 (fixed by #807).

### Root Cause

Two changes from #883 combine:

1. Serve-mode share proxies no longer bind an **eager local fallback** — on a cache miss they defer export assignment into `pendingShareLoads` until after federation init. Correct for cache reuse, but it makes seeding order load-bearing: any share proxy hit during seeding must already be cached.
2. The new runtime seed loop (`generateRuntimeSharedCacheSeedCode`) iterates `Object.entries(usedShared)`, whose key order is **alphabetical** (`getUsedShares().sort()`). `"widgets"` sorts before `"widgets/api"`, so seeding the package root `await share.get()` evaluates `widgets/index.js` while `widgets/api` is still unseeded — its proxy defers, its named exports stay `undefined`, and the root's top-level code crashes.

The dependency-first ordering from #807 (`orderSharedDependenciesFirst`) already exists but wasn't used by the new seed loop, and it didn't know that a package root implicitly depends on its own subpath shares.

### Fix

- `orderSharedDependenciesFirst` now treats a package's subpath share keys as dependencies of the package root: when visiting a root key, its subpath keys are visited (and therefore ordered) first. Subpath keys also resolve their dependency list through the root package's `package.json` when the subpath itself doesn't resolve to one.
- `generateRuntimeSharedCacheSeedCode` seeds in `getOrderedUsedShares()` order (embedded at codegen time) instead of `usedShared` insertion order, with a runtime fallback that appends any share keys discovered after codegen so nothing is skipped.

This preserves #883's cache-reuse semantics — no eager fallback is reintroduced; the fix only makes the seeding order match the evaluation-order requirements the deferred fallback created. For the example above the seed order becomes `['widgets/api', 'widgets']`, so by the time `widgets/index.js` evaluates, the `widgets/api` proxy hits a warm cache and applies its exports synchronously.

Known limitation: a genuine evaluation-time cycle (root imports subpath **and** subpath imports root) cannot be solved by ordering alone; that case was equally broken before this change.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -632,6 +632,97 @@ describe('virtualRemoteEntry', () => {
     expect(code).toContain('for (const pkg of Object.keys(usedShared))');
   });
 
+  it('orders package subpath shares discovered after remoteEntry codegen before their package root', async () => {
+    const shareItem = (name: string) => ({
+      name,
+      from: '',
+      version: '1.0.0',
+      scope: 'default',
+      shareConfig: {
+        singleton: true,
+        requiredVersion: '^1.0.0',
+        strictVersion: false,
+      },
+    });
+    normalizedSharedMock.mockReturnValue({
+      '@repro/shared-lib': shareItem('@repro/shared-lib'),
+    });
+    const mod = await import('../virtualRemoteEntry');
+
+    mod.getUsedShares().clear();
+    mod.addUsedShares('@repro/shared-lib');
+
+    const code = mod.generateRemoteEntry(
+      {
+        internalName: '__mfe_internal__host',
+        name: 'host',
+        filename: 'remoteEntry.js',
+        exposes: {},
+        remotes: {},
+        shared: normalizedSharedMock(),
+        runtimePlugins: [],
+        shareScope: 'default',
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

gioboa (commented):

Image

gioboa (approved):
Thanks @YurySolovyov for your help 👏

YurySolovyov:
Thanks for taking a look and all the follow ups!

PR #897, opened 2026-07-10
fix: solve runtime host shared singleton race

### LINKED ISSUE #895: [Bug]: Shared singleton React is duplicated when a pure `@module-federation/runtime` host provides it to a `@module-federation/vite` remote
## Describe the bug

When the **host is built with the runtime API only** (`@module-federation/runtime` / `@module-federation/enhanced/runtime`, no bundler plugin) and provides `react` as a shared singleton, a **remote built with `@module-federation/vite`** does not bind the host's React instance. Instead the remote falls back to its **own bundled React**, producing two React copies. React hooks then fail with `Invalid hook call` (and, when React subpaths are involved, `Cannot read properties of null (reading 'S')`).

This happens in **both `vite` (dev/serve) and `vite build` (preview)**.

## Environment

- `@module-federation/vite`: **1.16.14**
- `@module-federation/runtime` / `@module-federation/sdk`: **2.7.0**
- Host uses `@module-federation/enhanced`: **2.5.0** (re-exports the runtime)
- react / react-dom: **19.2.4**
- vite: **8.1.0** (rolldown-vite)
- node: **24.16.0**

## Reproduction

I did not create a separate reproduction repository. The scenario is reproducible using the example that already exists in this repository: **`examples/vite-runtime-register/`** (`vite-host` = runtime-only host, `vite-remote` = `@module-federation/vite` remote). The snippets below mirror that example.

**Host** — runtime only, no bundler plugin. Registers React as a shared singleton (eager `lib`), then loads a remote:

```js
import React from 'react';
import ReactDOM from 'react-dom';
import { createInstance } from '@module-federation/enhanced/runtime';

const mf = createInstance({
  name: 'hostApp',
  remotes: [],
  shared: {
    react: {
      version: React.version,
      scope: 'default',
      lib: () => React, // eager
      shareConfig: { singleton: true, requiredVersion: `^${React.version}` },
    },
    'react-dom': {
      version: ReactDOM.version,
      scope: 'default',
      lib: () => ReactDOM,
      shareConfig: { singleton: true, requiredVersion: `^${ReactDOM.version}` },
    },
  },
});

mf.registerRemotes([{ name: 'runtimeRemote', entry: 'http://localhost:4176/mf-manifest.json', type: 'module' }]);

const Card = (await mf.loadRemote('runtimeRemote/MessageCard')).default;
// render <Card /> inside the host's React tree
```

**Remote** — built with `@module-federation/vite`, exposes a component that uses a hook:

```js
// vite.config.js
federation({
  name: 'runtimeRemote',
  manifest: true,
  exposes: { './MessageCard': './src/MessageCard.jsx' },
  shared: {
    react: { singleton: true, requiredVersion: '^19.1.0' },
    'react-dom': { singleton: true, requiredVersion: '^19.1.0' },
  },
});
```

```jsx
// src/MessageCard.jsx
import { useState } from 'react';
export default function MessageCard() {
  const [count, setCount] = useState(0); // fails: dual React
  return <button onClick={() => setCount((c) => c + 1)}>count: {count}</button>;
}
```

**Steps** (using the in-repo example):

1. `pnpm install`
2. `pnpm --filter 'examples-vite-runtime-register*' --parallel run preview`
3. Open the host at `http://localhost:4175`, click **registerRemotes()**, then **loadRemote() component**.

The exposed component must call a React hook (e.g. `useState`) for the duplication to surface as a visible error; a hook-less component may render without an obvious crash even though two React copies exist.

## Expected behavior

The remote component binds the **host's** React instance (a single React across host and remote); hooks work.

## Actual behavior

The remote binds its **own** bundled React. React is duplicated. Rendering the exposed component throws:

- `Invalid hook call. Hooks can only be called inside of the body of a function component.`
- or `TypeError: Cannot read properties of null (reading 'S')` (when a React subpath such as `react-dom/client` is consumed but not shared).

`window.__hostReact__ !== window.__remoteReact__` (verified by exposing both).

## Root cause (investigated)

A vite remote does not consume normal shared modules through the runtime's `loadShare` at the import site. Instead, `import 'react'` is rewritten to a generated `__loadShare__react` virtual module that resolves React via a **vite-private cache** `globalThis.__mf_module_cache__.share`. Two facts combine to cause the bug:

1. **The `__loadShare__<pkg>` module is evaluated *before* the container's `init()` runs.** `init()` contains the "bridge" that copies the host's shared modules into `__mf_module_cache__.share`. So at the moment `__loadShare__react` reads the cache, it is still empty. (Confirmed by instrumenting the generated code: `__loadShare__react` read/fallback logs fire before the `init()` bridge logs.)

2. Depending on `command`/scope (`usesDeferredSingletonFallback` in `src/virtualModules/virtualShared_preBuild.ts`, introduced in commit `909d84e` / #883), the module takes one of two paths, and **both fail for a pure runtime host:**
   - **eager path** (`vite build` + default scope): binds the remote's own local React synchronously at evaluation time — before the `init()` bridge fills the cache.
   - **deferred path** (`vite` dev; or `vite build` + non-default scope): defers via `initPromise`, but the current implementation then **imports the local copy without re-reading the shared cache** (so it ignores the host copy the bridge placed), and — critically — the pending work is tracked in `__mfModuleCache.pendingShareLoads`, which is **only awaited by a vite-built host's entry bootstrap** (`src/plugins/pluginAddEntry.ts`). A pure runtime host loads the exposed module via the container's `getExposes`, which **does not await `pendingShareLoads`**, so the component renders before the deferred binding completes.

Relevant source:
- `src/virtualModules/virtualShared_preBuild.ts` — `writeLoadShareModule`, `generateLazyWorkspaceSingletonExports`, `usesDeferredSingletonFallback`.
- `src/virtualModules/virtualRemoteEntry.ts` — container `init()` bridge and `getExposes`.
- `src/plugins/pluginAddEntry.ts` — `pendingShareLoads` await (host entry bootstrap only).

## What was verified

- The bug reproduces in both dev and build with the setup above.
- Instrumentation confirmed `__loadShare__<pkg>` evaluates before the `init()` bridge, so the cache is empty at bind time.
- **Configuration alone does not fix it.** Using eager `lib`, sharing every consumed React subpath (`react`, `react/jsx-runtime`, `react-dom`, `react-dom/client`), and using a non-default `shareScope` still fails on the current code, because the deferred path does not re-read the cache and `getExposes` does not await `pendingShareLoads`.

## Possible fix (verified locally)

Two small plugin changes make the runtime-host scenario work without regressing the existing vite-host ↔ vite-remote behavior:

1. In the deferred path (`generateLazyWorkspaceSingletonExports`), after `initPromise` resolves, **re-read `__mf_module_cache__.share`** and prefer the host-provided copy; fall back to the local copy only if the cache is still empty.
2. In `getExposes` (`virtualRemoteEntry.ts`), **await `__mfModuleCache.pendingShareLoads`** before returning the exposed module, so a pure runtime host also waits for deferred shared bindings to complete.

With these two changes (plus the user-side prerequisites below), the remote binds the host's React. This was validated with a Playwright e2e (remote hook works + single React instance) and confirmed not to regress the existing `vite-vite` host/remote e2e. Removing either change reproduces the failure.

**User-side prerequisites for the fix:**
- Register shared modules with eager `lib` (the `init()` bridge only copies providers whose `lib` is set).
- Share every consumed React subpath: `react`, `react/jsx-runtime`, `react-dom`, `react-dom/client`.
- Use a non-default `shareScope` on the remote so `vite build` selects the deferred path.

## Additional notes / open questions

- With the fix, `vite build` + **default** `shareScope` still uses the eager path (kept that way to preserve standalone remote rendering), so the runtime-host scenario currently requires a **non-default** `shareScope`. Supporting the default scope too would require choosing eager/deferred at runtime (e.g. detecting whether an external provider exists), which is a larger change.
- The bridge skips providers registered lazily (`get` without `lib`) via `if (!provider.lib) continue`, so the host must use eager `lib`.

### Version

^6.4.3

### Reproduction

Reproduces via this repo's example `examples/vite-runtime-register/` (host = runtime-only, remote = @module-federation/vite); see the issue body for steps. 

### Relevant log output

```shell

```

### Validations

gioboa:
Close #895

Fixes shared singleton races for runtime-only hosts by rechecking the shared cache after initialization and awaiting pending shared loads before exposing remote modules.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -490,6 +490,7 @@ describe('virtualRemoteEntry', () => {
     expect(code).toContain('const {usedShared, usedRemotes} = await getLocalSharedImportMap()');
     expect(code).toContain('const exposesMap = await getExposesMap()');
     expect(code).toContain('const mfName = "host"');
+    expect(code).toContain('await Promise.all(__mfModuleCache.pendingShareLoads)');
     expect(code).toContain('share.shareConfig?.import !== false');
     expect(code).toContain('const versions = shared?.[pkg]');
     expect(code.indexOf('share.shareConfig?.import !== false')).toBeLessThan(
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -2128,6 +2128,9 @@ describe('writePreBuildLibPath', () => {
     // instead of a bare initPromise.then()
     expect(generatedCode).toContain('pendingShareLoads');
     expect(generatedCode).toContain('initPromise.then');
+    expect(generatedCode).toContain('__mfReadSharedCache(__mfModuleCache.share');
+    expect(generatedCode).toContain('if (exportModule !== undefined)');
+    expect(generatedCode).toContain('return import(');
 
     // Must use the ||= pattern for lazy initialization
     expect(generatedCode).toContain('||= []');
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -583,6 +583,9 @@ export function generateRemoteEntry(
   async function getExposes(moduleName) {
     const exposesMap = await getExposesMap()
     if (!(moduleName in exposesMap)) throw new Error(\`[Module Federation] Module \${moduleName} does not exist in container.\`)
+    if (__mfModuleCache.pendingShareLoads) {
+      await Promise.all(__mfModuleCache.pendingShareLoads)
+    }
     return (exposesMap[moduleName])().then(res => () => res)
   }
   export {
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -715,13 +715,18 @@ function generateLazyWorkspaceSingletonExports(
       if (import.meta.env.SSR) {
         ${applyLocalFallback}
       } else {
-        (__mfModuleCache.pendingShareLoads ||= []).push(initPromise.then(() =>
-          import(${escapeGeneratedStringLiteral(importSource)}).then((mod) => {
+        (__mfModuleCache.pendingShareLoads ||= []).push(initPromise.then(() => {
+          exportModule = __mfReadSharedCache(__mfModuleCache.share, ${cacheDescriptor});
+          if (exportModule !== undefined) {
+            __mfApplyLazyShareExports(exportModule);
+            return;
+          }
+          return import(${escapeGeneratedStringLiteral(importSource)}).then((mod) => {
             exportModule = __mfNormalizeShareModule(mod);
             __mfWriteSharedCache(__mfModuleCache.share, ${cacheDescriptor}, exportModule);
             __mfApplyLazyShareExports(exportModule);
-          })
-        ));
+          });
+        }));
       }
     } else {
       __mfApplyLazyShareExports(exportModule);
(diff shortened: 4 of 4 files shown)


gioboa (commented):

Image

PR #898, opened 2026-07-10
fix: support star re-exports in shared modules

### LINKED ISSUE #896: [Bug]: Federated module misses star re-exports when `import` is set on shared dependency
### Background

Vue ships two distinct builds ([docs](https://vuejs.org/guide/scaling-up/tooling.html#note-on-in-browser-template-compilation)):

- `vue/dist/vue.runtime.esm-bundler.js` — runtime-only. No template compiler included. Templates must be pre-compiled by a build tool (e.g. `@vitejs/plugin-vue`). This is the default exposed via `package.json` exports.
- `vue/dist/vue.esm-bundler.js` — full build. Includes the template compiler for in-browser template compilation.

When a host needs to share Vue with remotes that require in-browser template compilation, it must point the shared module at the full build via the `import` option.

### The bug

When `import` is set on a shared singleton in the federation config, the build fails with `[MISSING_EXPORT]` errors for every Vue export. The virtual loadShare module generated by `@module-federation/vite` ends up not re-exporting any of Vue's named exports.

This is not limited to the full build. The same failure also reproduces when `import` points at the default runtime build (`vue/dist/vue.runtime.esm-bundler.js`), which is what Vue's `package.json` `exports` field resolves to by default. It does not happen without `import` set.

**To reproduce:**

```ts
shared: {
    vue: {
        singleton: true,
        import: 'vue/dist/vue.esm-bundler.js', // <-- this causes the bug
        // import: 'vue/dist/vue.runtime.esm-bundler.js', // <-- same bug with the default export
    },
},
```

**Expected:** build/dev server succeeds.

**Actual (build):** build fails with 9 `[MISSING_EXPORT]` errors:

```
Build failed with 4 errors:

[MISSING_EXPORT] "createApp" is not exported by "\0virtual:mf:__mfe_internal__host__loadShare__vue__loadShare__.js".
   ╭─[ src/main.ts:3:10 ]
   │
 3 │ import { createApp } from "vue";
   │          ────┬────  
   │              ╰────── Missing export
───╯

[MISSING_EXPORT] "createElementVNode" is not exported by "\0virtual:mf:__mfe_internal__host__loadShare__vue__loadShare__.js".
   ╭─[ src/App.vue:2:10 ]
   │
 2 │ import { createElementVNode as _createElementVNode, openBlock as _openBlock, createElementBlock as _createElementBlock } from "vue"
   │          ─────────┬────────  
   │                   ╰────────── Missing export
───╯

[MISSING_EXPORT] "openBlock" is not exported by "\0virtual:mf:__mfe_internal__host__loadShare__vue__loadShare__.js".
...
```

**Actual (dev server):** the dev server starts, but the app fails to load. The browser console shows:

```
Uncaught (in promise) SyntaxError: The requested module '/@id/__x00__virtual:mf:__mfe_internal__host__loadShare__vue__loadShare__.js' does not provide an export named 'createElementBlock' (at App.vue:2:78)
```

### Investigation

Both Vue entry files re-export most of their public API via a star export from `@vue/runtime-dom`:

```js
// vue/dist/vue.esm-bundler.js
export * from '@vue/runtime-dom';
// ...
export { compileToFunction as compile };

// vue/dist/vue.runtime.esm-bundler.js
export * from '@vue/runtime-dom';
// ...
export { compile };
```

The missing exports (`createApp`, `createElementVNode`, `openBlock`, `createElementBlock`, etc.) are all provided through that `export * from '@vue/runtime-dom'` line. The virtual loadShare module generated by `@module-federation/vite` does not surface star re-exports from the shared entry.

By contrast, the explicit `export { compile }` at the bottom of each file is a direct named export on the entry module itself, and that export does come through. So the failure is specific to star re-exports, not to every export on the Vue entry file.

### Version

1.16.14

### Reproduction

https://github.com/levrik/vite-module-federation-repro

### Validations

gioboa:
Close #896

Fix shared module export detection for package-level export * re-exports, allowing Vue full builds to expose named exports correctly.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -82,6 +82,9 @@ vi.mock('../../utils/packageUtils', () => ({
   getPackageDetectionCwd: vi.fn(() => '/repo/apps/remote'),
   resolveImportPath: vi.fn(() => '/repo/node_modules/@module-federation/runtime/dist/index.js'),
   getInstalledPackageEntry: vi.fn((pkg: string, opts?: { cwd?: string }) => {
+    if (pkg === 'mock-package-star-dependency') {
+      return '/repo/apps/remote/node_modules/mock-package-star-dependency/index.js';
+    }
     if (pkg === 'mock-package-esm-only/stores' || pkg === 'mock-package-esm-only') {
       return '/repo/apps/remote/node_modules/mock-package-esm-only/dist/stores.js';
     }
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #900, opened 2026-07-10
fix: preserve requiredVersion false in shared config

tonoizer:
## Summary

- support the official `requiredVersion: false` shared configuration
- preserve `false` during normalization instead of replacing it with an inferred semver range
- align Vite behavior with the Module Federation 2.0 SDK, runtime, and Rspack implementation

## Testing

- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`
- [x] `pnpm test`
- [x] `pnpm test:integration`
- [x] `pnpm build`

## Risks

Low. The change only affects configurations that explicitly set `requiredVersion: false`. Omitted and string-valued requirements retain their existing behavior.

### DIFF
--- a/src/utils/__tests__/normalizeModuleFederationOption.test.ts
+++ b/src/utils/__tests__/normalizeModuleFederationOption.test.ts
@@ -288,6 +288,23 @@ describe('normalizeModuleFederationOption', () => {
       });
     });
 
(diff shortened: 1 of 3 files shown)


gioboa (approved):
Thanks @tonoizer you rock 🚀

PR #901, opened 2026-07-10
fix: preserve repeated runtime plugin option values

tonoizer:
## Summary

- distinguish repeated runtime plugin option values from true circular references
- serialize repeated objects, arrays, maps, and sets in each independent branch
- keep the existing `__circular__` marker for cycles in the active recursion path

## Testing

- [x] `pnpm exec vitest run src/utils/__tests__/serializeRuntimeOptions.test.ts`
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`
- [x] `pnpm test`
- [x] `pnpm test:integration`
- [x] `pnpm build`

## Risks

Low. Runtime plugin options containing repeated object references now preserve their values instead of replacing later occurrences with the string `__circular__`. True circular references retain the previous fallback.


gioboa (approved):
Thanks @tonoizer this is a good fix 🚀

PR #902, opened 2026-07-10
fix: identify var remote entry in production manifest

tonoizer:
## Summary

- identify `varRemoteEntry` as a `var` container in production manifests
- align production metadata with the dev manifest and generated global wrapper
- add regression coverage for the emitted manifest metadata

## Testing

- [x] `pnpm exec vitest run src/plugins/__tests__/pluginMFManifest.test.ts`
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`
- [x] `pnpm test`
- [x] `pnpm test:integration`
- [x] `pnpm build`

## Risks

Low. This changes only the reported type of the optional production `varRemoteEntry` metadata from `module` to `var`; the generated remote entry asset is unchanged.

### DIFF
--- a/src/plugins/__tests__/pluginMFManifest.test.ts
+++ b/src/plugins/__tests__/pluginMFManifest.test.ts
@@ -128,6 +128,7 @@ async function runGenerateBundleWithManifest(
     >;
     dts?: unknown;
     filename?: string;
+    varFilename?: string;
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks @tonoizer that's a nice fix 👏

PR #905, opened 2026-07-11
fix: fix shared dependency initialization order

### LINKED ISSUE #903: Standalone-host shared-dep bootstrap runs alphabetically, crashes React-touching shared packages sorting before "react"
### Bug: standalone-host shared-dependency bootstrap runs in alphabetical order, not dependency order — crashes any React-touching shared package that sorts before "react"

**Package:** `@module-federation/vite`
**Version:** 1.16.15 (confirmed latest; also reproduced on 1.16.14)

**Summary**

When a Vite app built with `federation({ shared: {...} })` runs in dev mode with no external host present (i.e. it bootstraps itself as its own MF host via the auto-injected `virtual:mf-html-entry-proxy` → `hostAutoInit` → `remoteEntry.init()` sequence), the generated `init()` populates every configured `shared` package by iterating `usedShared` in **plain alphabetical order**:

```js
// generated in virtual:mf-localSharedImportMap:<name>
const usedShared = {
  ${Array.from(getUsedShares()).sort().map(key => `...`).join(",")}
}
```

and then, inside the generated `remoteEntry` `init()`:

```js
for (const [pkg, share] of Object.entries(usedShared)) {
  ...
  const factory = await share.get(); // dynamically imports & evaluates the whole package
  ...
}
```

`share.get()` synchronously imports and evaluates the *entire* module graph of that shared package. If a shared package's name sorts alphabetically **before** `"react"` (e.g. an internal component-library package, `@tanstack/react-query` — both start with `@`), and that package touches any React API (`createContext`, `forwardRef`, etc.) at **module-evaluation time** (not just at render time — a completely standard pattern; `@tanstack/react-query`'s `QueryClientContext` and virtually every Radix UI primitive do this), its bundle executes *before* `"react"` gets its own turn in the same loop to populate the shared cache. At that point, the per-consumer "react" shared-import shim (also generated by this plugin) has not resolved yet, so the imported `React` binding is an empty/undefined stub, and the shared package throws:

```
TypeError: e.createContext is not a function
    at .../dist/chunks/utils-*.js  (React.createContext called at module scope)
```
or, once that specific call site is patched:
```
TypeError: React.forwardRef is not a function
    at .../@radix-ui/react-slot ... (Radix's createSlot(), also module-scope)
```
or, with the internal component-library package removed from `shared` entirely:
```
TypeError: React.createContext is not a function
    at .../@tanstack/react-query ... (QueryClientContext, also module-scope)
```

**This is order-dependent, not package-dependent** — whichever shared package alphabetically precedes `"react"` and touches a React API at import time will fail, regardless of which specific package it is.

**Reproduction**

1. Vite app with `@module-federation/vite`'s `federation()` plugin, `shared: { react: { singleton: true }, 'some-package-starting-before-r': { singleton: true } }`, where `some-package-starting-before-r` calls e.g. `React.createContext(...)` at module top level (or depends on something that does, like Radix or TanStack Query).
2. Run `vite dev` with **no external MF host** consuming this app (i.e. run it standalone — the app acts as its own host).
3. Load the page. `[Unhandled rejection] TypeError: <ReactAPI> is not a function` fires from the shared package's chunk.

**Also checked, for completeness:**
- The per-package `eager` field (as documented informally / expected by consumers building their own `SharedDependency` config) is **not read anywhere** in `normalizeShareItem()` — only `import`, `singleton`, `requiredVersion`, `strictVersion` are consumed. There is currently no working per-package "eager"/synchronous-load escape hatch to sidestep this ordering issue.
- Confirmed on both 1.16.15 (latest) and 1.16.14 — same behavior on both.

**Expected behavior**

Either:
- Shared packages should be initialized in an order that guarantees a package's own declared/inferred dependencies (at minimum `react`/`react-dom` for any package using them) resolve first, or
- The generated per-package "consume shared" virtual module should defer evaluation of the *importing* package's own module graph until its shared dependencies are ready (i.e. don't let `share.get()` synchronously trigger module evaluation of code that itself needs another not-yet-ready shared package), or
- A genuinely-functional `eager` flag should exist so `react`/`react-dom` can be guaranteed synchronously available before any other shared package is evaluated.

**Workaround status:** none found that doesn't require dropping React-dependent packages from `shared` entirely (which just shifts the failure to whichever remaining shared package sorts next).

gioboa:
Close #903

Ensures shared dependencies initialize in the correct dependency order.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -130,6 +130,18 @@ vi.mock('../../utils/packageUtils', () => {
       return match ? match[0] : packageString;
     },
     getInstalledPackageJson: (pkg: string) => {
+      if (pkg === '@repro/react-consumer') {
+        return {
+          path: '/repo/packages/react-consumer/package.json',
+          dir: '/repo/packages/react-consumer',
+          packageJson: {
+            name: '@repro/react-consumer',
+            dependencies: {
+              react: '^19.0.0',
+            },
+          },
+        };
+      }
       if (pkg === '@repro/core') {
         return {
           path: '/repo/packages/core/package.json',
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #907, opened 2026-07-11
fix: preserve remote aliases during resolution

tonoizer:
## Summary

- preserve the consumer-facing remote alias during Vite resolution
- escape aliases before using them in generated regular expressions
- register aliases in both loaded-first and version-first runtime initialization
- avoid repeating the remote-alias lookup during code generation
- add regression coverage for aliases that differ from container names

## Feedback addressed

The version-first `usedRemotes` configuration now includes `alias: key`, so imports such as `catalog/Button` resolve when the configured container name is `catalogContainer`.

## Verification

- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm test -- --run` — 691 passed, 1 skipped
- `pnpm test:integration -- --run` — 43 passed
- `pnpm build`
- `pnpm exec playwright test --project=vite-vite-host --project=vite-vite-remote` — 9 passed

No Rspack/API-parity changes are included.

### DIFF
--- a/src/plugins/__tests__/pluginProxyRemotes.test.ts
+++ b/src/plugins/__tests__/pluginProxyRemotes.test.ts
@@ -131,6 +131,49 @@ describe('pluginProxyRemotes', () => {
     expect(alias.find.test('scheduler/SchedulePanel')).toBe(true);
   });
 
+  it('matches and records the consumer alias when it differs from the container name', () => {
+    const plugin = pluginProxyRemotes(
+      normalizeModuleFederationOptions({
+        name: 'host',
+        remotes: {
+          catalog: {
+            type: 'module',
+            name: 'catalogContainer',
+            entry: 'http://example.com/remoteEntry.js',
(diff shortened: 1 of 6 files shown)


gioboa (changes requested):
This preserves the alias only for the loaded-first path. With the default version-first strategy, remotes are initialized from usedRemotes in virtualRemoteEntry.ts, where the generated remote config still omits alias.

As a result, with:
```
remotes: {
 catalog: {
  name: 'catalogContainer',
  entry: '...',
  type: 'module',
 },
}
```

import('catalog/Button') is intercepted by the plugin, but the runtime only registers catalogContainer, so loadRemote('catalog/Button') cannot resolve it.
Please include alias: key in the usedRemotes config as well, and add a regression test using the default version-first strategy.

gioboa (approved):
LGTM 👏 
@tonoizer you are on fire 🔥

PR #908, opened 2026-07-11
fix: report remote container names in manifest stats

tonoizer:
## Summary

- report the configured federation container name in manifest stats
- handle string-form remotes where `entryGlobalName` differs from the alias
- retain object-form remote names when normalization defaults `entryGlobalName` to the alias
- add regression coverage for both object-form and string-form remotes

## Feedback addressed

For `remote1: Button@https://...`, the manifest now reports `Button` as `federationContainerName` and `remote1` as `alias`.

## Verification

- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm test -- --run` — 689 passed, 1 skipped
- `pnpm test:integration -- --run` — 43 passed
- `pnpm build`
- `pnpm exec playwright test --project=vite-vite-host --project=vite-vite-remote` — 9 passed

No Rspack/API-parity changes are included.

### DIFF
--- a/src/plugins/__tests__/pluginMFManifest.test.ts
+++ b/src/plugins/__tests__/pluginMFManifest.test.ts
@@ -116,7 +116,7 @@ async function runGenerateBundleWithManifest(
   manifestOptions: unknown,
   runtime: {
     usedShares?: Set<string>;
-    usedRemotes?: Map<string, Set<string>>;
+    usedRemotes?: Map<string, Set<string>> | Record<string, Set<string>>;
     exposePaths?: Record<string, { import: string }>;
     shareItems?: Record<
       string,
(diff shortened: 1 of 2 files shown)


gioboa (changes requested):
This works for object-form remotes, but not for string-form remotes where the container name differs from the alias.

For example:
```
remotes: {
 remote1: 'Button@https://cdn.example.com/remoteEntry.js',
}
```

Normalization sets name to remote1 and entryGlobalName to Button, so this line reports the alias instead of the actual federation container name: federationContainerName: options.remotes[remoteKey].name
Could we handle entryGlobalName for string-form remotes and add a regression test for this case?

gioboa (approved):
👏 @tonoizer that's great

PR #912, opened 2026-07-13
fix: refresh nested remote dependencies before exposes

### LINKED ISSUE #911: [Bug]: Nested remote dependencies in exposed modules are not refreshed before generating virtual exposes
### Describe the bug

# Nested remote dependencies in exposed modules are not refreshed before generating virtual exposes

## Package

`@module-federation/vite`

## Affected versions

Confirmed in:

- `1.16.14`
- `1.16.16`

## Environment

- Vite 6
- Vue 3
- Vue Router 4
- `@module-federation/vite@1.16.16`
- `@module-federation/enhanced` runtime
- Development mode using `vite dev`
- Multiple Module Federation remotes

## Summary

When an exposed module imports another Module Federation remote, its nested remote dependencies are collected during the `buildStart()` hook and stored in `exposeRemoteDependencies`.

However, the cached dependency information may be incomplete or stale when the virtual exposes module is generated, especially during development startup.

The `load()` and `transform()` hooks call `generateExposes()` directly without refreshing the dependency graph:

```js
if (id === virtualExposesId) {
  return generateExposes(options, exposeRemoteDependencies, command);
}
```

As a result, the generated expose loader may not wait for nested remote dependencies to finish initializing before evaluating the exposed module.

In our Vue application, the first route navigation to a federated page fails with a pending Promise being passed to Vue Router as a navigation error.

Refreshing the exposed remote dependency graph immediately before calling `generateExposes()` fixes the problem.

## Example

Assume the following Module Federation topology:

```text
Host
 └── loads Remote A
      └── exposes ./example
           └── imports Remote B
```

Remote A exposes a Vue page, while that page imports functionality from Remote B.

### Remote B configuration

```ts
// remote-b/vite.config.ts

import vue from '@vitejs/plugin-vue';

export default defineConfig({
  plugins: [
    vue(),
    federation({
      name: 'remote-b',
      filename: 'remoteEntry.js',
      manifest: true,
      exposes: {
        './theme-core': './src/theme-core/index.ts',
      },
      shared: {
        vue: {
          singleton: true,
        },
      },
    }),
  ],
});
```

Remote B exposes a simple composable:

```ts
// remote-b/src/theme-core/index.ts

import { ref } from 'vue';

const theme = ref('light');

export function useThemeConfig() {
  return {
    theme,
  };
}
```

### Version

1.6.14

### Reproduction

master

### Relevant log output

```shell

```

### Validations

gioboa:
Close #911

Refresh nested remote dependencies lazily when the dependency graph is invalidated, ensuring virtual exposes wait for transitive remotes while avoiding repeated filesystem scans during development and builds.

### DIFF
--- a/src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
@@ -4,6 +4,7 @@ import type {
   ResolvedConfig,
   Rollup,
 } from 'vite';
+import { resolve } from 'node:path';
 import { describe, expect, it } from 'vitest';
 import { getDefaultMockOptions } from '../../utils/__tests__/helpers';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #914, opened 2026-07-13
fix: optimize shared singleton React in dev

### LINKED ISSUE #913: [Bug]: Shared singleton React is served as raw CommonJS in dev since 1.16.12
### Describe the bug

A minimal Vite 8 dev app fails on page load when React is configured as a shared singleton. The regression starts with **@module-federation/vite 1.16.12**; the same app works with 1.16.11.

No remote application, React Redux, or @vitejs/plugin-react is needed to reproduce it.

## Actual behavior

The generated Module Federation prebuild module imports React from the raw package entry:

~~~js
import * as __mfPrebuildNamespace from "/node_modules/react/index.js?v=...";
~~~

React's package entry is CommonJS and contains module.exports. Because the file is sent directly to the browser, startup fails with **ReferenceError: module is not defined**.

## Expected behavior

Shared singleton React should work in Vite dev mode without sending raw CommonJS to the browser and without creating a second React instance.

## Regression matrix

All runs used vite --force to clear dependency optimization state.

| @module-federation/vite | Result | Generated React import |
| --- | --- | --- |
| 1.16.11 | Works | /node_modules/.vite/deps/react.js?v=... |
| 1.16.12 | Fails | /node_modules/react/index.js?v=... |
| 1.16.16 | Fails | /node_modules/react/index.js?v=... |
| 1.16.12 with optimizeDeps.include: ['react'] | Works | /node_modules/.vite/deps/react.js?v=... |

## Pinpointed regression

This was introduced by #861, landed as commit [3bfe17c](https://github.com/module-federation/vite/commit/3bfe17cc20aabd981f9225d10a5363dc2cbf9dea), and first released in 1.16.12.

The dev configuration changed from excluding root React only when React Redux is a direct dependency:

~~~ts
isLitShare(key) ||
  (key === 'react' && hasPackageDependency('react-redux', root))
~~~

to excluding it for every shared singleton React configuration:

~~~ts
isLitShare(key) ||
  (key === 'react' && shareItem.shareConfig?.singleton === true)
~~~

That adds React to optimizeDeps.exclude. Vite then skips dependency optimization, while Module Federation's generated prebuild module still imports the root React entry. The resolved URL changes from the CommonJS-compatible /node_modules/.vite/deps/react.js wrapper to raw /node_modules/react/index.js.

The configuration-level test added in #861 verifies that React is excluded, but does not exercise the generated module in a browser, so it misses this runtime failure.

Explicitly adding React to optimizeDeps.include proves the cause, but is not a complete fix because it may reintroduce the duplicate-React behavior that #861 intended to address. The generated local provider/fallback needs CommonJS interop while preserving singleton routing.

## Environment

- @module-federation/vite: 1.16.12 (first failing), 1.16.16 (still failing)
- Working comparison: 1.16.11
- Vite: 8.0.16
- React / React DOM: 18.3.1
- @vitejs/plugin-react: not installed
- Node: 24.18.0
- npm: 11.16.0
- OS: macOS 26.5.1 arm64

Related: #812 documented the same runtime consequence under a different React Redux/plugin-ordering trigger; #858 describes the duplicate-React issue that #861 intended to solve.

This is distinct from #903 / #905: that issue concerns initialization order between React and another shared package. This reproduction contains only React and React DOM and fails because the browser receives React's raw CommonJS entry.

### Version

8.0.16

### Reproduction

https://github.com/veithlukas/mf-vite-react-singleton-regression

### Relevant log output

```shell
[vite] (client) [Unhandled rejection] ReferenceError: module is not defined
 > node_modules/react/index.js?v=da7f0748:6:2

Generated prebuild module in 1.16.12 and 1.16.16:

import * as __mfPrebuildNamespace from "/node_modules/react/index.js?v=...";

Generated prebuild module in 1.16.11:

import __vite__cjsImport0_react from "/node_modules/.vite/deps/react.js?v=...";
```

### Validations

dmchoi77:
Fixes #913

Follow-up to #861 and #862

## Summary

This restores Vite dependency optimization for shared singleton React in dev mode.

#861 excluded every shared singleton React entry from `optimizeDeps` to prevent a remote-local React copy from bypassing Module Federation singleton sharing. However, local shared-provider fallbacks still need Vite's dependency optimizer to convert React's CommonJS package entry into browser-compatible ESM.

Without that conversion, the generated React prebuild module imports the raw package entry:

```txt
/node_modules/react/index.js
```

The browser then evaluates React's CommonJS `module.exports` code directly and fails with:

```txt
ReferenceError: module is not defined
```

This change keeps singleton identity in the federation `loadShare` and shared-cache layers introduced by the #862 follow-up, while allowing Vite to handle CommonJS-to-ESM interop.

## Problem

The fix in #861 used `optimizeDeps.exclude` as a singleton identity boundary:

```ts
isLitShare(key) ||
  (key === 'react' && shareItem.shareConfig?.singleton === true)
```

That avoids Vite pre-bundling React, but it also prevents Vite from generating its CommonJS interoperability wrapper.

As a result, the local React provider changes from:

```txt
/node_modules/.vite/deps/react.js
```

to:

```txt
/node_modules/react/index.js
```

The latter is raw CommonJS and cannot execute directly in the browser.

Explicitly adding React to `optimizeDeps.include` confirms the interop issue, but the complete fix must also preserve the singleton behavior that #861 intended to protect.

Singleton selection and dependency representation are separate responsibilities:

- Vite dependency optimization provides CommonJS-to-ESM interop.
- Module Federation's `loadShare` and shared cache preserve singleton identity.

## Changes

- Removed the singleton React exception from the dev `optimizeDeps` bypass condition.
- Kept the existing Lit optimization bypass unchanged.
- Allowed shared singleton React to use Vite's optimized ESM provider.
- Preserved the existing include-wins normalization behavior.
- Updated the React `optimizeDeps` unit tests for:
  - `react-redux` installed,
  - `react-redux` absent,
  - explicit React inclusion.
- Added explicit coverage ensuring Lit remains excluded.
- Added a dev integration regression test that:
  - verifies the React prebuild provider imports optimized `/deps/react.js`,
  - verifies raw `/react/index.js` is not exposed,
  - starts host and remote Vite servers in separate processes and cache directories,
  - renders a remote React hook component with the host's `react-dom`,
  - verifies host and remote React objects have strict identity,
  - verifies no `Invalid hook call`, `module is not defined`, or browser `pageerror` occurs.
- Generates the integration fixture in a temporary directory during the test so no permanent example files are required.

## Tests

Added coverage in `src/__tests__/index.test.ts` for:

- shared singleton React with `react-redux`,
- shared singleton React without `react-redux`,
- explicit React optimization precedence,
- the existing Lit optimization exception.

Added `integration/dev-singleton-react.test.ts` coverage for:

- optimized React provider generation,
- raw CommonJS React entry prevention,
- host/remote singleton identity,
- remote `useState` hook rendering through host `react-dom`,
- browser runtime errors.

The integration test uses the root Vite installation, so the existing integration CI matrix exercises the same test with Vite 5, 6, 7, and 8.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -458,12 +458,13 @@ export default __mfShared.default ?? __mfShared;`,
             const optimizeDeps = (config.optimizeDeps ??= {});
             optimizeDeps.include ??= [];
             optimizeDeps.exclude ??= [];
-            // Some packages must stay outside dependency optimization because
-            // their submodules rely on parent initialization order. Other
-            // shared deps should remain optimizable so Vite can apply CJS
-            // interop for transitive dependencies used by prebuild fallbacks.
-            const shouldBypassOptimizeDep =
-              isLitShare(key) || (key === 'react' && shareItem.shareConfig?.singleton === true);
+            // Lit must stay outside dependency optimization because its
+            // submodules rely on parent initialization order. Other shared
+            // deps, including singleton React, must remain optimizable so
+            // local prebuild fallbacks receive Vite's CJS-to-ESM interop.
+            // Singleton identity is enforced by the federation share cache
+            // and loadShare proxy, independently from dependency optimization.
+            const shouldBypassOptimizeDep = isLitShare(key);
             if (optimizeDeps.include.includes(key)) {
               optimizeDeps.exclude = optimizeDeps.exclude.filter((dep) => dep !== key);
             } else if (shouldBypassOptimizeDep) {
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -1,0 +1,324 @@
+import { spawn, type ChildProcess } from 'node:child_process';
+import { once } from 'node:events';
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { createServer as createNetServer } from 'node:net';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { chromium } from '@playwright/test';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { createServer as createViteServer, version as viteVersion } from 'vite';
+import { federation } from '../src';
+
+const REPO_ROOT = path.resolve(import.meta.dirname, '..');
+const FIXTURE_PARENT = path.join(REPO_ROOT, 'examples/vite-webpack-rspack/remote');
+const VITE_CLI = path.join(REPO_ROOT, 'node_modules/vite/bin/vite.js');
+
+const cleanupTasks: Array<() => Promise<void>> = [];
+let fixtureRoot: string;
+
+beforeEach(async () => {
+  fixtureRoot = await mkdtemp(path.join(FIXTURE_PARENT, '.issue-913-'));
+  await mkdir(path.join(fixtureRoot, 'src'));
+  const pluginEntry = pathToFileURL(path.join(REPO_ROOT, 'src/index.ts')).href;
+  await Promise.all([
+    writeFile(
+      path.join(fixtureRoot, 'index.html'),
+      '<div id="root"></div>\n<script type="module" src="/src/main.js"></script>\n'
+    ),
+    writeFile(
+      path.join(fixtureRoot, 'src/main.js'),
+      `import React from 'react';
+import { createRoot } from 'react-dom/client';
+
+window.__issue913HostReact = React;
+const { default: RemoteHookComponent } = await import('issue913Remote/HookComponent');
+createRoot(document.querySelector('#root')).render(
+  React.createElement(RemoteHookComponent, { hostReact: React })
+);
+`
+    ),
+    writeFile(
+      path.join(fixtureRoot, 'src/RemoteHookComponent.js'),
+      `import React, { useState } from 'react';
+
+window.__issue913RemoteReact = React;
+export default function RemoteHookComponent({ hostReact }) {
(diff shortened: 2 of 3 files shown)


dmchoi77:
@gioboa I’ve opened this PR to address #913 while preserving the React singleton behavior addressed by #861 and #862. It also includes a browser-level regression test covering the optimized React provider and host/remote React identity.
When you have a chance, I’d really appreciate it if you could take a look.
Thank you!

src/index.ts
```diff
@@ -458,12 +458,13 @@ export default __mfShared.default ?? __mfShared;`,
...
-            // interop for transitive dependencies used by prebuild fallbacks.
-            const shouldBypassOptimizeDep =
-              isLitShare(key) || (key === 'react' && shareItem.shareConfig?.singleton === true);
+            // Lit must stay outside dependency optimization because its
+            // submodules rely on parent initialization order. Other shared
+            // deps, including singleton React, must remain optimizable so
+            // local prebuild fallbacks receive Vite's CJS-to-ESM interop.
+            // Singleton identity is enforced by the federation share cache
+            // and loadShare proxy, independently from dependency optimization.
+            const shouldBypassOptimizeDep = isLitShare(key);
```
gioboa, line 467: Is #792 still working?

dmchoi77: Confirmed. #792 still works correctly with this PR—useSelector renders without an invalid hook call.

gioboa (commented):

Image

gioboa (approved):
Thanks for the quick fix 💯  @dmchoi77

PR #916, opened 2026-07-13
fix: handle circular SSR module graphs

tonoizer:
## Summary
- reserve deterministic temp-file paths before traversing remote SSR imports
- avoid awaiting an ancestor's in-flight fetch when an ESM graph contains a cycle
- wait for every discovered module writer before importing the completed graph
- clear reserved path entries together with the existing SSR loader cache

## Testing
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`
- [x] `pnpm test` (746 passed, 1 skipped)
- [x] `pnpm build`
- [x] packaged-loader reproduction: a two-module circular SSR graph resolves instead of timing out

## Risks
- The change is scoped to the HTTP-to-temp-file SSR loading strategy. It preserves existing deterministic filenames and cache keys, but changes in-flight cache hits to use a reserved destination path so circular graphs can complete.

### DIFF
--- a/src/utils/__tests__/ssrEntryLoader.test.ts
+++ b/src/utils/__tests__/ssrEntryLoader.test.ts
@@ -760,6 +760,38 @@ describe('ssrEntryLoaderPlugin — code transformation', () => {
     );
   });
 
+  it('finishes fetching circular relative import graphs', async () => {
+    const fsMock = await import('fs');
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

gioboa (approved):
@tonoizer that's awesome 💪 Thank you

PR #917, opened 2026-07-13
fix: bound SSR remote fetches

tonoizer:
## Summary

- abort SSR manifest, candidate probe, module graph, VM-strategy, and ModuleRunner transport requests after a configurable timeout
- default each request to a 10-second limit while allowing `fetchTimeoutMs: 0` to opt out
- evict failed manifest, entry-resolution, and temp-module cache records so a transient timeout can be retried

## Why

A remote that accepts a connection but never completes its response can currently leave server rendering pending indefinitely. The pending promise is also cached in several paths, so later renders remain stuck even after the remote recovers.

## Verification

- `pnpm vitest run src/utils/__tests__/ssrEntryLoader.test.ts src/utils/__tests__/ssrVmStrategy.test.ts`
- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm test` (747 passed, 1 skipped)
- `pnpm build`

### DIFF
--- a/src/utils/__tests__/ssrEntryLoader.test.ts
+++ b/src/utils/__tests__/ssrEntryLoader.test.ts
@@ -21,7 +21,7 @@ type FetchEntry = {
 };
 
 function makeFetchMock(responses: Record<string, FetchEntry>) {
-  return vi.fn(async (url: string, _options?: { method?: string }) => {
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Image

gioboa (approved):
Thank you @tonoizer well done 👍

PR #918, opened 2026-07-14
fix: resolve lazy shared providers

Nsttt:
## Summary

- materialize lazy/get-only shared providers through the Module Federation Runtime before exposing remote modules
- keep singleton proxy exports live when runtime selection replaces a local fallback
- seed root-host singletons before version-first remote initialization, while deferring child fallbacks to real parent providers
- preserve provider provenance, cache ownership, runtime plugin hooks, and circular-init scope
- preserve normal `beforeLoadShare` → `resolveShare` → `afterLoadShare` ordering while honoring custom provider selection
- keep unknown and transpiled CommonJS export surfaces on one coherent module instance
- cover get-only React sharing with a browser regression that executes `useState`

## Root cause

The deployed host supplied React 18 as an unloaded `get` provider, while the layout remote bundled React 17. The Vite remote bridge only copied providers with an existing `lib`, so it skipped the host provider and seeded its local React into the Vite cache. The Runtime share scope selected host React, but the remote's generated imports still resolved to local React; ReactDOM then rendered hooks from a different React instance and raised minified React error #321.

## Verification

- `pnpm fmt.check`
- `pnpm typecheck`
- `pnpm test` — 816 passed, 1 skipped
- `pnpm test:integration` — 45 passed
- `pnpm build`
- Playwright browser matrix — 21 passed across Vite, Webpack, Rspack, and runtime-register scenarios
- deployed-chunk control — upstream reproduces the null `useState` crash; this branch resolves the remote to the host React export identity and mounts successfully

### DIFF
--- a/e2e/vite-runtime-register/runtime-register.spec.ts
+++ b/e2e/vite-runtime-register/runtime-register.spec.ts
@@ -1,0 +1,27 @@
+import { expect, test } from '@playwright/test';
+
+test('uses a runtime host get-only React singleton', async ({ page }) => {
+  const pageErrors: Error[] = [];
+  page.on('pageerror', (error) => pageErrors.push(error));
+
+  await page.goto('/');
+  await page.getByRole('button', { name: 'registerRemotes()' }).click();
+  await page.getByRole('button', { name: 'loadRemote() component' }).click();
+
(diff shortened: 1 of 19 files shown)


gioboa:

image

gioboa (approved):
Thanks that's a great improvement @Nsttt

PR #921, opened 2026-07-14
feat: support multiple provider share scopes

tonoizer:
## Summary
- accept `string | string[]` for top-level and per-remote `shareScope` configuration
- initialize every configured provider scope in browser and SSR remote entries
- preserve the existing single-scope path and the current circular-init/lazy-provider protections
- cover normalization and browser/SSR code generation for multiple scopes

## Testing
- [x] `pnpm test` (824 passed, 1 skipped)
- [x] `pnpm run typecheck`
- [x] `pnpm run fmt.check`
- [x] `pnpm run build`

## Risks
- The change touches generated remote-entry initialization. Array configuration uses the new multi-scope path; string configuration retains the existing path and is covered by the full source test suite.

### DIFF
--- a/src/utils/__tests__/normalizeModuleFederationOption.test.ts
+++ b/src/utils/__tests__/normalizeModuleFederationOption.test.ts
@@ -59,6 +59,15 @@ describe('normalizeModuleFederationOption', () => {
     });
   });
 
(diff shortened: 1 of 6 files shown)


gioboa (approved):
Thanks @tonoizer 👍

PR #922, opened 2026-07-14
fix: solve errors in the examples

### LINKED ISSUE #920: [Bug]: TypeError: Cannot read properties of undefined (reading 'S') when using `exposes` in Module Federation config
### Describe the bug

After the latest release, running a Vite app that uses Module Federation with an `exposes` 
config throws the following runtime error in **dev mode only**:

Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'S')

The error does NOT occur when building the app (`vite build`) — only when running 
the dev server (`vite dev`). Removing `exposes` entirely from the Module Federation 
plugin config resolves the issue in dev mode, confirming it is directly tied to 
the exposed module resolution in the dev server.

This appears to be a regression introduced in the latest release.

### Version

1.17.1

### Reproduction

NA

### Relevant log output

```shell
TypeError: Cannot read properties of undefined (reading 'S') when using `exposes` in Module Federation config
```

### Validations

gioboa:
Close #920

### DIFF
--- a/examples/vite-webpack-rspack/dynamic-remote/vite.config.js
+++ b/examples/vite-webpack-rspack/dynamic-remote/vite.config.js
@@ -8,7 +8,6 @@ export default defineConfig({
     react(),
     federation({
       name: "dynamicRemote",
-      hostInitInjectLocation: "entry",
       filename: "remoteEntry.js",
       // Modules to expose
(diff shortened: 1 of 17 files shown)


gioboa (commented):

Image

PR #925, opened 2026-07-14
fix: order React roots before subpaths

### LINKED ISSUE #924: [Bug]: TypeError: can't access property "ReactCurrentDispatcher", ReactSharedInternals is undefined
### Describe the bug

  After upgrading `@module-federation/vite` from `1.16.12` to `1.16.16`, a React 18 Vite remote fails during development with:

  ```text
  Uncaught (in promise) TypeError: can't access property "ReactCurrentDispatcher", ReactSharedInternals is undefined
  ```

  All installed copies of `react` and `react-dom` resolve to `18.3.1`, so this does not appear to be caused by duplicate or mismatched React versions.

  The generated federation bootstrap initializes React subpath shares before the root `react` package:

  ```js
  const __mfHostInitShareOrder = [
    'react/compiler-runtime',
    'react/jsx-dev-runtime',
    'react/jsx-runtime',
    'react',
    'react-dom/client',
    'react-dom/server',
    'react-dom/server.browser',
    'react-dom'
  ];
  ```

  When `react/jsx-runtime` is evaluated, React's federation share has not been initialized. The React 18 JSX runtime then executes:

  ```js
  var ReactSharedInternals =
    React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;

  var ReactCurrentDispatcher =
    ReactSharedInternals.ReactCurrentDispatcher;
  ```

  At that point, `ReactSharedInternals` is `undefined`, causing the application to crash.

  ### Regression

  - `@module-federation/vite@1.16.12`: works
  - `@module-federation/vite@1.16.16`: fails

  I have not bisected the intermediate releases to identify the first affected version.

  ### Versions

  ```text
  @module-federation/vite: 1.16.16
  vite: 8.0.16
  react: 18.3.1
  react-dom: 18.3.1
  node: 24.15.0
  ```

### Minimal reproduction

  ```js

  export default defineConfig({
    plugins: [
      react(),
      federation({
        name: 'example-remote',
        filename: 'remoteEntry.js',
        dts: false,
        exposes: {
          './App': './src/App.jsx'
        },
        shared: {
          react: {
            singleton: true,
            requiredVersion: '^18.3.1'
          },
          'react-dom': {
            singleton: true,
            requiredVersion: '^18.3.1'
          },
          'react/jsx-runtime': {
            singleton: true,
            requiredVersion: '^18.3.1'
          }
        }
      })
    ]
  });
  ```

  The app uses an asynchronous bootstrap:

  ```js
  // src/main.js
  import('./bootstrap.jsx');
  ```

  ```jsx
  // src/bootstrap.jsx
  import React from 'react';
  import { createRoot } from 'react-dom/client';

  createRoot(document.getElementById('root')).render(
    <React.StrictMode>
      <div>Module Federation remote</div>
    </React.StrictMode>
  );
  ```

  Run:

  ```shell
  vite --open
  ```

  ### Expected behavior

  Package roots should be initialized before their subpath exports:

  ```text
  react
  react/compiler-runtime
  react/jsx-dev-runtime
  react/jsx-runtime
  react-dom
  react-dom/client
  react-dom/server
  react-dom/server.browser
  ```

  The page should render without a React internals error.

  ### Actual behavior

  React subpaths are initialized before `react`, and ReactDOM subpaths are initialized before `react-dom`. Evaluating `react/jsx-runtime`
  consequently throws because the root React share has not populated its exports.

  ### Suspected cause

  The ordering behavior appears related to the subpath-ordering logic added after `1.16.12`.

  In `1.16.16`, `getOrderedUsedShares()` collects package subpaths and visits them before adding the root package to the result:

  ```js
  if (pkg === packageName) {
    (subpathKeysByPackageName.get(packageName) || []).forEach(visit);
  }

  ordered.push(pkg);
  ```

  This overrides the earlier React priority sort and effectively models the root package as depending on all its subpaths. The actual runtime
  dependency is the reverse: `react/jsx-runtime` requires the initialized `react` package.

  Could subpath shares explicitly depend on their package root, or otherwise be ordered after the root package?

### Version

8.0.16

### Reproduction

See "Minimal reproduction" in above description

### Relevant log output

```shell

```

### Validations

gioboa:
Close #924

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -833,6 +833,43 @@ describe('virtualRemoteEntry', () => {
     expect(hostInit).toContain('const __mfHostInitShareOrder = ["react","@repro/react-consumer"]');
   });
 
+  it('orders React package roots before their subpath shares', async () => {
+    const share = (name: string) => ({
+      name,
+      version: '18.3.1',
+      scope: 'default',
+      shareConfig: { singleton: true, strictVersion: false },
+    });
+    normalizedSharedMock.mockReturnValue({
+      'react/jsx-runtime': share('react/jsx-runtime'),
+      react: share('react'),
+      'react-dom/client': share('react-dom/client'),
+      'react-dom': share('react-dom'),
+    });
+
+    const mod = await import('../virtualRemoteEntry');
+    mod.getUsedShares().clear();
+    mod.addUsedShares('react/jsx-runtime');
+    mod.addUsedShares('react');
+    mod.addUsedShares('react-dom/client');
+    mod.addUsedShares('react-dom');
+
+    const hostInit = mod.generateHostAutoInitCode('"virtual:remoteEntry"', 'serve');
+    const orderMarker = 'const __mfHostInitShareOrder = ';
+    const orderStart = hostInit.indexOf(orderMarker);
+    const orderLineEnd = hostInit.indexOf('\n', orderStart);
+    const orderSource =
+      orderStart === -1 || orderLineEnd === -1
+        ? '[]'
+        : hostInit.slice(orderStart + orderMarker.length, orderLineEnd).trim();
+    const order = JSON.parse(
+      orderSource.endsWith(';') ? orderSource.slice(0, -1) : orderSource
+    ) as string[];
+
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #926, opened 2026-07-14
fix: isolate generated federation configuration

tonoizer:
## Summary
- bind remote wrappers, host-init modules, local shared maps, manifests, and expose preloads to the normalized options of their owning plugin instance
- partition virtual IDs, usage registries, SSR bootstrap remotes, explicit-share metadata, and dev invalidators by federation instance
- keep tree-shaking provider output and shared-cache seed ownership scoped to the captured configuration
- add production-shaped multi-instance regressions, including same-name instances and two dev servers

## Testing
- [x] `pnpm test` — 837 passed, 1 skipped
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`
- [x] `pnpm build`
- [x] `pnpm build:shared-lib`
- [x] `pnpm playwright test` — 21 passed
- [x] manual production-preview smoke — host and remote demos rendered; both shared counters incremented; no browser page errors
- [x] `git diff --check`

## Risks
- separate federation instances intentionally no longer reuse generated wrappers, host-init modules, local shared maps, or mutable usage/invalidation state
- legacy no-options helper fallbacks remain available for direct callers and existing single-instance behavior
- owner suffixes change internal virtual IDs only; emitted public container names and remote-entry contracts are unchanged


tonoizer:
The follow-up ownership fixes are pushed and the PR is ready. GitHub has placed the new fork workflow runs behind maintainer approval (`action_required`). Local verification is green: 837 unit tests passed (1 skipped), typecheck/format/build passed, all 21 Playwright scenarios passed, and the production-preview manual smoke reported no page errors. A maintainer approval of the pending Actions runs will allow the official CI matrix to execute.

tonoizer:
Updated the integration assertions in f89e402 to match the intentional per-instance `__mf_owner__<n>` suffix while still checking the complete direct and transitive remote IDs.

Local verification: full integration suite (45/45), full unit suite (837 passed, 1 skipped), typecheck, formatting, and build all pass. The new upstream workflows are currently waiting for fork-run approval; once approved, the Integration Test workflow will re-run the Vite 5/6/7/8 matrix.

gioboa (changes requested):
The overall direction makes sense: multiple federation instances need isolated
virtual-module IDs, usage registries, manifests, and host-init state.

However, I don’t think the isolation is complete yet. Several shared-module caches
remain process-global and are still keyed only by package name:

- src/virtualModules/virtualShared_preBuild.ts:802-807
- src/virtualModules/virtualShared_preBuild.ts:1085-1089
- src/virtualModules/virtualShared_preBuild.ts:1353-1362

These functions are invoked by each federation instance, but writeLoadShareModule() and the prebuild/tree-shaking helpers do not receive the owning NormalizedModuleFederationOptions. Two instances sharing the same dependency can therefore overwrite each other’s generated virtual module code, cache owner, tree-shaking state, or virtual IDs.

The build-time runtime-init helpers also still use the global virtualRuntimeInitStatus ID:

- src/virtualModules/virtualRuntimeInitStatus.ts:163-183
- src/virtualModules/virtualRemoteEntry.ts:992-996
- src/virtualModules/virtualShared_preBuild.ts:1396-1401

As a result, separate federation instances may still share the same initPromise/
initResolve global in production builds. SSR remote configuration has a similar
issue because setSsrRemotes() stores one process-global remote list.

tonoizer:
@gioboa Thanks for catching the remaining process-global ownership paths. Addressed in 6f68e26.

The follow-up now:
- scopes loadShare, prebuild, provider materialization, and tree-shaking analysis by the owning NormalizedModuleFederationOptions, including separate IDs for same-name instances
- uses a per-instance runtime-init barrier consistently in build and serve
- passes SSR remotes per instance and keeps the barrier owner separate from the host-init import
- preserves the legacy no-options helper fallbacks
- adds same-package/same-name regressions covering generated module ownership, build and serve barriers, SSR remote isolation, and tree-shaking provider isolation

Verification on the pushed commit is green: 860 unit tests passed (1 skipped), 45 integration tests passed, all 21 Playwright tests passed in a clean CI-style run, and typecheck, formatting, package build, shared-library build, and git diff checks passed.

Would appreciate another look when you have a chance.

gioboa (commented):
The instance scoping is much more complete now, but two process-global paths remain:

- __mf_module_cache__ is shared across federation instances, while remote cache keys use only the remote ID (remote/App). Two instances with the same alias can reuse each other’s pending promise or loaded module. Please scope the runtime cache or include the federation owner in its keys.

- pluginVarRemoteEntry still reads options through getNormalizeModuleFederationOptions() instead of capturing its owning normalized options. Multiple federation instances could therefore generate a var entry using the wrong name or filename.

- add two federation configs in the vite-vite example to validate this fix e2e

tonoizer:
@gioboa Thanks — the remaining paths are addressed in 0ea26f4.

- Remote pending/loaded cache entries are now keyed by the owning federation instance, and each instance also registers/loads through an owner-qualified internal runtime alias while preserving the public import ID. The runtime alias was necessary because the two-config E2E reproduced a second process-global collision below the JS cache layer.
- `pluginVarRemoteEntry` now captures the normalized options passed by its owning `federation()` call; it no longer rereads mutable global options during bundle generation.
- The `vite-vite` example now has two host federation configs and two remote federation configs. Both host configs consume the identical public ID `@namespace/viteViteRemote/InstanceMarker`, and the Playwright assertion verifies two distinct owner-scoped cache entries containing the primary and secondary modules. It also checks that both captured var entries generate the correct globals.

Full local verification on the pushed commit is green: formatting, typecheck, package build, 864 unit tests passed (1 skipped), 45/45 integration tests, and all 23 Playwright tests. I also ran an independent diff review; it found no blocking issues.

Would appreciate one final look when you have a chance.

gioboa (commented):

Image

gioboa (approved):
That's a really good fix @tonoizer 💯

PR #928, opened 2026-07-14
fix: constrain development runner boundary

tonoizer:
## Summary
- stop the federation runner middleware from overriding the Vite server CORS policy with a wildcard origin
- allow only supported `fetchModule` option keys and value types
- require `startOffset` to be a non-negative safe integer capped at 1 MiB
- add regressions for CORS ownership and rejected or accepted runner options

## Testing
- [x] `pnpm test src/plugins/__tests__/pluginSSRRemoteEntry.test.ts`
- [x] `pnpm test` — 840 passed, 1 skipped
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`
- [x] `pnpm build`
- [x] `pnpm build:shared-lib`
- [x] `pnpm playwright test` — 21 passed locally
- [x] manual production-preview smoke — federated host/remote content rendered with no browser page errors
- [x] GitHub unit, Playwright, Vite 5–8 integration, Node 22/24/25 package, lint, title, and publish checks
- [x] `git diff --check`

## Risks
- browser clients that relied on the runner route wildcard must configure the intended origin through Vite CORS settings
- server-to-server SSR requests are not subject to browser CORS enforcement

### DIFF
--- a/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
@@ -650,6 +650,55 @@ describe('pluginSSRRemoteEntry', () => {
       expect(JSON.parse(res.body)).toEqual({ error: { message: 'Invalid runner invoke' } });
     });
 
+    it.each([
+      '/@id/__x00__/../../etc/passwd',
+      '/@id/__x00__../../etc/passwd',
+      '/@id/__x00__..%2F..%2Fetc%2Fpasswd',
+    ])('rejects traversal hidden behind a Vite-encoded null prefix: %s', async (unsafeId) => {
+      const handleInvoke = vi.fn().mockResolvedValue({ result: { code: 'escaped' } });
+
+      const res = await invokeRunnerMiddleware(
+        {
+          type: 'custom',
+          event: 'vite:invoke',
+          data: {
+            id: 'send',
+            name: 'fetchModule',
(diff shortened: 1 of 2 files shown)


gioboa (changes requested):
This check allows every ID that starts with \0 to bypass validation:

```
if (!decoded || rawDecoded.startsWith('\0') || decoded.startsWith('virtual:')) {
   return !!decoded;
}
```

Vite encodes these IDs as /@id/__x00__/..., so a value such as /@id/__x00__/../../etc/passwd could avoid the filesystem-boundary checks.

Please restrict null-prefixed IDs to known internal virtual modules, or validate the decoded path before accepting it. Please also add a regression test for this case

tonoizer:
Addressed the requested null-prefix validation change in bdcb01c. Null-prefixed IDs no longer bypass URL/traversal/filesystem validation. Regression tests cover the exact `/@id/__x00__/../../etc/passwd` case, canonical and percent-encoded variants, and preservation of a legitimate Vite-encoded internal ID.

Local verification: focused runner suite (55/55), full unit suite (843 passed, 1 skipped), full integration suite (45/45), typecheck, formatting, and build all pass. The new upstream workflows are currently waiting for fork-run approval. @gioboa, this is ready for re-review after those workflows are approved.

gioboa (approved):

Image

gioboa (approved):
it looks great to me @tonoizer

PR #933, opened 2026-07-15
fix: prevent Vite dependency optimization waves

### LINKED ISSUE #923: [Bug]: Dev-server cold start triggers multi-wave "optimized dependencies changed. reloading" cascade for MF shared singletons
### Environment

| | |
|---|---|
| `@module-federation/vite` | 1.17.1 |
| Vite | 8.x (Rolldown mode) |
| Framework | React 19 |
| Package manager | pnpm (strict/isolated `node_modules`) |

---

### What happens

On every cold start, the dev server triggers **4–5 consecutive "optimized dependencies changed. reloading" waves** for packages declared in the MF `shared` config. Each wave forces a full browser reload. Total settle time from server start to final reload is **3–4 minutes** on a reasonably fast machine.

Example wave sequence (varies by session):

1. `axios`
2. `@tanstack/react-query`, `zustand`
3. `zod`, UI component library packages
4. CSS utility packages (`clsx`, `tailwind-merge`, etc.)
5. `react-router-dom` — plus re-optimization of packages already seen in waves 1–3

Every package in these waves is explicitly declared as a `singleton` in the shell's `shared` config. Several (including `react-router-dom` and `@tanstack/react-query`) are also statically imported in `bootstrap.tsx` — the shell's earliest real entry point. Moving imports earlier makes no difference.

---

### Minimal shared config shape that reproduces it

```ts
federation({
  shared: {
    react: { singleton: true },
    'react-router-dom': { singleton: true },
    '@tanstack/react-query': { singleton: true },
    zustand: { singleton: true },
    axios: { singleton: true },
  },
});
```

With `bootstrap.tsx`:

```ts
// All already statically imported at the earliest entry point — no effect on wave timing
import 'react-router-dom';
import '@tanstack/react-query';
import 'zustand';
```

---

### Root cause (traced against plugin source)

The plugin's dev-mode resolver hook (`module-federation:optimize-shared-*`) only wires a shared specifier into the optimizer **the first time that hook actually fires** for that specifier. This is entirely decoupled from Vite's normal static-import graph scan.

The consequence: shared singletons are "discovered" lazily as the MF runtime resolves each specifier at runtime — some only when a `React.lazy` remote route first renders — triggering fresh optimizer passes and browser reloads for each new batch.

---

### What was tried (all ineffective)

| Attempt | Result |
|---------|--------|
| `optimizeDeps.include` with all shared package names | Zero change in wave count or timing. Also causes `Failed to resolve dependency` warnings for transitive deps not directly declared in the consuming app's own `package.json` (pnpm strict layout — resolution is project-root-relative, not importer-relative) |
| Narrowing shared specifiers to specific subpaths | Reduces transitive dep surface per package, marginal improvement, does not eliminate waves |
| `eager: true` on shared entries | Causes concurrent optimizer crash (`TypeError: Cannot read properties of undefined (reading 'browserHash')` in `commitProcessing`/`runOptimizer`) — reverted |
| Upgrading to 1.17.1 (latest) | No change in wave count or timing |

---

### Expected behavior

Packages declared as `shared` singletons should be registered into Vite's optimizer at `configResolved` time — before the first browser request — so they are included in the initial optimizer pass rather than discovered across multiple runtime waves.

---

Is there a planned mechanism to pre-seed the optimizer with all `shared` singletons at config time? The plugin already pushes literal `shared` keys into `optimizeDeps.include` in its `config()` hook — but the MF resolver hook's own first-fire timing gate means these still arrive late in practice.

Would a `federation({ preloadShared: true })` flag (or equivalent) be feasible to force eager registration of all declared shared entries before the first module request?

### Version

1.17.1

### Reproduction

N/A

### Relevant log output

```shell

```

### Validations

gioboa:
Close #923 

Fix Vite development dependency reload waves by pre-scanning linked shared packages and federation exposes, while preventing duplicate workspace singleton imports that caused __mfLocalShare syntax errors.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1414,6 +1414,57 @@ describe('vite:module-federation-early-init', () => {
     expect(virtualModule?.code).toContain('jsx');
   });
 
+  it('pre-seeds transitive shared dependencies for the dev optimizer', () => {
+    const plugin = (
+      federation({
+        name: 'host',
+        filename: 'remoteEntry.js',
+        shared: {
+          '@vite-vite/shared-lib': { singleton: true },
+        },
+      }) as Plugin[]
+    ).find((entry) => entry.name === 'vite:module-federation-early-init');
+    if (!plugin) throw new Error('vite:module-federation-early-init plugin not found');
+
+    const config: any = {
+      // Use the real vite-vite host fixture so this covers pnpm's isolated
+      // workspace dependency layout rather than a synthetic package graph.
+      root: REACT_EXAMPLE_ROOT,
+      optimizeDeps: { include: [], exclude: [], entries: ['custom-entry.ts'] },
+    };
+
+    runConfig(plugin, {} as ConfigPluginContext, config, {
+      command: 'serve',
+      mode: 'test',
+    });
+
+    // The linked shared library must be part of Vite's initial optimizer
+    // inputs, so its transitive imports are discovered before browser startup.
+    expect(config.optimizeDeps.entries).toContain(
+      path.join(REACT_EXAMPLE_ROOT, '..', 'shared-lib', 'src', 'index.tsx')
+    );
+    expect(config.optimizeDeps.entries).toContain('custom-entry.ts');
+  });
+
+  it('does not create optimizer entries for missing exposes', () => {
+    const plugin = (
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Image

PR #937, opened 2026-07-16
fix: fix scan complete flag when `default` gets reexported

smeng9:
#918 introduced a regression such that `default` and `__esModule` (for cjs interop) keywords cause `scanState.complete` to be false.

If the scan state is incomplete, it drops all exports. Then we would see errors like #15 5.442 [MISSING_EXPORT] "Stack" is not exported by "\0virtual:mf:__mfe_internal___mf_0_ when using @mui/material package.


smeng9:
The generated code should be export { __mfDefaultExport as default };
    export { __mf_0 as buttonBase, __mf_1 as buttonClasses };   // ← explicit named exports
So module-federation can resolve these imports, rather than export * from "/repo/packages/default-reexport-list.js"   // ← blind wildcard fallback

gioboa (commented):

Image

gioboa (approved):
Thanks @smeng9 for this great fix.

PR #939, opened 2026-07-16
fix: fix Vite 7 shared dependency optimization

### LINKED ISSUE #936: [Bug]: optimized deps cannot resolve virtual loadShare subpath module
### Describe the bug

This is a regression of https://github.com/module-federation/vite/issues/768. 

Seeing this in v1.16.12, but also latest v1.18.1. Updated the previous reproduction to those versions, seeing the same issue reproducible there again:

Image

### Version

1.18.1

### Reproduction

https://github.com/simonihmig/mfevite-optimize-deps-reproduction

### Relevant log output

```shell

```

### Validations

gioboa:
Close #936

Use virtual IDs for Vite 7 esbuild optimization, preserve Vite 8 Rolldown behavior.


gioboa (commented):

Image

PR #940, opened 2026-07-17
fix: only add modulePreload tags when not disabled

simonihmig:
The plugin should follow Vite's config (`build.modulePreload`) and not generate preload tags when disabled. We have our own plugin to generate these, and the two would conflict.

Related: https://github.com/module-federation/vite/pull/810

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -1,3 +1,6 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
 import type {
   ConfigEnv,
   ConfigPluginContext,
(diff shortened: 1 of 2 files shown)


src/plugins/__tests__/pluginAddEntry.test.ts
```diff
@@ -10,9 +13,6 @@ import type {
   ViteDevServer,
 } from 'vite';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import fs from 'fs';
-import os from 'os';
-import path from 'path';
```
simonihmig, line 15: Unrelated, caused  by https://github.com/module-federation/vite/blob/main/.vscode/settings.json#L3 I believe...

gioboa (approved):
Thanks @simonihmig nice catch 👏

PR #943, opened 2026-07-18
feat: support experiments.externalRuntime / provideExternalRuntime

tonoizer:
## Summary
- Add `experiments.externalRuntime` and `experiments.provideExternalRuntime` to `@module-federation/vite`, matching Rspack/webpack MF 2.0 option shapes.
- Pure-consumer hosts inject `@module-federation/inject-external-runtime-core-plugin` to publish `runtime-core` on `globalThis._FEDERATION_RUNTIME_CORE`.
- Remotes with `externalRuntime` rewrite `@module-federation/runtime-core` to a virtual shim that reads that global (with introspected named exports for Rollup/Rolldown), shrinking remote bundles.

## Testing
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm test:integration` (includes `integration/external-runtime.test.ts`)
- [x] `pnpm build`
- [ ] `pnpm run e2e:external` (opt-in Playwright: host provide + remote external + manifest HTTP check)

## Risks
- Remotes with `externalRuntime: true` will fail at runtime unless a host provides `_FEDERATION_RUNTIME_CORE` via `provideExternalRuntime`.
- `provideExternalRuntime` with any `exposes` throws (intentional parity with Rspack).
- New runtime deps (`inject-external-runtime-core-plugin`, `runtime-core`, `runtime-tools`) are pinned to the current MF 2.8.0 line.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,10 @@
       "types": "./lib/utils/ssrEntryLoader.d.ts",
       "default": "./lib/utils/ssrEntryLoader.js"
     },
+    "./injectExternalRuntimeCorePlugin": {
+      "types": "./lib/utils/injectExternalRuntimeCorePlugin.d.ts",
+      "default": "./lib/utils/injectExternalRuntimeCorePlugin.js"
+    },
     "./package.json": "./package.json"
   },
   "engines": {
@@ -74,10 +78,7 @@
   },
   "dependencies": {
     "@module-federation/dts-plugin": "2.8.0",
-    "@module-federation/inject-external-runtime-core-plugin": "2.8.0",
     "@module-federation/runtime": "2.8.0",
-    "@module-federation/runtime-core": "2.8.0",
-    "@module-federation/runtime-tools": "2.8.0",
     "@module-federation/sdk": "2.8.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
     "dev-vv": "pnpm clean && pnpm -filter 'examples-vite-vite*' run dev",
     "preview-vv": "pnpm clean && pnpm run build:shared-lib && pnpm -filter 'examples-vite-vite*' --parallel run preview",
     "preview-vv:ci": "pnpm run build:shared-lib && pnpm -filter 'examples-vite-vite*' --parallel run preview",
+    "preview-vv:external": "pnpm clean && pnpm run build:shared-lib && EXTERNAL_RUNTIME=1 pnpm -filter 'examples-vite-vite*' --parallel run preview",
+    "preview-vv:external:ci": "pnpm run build:shared-lib && EXTERNAL_RUNTIME=1 pnpm -filter 'examples-vite-vite*' --parallel run preview",
+    "e2e:external": "playwright test --config=playwright.external-runtime.config.ts",
     "build:shared-lib": "pnpm --filter @vite-vite/shared-lib run build",
     "typecheck": "tsc -p tsconfig.json",
     "multi-example:ci": "pnpm -filter 'multi-example-*' --parallel run start",
@@ -71,7 +74,10 @@
   },
   "dependencies": {
     "@module-federation/dts-plugin": "2.8.0",
+    "@module-federation/inject-external-runtime-core-plugin": "2.8.0",
     "@module-federation/runtime": "2.8.0",
+    "@module-federation/runtime-core": "2.8.0",
+    "@module-federation/runtime-tools": "2.8.0",
     "@module-federation/sdk": "2.8.0"
   },
   "devDependencies": {
--- a/src/plugins/pluginExternalRuntimeCore.ts
+++ b/src/plugins/pluginExternalRuntimeCore.ts
@@ -32,12 +32,18 @@ export function isSsrRemoteRuntimeImporter(importer: string | undefined): boolea
  * syntheticNamedExports the way Rollup does, and `@module-federation/runtime`
  * statically imports many symbols from runtime-core.
  */
-export function collectRuntimeCoreExportNames(
+export interface RuntimeCoreExportShape {
+  name: string;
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -71,7 +74,10 @@
   },
   "dependencies": {
     "@module-federation/dts-plugin": "2.8.0",
+    "@module-federation/inject-external-runtime-core-plugin": "2.8.0",
```
gioboa: I'm against adding a new dependency, is there a way to avoid these new ones?

src/plugins/pluginExternalRuntimeCore.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export async function resolveRuntimeCoreExportNames(): Promise<string[]> {
+  if (cachedExportNames) return cachedExportNames;
+  try {
+    const resolved = resolveImportPath(RUNTIME_CORE_PACKAGE);
+    const runtimeCore = (await import(pathToFileURL(resolved).href)) as Record<string, unknown>;
+    cachedExportNames = collectRuntimeCoreExportNames(runtimeCore);
+  } catch {
+    try {
+      const runtimeCore = (await import(RUNTIME_CORE_PACKAGE)) as Record<string, unknown>;
```
dmchoi77: It looks like we may be able to avoid a direct dependency on `@module-federation/runtime-core` by using `@module-federation/runtime/core` for export introspection instead.

package.json
tonoizer: Good call — resolved in d770247.

`@module-federation/inject-external-runtime-core-plugin` (and its `runtime-tools` peer) are no longer direct dependencies. `provideExternalRuntime` now injects a local equivalent at `@module-federation/vite/injectExternalRuntimeCorePlugin`, same approach as `ssrEntryLoader`. Behavior is unchanged: it still publishes `runtime-core` on `globalThis._FEDERATION_RUNTIME_CORE` in `beforeInit`.

If someone still wants the official package, they can add it manually via `runtimePlugins` — we still recognize that id and won't double-inject.

src/plugins/pluginExternalRuntimeCore.ts
tonoizer: Agreed — fixed in d770247.

Export introspection now uses `@module-federation/runtime/core` (already available via our existing `@module-federation/runtime` dependency). Confirmed the named export set matches `@module-federation/runtime-core`, so the direct `runtime-core` dependency is gone.

We still rewrite remotes' imports of `@module-federation/runtime-core` to the virtual shim — that's the import path remotes use at runtime, not a package we need to declare ourselves.

src/plugins/pluginExternalRuntimeCore.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export function collectRuntimeCoreExportNames(
+  runtimeCoreModule: Record<string, unknown>
+): string[] {
+  return Object.keys(runtimeCoreModule)
+    .filter((key) => key !== 'default' && key !== '__esModule')
+    .sort();
+}
+
+export function buildExternalRuntimeCoreShimCode(exportNames: string[]): string {
```
dmchoi77: I think this can still fail in Vite dev because import-time ordering is not guaranteed there.

`pluginExternalRuntimeCore` rewrites `@module-federation/runtime-core` to a shim that throws when
`globalThis._FEDERATION_RUNTIME_CORE` is missing.
In dev, remote graph modules can be evaluated before host injection runs, so this can still crash with
`_FEDERATION_RUNTIME_CORE is missing` even when config is otherwise valid.

Could we make this path more resilient, for example by avoiding a hard throw here (or by guaranteeing host global
registration before any remote graph evaluation), and add a dev-mode regression test in this branch?

tonoizer: Good catch — addressed in 3e16678.

Two complementary fixes for the Vite dev ordering race:



1. Lazy shim — no longer reads/throws at module evaluation; named exports are lazy Proxies that look up globalThis._FEDERATION_RUNTIME_CORE only on access.
1. Eager host publish — injectExternalRuntimeCorePlugin publishes runtime-core as a module side effect (metadata still applied in beforeInit).

Regression coverage added in pluginExternalRuntimeCore.test.ts, injectExternalRuntimeCorePlugin.test.ts, and integration/external-runtime.test.ts.

dmchoi77: [`buildExternalRuntimeCoreShimCode`](https://github.com/module-federation/vite/blob/3e166780f8752c55c5a12561433a8500c2f3f375/src/plugins/pluginExternalRuntimeCore.ts#L49)

The deferred global read looks good, but wrapping every named export as a function Proxy may alter runtime-core semantics. Some exports exposed through `@module-federation/runtime/core` are objects (`Global`, `CurrentGlobal`, `helpers`, `types`), not functions.

For example, with the real module:

```ts
import { Global } from '@module-federation/runtime/core';

typeof Global; // "object"
```

With the generated shim for `@module-federation/runtime-core`, `Global` is produced by `__mfCreateLazyRuntimeCoreExport("Global")`, so it becomes a function Proxy:

```ts
import { Global } from '@module-federation/runtime-core';

typeof Global; // "function"
```

That can affect type checks, identity checks, or object introspection. Would it be possible to preserve callable vs non-callable export shapes separately?

gioboa: is this last comment solved?

gioboa: cc @tonoizer

gioboa (approved):
Thanks @tonoizer this looks great 🚀

dmchoi77 (approved):

gioboa (commented):

Image

PR #944, opened 2026-07-18
fix: harden SSR fetch abort fallback and memoize manifest graphs

tonoizer:
## Summary
- stop retrying aborted/timed-out localhost SSR fetches over IPv6 loopback so configured timeouts are not doubled
- fall back to the temp-file writer promise when a path reservation is missing on cache hit
- memoize per-chunk static-graph analysis during manifest asset collection so co-located exposes reuse one traversal

## Testing
- [x] `pnpm exec vitest run src/utils/__tests__/fetchWithTimeout.test.ts src/utils/__tests__/cssModuleHelpers.test.ts src/utils/__tests__/ssrEntryLoader.test.ts`
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`

## Risks
- localhost IPv6 fallback no longer retries abort/timeout failures (intentional)
- manifest chunk analysis is cached only within one `processModuleAssets` call; output must remain identical for shared starting chunks

## Related
- Pair with #945 (SSR response body size limit). **Merge this PR first** — #945 is already rebased onto this branch and depends on the abort/timeout non-retry guard plus the temp-file reservation fallback.

PR #945, opened 2026-07-18
fix: bound SSR remote response body size

tonoizer:
## Summary
- cap SSR remote response bodies at 10 MiB by default via `readResponseTextBounded`
- reject early when `Content-Length` exceeds the limit, and abort streamed reads once the budget is exceeded
- add `fetchMaxBytes` on the SSR entry loader (`0` disables the limit), and apply the same bound to temp-file loads, manifest JSON, and the VM strategy
- include `fetchMaxBytes` in process-level SSR cache keys (manifest, entry resolution, temp-file, VM) so a looser limit cannot satisfy a stricter later load
- rethrow `SsrFetchBodyTooLargeError` from manifest discovery so oversized manifests fail closed instead of falling through to URL conventions

## Testing
- [x] `pnpm exec vitest run src/utils/__tests__/fetchWithTimeout.test.ts src/utils/__tests__/ssrEntryLoader.test.ts src/utils/__tests__/ssrVmStrategy.test.ts`
- [x] `pnpm typecheck`
- [x] `pnpm fmt.check`

## Risks
- remotes larger than 10 MiB fail closed unless hosts raise `fetchMaxBytes` or set it to `0`
- oversized failures now surface as `SsrFetchBodyTooLargeError` instead of falling through to a null load (modules **and** manifests)

## Related
- **Depends on #944** — this branch is rebased onto #944 so `fetchWithTimeout` keeps the abort/timeout localhost non-retry guard and the temp-file reservation fallback.
- Prefer merging **#944 first**, then this PR (GitHub will drop the shared commits from the diff once #944 is on `main`). Until then, this PR’s commit list intentionally includes #944.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,28 @@ You can specify the place the host initialization file is injected with the **ho
 The **moduleParseTimeout** option allows you to configure the maximum time to wait for module parsing during the build process.
 The **moduleParseIdleTimeout** option is an alternative that resets the timer on every parsed module. It only fires when there has been no module activity for the configured duration, making it suitable for large codebases where the total build time exceeds the fixed timeout.
 
+## Runtime capability optimization
+
+Runtime features that a build never uses can be removed at build time:
+
+```ts
+federation({
+  name: "remote",
+  exposes: {
+    "./remote-app": "./src/App.vue",
+  },
+  disableRemote: true,
+  disableShared: true,
+  disableSnapshot: true,
+});
+```
+
+- `disableRemote` removes remote-consumption support. Do not enable it when `remotes` are configured.
+- `disableShared` removes shared-dependency support. Do not enable it when `shared` dependencies are configured.
+- `disableSnapshot` removes snapshot support, including manifest-based remotes, preload, dynamic type hints, HMR, and devtools integration.
+
+All three options default to `false`, except `disableSnapshot`, which defaults to `true` for Node/SSR builds. An explicitly configured Vite `define` value takes precedence over the corresponding option.
+
 ## Load the Remote App
 
 In your host app, you can now import and use the remote app with **defineAsyncComponent**
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -280,7 +280,7 @@ describe(`singleton React dev fallback (Vite ${viteVersion})`, () => {
     const requestFailures = new Set<string>();
     const errorResponses = new Set<string>();
     const requests = new Set<string>();
-    page.on('pageerror', (error) => pageErrors.add(error.message));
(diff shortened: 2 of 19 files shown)


gioboa:
Apply the configured response-size limit to ModuleRunner requests, preventing oversized SSR responses from bypassing the protection used by other SSR loading paths.

gioboa (changes requested):
[Astro example](https://github.com/gioboa/module-federation-vite-examples/actions/runs/29651674858/job/88099340421#step:10:210) is failing. Before merging we need to check it.

tonoizer:
@gioboa thanks for the Astro catch — fixed in 3935445.

The Astro RUNTIME-004 came from `#926` scoping `loadRemote` request ids; `@module-federation/astro` still resolves `localRemotes` by the configured alias. We now keep `loadRemote`/alias on the configured key and only scope the unique remote `name`.

Locally verified: Astro host+remote `pnpm build` succeeds with this commit. ModuleRunner body-size limiting was already in dce9539.

tonoizer (commented):
Addressed the Astro failure in 3935445.

**Root cause:** `#926` scoped both the remote `name` and the `loadRemote` request id / `alias`. `@module-federation/astro` looks up `ssr.localRemotes` (and builds its SSR `createInstance` remotes) by the configured key (`astro_remote`). After scoping, the request became `__mfe_internal__…__mf_owner__N__astro_remote/…`, so localRemotes missed and createInstance registered plain remotes → RUNTIME-004.

**Fix:** keep `loadRemote` ids and remote `alias` as the configured key; scope only the unique remote `name` for instance isolation.

**Verified locally:** unit tests + Astro host/remote `pnpm build` (host + remote cards prerender into `index.html`).

**ModuleRunner size limit:** already covered in dce9539.

gioboa:
There was an error in the Astro example, I fixed it. you can revert your changes.

tonoizer:
@gioboa thanks — reverted the Astro `loadRemote` alias workaround in d0d780c.

That cherry-pick was what broke Integration (vite@5–8) (`host-build` / `remote-dependency-pending` expected scoped request ids). With the Astro example fixed on your side, we no longer need the plugin-side change.

tonoizer:
CI workflows for the revert commit are queued but awaiting maintainer approval (`action_required` on the fork PR). Locally after the revert:

- `integration/host-build.test.ts` + `remote-dependency-pending.test.ts` pass
- full `pnpm test:integration` — 12 files / 45 tests pass

gioboa (commented):

Image

gioboa (approved):
Thanks for your help @tonoizer 🚀

tonoizer:
Happy to help

PR #946, opened 2026-07-18
fix: skip unavailable React compiler runtime

dmchoi77:
# fix: skip unavailable React compiler runtime

## Summary

Prevent `react/compiler-runtime` from being registered as a shared dependency when it is not exported by the project's installed React version.

## Problem

When `react` is configured as a shared dependency, the plugin pre-registers known React subpaths.

React 18 does not export `react/compiler-runtime`. However, the plugin still registered it in the federation `usedShares` collection and generated its virtual modules, even after adding it to `optimizeDeps.exclude`.

In a pnpm workspace containing both React 18 and React 19, the unresolved import could resolve to a React 19 installation from another workspace package. This caused React 19's CommonJS `compiler-runtime.js` to be served directly to the browser, resulting in:

```text
ReferenceError: module is not defined
```

## Changes

- Check whether `react/compiler-runtime` can be resolved from the project root.
- When unavailable:
  - Keep it in `optimizeDeps.exclude`.
  - Do not register it in federation `usedShares`.
  - Do not generate its `loadShare` or prebuild virtual modules.
- Preserve the existing behavior for React versions that export `react/compiler-runtime`.
- Include the browser error stack in integration test diagnostics.

## Tests

- Added a unit test verifying that `react/compiler-runtime` is not registered for a React 18 project.
- Updated the singleton React integration test to verify that `react/compiler-runtime.js` is not incorrectly requested.
- Preserved the existing assertion that the host and remote use the same React singleton instance.

Validation results:

- `pnpm test`: 867 passed, 1 skipped
- `pnpm test:integration`: 45 passed
- `pnpm typecheck`: passed
- `pnpm build`: passed
- `pnpm fmt.check`: passed

## Impact

This change only affects projects whose installed React version does not export `react/compiler-runtime`. Registration behavior for other common shared subpaths remains unchanged.

### DIFF
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -280,7 +280,7 @@ describe(`singleton React dev fallback (Vite ${viteVersion})`, () => {
     const requestFailures = new Set<string>();
     const errorResponses = new Set<string>();
     const requests = new Set<string>();
-    page.on('pageerror', (error) => pageErrors.add(error.message));
+    page.on('pageerror', (error) => pageErrors.add(error.stack || error.message));
     page.on('console', (message) => {
       if (message.type() === 'error') consoleErrors.add(message.text());
     });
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Image

gioboa (approved):
it's great @dmchoi77 Thanks

PR #948, opened 2026-07-19
fix: exclude unoptimizable raw JSX shares

gioboa:
Speeds up dev-server cold start for shared deps that resolve to raw .jsx/.tsx workspace source, common in monorepos where a shared package has no prebuilt dist/. Vite's optimizer never bundles .jsx/.tsx, so forcing them into optimizeDeps.include left them stuck unresolved (triggering repeated Cannot optimize dependency warnings and, in some cases, indefinite hangs) instead of routing them to exclude like other unoptimizable shares already are.


gioboa (commented):

Image

PR #950, opened 2026-07-20
fix: fix React HMR with base paths

### LINKED ISSUE #949: [Bug]: HMR not working when remote has context path
### Describe the bug

I cloned the examples repo. At the react part I modify two things to reflect my real app code:

1. change remote components export with function name e.g. `export default () => {` to 
`export default function App() {`
2. change the context path of remote. add `base: "/aaa"` at remote config and point to `http://localhost:4174/aaa/remoteEntry.js` at host config

Run `pnpm dev`. Open browser to `http://localhost:4173`.  Live change text at `Counter.tsx` won't get update.
I looked into the socket message it did receive 'js-update' message, but maybe the path is wrong?

```
{"type":"update","updates":[{"type":"js-update","timestamp":1784541804105,"path":"/src/components/Counter.tsx","acceptedPath":"/src/components/Counter.tsx","explicitImportRequired":false,"isWithinCircularImport":false}]}
```

If I change all the component to use `export default () => {`, HMR will do a full page reload.

### Version

8.1.0

### Reproduction

https://github.com/jackysee/module-federation-vite-examples-hmr-ctx-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #949

Fix React remote HMR with non-root Vite base paths by matching refresh endpoints by suffix and resolving the local refresh runtime relative to the remote URL.

### DIFF
--- a/src/plugins/hmr/__tests__/react.test.ts
+++ b/src/plugins/hmr/__tests__/react.test.ts
@@ -11,10 +11,11 @@ type Middleware = (
   next: () => void
 ) => void;
 
-function createCtx(): { ctx: AdapterContext; middlewares: Middleware[] } {
+function createCtx(base = '/'): { ctx: AdapterContext; middlewares: Middleware[] } {
   const middlewares: Middleware[] = [];
   const ctx: AdapterContext = {
     server: {
+      config: { base, root: process.cwd() },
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #953, opened 2026-07-21
fix: fix up import-false shared provider selection

### LINKED ISSUE #952: [Bug]: import: false shared registers a throwing provider that wins version selection over a runtime-only host
### Describe the bug

Per the docs, `shared: { pkg: { import: false } }` means *"this shared is not packaged into the remote, and only the shared provided by the consumer/host will be used."*
  
  In practice, the Vite plugin registers the remote's `import: false` shared as a **competing provider** in the share scope, keyed under the remote's own build-time version, whose `get()` only throws `must be provided by host`.
  During share resolution this throwing provider **wins version selection** against a runtime-only host that registered the same package without an explicit version (default sentinel `"0"`), so the remote resolves to its own
  throwing stub instead of the host instance.
  
  Runtime error:
  
  ```
  [Module Federation] Failed to bridge external shared module "@my-company/components"
  Error: [Module Federation] Shared module '@my-company/components' must be provided by host
      at Object.get (virtual:mf-localSharedImportMap:...)
      at SharedHandler.loadShare (...)
      at __mfLoadRuntimeShare (virtual:mf-REMOTE_ENTRY_ID:...)
      at __mfBridgeExternalSharedProvider (...)
  ```
  
  ## Setup
  
  - **Host**: runtime-only host (Storybook) consuming the remote via `@module-federation/runtime` `createInstance(...)` — it does NOT use the Vite federation plugin. It provides the shared singletons eagerly:
  
  ```ts
  createInstance({
    name: 'host',
    remotes: [...],
    shared: {
      '@my-company/components': {
        lib: () => CompanyComponents,          // eagerly provided, loaded: true
        shareConfig: { singleton: true, requiredVersion: false },
        // NOTE: no `version` -> runtime registers it under version "0"
      },
    },
  })
  ```
  
  - **Remote**: built with the Vite plugin:
  
  ```ts
  federation({
    name: 'remote',
    exposes: { './contact-form': './src/form/contact-form.tsx' },
    shared: {
      '@sabdx/components': { singleton: true, import: false },
    },
  })
  ```
  
  `@my-component/components` establishes a React context (`FormProvider`); the exposed component calls `useForm()`.
  
  ## Root cause (traced)
  
  1. `generateLocalSharedImportMap` (lib/index.js) emits, for an `import: false` share, a provider registered in the scope under the remote's build version:
  
  ```js
  {
    name: "@my-company/components",
    version: shareItem.version,   // remote build version, e.g. "0.0.0-mr-..."
    loaded: false,
    from: "<remote>",
    async get () {
      if (true /* import === false */) {
        throw new Error("[Module Federation] Shared module '...' must be provided by host");
      }
      ...
    },
    shareConfig: { singleton: true, requiredVersion: "*", import: false, ... },
  }
  ```
  
  2. The host's provider is registered under version `"0"` (runtime default when no `version` is given), `loaded: true`.
  
  3. Share resolution (`getRegisteredShare` -> `findSingletonVersionOrderByLoaded` / `findSingletonVersionOrderByVersion` in `@module-federation/runtime-core/dist/utils/share.js`) selects by version once both providers are
  `loaded`/`loading`. During the bridge the remote stub is `loading`, so version comparison decides: the remote's real version wins over the host's `"0"`.
  
  4. The selected provider is the remote's `import: false` stub, whose `get()` throws `must be provided by host` — even though the host DID provide the module.
  
  So the `import: false` stub competes in (and wins) version selection instead of being a pure no-op consumer request, contradicting the documented behavior.
  
  ## Expected behavior
  
  With `import: false`, the remote must not register a provider that can win selection and throw. It should resolve exclusively to a host-provided provider for that package, regardless of the version the host registered it under
  (docs: "only the shared provided by the consumer will be used"). At minimum, a host-provided (loaded) provider must be preferred over the remote's own throwing `import: false` stub.
  
  ## Actual behavior
  
  The remote resolves to its own `import: false` stub and throws `must be provided by host`, even though the host provides the module (just under version `"0"`).
  
  ## Workaround
  
 Have the host register the shared with an explicit `version` >= the remote's version (e.g. `version: React.version`, or the package `package.json` version), so version selection resolves to the host provider. Required today;
  without it the host stays on sentinel `"0"` and always loses selection.
  
  ## Environment
  
  - `@module-federation/vite`: 1.18.2
  - `@module-federation/runtime`: 2.8.0
  - `@module-federation/runtime-core`: 2.8.0
  - Vite: 8.x (Rolldown)
  - Mode: dev (serve)
  
  ## Notes
  
  - A runtime-only host (e.g. Storybook) participating via `@module-federation/runtime` `createInstance` cannot easily declare a package version at browser runtime (unlike `React.version`, most packages do not expose their version),
  which makes the workaround awkward.
  - `import: false` is expected to be the clean way to express "always use the host's instance", but is currently not honored in the Vite dev implementation for a runtime-only host.

### Version

1.18.2

### Reproduction

-

### Relevant log output

```shell

```

### Validations

gioboa:
Close #952

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -3575,6 +3575,31 @@ describe('virtualRemoteEntry', () => {
     expect(hostGet).not.toHaveBeenCalled();
   });
 
+  it('selects a runtime-only host provider for import:false regardless of its sentinel version', async () => {
+    const selectExternalProvider = await getExternalSharedProviderSelector();
+    const hostProvider = {
+      from: 'runtime-host',
+      lib: () => ({ marker: 'host' }),
+      loaded: true,
+    };
+    const localStub = {
+      version: '1.2.3',
+      from: 'remote',
+      scope: ['default'],
+      loaded: false,
+      shareConfig: {
+        singleton: true,
+        import: false,
+        requiredVersion: '*',
+        strictVersion: false,
+      },
+    };
+
+    expect(
+      selectExternalProvider({ '0': hostProvider }, 'host-only-package', localStub, 'version-first')
+    ).toBe(hostProvider);
+  });
+
   it.each([false, true])(
     'keeps a loaded host singleton ahead of a newer local fallback with version-first (local loaded: %s)',
     async (localLoaded) => {
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -488,7 +488,7 @@ export const externalSharedProviderSelectionHelperCode = `const __mfSelectExtern
             const candidates = Object.fromEntries(
               Object.entries(versions || {}).filter(([, provider]) => !isLocalProvider(provider))
             );
-            if (localShare?.version) {
+            if (localShare?.version && localShare.shareConfig?.import !== false) {
               const sameVersionProvider = candidates[localShare.version];
               // Runtime registration keeps an existing same-version record,
               // even when it has only a getter and is not loaded yet. Model


gioboa (commented):

Image

PR #955, opened 2026-07-21
fix: preserve shared packages’ internal imports

### LINKED ISSUE #954: [Bug]: Shared singleton with self-referencing subpath exports (zustand) yields empty optimized dep chunk in dev — regression in 1.16.9, still broken in 1.19.0
### Describe the bug

Sharing **zustand** as a singleton breaks every dev-mode load from **1.16.9 through 1.19.0 (latest)**: `TypeError: create is not a function` on the host and on the remote standalone, cold or warm cache — a manual reload does not recover (this is not the optimizer-wave/cache-timing class from #923).

**Root cause, traced in dev:** during dependency optimization the plugin rewrites zustand's *internal* subpath self-re-exports (its `index.mjs` re-exports `./react.mjs` / `./vanilla.mjs`, i.e. `zustand/react` and `zustand/vanilla`) into `mf-shared:` virtual loadShare modules. The optimized `deps/zustand.js` chunk then contains **no export statement at all** — importers receive an empty namespace, so `import { create } from 'zustand'` is `undefined`. The share cache later holds a healthy module, but the virtual module's live bindings never update.

This should generalize to **any shared package whose root entry re-exports its own subpaths**, not just zustand.

### Reproduction

https://github.com/razchiriac/mf-vite-923-repro — minimal two-app pnpm workspace (Vite 7.3.1 host + React 19.2.4 remote loaded via `React.lazy`, `shared` all `{ singleton: true, requiredVersion: false }`: react, react-dom, react/jsx-runtime, react/jsx-dev-runtime, react-router, @tanstack/react-query, zustand, react-hook-form, zod, luxon, react-markdown). We run this shape in production (Vite host shell + React remotes over pnpm).

Steps: `pnpm install` → start remote, start host (cold: remove `node_modules/.vite` first) → open the host. The README has per-version server logs and browser console captured with headless Chrome, and the broken optimized chunk verbatim (`evidence/1.19.0-broken-optimized-zustand-chunk.js` is inlined there).

### Bisect

- **1.16.8: good** (zustand shared, everything works)
- **1.16.9: bad** — first broken release; its changelog is mostly workspace-singleton share fixes (#819, #825, #828), which are the suspects
- 1.16.11, 1.17.1, 1.19.0: still broken
- 1.19.0 with only `zustand` removed from `shared`: clean

### Related

Same dev-mode shared-singleton family as #913, #858, #895 (all closed); found while building the reproduction @gioboa asked for on #923 (comment with the full version matrix posted there).

Since this class keeps regressing and is dev-mode-only, I'd be glad to also contribute a dev-mode Playwright e2e suite (cold-start two-server scenario asserting no reload waves, no 504s, and working named exports for shared singletons) — the capture harness behind the README tables is already close to that shape. Happy to help land the fix itself too.

### System info

Node v22.16.0, pnpm 10.28.1, Vite 7.3.1, React 19.2.4, macOS arm64, Chrome stable. Exact versions pinned via the repro's committed `pnpm-lock.yaml` (plugin at 1.19.0).

gioboa:
Close #954

Prevent shared packages’ internal subpath imports from being incorrectly proxied during Vite 7 dependency optimisation.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1091,6 +1091,39 @@ describe('vite:module-federation-early-init', () => {
     ).toBeUndefined();
   });
 
+  it('does not proxy a shared package self-referencing its own subpath', () => {
+    const plugin = (
+      federation({
+        name: 'host',
+        shared: { zustand: { singleton: true } },
+      }) as Plugin[]
+    ).find((entry) => entry.name === 'vite:module-federation-early-init');
+    if (!plugin) throw new Error('vite:module-federation-early-init plugin not found');
+
+    const config: any = { root: process.cwd(), optimizeDeps: { include: [] } };
+    runConfig(plugin, { meta: {} } as ConfigPluginContext, config, {
+      command: 'serve',
+      mode: 'test',
+    });
+
+    const optimizeSharedProxy = config.optimizeDeps.esbuildOptions.plugins.find(
+      (entry: any) => entry.name === 'module-federation:optimize-shared-proxy'
+    );
+    const onResolveHandlers: any[] = [];
+    optimizeSharedProxy.setup({
+      onResolve: (_options: unknown, handler: unknown) => onResolveHandlers.push(handler),
+      onLoad: () => undefined,
+    });
+
+    expect(
+      onResolveHandlers[1]({
+        path: 'zustand/react',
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #957, opened 2026-07-22
fix: preserve legacy shared subpath exports

Nsttt:
## Summary

Preserve exact package subpaths during ESM-oriented export inspection for legacy packages without an `exports` map. This prevents React 17 `react/jsx-runtime` from being inspected as the root `react/index.js` module and generating incorrect load-share named exports.

## Changes

- Resolve legacy package subpaths through the project resolver when no `exports` map exists.
- Add package-resolution coverage for packages with a root `main` and separate runtime subpath.
- Add React 17 JSX runtime wrapper coverage for `Fragment`, `jsx`, and `jsxs`.

## Testing

- `pnpm test` (898 passed, 1 skipped)
- `pnpm typecheck`
- `pnpm build`
- `pnpm fmt.check`
- `git diff --check`


gioboa (commented):

Image

gioboa (approved):
Thanks @Nsttt great fix

PR #960, opened 2026-07-22
fix: fix up context paths and HMR

### LINKED ISSUE #959: [Bug]: Module Federation failed when host is at context path
### Describe the bug

I cloned the example repo, and set both host and remote with context path. Like host is `http://localhost:4173/bbb` and remote is `http://localhost:4174/aaa`.  Go to browser, it would fail with `@react-refresh:3  GET http://localhost:4173/@react-refresh 404 (Not Found)`. 

### Version

8.1.0

### Reproduction

https://github.com/jackysee/module-federation-vite-examples-hmr-ctx-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #959

Fix Vite context-path support for Module Federation HMR and bootstrap by preserving configured base paths, correctly resolving host remote entries, and preventing HTML proxy IDs from being misclassified as host auto-init modules.


gioboa (commented):

Image

PR #963, opened 2026-07-23
fix: fix cold root remote imports

### LINKED ISSUE #961: [Bug]: Module exposed on root cannot be statically imported from host
### Describe the bug

A remote module exposed on the root entry (`.`) cannot be statically imported from the host.

`packages/remote/vite.config.ts` exposes:

```ts
exposes: {
    '.': './src/index.ts',
},
```

`packages/host/src/main.ts` consumes it with a normal static import:

```ts
import { helloWorld } from 'remote';

helloWorld();
```

**To reproduce:**

**Expected:** the page loads and `helloWorld()` runs, logging to the console.

**Actual:** the host fails to load. The browser console shows:

```
(index):1 Uncaught (in promise) Promise {<pending>}
```

**Workarounds that avoid the bug:**

- Using a dynamic import instead of a static one works:
  ```ts
  const { helloWorld } = await import('remote');
  helloWorld();
  ```
- Exposing on a subpath instead of the root also works:
  ```ts
  exposes: {
      './helloWorld': './src/index.ts',
  },
  ```
  ```ts
  import { helloWorld } from 'remote/helloWorld';
  ```

So the failure is specific to a *static* import of a *root* (`.`) expose.

### Version

1.19.1

### Reproduction

https://github.com/levrik/vite-module-federation-repro/tree/static-root-expose

### Validations

gioboa:
Close #961

Preloads root and subpath remote imports before host entry execution, preventing cold-start Promise errors.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -750,7 +750,7 @@ describe('pluginAddEntry', () => {
     expect(result).toContain('entry=%2F_nuxt%2Fentry.async.js');
   });
 
-  it('preloads scoped remote subpaths but skips the bare scoped remote key', async () => {
+  it('preloads scoped remote subpaths and the bare scoped remote key', async () => {
     const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mf-add-entry-scoped-'));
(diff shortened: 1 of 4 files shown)


gioboa (commented):

Image

PR #964, opened 2026-07-23
fix: fix React DOM optimizer race

### LINKED ISSUE #962: [Bug]: Cold-start optimizer race serves incompatible react-dom and react-dom/client generations
## Description

  During a cold Vite dev-server start, `@module-federation/vite@1.19.1` can cause `react-dom` and `react-dom/client` to be served from incompatible dependency-optimizer generations.

  The browser receives a `react-dom.js` module that does not export Vite's generated private binding `t`, followed by a `react-dom_client.js` module that imports that binding.

  This causes the application to fail during startup with:

  ```text
  The requested module '/node_modules/.vite/deps/react-dom.js?v=<hash>'
  does not provide an export named 't'
  ```

  ## Minimal reproduction

  https://github.com/francesc79/module-federation-vite-react-dom-repro

  ## Steps to reproduce

  ```bash
  git clone https://github.com/francesc79/module-federation-vite-react-dom-repro.git
  cd module-federation-vite-react-dom-repro
  corepack enable
  pnpm install
  pnpm repro
  ```

  The command:

  1. clears only Vite's dependency cache;
  2. starts Chrome before the Vite server binds its port;
  3. captures the first optimized React DOM modules served to the browser;
  4. checks whether their generated imports and exports are compatible.

  When the problem is reproduced, the command intentionally exits with code `1` and prints:

  ```text
  react-dom.js exports 't': false
  react-dom_client.js imports 't': true
  REPRODUCED: mixed React DOM optimizer generations were served.
  ```

  ## Actual behavior

  The optimized `react-dom/client` entry imports a generated private binding from `react-dom`, but the `react-dom` module previously served to the browser does not export that binding.

  The browser module graph is therefore internally inconsistent and the application cannot start.

  ## Expected behavior

  `@module-federation/vite` should not allow `react-dom` and its subpath entries to be served from incompatible optimizer generations.

  A `react-dom/client` optimized entry must never import generated bindings that are absent from the `react-dom` module already served to the same browser module graph.

  ## Negative control

  Run:

  ```bash
  pnpm repro:control
  ```

  The control uses:

  - the same application source;
  - the same import order and delay;
  - the same Vite version;
  - the same cold dependency cache;
  - the same browser startup and polling sequence.

  The only relevant difference is that the Module Federation plugin is disabled.

  The control renders successfully and exits with code `0`:

  ```text
  Mode: control without Module Federation
  Page errors: none
  react-dom.js exports 't': true
  react-dom_client.js imports 't': true
  PASS: application rendered and React DOM optimizer modules are compatible.
  ```

  ## Versions

  ```text
  Node.js 24.15.0
  pnpm 11.1.1
  @module-federation/vite 1.19.1
  @module-federation/enhanced 2.8.0
  Vite 8.0.16
  React 19.2.7
  React DOM 19.2.7
  ```

  ## Notes about the deterministic trigger

  In the original application, the late discovery of `react-dom/client` occurred naturally while Vite was still scanning a larger Module Federation expose graph.

  The minimal reproduction contains a small test-only Vite config probe that removes only `react-dom/*` subpaths from the resolved initial `optimizeDeps.include` list, while leaving the
  root `react-dom` entry untouched.

  It does not exclude, alias, mock, or replace any React DOM module. `react-dom/client` is still discovered and optimized normally by Vite at runtime.

  This probe only makes the optimizer-wave ordering deterministic. It is test instrumentation and is not a proposed application workaround.

### Version

8.0.16

### Reproduction

 https://github.com/francesc79/module-federation-vite-react-dom-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #962 

Ensures react-dom and subpaths like react-dom/client are optimized together, preventing incompatible dependency generations during cold starts.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -551,6 +551,8 @@ export default __mfShared.default ?? __mfShared;`,
               addUsedShares(subpath, options);
               if (canResolveSubpath) {
                 optimizeDeps.include.push(subpath);
+                // Prevent subpaths like react-dom/client from using a later, incompatible optimizer generation.
+                if (key === 'react-dom') optimizeDeps.include.push(`${key} > ${subpath}`);
               } else {
                 optimizeDeps.exclude.push(subpath);
               }


gioboa (commented):
@francesc79 Can you double check this fix?

francesc79:
Sure! I'll double-check it and let you know.

francesc79:
I tested the published PR artifact from commit `0b5a9bf`:

  ```bash
  pnpm add -D https://pkg.pr.new/@module-federation/vite@964
  ```

  ## Minimal reproduction

  The result is positive on the public minimal reproduction:

  - 6/6 cold-cache runs passed;
  - no page errors;
  - `react-dom.js` exported the required `t` binding;
  - `pnpm repro:control` passed;
  - typecheck and production build passed.

  The resolved configuration contained the new nested entries:

  ```text
  react-dom
  react-dom > react-dom/client
  react-dom > react-dom/server
  react-dom > react-dom/server.browser
  ```

  ## Original project

  Unfortunately, the same PR artifact still reproduces the original failure in the larger application graph.

  Environment:

  ```text
  Linux
  Node.js 24.15.0
  pnpm 11.1.1
  Vite 8.0.16
  Google Chrome 150.0.7871.128
  ```

  I ran two cold starts with Chrome polling before Vite bound the port. Both failed with the original error:

  ```text
  The requested module '/node_modules/.vite/deps/react-dom.js?v=<hash>'
  does not provide an export named 't'
  ```

  Example result:

  ```text
  react-dom.js exports 't': false

  new dependencies found: react-dom
  new dependencies found: react-dom/client
  ```

  The resolved configuration already contained both the direct entries and the nested selectors introduced by the PR:

  ```text
  react-dom
  react-dom/client
  react-dom > react-dom/client
  react-dom/server
  react-dom > react-dom/server
  react-dom/server.browser
  react-dom > react-dom/server.browser
  ```

  React and React DOM were not present in `optimizeDeps.exclude`.

  Therefore, the PR fixes the reduced test case, but it does not yet prevent the optimizer race in the original larger graph. Vite still publishes `react-dom` and discovers `react-dom/
  client` in separate optimizer waves.

  I do not think the PR should be considered fully verified yet. The regression test needs to preserve the early browser navigation and reproduce the actual late sequence:

  ```text
  react-dom
  react-dom/client
  ```

gioboa:
I see. I need the larger failing project or a reproduction preserving its exact optimizer sequence.
Current minimal repro no longer exercises the remaining race, so fixing blindly risks another false positive.

gioboa (commented):

Image

gioboa:
> The result is positive on the public minimal reproduction

I am merging this because it resolves the original issue.

PR #965, opened 2026-07-23
fix: fix React DOM optimizer race

gioboa:
Prevents React DOM optimizer races by routing internal shared imports through stable Federation proxies across esbuild and Rolldown dependency generations.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -919,6 +919,33 @@ describe('vite:module-federation-early-init', () => {
     );
   });
 
+  it('decouples Rolldown react-dom/client from the react-dom optimizer entry', () => {
+    const plugin = (
+      federation({ name: 'host', shared: { 'react-dom': { singleton: true } } }) as Plugin[]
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

gioboa:
@francesc79 This version should solve all the edge cases. Can you try this please?

francesc79:
@gioboa Thank you very much for the fix and for taking the time to investigate this! 

  I tested the artifact from PR #965 both on the minimal reproduction and on the original, much larger application.

  I also verified the real cold-start scenario by deleting Vite’s dependency cache and restarting the development server. During startup, Vite regenerated the optimizer dependencies, but
  React DOM remained coherent:

  - `react-dom`, `react-dom/client`, and the new Federation proxy ended up in the same optimizer generation;
  - `react-dom/client` correctly imports the stable `module-federation_optimized-require-react-dom` proxy;
  - the private `t` binding mismatch no longer occurs;
  - the application renders successfully, including after subsequent reloads.

  So PR #965 appears to solve the issue and the edge case from the original application as well.

  Thanks again!

PR #967, opened 2026-07-24
fix: preserve base path in development

### LINKED ISSUE #966: [Bug]: Base path missing from imports in `virtual:mf-html-entry-proxy` behind prefix-stripping proxy setup
#### Background

The Vite dev server is configured with a base path (e.g. `base: '/subpath'`) and the Module Federation Vite plugin enabled as a host (no remotes are needed to reproduce this, an otherwise-empty host config is enough).

The dev server also has a proxy rule of its own: any incoming request that doesn't already start with the base path gets rewritten to prepend it, then re-requested against itself. This lets the dev server respond correctly whether it's hit at its root or at the base path directly. Since you cannot configure the root serve path separately from the `base` path used in JS files, that's the only option I found.

In front of that dev server sits a separate reverse proxy, reachable externally under the subpath, but stripping that prefix before forwarding requests on to the dev server's origin.

I don't know if that's a common setup but I didn't find any other way to serve a Vite dev server behind a prefix-stripping proxy. It basically matches the production setup which handles this just fine. Also raw Vite has no issues with such a setup so I guess something isn't working correctly in this plugin.

#### The bug

When a request reaches Vite without the `/subpath` prefix already present in its path — which happens whenever it's routed through the standalone reverse proxy (which strips the prefix) — the `virtual:mf-html-entry-proxy` module that `@module-federation/vite` generates (the module federation host-init bootstrap) omits the `/subpath` base prefix from its own import specifiers. Fetching that same module directly against Vite (where the request path still carries `/subpath`) produces the correct, prefixed imports.

**To reproduce:**

Fetch the entry-proxy virtual module through a prefix-stripping proxy:
   ```bash
   curl "http://localhost:5175/subpath/@id/__x00__virtual:mf-html-entry-proxy?init=%2F%40id%2Fvirtual%3Amf%3A__mfe_internal__host__mf_owner__1__H_A_I__hostAutoInit__H_A_I__.js&entry=%2Fsrc%2Fmain.tsx"
   ```

**Expected** (this is what you get fetching the same URL directly against Vite on port 5174 instead of through the proxy): imports keep the `/subpath` prefix.

```js
(async () => {
  const __mfHostInit = await import("/subpath/@id/virtual:mf:__mfe_internal__host__mf_owner__1__H_A_I__hostAutoInit__H_A_I__.js");
  await __mfHostInit.__tla;
  const { initHost } = __mfHostInit;
  await initHost();
  if (__mfModuleCache.pendingShareLoads) {
    await Promise.all(__mfModuleCache.pendingShareLoads);
  }
})().then(() => import("/subpath/src/main.tsx"));
```

**Actual:** through the standalone proxy, the `/subpath` prefix is missing entirely.

```js
(async () => {
  const __mfHostInit = await import("/@id/virtual:mf:__mfe_internal__host__mf_owner__1__H_A_I__hostAutoInit__H_A_I__.js");
  await __mfHostInit.__tla;
  const { initHost } = __mfHostInit;
  await initHost();
  if (__mfModuleCache.pendingShareLoads) {
    await Promise.all(__mfModuleCache.pendingShareLoads);
  }
})().then(() => import("/src/main.tsx"));
```

Since the base prefix is missing, the browser ends up requesting these modules at the wrong (un-proxied) paths, and the app fails to load.

### Version

1.19.1

### Reproduction

https://github.com/levrik/vite-module-federation-repro/tree/proxy-prefix-bug

### Validations

gioboa:
Close #966

Preserve Vite’s configured base path in dev HTML proxy imports.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -207,6 +207,42 @@ describe('pluginAddEntry', () => {
     expect(next).toHaveBeenCalledOnce();
   });
 
+  it('serves html proxy imports with the configured base', () => {
+    const [servePlugin] = addEntry({
+      entryName: 'hostInit',
+      entryPath: 'virtual:mf-host-init',
+      inject: 'html',
+    });
+    const handlers: Function[] = [];
+
+    runConfig(
+      servePlugin,
+      {} as ConfigPluginContext,
+      {},
+      { command: 'serve', mode: 'development' }
+    );
+    runConfigResolved(servePlugin, {
+      root: '/repo',
+      base: '/subpath/',
+    } as unknown as ResolvedConfig);
+    runConfigureServer(servePlugin, {
+      middlewares: { use: (handler: Function) => handlers.push(handler) },
+    } as unknown as ViteDevServer);
+
+    const end = vi.fn();
+    handlers[0](
+      {
+        url: `${toViteEncodedId('virtual:mf-html-entry-proxy')}?init=%2F%40id%2Fvirtual%3Amf-host-init&entry=%2Fsrc%2Fmain.tsx`,
+      },
+      { setHeader: vi.fn(), end },
+      vi.fn()
+    );
+
+    const code = end.mock.calls[0][0] as string;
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #973, opened 2026-07-24
fix: heal stale owner loadShare ids

### LINKED ISSUE #970: [Bug]: mf_owner generation mismatch on cold start - dependency optimizer bundles twice, orphaning virtual loadShare modules
### Describe the bug

On a cold dev-server start (fresh `node_modules/.vite`), Vite's dependency optimizer runs its bundling step twice in the same process (visible in Vite's own log as two separate `[optimizer] bundling dependencies...` lines for one boot). This appears to correspond to `federation()` being invoked twice - once per optimizer pass - and each invocation gets a distinct "owner" generation for its shared virtual modules. Whichever generation's output was written to `node_modules/.vite/deps/*.js` becomes unreachable once the live dev server starts serving requests under the *other* generation, producing a permanent `Failed to resolve import "virtual:mf:...__mf_owner__N__loadShare__..."` error.

I want to be upfront that I'm not 100% certain this is a single, isolated root cause rather than one manifestation of the broader dev-mode optimizer/shared-module problem space already tracked in #708, #768, #923, #954, and #311 - the error shape in #708/#768 is very close to what I'm seeing, just without the `mf_owner__N` segment. I'm filing this because the specific mechanism below is something I could pin down precisely, and because it bisects to an exact version boundary, which might help narrow down whether it's the same root cause or a distinct one.

### What I was able to verify directly

**1. The `mf_owner` scheme was introduced between 1.18.0 and 1.19.0.**

I downloaded the published tarballs for `@module-federation/vite` from 1.12.0 through 1.19.1 and grepped `lib/index.js` for the string `mf_owner`:

| Version | `mf_owner` present |
|---|---|
| 1.12.0 – 1.18.0 | No |
| 1.19.0, 1.19.1 | Yes |

**2. The mechanism, from reading the built `lib/index.js` in 1.19.1:**

- `normalizeModuleFederationOptions(options)` builds a new normalized options object on every call and assigns it to a module-scoped `config` variable (`return config = normalized`).
- `getSharedVirtualModuleState(options)` caches per-`options` state (including the `ownerKey`, e.g. `${internalName}__mf_owner__${id}`) in a `WeakMap` keyed by object identity, and mints a new id via a monotonically-incrementing counter (`nextSharedVirtualModuleOwnerId++`) whenever that exact object reference hasn't been seen before.

Since a brand-new `options` object is created on every `federation()` call, two calls within the same process can never share an owner id - each gets its own generation, even if they're conceptually "the same" config.

**3. Reproduction with real logs.**

Setup: one host (`core`) + one remote (`project_manager`), bidirectional (each lists the other in `remotes`), several shared singletons (`react`, `react-dom`, `zustand`, `react-router-dom`, etc.), `@module-federation/vite@1.19.1`, `vite@8.1.5`, Windows, pnpm workspace. On a cold start with `node_modules/.vite` removed:

```
[vite] (client) [optimizer] scanning dependencies...
[vite] (client) [optimizer] bundling dependencies...
[vite] (client) [optimizer] bundling dependencies...
[vite] (client) dependency optimized: @azure/msal-browser
[vite] (client) optimized dependencies changed. reloading
```

```
[vite] (client) Pre-transform error: Failed to resolve import "virtual:mf:__mfe_internal__core__mf_owner__2__loadShare__zustand_mf_1_vanilla__loadShare__.js" from "node_modules/.vite/deps/zustand.js?v=1dc80c4e". Does the file exist?
  Plugin: vite:import-analysis
  File: .../core/node_modules/.vite/deps/zustand.js?v=1dc80c4e:1:15
  1  |  export * from "virtual:mf:__mfe_internal__core__mf_owner__2__loadShare__zustand_mf_1_vanilla__loadShare__.js";
  2  |  export * from "virtual:mf:__mfe_internal__core__mf_owner__2__loadShare__zustand_mf_1_react__loadShare__.js";
```

This reproduced identically and deterministically across 5 separate cold starts over roughly 1.5 hours of testing - always the double `bundling dependencies` line, always the same class of orphaned-generation error.

**4. Workaround that resolves it for us.**

Setting `optimizeDeps.noDiscovery: true` (with every shared dependency already listed in `optimizeDeps.include`, per Vite's own documented requirement for that option) eliminates the `[optimizer] scanning dependencies...` line entirely, leaves exactly one `bundling dependencies...` pass, and the orphaned-generation error is gone. I verified this in a reduced reproduction (just the host + one remote, cut down from a larger 7-app workspace) to rule out app-count/complexity as the variable - the double-bundling-pass and the fix both reproduce the same way in the minimal 2-app case.

I have not yet tested pinning back to `1.18.0` as an alternative to the `noDiscovery` workaround, so I can't say whether that also avoids it, or whether `1.18.0` has other issues this scheme was meant to fix.

### Version

@module-federation/vite: 1.19.1
vite: 8.1.5

### Reproduction

I don't have a public repro repository yet - the setup above is inside a private workspace. I can put together a minimal public reproduction (two Vite apps, host + remote, matching the shared-singleton shape above) if that would help narrow this down further; let me know and I'll follow up with one.

### Relevant log output

See the log excerpts in the "Reproduction with real logs" section above.

### Validations

- [x] Read the docs.
- [x] Checked for an existing issue reporting the same bug - closest matches are #708, #768, #923, #954, and #311, none of which mention the `mf_owner` generation-mismatch mechanism specifically, so filing this as a distinct report rather than assuming it's a duplicate.

gioboa:
Close #970 

When resolveId receives a loadShare id whose owner generation no longer exists, it now extracts the instance name and package from the id.
It verifies both belong to the current federation instance, and resolves to that instance's current loadShare module instead of failing.
Ids minted by a different federation instance are never touched, so instance isolation is preserved.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -79,6 +79,7 @@ import { addUsedShares } from './virtualModules/virtualRemoteEntry';
 import { addUsedRemote } from './virtualModules/virtualRemotes';
 import { getRuntimeInitStatusImportId } from './virtualModules/virtualRuntimeInitStatus';
 import {
+  findCurrentLoadShareForStaleOwnerId,
   getLoadShareModulePath,
   materializeCachedLoadShareModule,
   prependWorkspaceSingletonSsrImport,
@@ -779,7 +780,9 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
             writeLocalSharedImportMap: () => writeLocalSharedImportMap(options),
             federationOptions: options,
           });
-          virtualModule = VirtualModule.findById(id);
+          virtualModule =
+            VirtualModule.findById(id) ??
+            findCurrentLoadShareForStaleOwnerId(id, options.shared, findSharedKey, options);
         }
         if (!virtualModule) return;
         return virtualModule.getResolvedId();
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -139,6 +139,7 @@ vi.mock('../../utils/packageUtils', () => ({
 }));
 
 vi.mock('../../utils/VirtualModule', () => ({
+  MF_OWNER_INFIX: '__mf_owner__',
   default: class MockVirtualModule {
     static findModule(tag: string, value: string) {
       const [, name] = value.split(tag);
--- a/src/utils/VirtualModule.ts
+++ b/src/utils/VirtualModule.ts
@@ -32,6 +32,10 @@ export const VITE_ID_PREFIX = '/@id/';
 export const VITE_NULL_BYTE_PLACEHOLDER = '__x00__';
 export const VITE_ENCODED_NULL_BYTE_PREFIX = `${VITE_ID_PREFIX}${VITE_NULL_BYTE_PLACEHOLDER}`;
 
+// Separates a federation instance's internalName from its per-generation
+// owner id inside scoped virtual module names.
+export const MF_OWNER_INFIX = '__mf_owner__';
+
 function escapeRegExp(value: string): string {
   return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -289,6 +289,7 @@ function getRuntimeDeferredResolutionCode(code: string) {
 
 vi.mock('../../utils/VirtualModule', () => {
   return {
+    MF_OWNER_INFIX: '__mf_owner__',
     default: class MockVirtualModule {
       name: string;
 
--- a/src/virtualModules/__tests__/virtualRuntimeInitStatus.test.ts
+++ b/src/virtualModules/__tests__/virtualRuntimeInitStatus.test.ts
@@ -5,6 +5,7 @@ const { writeSyncSpy } = vi.hoisted(() => ({
 }));
 
 vi.mock('../../utils/VirtualModule', () => ({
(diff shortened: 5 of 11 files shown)


gioboa (commented):

Image

imtaotao:
Thanks for fixing stale owner loadShare IDs in #973.

  We are currently using @module-federation/vite@1.20.7 with Vite 8.2.1, so the fix from #973 is already included. However, we still
  occasionally encounter a similar issue after restarting the Remote dev server or during HMR:

  - Vite reports a pre-transform or filesystem read error for an ID such as \0virtual:mf:...__mf_owner__N__loadShare__....
  - A second hard refresh usually succeeds.
  - The Remote is loaded cross-origin by a runtime/webpack Host.
  - The Remote uses multiple Host-provided shared dependencies configured with import: false.

  From the implementation, it appears that #973 heals an old owner-generation ID by mapping it to the current loadShare virtual module when
  resolveId runs.

  We are wondering whether there may be another path where Vite reuses a stale node from its module graph and invokes load without running
  resolveId again. In that case, VirtualModule.findById(oldId) would not find the module, and the virtual ID could eventually fall through
  to filesystem loading.

  We currently recover from this with a dev-only workaround that:

  1. Detects a stale loadShare virtual ID.
  2. Invalidates the stale module-graph node.
  3. Resolves and loads the current loadShare virtual module through Vite’s plugin container.
  4. Returns the current module source from the load hook.

  We do not yet have a stable minimal reproduction, and we have not added diagnostics to confirm that the load fallback is definitely the
  path recovering the failure.

  Could Vite bypass resolveId and invoke load directly for a reused module-graph entry in this situation?

  Would you prefer that we open a separate issue with additional diagnostics, or would a regression test and PR covering the direct
  load(staleId) path be useful?

gioboa (commented):
Hi @imtaotao let me check and drop a PR.

PR #975, opened 2026-07-25
fix: optimize ESM-only shared singletons in dev (avoid raw CommonJS)

### LINKED ISSUE #974: [Bug]: ESM-only shared singletons served as raw CommonJS in dev — canResolveSharedSubpath rejects exports-map packages (follow-up to #913)
### Describe the bug

Follow-up to #913 / #914. A shared **singleton that is an ESM-only package** (an `exports` map with no CommonJS `require` condition) is excluded from Vite's dev dependency optimization and served to the browser as **raw CommonJS**. Its transitive CJS dependencies then fail to evaluate as ESM, the shared scope never populates, `initHost()` rejects, and the app blanks in dev. The production build works.

#914 fixed the `react` case by dropping the singleton bypass. In the current code (1.19.1) the bypass decision is:

```js
const shouldBypassOptimizeDep = isLitShare(key) || !canResolveSharedSubpath(key, root);
```

and `canResolveSharedSubpath` resolves the entry with Node's **CJS** `require.resolve`:

```js
function canResolveSharedSubpath(subpath, projectRoot) {
  try {
    return isViteOptimizableEntry(
      createRequire(pathToFileURL(path.join(projectRoot, "package.json"))).resolve(subpath)
    );
  } catch {
    return false;
  }
}
```

For an ESM-only package (its `package.json` `exports` declares only `import`/`types`, no `require`), `require.resolve` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, so `canResolveSharedSubpath` returns `false` → the package is pushed to `optimizeDeps.exclude`. Its generated dev prebuild module then imports the raw package source, and its transitive CommonJS deps (e.g. `react-aria` / `react-stately` → `use-sync-external-store`) are served raw. Startup fails with, e.g.:

> SyntaxError: The requested module `…/use-sync-external-store/shim/index.js` does not provide an export named 'useSyncExternalStore'

Vite's own resolver handles the `exports` map fine — only the plugin's CJS `require.resolve` gate rejects it. `react` / `react-dom` (which do declare a `require` condition) are unaffected; only ESM-only shared singletons hit this.

### Reproduction

1. `federation({ shared: { '<esm-only-ui-pkg>': { singleton: true } } })`, where `<esm-only-ui-pkg>` has an `exports` map with **no `require` condition** and pulls a CJS transitive dependency with named exports (e.g. anything depending on `use-sync-external-store`).
2. Run in Vite dev (`vite --force`).

The app blanks. `globalThis.__mf_module_cache__.share` stays empty and `initHost()` rejects on the raw CJS import. Confirmed the excluded package's prebuild imports `/node_modules/<pkg>/…` (raw) instead of `/node_modules/.vite/deps/<pkg>.js` (the interop wrapper).

**Workaround:** add the package to the host's `optimizeDeps.include` (mirrors #913's `optimizeDeps.include: ['react']` workaround).

### Proposed fix

Treat `ERR_PACKAGE_PATH_NOT_EXPORTED` (package resolvable, just not via CJS) as optimizable, so Vite's own resolver takes over:

```diff
 function canResolveSharedSubpath(subpath, projectRoot) {
   try {
     return isViteOptimizableEntry(
       createRequire(pathToFileURL(path.join(projectRoot, "package.json"))).resolve(subpath)
     );
-  } catch {
+  } catch (error) {
+    // ESM-only packages (an "exports" map with no CommonJS "require" condition) throw
+    // ERR_PACKAGE_PATH_NOT_EXPORTED under require.resolve even though Vite can resolve
+    // them. Excluding them serves their raw CJS sub-dependencies in dev. Treat as
+    // optimizable so Vite pre-bundles them (and their transitive CJS deps) with interop.
+    if (error && error.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED') return true;
     return false;
   }
 }
```

Verified against this exact scenario: with the change the shared scope populates and both the host app and a federated remote render in dev, with **no** `optimizeDeps.include` workaround. Happy to open a PR (with a dev integration test alongside the #914 one) if useful.

### Versions

| | |
| --- | --- |
| `@module-federation/vite` | 1.19.1 (latest) |
| `@module-federation/runtime` | 2.8.0 |
| `vite` | 6.4.x |
| `react` / `react-dom` | 19.x |
| package manager | bun (isolated `node_modules`) |

Note: bun's isolated layout leaves the ESM-only package's transitive CJS deps non-hoisted, which is why they surface as raw un-optimized modules rather than being incidentally pre-bundled.

cosmonova:
## What

Fixes #974.

`canResolveSharedSubpath()` decides whether a shared dependency stays in Vite's dev dependency optimization. It resolves the entry with Node's **CJS** `require.resolve()`:

```ts
return isViteOptimizableEntry(createRequire(...).resolve(subpath));
```

For an **ESM-only package** — one whose `package.json` `exports` map declares only an `import` condition (no `require`/`default`) — `require.resolve()` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, so the current `catch { return false }` reports it as unresolvable. The shared singleton is then pushed to `optimizeDeps.exclude` and served as **raw CommonJS**, and its transitive CommonJS dependencies (e.g. `react-aria`/`react-stately` → `use-sync-external-store`) are served raw too. In dev the shared scope never populates, `initHost()` rejects, and the app blanks. The production build is unaffected. Vite's own resolver handles these packages fine — only this CJS gate rejects them.

This is the ESM-only variant of the same class as #913 / #914 (which handled `react`).

## Fix

Treat a package **main entry** that throws `ERR_PACKAGE_PATH_NOT_EXPORTED` as optimizable, so Vite pre-bundles it (and its transitive CJS deps) with interop.

The scope matters: the same error code is thrown both for an ESM-only package that *does* exist and for a **subpath that isn't exported at all** (e.g. `react/compiler-runtime` on React versions that predate it), which genuinely can't be optimized and must stay excluded. Node's stable `import.meta.resolve` can't resolve from an arbitrary project root, so the fix is scoped to **bare specifiers** — a package's main entry — leaving missing subpaths conservatively excluded (preserving the existing `does not register react/compiler-runtime …` behavior).

## Test

`src/__tests__/index.test.ts`: a config-level test that creates a temp ESM-only package (`exports: { ".": { "import": "./index.js" } }`, so CJS `require.resolve` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`) and asserts it lands in `optimizeDeps.include`, not `exclude`. The full file passes (82/82), including the existing `react/compiler-runtime` exclusion test.

Verified end-to-end against a real host + federated remote using `@module-federation/vite@1.19.1`, React 19, and an ESM-only design-system singleton under bun's isolated `node_modules`: with the change the shared scope populates and both the host and the remote render in dev, with no `optimizeDeps.include` workaround.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1,3 +1,5 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
 import path from 'node:path';
 import type {
   ConfigEnv,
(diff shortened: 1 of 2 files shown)


dmchoi77:
One concern: limiting this fallback to bare specifiers protects missing subpaths, but `ERR_PACKAGE_PATH_NOT_EXPORTED` still does not guarantee that the package main is optimizable.

A package with no `"."` export, or an import-only `"."` entry pointing to `.tsx`, would also return `true` here. Could we resolve the exact `"."` import target and apply `isViteOptimizableEntry()` before including it?
Negative tests for these two cases would help.

gioboa (changes requested):
Thanks @cosmonova 
@dmchoi77 ’s concern is valid.

gioboa (approved):
Thanks @cosmonova 👏

gioboa:

image

PR #977, opened 2026-07-26
feat: add `hostInitEntry` for explicit entrypoint ID configuration

### LINKED ISSUE #976: [Bug]: Incompatible with React Router v7+ framework mode (react router Vite plugin)
### Describe the bug

This vite plugin treats all Vite inputs as entrypoints to wrap with runtime code. This causes:
- Multiple instances of the runtime
- Various non-entrypoint bundle requests to be performed with the query string request guard which interferes with handling of the requests during development, such as server modules code isolation (loaders, etc) being requested from the client, failing the server module isolation check.

Image

Only the first request in the above screenshot is the client entrypoint for a React Router application. The rest are route or code-split bundles, which should not be wrapped.

### Version

v8.1.3

### Reproduction

https://github.com/WoodyWoodsta/react-router-mod-fed-entrypoint-repro-custom-server

### Relevant log output

```shell

```

### Validations

WoodyWoodsta:
Fixes #976.

This allows developers to specify the client entrypoint file (in react router, this is entry.client.tsx).

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,11 +167,6 @@ export default defineConfig({
       // initHost() completes before hydrateRoot and @module-federation/bridge-react
       // remotes render on first paint.
       hostInitInjectLocation: "html", // or "entry"
-      // When `hostInitInjectLocation` is "entry", optionally wrap exactly this
-      // client module instead of discovering one from Vite inputs. Useful for
-      // frameworks such as React Router that expose route modules as additional
-      // Vite inputs.
-      // hostInitEntry: "./src/entry.client.tsx",
       // Controls whether all CSS assets from the bundle should be added to every exposed module.
       // When false (default), the plugin will not process any CSS assets.
       // When true, all CSS assets are bundled into every exposed module.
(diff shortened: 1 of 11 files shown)


gioboa (commented):
Thanks @WoodyWoodsta 
I’d prefer not to change the public API for a single scenario. If you agree, we can handle it internally this way instead.
Let me know if it's working for you.

gioboa (commented):

Image

WoodyWoodsta:
No worries @gioboa. I can understand not wanting to change the public API, and I noticed that the plugin special-cases some other frameworks internally.

I was hoping that the public configuration meant that any framework or custom application with a single client entrypoint but multiple Vite inputs would benefit from being able to configure that without a change made to the plugin. So I don't fully agree with the proposed approach.

Equally, it's up to you, so happy to go ahead if it means getting React Router framework mode working.

gioboa:
is this pre-release working for you @WoodyWoodsta  ?

WoodyWoodsta:
Have just checked out now, and it's working for me. Thank you!

gioboa (commented):
Thanks for your help @WoodyWoodsta 💪

PR #979, opened 2026-07-26
fix: preserve singleton sharing for TypeScript generic exports

### LINKED ISSUE #978: [Bug]: `singleton: true` shared modules are duplicated across multiple remotes registered via `registerRemotes()`
### Description

In `@module-federation/vite` >= 1.17.1, declaring a shared module with `{ singleton: true }` fails when a host dynamically registers 2+ remotes using `registerRemotes()`. Each remote silently initializes its own local share scope instead of joining the host's share scope, creating duplicate instances of module-level state (e.g. WeakMaps, singleton stores).

### Bisect Results

- **`@module-federation/vite@1.17.0`**: ✅ **PASS** (`host.instance === remote1.instance === remote2.instance`)
- **`@module-federation/vite@1.17.1`**: ❌ **FAIL** (Host, remote1, and remote2 get 3 separate module instances)
- **`@module-federation/vite@1.19.1`**: ❌ **FAIL** (Reproduces on latest)

### Root Cause Analysis

Likely introduced in PR [#918](https://github.com/module-federation/vite/pull/918) (*"fix: resolve lazy shared providers"*) - at src/virtualModules/virtualRemoteEntry.ts:

```ts
const federationInstances = globalThis.__FEDERATION__?.__INSTANCES__ || [];
const scopeRoot = federationInstances.find((instance) => 
  instance?.options?.name === initRootName && instance?.shareScopeMap?.['default'] === shared
) || federationInstances.find((instance) => 
  instance?.options?.name !== mfName && instance?.shareScopeMap?.['default'] === shared
);
```

When multiple remotes are registered simultaneously, the object reference equality check `instance?.shareScopeMap?.['default'] === shared` or `instance?.options?.name === initRootName` fails to match the owning federation host instance. When `scopeRoot` is `undefined`, the remote falls back to creating a new local share scope rather than reusing the host's singleton scope.

### Minimal Reproduction Repository

Reproduction Repo: https://github.com/kingston/module-federation-singleton-repro

#### Repro Steps:
1. Clone the reproduction repo.
2. Run `pnpm install`.
3. Run `pnpm dev` (or `pnpm build && pnpm preview`).
4. Open `http://localhost:5100` in the browser and inspect the rendered test object or console output.

##### `@module-federation/vite@1.17.0`:
```json
{
  "host": { "instanceId": "shared-lib-instance-4lnnh10", "refLookup": { "entityName": "User" } },
  "remote1": { "instanceId": "shared-lib-instance-4lnnh10", "refLookup": { "entityName": "User" } },
  "remote2": { "instanceId": "shared-lib-instance-4lnnh10", "refLookup": { "entityName": "User" } },
  "success": true,
  "errors": []
}
```

##### `@module-federation/vite@1.17.1` / `@1.19.1`:
```json
{
  "host": { "instanceId": "shared-lib-instance-kqfmepw", "refLookup": { "entityName": "User" } },
  "remote1": { "instanceId": "shared-lib-instance-2cm1w7p" },
  "remote2": { "instanceId": "shared-lib-instance-v5ayc9b" },
  "success": false,
  "errors": [
    "Remote1 instanceId mismatch: expected shared-lib-instance-kqfmepw, got shared-lib-instance-2cm1w7p",
    "Remote2 instanceId mismatch: expected shared-lib-instance-kqfmepw, got shared-lib-instance-v5ayc9b",
    "Remote1 failed WeakMap lookup (got undefined)",
    "Remote2 failed WeakMap lookup (got undefined)"
  ]
}
```

### Environment
- OS: macOS
- Node.js: v24
- `@module-federation/vite`: 1.17.1 - 1.19.1
- `@module-federation/enhanced`: 2.7.0

### Version

8.1.5

### Reproduction

https://github.com/kingston/module-federation-singleton-repro

### Relevant log output

```shell

```

### Validations

dmchoi77:
## Summary

Fixes singleton duplication when a shared TypeScript source exports a variable initialized with generic syntax containing a comma, for example:

```ts
export const registry = new WeakMap<object, any>();
```

Closes #978.

## Root cause

The regex-based named-export scan uses `hasTopLevelDeclaratorComma` to determine whether an exported variable statement contains multiple declarators.

The scanner tracked `()`, `[]`, and `{}`, but did not account for TypeScript type arguments. As a result, the comma in `<object, any>` was interpreted as a top-level variable declarator separator.

This marked named-export detection as incomplete and made the shared proxy fall back to:

```ts
export * from "<local module>";
```

That fallback bypassed the live named-export bridge backed by the runtime-selected shared instance, allowing the host and runtime-registered remotes to evaluate separate copies of the shared module.

## Changes

- Add an internal `typeArgumentScanner` utility that recognizes balanced, type-like `<...>` ranges containing commas.
- Reuse the existing `createCodePositionMap` result so strings, comments, template literals, and regular expressions remain excluded from syntax scanning.
- Skip generic commas while scanning exported variable declarations.
- Preserve the existing fail-closed behavior for:
  - actual multi-declarator exports;
  - relational `<` and `>` expressions;
  - incomplete or ambiguous angle-bracket syntax.
- Add focused utility tests covering:
  - generic constructors and function calls;
  - nested type arguments;
  - generic arrow functions;
  - TypeScript assertion keywords;
  - relational and malformed syntax.
- Add generated-proxy regression coverage.
- Add a dev integration test with one host and two runtime-registered remotes sharing the same module ID and `WeakMap` state.

No dependencies or lockfile entries were added.

## Why this approach

- It fixes the incorrect comma classification at the scanner boundary instead of special-casing `WeakMap`.
- It avoids introducing a TypeScript parser dependency for this synchronous source-inspection path.
- The scanner is intentionally conservative: uncertain syntax returns `undefined`, preserving the existing safe fallback.
- The syntax-specific logic is isolated in `src/utils/typeArgumentScanner.ts`; `virtualShared_preBuild.ts` only integrates the result.
- There are no public API, configuration, or type-surface changes.

## Review guide

The change can be reviewed in this order:

1. `src/utils/typeArgumentScanner.ts`
   - Start-context validation.
   - Balanced group and angle-bracket tracking.
   - Valid follower checks.
   - Fail-closed exits for ambiguous syntax.
2. `src/virtualModules/virtualShared_preBuild.ts`
   - The scanner is invoked only when `<` appears while checking an exported variable declaration.
   - A recognized type-argument range is skipped before top-level comma detection continues.
3. `src/utils/__tests__/typeArgumentScanner.test.ts`
   - Direct scanner contract and ambiguity coverage.
4. `src/virtualModules/__tests__/virtualShared_preBuild.test.ts`
   - Generated shared-proxy behavior and real declarator fallback.
5. `integration/dev-singleton-react.test.ts`
   - End-to-end module identity and shared `WeakMap` state across the host and two registered remotes.

## Verification

- `pnpm test`
  - 45 test files passed.
  - 939 tests passed, 1 skipped.
- `pnpm typecheck`
- `pnpm build`
- `npx oxfmt --check src`
- Targeted integration regression:

```sh
pnpm exec vitest run integration/dev-singleton-react.test.ts \
  -t "runtime-registered TypeScript source singleton"
```

The integration test verifies that:

- the host and both remotes observe the same module instance ID;
- both remotes read the value registered in the host's `WeakMap`;
- no browser runtime errors are emitted.

### DIFF
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -373,7 +373,7 @@ export const getRemoteInfo = (key: object) => ({
       name: 'issue978RemoteOne',
       filename: 'remoteEntry.js',
       manifest: true,
-      exposes: { './info': './src/RemoteOne.ts' },
+      exposes: { './info': path.join(fixtureRoot, 'src/RemoteOne.ts') },
       dts: false,
       shared,
     });
@@ -381,7 +381,7 @@ export const getRemoteInfo = (key: object) => ({
       name: 'issue978RemoteTwo',
       filename: 'remoteEntry.js',
       manifest: true,
-      exposes: { './info': './src/RemoteTwo.ts' },
+      exposes: { './info': path.join(fixtureRoot, 'src/RemoteTwo.ts') },
       dts: false,
       shared,
     });
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -1,12 +1,12 @@
+import { chromium } from '@playwright/test';
 import { spawn, type ChildProcess } from 'node:child_process';
 import { once } from 'node:events';
 import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { createServer as createNetServer } from 'node:net';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
-import { chromium } from '@playwright/test';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { createServer as createViteServer, version as viteVersion } from 'vite';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { federation } from '../src';
 
 const REPO_ROOT = path.resolve(import.meta.dirname, '..');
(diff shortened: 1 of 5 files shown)


gioboa (commented):

Image

gioboa (approved):
Great fix @dmchoi77 🙏

PR #980, opened 2026-07-26
fix: resolve shared exports using per-environment conditions

dmchoi77:
### Summary

This PR makes shared named-export analysis use the export conditions of the active Vite build environment.

Node SSR builds now inspect the Node entry of conditional packages such as `react-dom/server`, while client and webworker graphs retain their respective condition sets. The change is internal and does not add plugin options or alter the public API.

### Problem

Shared proxy generation always inspected packages with:

```ts
['browser', 'import', 'module', 'default']
```

During a Node SSR production build, this caused the plugin to analyze `react-dom/server.browser.js` even though Vite bundled `react-dom/server.node.js`.

As a result, the generated shared proxy omitted `renderToPipeableStream`, causing the SSR build or emitted server module to fail.

### Root cause

Shared virtual modules were generated and cached without being scoped to the active build environment. As a result, a module first generated with client conditions could be reused by an SSR graph.

This is especially visible with Vite's Environment API, where client and SSR graphs can coexist in the same process.

### Changes

- Select shared export conditions from the active Vite environment.
  - Vite 8: use `this.environment.config.resolve.conditions`.
  - Vite 5–7: derive conditions from `build.ssr`, `ssr.target`, and configured resolve conditions.
- Treat both boolean and entry-string forms of `build.ssr` as SSR builds.
- Regenerate owned `loadShare` and prebuild virtual modules when they are loaded, preventing one environment from reusing another environment's export surface.
- Apply the same conditions to package roots and `shareConfig.import` package sources.
- Preserve the existing browser-first behavior when no environment conditions are supplied.
- Use worker/browser fallback conditions for webworker SSR instead of Node conditions.
- Add a React 19 production SSR regression test by reusing an existing workspace package, without changing workspace configuration, dependencies, or the lockfile.

### Review guide

1. `src/utils/sharedExportConditions.ts`
   - Condition selection for Vite environments and Vite 5–7 fallback behavior.
2. `src/virtualModules/virtualShared_preBuild.ts`
   - Condition-aware package entry resolution and named-export analysis.
3. `src/index.ts`
   - Per-environment virtual-module refresh and ownership checks.
4. Tests
   - Unit coverage for client, Node SSR, webworker SSR, `shareConfig.import`, and string-valued `build.ssr`.
   - Integration coverage that builds and imports a React 19 SSR bundle and verifies `renderToPipeableStream`.

### Validation

- `pnpm test` — 937 passed, 1 skipped
- `pnpm test:integration` — 52 passed
- `pnpm typecheck`
- `pnpm build`
- `pnpm fmt.check`
- `pnpm exec oxfmt --check integration/ssr-shared-exports.test.ts`
- React 19 SSR regression passed with:
  - Vite 5.4.21
  - Vite 6.4.2
  - Vite 7.3.2
  - Vite 8.1.0

### DIFF
--- a/integration/ssr-shared-exports.test.ts
+++ b/integration/ssr-shared-exports.test.ts
@@ -1,0 +1,75 @@
+import { mkdtemp, rm } from 'node:fs/promises';
+import { resolve } from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { build, type Plugin, type Rollup } from 'vite';
+import { afterEach, describe, expect, it } from 'vitest';
+import { federation } from '../src';
+
+const outputDirs: string[] = [];
+// Reuse an existing React 19 workspace package so this regression test does not
+// need its own package.json, workspace entry, or lockfile importer.
+const reactWorkspaceRoot = resolve(import.meta.dirname, '../examples/vite-vite/vite-remote');
+const ssrEntryId = 'virtual:ssr-shared-exports-entry';
+const resolvedSsrEntryId = `\0${ssrEntryId}`;
+
+const ssrEntryPlugin: Plugin = {
+  name: 'ssr-shared-exports-entry',
+  resolveId(id) {
+    if (id === ssrEntryId) return resolvedSsrEntryId;
+  },
+  load(id) {
+    if (id === resolvedSsrEntryId) {
+      return "export { renderToPipeableStream } from 'react-dom/server';";
+    }
+  },
+};
+
+afterEach(async () => {
+  await Promise.all(outputDirs.splice(0).map((dir) => rm(dir, { recursive: true, force: true })));
+});
+
+describe('SSR shared conditional exports', () => {
+  it('builds and executes the node export surface of react-dom/server', async () => {
+    const outDir = await mkdtemp(resolve(reactWorkspaceRoot, '.mf-vite-ssr-'));
+    outputDirs.push(outDir);
+
+    const result = await build({
+      root: reactWorkspaceRoot,
+      configFile: false,
+      logLevel: 'silent',
+      plugins: [
+        ssrEntryPlugin,
+        federation({
+          name: 'ssrHost',
+          shared: {
+            'react-dom': { singleton: true },
+          },
+          dts: false,
+        }),
+      ],
+      build: {
+        ssr: true,
+        outDir,
+        write: true,
+        minify: false,
+        target: 'node20',
+        rollupOptions: {
+          input: ssrEntryId,
+        },
+      },
+    });
+    expect(Array.isArray(result), 'Expected a single RollupOutput, not an array').toBe(false);
+    const output = result as Rollup.RollupOutput;
+
+    const entryChunk = output.output.find(
+      (item): item is Rollup.OutputChunk =>
+        item.type === 'chunk' && item.isEntry && item.facadeModuleId === resolvedSsrEntryId
+    );
+    expect(entryChunk, 'Expected the virtual SSR entry chunk').toBeDefined();
+
+    const serverEntry = await import(
+      `${pathToFileURL(resolve(outDir, entryChunk!.fileName)).href}?test=${Date.now()}`
+    );
+    expect(serverEntry.renderToPipeableStream).toBeTypeOf('function');
+  });
+});
--- a/src/index.ts
+++ b/src/index.ts
@@ -765,6 +765,7 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
   let command: string;
   let desiredRolldownOutput: OutputNameOptions[] | undefined;
   let isSsrBuild = false;
+  let isProduction = false;
   let rootResolveConditions: string[] | undefined;
   let ssrResolveConditions: string[] | undefined;
   let ssrTarget: 'node' | 'webworker' = 'node';
@@ -778,6 +779,7 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
       config?: {
         consumer?: string;
         build?: { ssr?: boolean | string };
+        isProduction?: boolean;
         resolve?: { conditions?: string[] };
       };
     };
@@ -794,6 +796,7 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
       environment?.name === 'server';
     return getSharedExportConditions({
       environmentConditions: environment?.config?.resolve?.conditions,
+      isProduction: environment?.config?.isProduction ?? isProduction,
       isSsr,
       rootConditions: rootResolveConditions,
       ssrConditions: ssrResolveConditions,
@@ -936,7 +939,8 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
           : undefined;
         ssrTarget = config.ssr?.target ?? 'node';
       },
-      configResolved() {
+      configResolved(config: ResolvedConfig) {
+        isProduction = config.isProduction;
         const ssrCapabilities = getSsrCapabilities(
           parseInt(viteVersion, 10),
           command as 'serve' | 'build',
--- a/src/index.ts
+++ b/src/index.ts
@@ -30,6 +30,7 @@ import {
 } from './utils/controlChunkSanitizer';
 import { isTestEnv } from './utils/isTestEnv';
 import { createModuleFederationError, mfWarn } from './utils/logger';
+import { getSharedExportConditions } from './utils/sharedExportConditions';
 import type {
   ModuleFederationOptions,
   NormalizedModuleFederationOptions,
@@ -80,7 +81,10 @@ import { addUsedRemote } from './virtualModules/virtualRemotes';
 import { getRuntimeInitStatusImportId } from './virtualModules/virtualRuntimeInitStatus';
 import {
   findCurrentLoadShareForStaleOwnerId,
+  getCachedLoadSharePkg,
+  getCachedPreBuildPkg,
   getLoadShareModulePath,
+  getPreBuildLibImportId,
   materializeCachedLoadShareModule,
   prependWorkspaceSingletonSsrImport,
   writeLoadShareModule,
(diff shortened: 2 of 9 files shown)


src/utils/sharedExportConditions.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+function appendConditions(
+  conditions: readonly string[],
+  fallbackConditions: readonly string[]
+): string[] {
+  return [...new Set([...conditions, ...fallbackConditions])];
+}
+
+export function getSharedExportConditions({
```
dmchoi77, line 41: Environment-provided conditions take precedence. When they are unavailable, the fallback conditions are selected for the client, Node SSR, or webworker environment.

The order is intentional because it determines which conditional package entry is inspected.

src/virtualModules/virtualShared_preBuild.ts
```diff
@@ -583,10 +591,14 @@ function getPackageNamedExports(pkg: string): string[] | undefined {
...
-export function getSharedNamedExports(pkg: string, shareItem?: ShareItem): string[] | undefined {
+export function getSharedNamedExports(
+  pkg: string,
+  shareItem?: ShareItem,
+  exportConditions = DEFAULT_SHARED_EXPORT_CONDITIONS
+): string[] | undefined {
   const configuredImport = shareItem?.shareConfig.import;
   if (typeof configuredImport === 'string') {
-    const configuredImportPath = resolveConfiguredImportPath(configuredImport);
+    const configuredImportPath = resolveConfiguredImportPath(configuredImport, exportConditions);
```
dmchoi77, line 601: `shareConfig.import` is an authoritative alternative to the package root, so it must be resolved with the same environment conditions.

Otherwise, the configured source and the generated shared proxy could expose different named-export surfaces.

integration/ssr-shared-exports.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const entryChunk = output.output.find(
+      (item): item is Rollup.OutputChunk =>
+        item.type === 'chunk' && item.isEntry && item.facadeModuleId === resolvedSsrEntryId
+    );
+    expect(entryChunk, 'Expected the virtual SSR entry chunk').toBeDefined();
+
+    const serverEntry = await import(
+      `${pathToFileURL(resolve(outDir, entryChunk!.fileName)).href}?test=${Date.now()}`
+    );
+    expect(serverEntry.renderToPipeableStream).toBeTypeOf('function');
```
dmchoi77, line 73: `renderToPipeableStream` is the regression signal because it is exposed by the Node entry of `react-dom/server`, but not by the browser entry inspected before this fix.

This assertion therefore verifies both that the SSR bundle executes and that shared named-export analysis selected the correct conditional entry.

src/utils/sharedExportConditions.ts
```diff
@@ -0,0 +1,48 @@
...
+  environmentConditions,
+  isSsr,
+  rootConditions,
+  ssrConditions,
+  ssrTarget = 'node',
+}: SharedExportConditionOptions): string[] {
+  if (environmentConditions !== undefined) {
```
gioboa, line 49: Vite’s resolved conditions still contain the special development|production placeholder. 
Vite normally replaces this with development or production immediately before package resolution, but our resolver receives it literally and cannot match either export condition. 
That can make named-export detection inspect the default or wrong entry. Please pass the resolved production mode (config.isProduction) into this logic and expand the placeholder before calling getInstalledPackageEntry. A test where the development and production entries expose different named exports would cover this.

src/index.ts
```diff
@@ -829,8 +926,15 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
     {
       name: 'vite:module-federation-config',
       enforce: 'pre',
-      config(_config: UserConfig, env: ConfigEnv) {
+      config(config: UserConfig, env: ConfigEnv) {
         command = env.command;
+        rootResolveConditions = config.resolve?.conditions
```
gioboa, line 936: These condition values are captured in an enforce: 'pre' config hook. Any framework or plugin that adds resolve.conditions, ssr.resolve.conditions, or ssr.target in a later config hook will therefore be ignored by the Vite 5–7 fallback path. Please capture these values from the final ResolvedConfig in configResolved(config) instead, where all plugin config hooks have already been applied. A test with a second plugin contributing an SSR condition would demonstrate the issue.

gioboa (changes requested):

src/utils/sharedExportConditions.ts
dmchoi77: Thanks for catching this. Fixed in [ad1a22e](https://github.com/module-federation/vite/pull/980/commits/ad1a22e7b19429303364ec801ce1fc70612999cc).

Shared-export condition selection now receives `isProduction` from `this.environment.config.isProduction` when an active Vite environment is available, and from the final `ResolvedConfig.isProduction` for the Vite 5–7 fallback.

`getSharedExportConditions` expands Vite's `development|production` placeholder to the concrete `development` or `production` condition before the resulting conditions are passed to `getInstalledPackageEntry`.

I also added regression tests for both modes. The development and production entries expose different named exports, and the tests verify that named-export scanning selects the corresponding entry.

src/index.ts
dmchoi77: Thanks, fixed in [27ca84d](https://github.com/module-federation/vite/pull/980/commits/27ca84d0877abb1d93bdc4205282e40cb649df9d).

The `enforce: 'pre'` config hook now captures only `command`. `resolve.conditions`, `ssr.resolve.conditions`, and `ssr.target` are copied from the final `ResolvedConfig` in `configResolved`, after all plugin config hooks have been applied.

I added a regression test that simulates a second plugin contributing root and SSR conditions together with a `webworker` target. It verifies that the Vite 5–7 load-hook fallback receives all values from the final resolved configuration.

gioboa (commented):
Can you have a look at [additional_changes.patch](https://github.com/user-attachments/files/30393514/additional_changes.patch) ?

This change aligns the resolver with Vite/Node behavior and adds a regression test.

dmchoi77:
@gioboa Thanks, I’ll take a look. The attachment currently returns `Not Found` for me, including while signed in. 
Could you re-upload the patch or paste the diff?

gioboa:
```
diff --git a/src/utils/__tests__/packageUtils.test.ts b/src/utils/__tests__/packageUtils.test.ts
index aeede46..fe6f257 100644
--- a/src/utils/__tests__/packageUtils.test.ts
+++ b/src/utils/__tests__/packageUtils.test.ts
@@ -116,6 +116,43 @@ describe('getInstalledPackageJson', () => {
     expect(entry).toBe(path.join(packageDir, 'dist/browser.js'));
   });
 
+  it('respects package export key order when multiple conditions are active', () => {
+    const packageName = 'mf-test-conditional-key-order';
+    const root = mkdtempSync(path.join(tmpdir(), 'mf-vite-conditional-order-'));
+    tempDirs.push(root);
+
+    const hostDir = path.join(root, 'apps/host');
+    const packageDir = path.join(hostDir, 'node_modules', packageName);
+    mkdirSync(path.join(packageDir, 'dist'), { recursive: true });
+    writeFileSync(path.join(hostDir, 'package.json'), JSON.stringify({ name: 'host' }));
+    writeFileSync(
+      path.join(packageDir, 'package.json'),
+      JSON.stringify({
+        name: packageName,
+        exports: {
+          '.': {
+            browser: './dist/browser.js',
+            worker: './dist/worker.js',
+            default: './dist/default.js',
+          },
+        },
+      })
+    );
+    writeFileSync(path.join(packageDir, 'dist/browser.js'), 'export const browserOnly = true;');
+    writeFileSync(path.join(packageDir, 'dist/worker.js'), 'export const workerOnly = true;');
+    writeFileSync(path.join(packageDir, 'dist/default.js'), 'export const defaultOnly = true;');
+
+    const entry = getInstalledPackageEntry(packageName, {
+      cwd: hostDir,
+      conditions: ['worker', 'browser', 'import', 'default'],
+      resolveSubpathWithRequire: false,
+    });
+
+    // Conditional exports are matched in package.json key order. Although both
+    // worker and browser are active, browser appears first and must win.
+    expect(entry).toBe(path.join(packageDir, 'dist/browser.js'));
+  });
+
   it('preserves legacy package subpaths when ESM condition resolution is requested', () => {
     const packageName = 'mf-test-legacy-subpath';
     const root = mkdtempSync(path.join(tmpdir(), 'mf-vite-legacy-subpath-'));
diff --git a/src/utils/packageUtils.ts b/src/utils/packageUtils.ts
index fef7320..11945a2 100644
--- a/src/utils/packageUtils.ts
+++ b/src/utils/packageUtils.ts
@@ -66,21 +66,36 @@ const DEFAULT_EXPORT_CONDITIONS = ['browser', 'import', 'module', 'default', 're
 function resolveExportsEntry(
   exportsField: unknown,
   conditions = DEFAULT_EXPORT_CONDITIONS
+): string | undefined {
+  return resolveExportsEntryWithConditions(exportsField, new Set(conditions));
+}
+
+function resolveExportsEntryWithConditions(
+  exportsField: unknown,
+  conditions: ReadonlySet<string>
 ): string | undefined {
   if (typeof exportsField === 'string') return exportsField;
   if (!exportsField || typeof exportsField !== 'object') return undefined;
-  const record = exportsField as Record<string, unknown>;
-  const rootExport = record['.'];
-  if (rootExport) return resolveExportsEntry(rootExport);
 
-  for (const condition of conditions) {
-    const target = resolveExportsEntry(record[condition], conditions);
-    if (target) return target;
+  if (Array.isArray(exportsField)) {
+    for (const target of exportsField) {
+      const resolved = resolveExportsEntryWithConditions(target, conditions);
+      if (resolved) return resolved;
+    }
+    return undefined;
   }
 
-  for (const target of Object.values(record)) {
-    const resolved = resolveExportsEntry(target, conditions);
-    if (resolved) return resolved;
+  const record = exportsField as Record<string, unknown>;
+  const rootExport = record['.'];
+  if (rootExport) return resolveExportsEntryWithConditions(rootExport, conditions);
+
+  // Conditional exports are matched in package.json key order. Conditions
+  // identify the active branches, but their own order does not affect which
+  // branch wins. `default` is always eligible as the universal fallback.
+  for (const [condition, value] of Object.entries(record)) {
+    if (condition !== 'default' && !conditions.has(condition)) continue;
+    const target = resolveExportsEntryWithConditions(value, conditions);
+    if (target) return target;
   }
 
   return undefined;

```

dmchoi77:
@gioboa Thanks for sharing the patch. I reviewed and applied it in [9cd5f88](https://github.com/module-federation/vite/pull/980/commits/9cd5f88c7a053d139aaea82ba9fc0f6a06be0add).

Conditional export resolution now treats conditions as an active set and follows `package.json` key order, matching Vite/Node condition precedence. Array fallback order is preserved, and `default` remains universally eligible.

While reviewing the change, I found one related edge case: the previous default condition list contained both `import` and `require`. Once converted to a set, both became active, so packages declaring `require` first could incorrectly select the CommonJS entry during ESM analysis. I removed `require` from the default ESM conditions and added regression coverage for both cases.

gioboa (approved):
Thanks @dmchoi77 it looks good to me 👏

dmchoi77:
@gioboa Thanks for your reviewing😄

PR #981, opened 2026-07-26
chore: support React SSR islands

gioboa:
Adds experimental React SSR islands with server rendering and client hydration. 

> [!CAUTION] 
> Do not use in production; the API will change.


gioboa (commented):

Image

PR #982, opened 2026-07-27
fix: propagate conditions through re-exports

gioboa:
Propagates active Vite export conditions through recursive package star re-exports, ensuring SSR and worker builds inspect the same conditional entries that Vite bundles.


gioboa (commented):

Image

PR #987, opened 2026-07-28
fix: ignore member properties during CommonJS detection

### LINKED ISSUE #986: [Bug]: `import: false` misclassifies ESM packages containing internal `t.exports` and drops named exports
### Describe the bug

When an installed browser ESM package is configured as a shared dependency with `import: false`, `@module-federation/vite` may incorrectly classify the package as CommonJS if its bundle
  contains an internal assignment such as:

  ```js
  t.exports = { internal: true };

  The package still has a valid ESM export:

  export const namedValue = 'named ESM export';

  However, Module Federation generates a virtual loadShare module without namedValue, causing the production build to fail with MISSING_EXPORT.

  The plugin also reports that the package is not installed locally, even though it is installed as a devDependency.

  ## Version

  - @module-federation/vite: 1.20.0
  - Vite: 8.0.16
  - Node.js: 24.15.0
  - pnpm: 11.1.1

  ## Reproduction

  https://github.com/francesc79/mf-vite-import-false-cjs-detection-repro

  The reproduction uses only Vite and a small workspace ESM package. It does not use React, private packages, or an application framework.

  ## Steps to reproduce

  git clone https://github.com/francesc79/mf-vite-import-false-cjs-detection-repro.git
  cd mf-vite-import-false-cjs-detection-repro
  corepack enable
  pnpm install
  pnpm repro

  No host, server, or browser is required. The problem occurs during the production build.

  ## Expected behavior

  Module Federation should detect the package's namedValue ESM export and generate a host-provided shared module exposing it.

  The production build should succeed.

  ## Actual behavior

  The build emits this warning:

  [Module Federation] Shared dependency "@repro/browser-esm-with-internal-cjs" has import: false but is not installed locally.
    Named imports (e.g. import { ... } from '@repro/browser-esm-with-internal-cjs') will not work in production builds.

  The package is already installed as a devDependency.

  The build then fails with:

  [MISSING_EXPORT] "namedValue" is not exported by
  "\0virtual:mf:__mfe_internal__reproRemote__mf_owner__1__loadShare___mf_0_repro_mf_1_browser_mf_2_esm_mf_2_with_mf_2_internal_mf_2_cjs__loadShare__.js".

  ## Minimal trigger

  The shared package contains:

  const internalModule = { exports: {} };

  ((t) => {
    t.exports = { internal: true };
  })(internalModule);

  window.__mfViteImportFalseRepro = internalModule.exports;

  export const namedValue = 'named ESM export';

  It is configured as:

  shared: {
    '@repro/browser-esm-with-internal-cjs': {
      singleton: true,
      import: false,
    },
  }

  Removing only the t.exports = ... assignment makes the build succeed.

  ## Suspected cause

  hasCommonJsExports() uses a pattern that matches any identifier named exports, including property access such as t.exports:

  /\bmodule\s*(?:\.exports|\[\s*['"]exports['"]\s*\])|\bexports\s*(?:\.|\[|[,)]|=(?!=|>))/g

  As a result, the ESM entry is classified as CommonJS.

  The plugin then tries to require() the browser ESM entry to determine its runtime exports. Since the browser package references window, that evaluation fails. Named-export detection
  returns undefined, and the generated virtual module exposes only the default export.

  The CommonJS detection should distinguish the global CommonJS exports identifier from member properties such as t.exports.


### Version

8.0.16

### Reproduction

https://github.com/francesc79/mf-vite-import-false-cjs-detection-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #986 

Prevent .exports properties from misclassifying ESM as CommonJS.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -248,6 +248,9 @@ vi.mock('../../utils/packageUtils', () => ({
     if (pkg === 'mock-package-cjs-comment') {
       return '/repo/apps/remote/node_modules/mock-package-cjs-comment/index.js';
     }
+    if (pkg === 'mock-package-browser-esm-internal-cjs') {
+      return '/repo/apps/remote/node_modules/mock-package-browser-esm-internal-cjs/index.js';
+    }
     if (pkg === 'workspace-shared-lib') {
       return '/repo/packages/workspace-shared-lib/src/index.tsx';
     }
@@ -458,6 +461,8 @@ vi.mock('fs', () => ({
       filePath.endsWith('/mock-package-dual-shape/dist/browser.js') ||
       filePath.endsWith('node_modules/mock-package-cjs-comment/index.js') ||
       filePath.endsWith('/mock-package-cjs-comment/index.js') ||
+      filePath.endsWith('node_modules/mock-package-browser-esm-internal-cjs/index.js') ||
+      filePath.endsWith('/mock-package-browser-esm-internal-cjs/index.js') ||
       filePath.endsWith('node_modules/react/jsx-runtime.js') ||
       filePath.endsWith('/repo/packages/workspace-shared-lib/package.json') ||
       filePath.endsWith('/repo/packages/workspace-unknown-exports/package.json') ||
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #988, opened 2026-07-28
chore: add opt-in React island adapters

gioboa:
Improve experimental React SSR islands with server rendering and client hydration. 

> [!CAUTION] 
> Do not use in production; the API will change.


gioboa (commented):

Image

PR #991, opened 2026-07-29
fix: tolerate trailing commas in export lists during named-export detection

### LINKED ISSUE #990: Regression in 1.17.1: named-export detection fails when a shared package's entry has a trailing comma in an export list
**Environment:**
- `@module-federation/vite` 1.20.0 (regressed in **1.17.1**; works through 1.17.0)
- `vite` 8.1.5 (Rolldown)

## Summary

When a `shared` entry uses `import: false`, the plugin statically scans the host-provided package's ESM entry to detect its named exports. If that entry contains an `export { ... }` list with a trailing comma — which Prettier emits by default for multi-line lists (`trailingComma: "all"` since v3) — the scan is marked incomplete and detection returns nothing. The build then fails with `MISSING_EXPORT` for every named import of that share, and a misleading warning claims the package is not installed.

## Reproduction

A shared dependency whose published entry looks like:

```js
// node_modules/some-lib/index.js
export const A = 1;
export const B = 2;
export {
  A as C,
};
```

```ts
// vite.config.ts
federation({
  name: 'remote',
  exposes: { './thing': './src/thing.ts' },
  shared: {
    'some-lib': { singleton: true, import: false },
  },
})
```

```ts
// src/thing.ts
import { A } from 'some-lib';
```

## Actual behavior

```
[Module Federation] Shared dependency "some-lib" has import: false but is not installed locally.
  Named imports (e.g. import { ... } from 'some-lib') will not work in production builds.
...
[MISSING_EXPORT] "A" is not exported by "\0virtual:mf:...__loadShare__some_mf_2_lib__loadShare__.js".
```

The package *is* installed and resolves fine; the warning fires because export detection returned `undefined`, not because resolution failed.

## Cause / regression

`getNamedExportsViaRegex` splits the `export {...}` specifier list on `","`; a trailing comma produces an empty final specifier that fails the specifier regex. Through **1.17.0** this was tolerated:

```js
const asMatch = trimmed.match(exportSpecifierRegex);
if (!asMatch) continue;
```

Since **1.17.1** the same branch poisons the whole scan:

```js
if (!asMatch) {
  scanState.complete = false;   // <-- empty string from the trailing comma lands here
  continue;
}
```

so an otherwise fully detected export list is discarded. Removing just the trailing comma from the shared package's entry makes detection succeed and the build pass.

## Expected behavior

Trailing commas are valid ES module syntax and Prettier-default formatting, so any Prettier-formatted package with a multi-line export list can trigger this. Empty specifiers from trailing commas should be skipped (as before 1.17.1) rather than marking the scan incomplete. It would also help to distinguish the "package not installed" warning from "exports could not be statically detected."

mstijak:
Fixes #990

## Problem

Since 1.17.1, named-export detection marks the scan incomplete when an `export { ... }` list contains a trailing comma — Prettier's default formatting for multi-line lists (`trailingComma: "all"`). Splitting the specifier list on `","` yields an empty final specifier, which fails the specifier regex and lands in the branch that sets `scanState.complete = false`, discarding an otherwise fully detected export list. For `import: false` shares this surfaces as `MISSING_EXPORT` build errors plus a misleading "is not installed locally" warning.

Through 1.17.0 the same branch was `if (!asMatch) continue;`, so trailing commas were tolerated.

## Fix

Skip empty specifiers before matching, so a trailing comma is treated as the valid syntax it is, while genuinely unparseable specifiers still mark the scan incomplete.

Includes a regression test with a Prettier-shaped fixture (`export {\n  first as alias,\n};`) asserting the export list is still detected as complete.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -494,6 +494,7 @@ vi.mock('fs', () => ({
       filePath.endsWith('/repo/packages/multi-declarator.js') ||
       filePath.endsWith('/repo/packages/postfix-multi-declarator.js') ||
       filePath.endsWith('/repo/packages/string-export-list.js') ||
+      filePath.endsWith('/repo/packages/trailing-comma-export-list.js') ||
       filePath.endsWith('/repo/packages/default-reexport-list.js') ||
       filePath.endsWith('/repo/packages/string-namespace-export.js') ||
       filePath.endsWith('/repo/packages/nested-destructuring.js') ||
@@ -596,6 +597,14 @@ export const assertedRegistry = <Map<object, any>>new Map();`;
     }
     if (filePath.endsWith('/repo/packages/string-export-list.js')) {
       return 'const foo = 1; export { foo as "a-b", foo as valid };';
+    }
+    if (filePath.endsWith('/repo/packages/trailing-comma-export-list.js')) {
+      // Prettier's default `trailingComma` formatting for a multi-line export list.
+      return `export const first = 1;
+export const second = 2;
+export {
+  first as alias,
+};`;
     }
     if (filePath.endsWith('/repo/packages/default-reexport-list.js')) {
       // MUI's per-component barrel shape: a bare `export { default }` re-export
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks for your help 🙏 @mstijak

PR #994, opened 2026-07-29
perf: reduce import:false shared consumer chunk size

dmchoi77:
## Summary

Limit the generated ESM consumer module for `import: false` shared dependencies to the named exports confirmed by production module graph analysis. Existing `treeShaking.usedExports` values are included when configured.

Previously, every named export binding detected from the local package was generated even when the application used only a subset. This change removes unused bindings when static analysis is reliable, reducing the consumer chunk size.

No new user-facing option or Module Federation runtime contract change is introduced.

## Changes

- Extend the existing AST-based shared import analysis to `import: false` consumers.
- Union export usage across application modules, exposed modules, and explicit Rollup/Rolldown inputs.
- Remove unused named export bindings only after the production module graph completes successfully.
- Preserve the complete locally detected named-export surface when analysis is not reliable.
- Isolate parse state per federation plugin instance and reset it for every build.

## Behavior

| Usage | Generated result |
| --- | --- |
| Static named imports | Only the union of analyzed exports and configured `usedExports` |
| Namespace, star, or dynamic usage; incomplete or mismatched analysis | Complete locally detected named-export surface |
| Local export detection is unavailable | Existing default-only fallback and warning |

## Tests

- [x] `pnpm test` — 48 files, 990 passed, 1 skipped
- [x] `pnpm exec vitest run integration/shared-deps.test.ts` — 11 passed
- [x] `pnpm typecheck`
- [x] `pnpm build`

The added coverage verifies:

- retaining the used `init` binding while omitting the unused `value1`, `value2`, and `value3` bindings
- unioning export usage across exposes and explicit Rollup inputs
- preserving the complete export surface for namespace usage or incomplete analysis
- isolating parse state across federation plugin instances and resetting it across consecutive builds

## Risk

Export usage is finalized only after successful module graph completion. Incomplete or uncertain analysis retains the complete locally detected export surface.

### DIFF
--- a/integration/fixtures/shared-remote/exposed-namespace.js
+++ b/integration/fixtures/shared-remote/exposed-namespace.js
@@ -1,0 +1,3 @@
+import * as shared from 'mock-shared-dep';
+
+export const sharedNamespace = shared;
--- a/integration/fixtures/shared-remote/exposed-secondary.js
+++ b/integration/fixtures/shared-remote/exposed-secondary.js
@@ -1,0 +1,3 @@
+import { value2 } from 'mock-shared-dep';
+
+export const secondaryValue = value2;
--- a/integration/fixtures/shared-remote/vendor/mock-shared-dep/index.js
+++ b/integration/fixtures/shared-remote/vendor/mock-shared-dep/index.js
@@ -1,1 +1,4 @@
 export const init = 'ready';
+export const value1 = 'one';
+export const value2 = 'two';
+export const value3 = 'three';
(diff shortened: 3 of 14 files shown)


gioboa (commented):

Image

gioboa (approved):
That's simply amazing @dmchoi77 👏

dmchoi77:
@gioboa I found an edge case in this change and opened a follow-up: #995.

When the module graph contains externals, the parse barrier never completes, so the `import: false` export analysis is discarded on every build. 
The output is identical to pre-#994, but the build waits out `moduleParseIdleTimeout` before falling back — a fixed
+10s for no benefit. None of our fixtures declare an external, which is why CI stayed green.

The follow-up drains the barrier correctly, adds an integration test that fails on `main`, and makes the discarded-analysis path log a warning instead of staying silent.
Review would be appreciated.

PR #995, opened 2026-07-30
fix: complete import:false shared export analysis when the graph contains externals

dmchoi77:
Follow-up to #994.

## Summary

When a Rolldown graph contains external modules, #994's parse barrier never completes, so the export analysis is discarded on every build and the full named-export surface is emitted — while still paying the full `moduleParseIdleTimeout` (10s by default) waiting for it.

A silent regression: no error, no failed build, output byte-identical to pre-#994. Measured on a federated remote with 7 `import: false` shared deps:

|                              |   1.18.2 | #994 as merged |      This PR |
| ---------------------------- | -------: | -------------: | -----------: |
| Shared consumer chunks, gzip | 17.85 KB |       17.85 KB |  **8.46 KB** |
| Build time                   |    6.15s |         16.50s | **6.0–6.8s** |

## Root cause

`addPendingIds`, added in #994, seeds `parseStartSet` from `module.importedIds`.
It runs only when `importedIdResolutions` is unavailable — the Rolldown case;
otherwise the pre-existing `addPendingResolutions` path runs and filters on `!resolution.external`. 
The new path has no equivalent filter, because its guard assumes externals are absent from the module graph:

```ts
const moduleInfo = this.getModuleInfo(pendingId);
// External ids have no ModuleInfo. ...
if (moduleInfo && !excludeFn(pendingId)) {
  controller.parseStartSet.add(pendingId);
}
```

Rolldown returns a `ModuleInfo` stub for external ids and has no `isExternal` field at all (`rolldown@1.1.5`, `src/types/module-info.d.ts`). Externals pass the guard, are tracked as pending, and never reach `moduleParsed` under that id, so `parseStartSet` never drains.

No field can distinguish them at seed time — an external stub and a real not-yet-loaded module both report `code=null exports=0 importers=0 isEntry=false meta={}`. (The stub also carries `invalidate`, but that is `moduleOptionMap` bookkeeping that ordinary modules get too.)

**Why this now costs 10s.** Before #994, `parsePromise` was only consumed by `generateRemoteEntry`; nothing waited on it, so an undrained set was harmless and `buildEnd` resolved it. #994 made the `__loadShare__` `load` hook await it, which also puts `buildEnd` out of reach — it cannot fire while a `load` hook is blocked.
`moduleParseIdleTimeout` is the only exit left, and it resolves with `complete: false`, which discards the analysis anyway.

### Reproduction

Three conditions together:

1. `import: false` shared deps — otherwise no `load` hook awaits the barrier.
2. Rolldown — Rollup-style `importedIdResolutions` never reaches the new path.
3. An external imported by **more than one** module. `getModuleInfo` returns
   `null` the first time an external is seen, so a single importer never triggers this; the second importer's `moduleParsed` gets the stub.

Both `build.rollupOptions.external` and a `resolveId` hook returning `{ external: true }` satisfy (3). On the remote measured above the barrier reported `Tracked modules: 1367/1374` — 7 externals were enough to throw away an otherwise complete analysis.

## Changes

In `src/plugins/pluginModuleParseEnd.ts`:

- Track ids seeded from `importedIds` in a separate `speculativeSet`, promoted to confirmed as soon as `load` or `moduleParsed` observes them.
- When every outstanding id is still speculative, arm one additional quiet window (`SPECULATIVE_SETTLE_TIMEOUT`, 250ms) instead of completing or stalling. Any load/parse activity cancels it.
- If it elapses with the graph still quiet, those ids cannot be modules this build will parse — drop them and resolve `graph-complete`.

And in `src/index.ts`, warn once per build when the analysis is discarded for any remaining reason. The fallback is safe but was entirely silent — the only signal was the idle-timeout warning, whose text ("Some shared/remote dependencies may be missing") never mentions shared export analysis, which is why this went unnoticed:

```
[Module Federation] import: false shared export analysis was discarded
(reason: idle-timeout) — falling back to the complete export surface, so shared
consumers keep every detected named export. If the build is simply slow,
increasing moduleParseIdleTimeout may let the analysis finish.
```

No new user-facing option, no runtime contract change.

The race #994 guards against is preserved: the window only elapses after the whole graph has gone silent, and worst case degrades to the pre-#994 barrier rather than a 10s stall. 
Collapsing it into the existing 10ms settle debounce was tried and rejected — it breaks #994's own`tracks resolved imported ids when resolution metadata is unavailable` test.

## Tests

- [x] `pnpm fmt.check` · `pnpm typecheck` · `pnpm build`
- [x] `pnpm test` — 48 files, 993 passed, 1 skipped
- [x] `pnpm test:integration` — 14 files, 57 passed
- [x] `pnpm playwright test` — 23 passed

Added:

- `integration/shared-deps.test.ts` — exposed module importing both an `import: false` shared dep and an external that a second module also imports.
  Confirmed to fail on current `main` and pass with this change.
- `pluginModuleParseEnd.test.ts` — completing when the only pending ids are external stubs; not dropping a speculative id that loads within the window.
- `integration/shared-deps.test.ts` — a forced discard warns exactly once and still emits the complete export surface.

On the real build: consecutive builds are byte-identical with no idle-timeout warning (the window is timing-based, so this was checked explicitly), and no export is lost — the reduced consumer for the largest package keeps 18 of 1419 detected exports, covering all 12 bindings its bundled importers actually use(49.66 KB → 2.45 KB raw, 9.30 KB → 1.11 KB gzip).

### DIFF
--- a/integration/fixtures/shared-remote/exposed-with-external.js
+++ b/integration/fixtures/shared-remote/exposed-with-external.js
@@ -1,0 +1,8 @@
+import { init } from 'mock-shared-dep';
+import { externalHelper } from 'mock-external-dep';
+// A second importer of the same external: by the time this module is parsed the
+// external is already registered in the graph, so `getModuleInfo` returns a
+// stub for it instead of null.
+import { consumed } from './external-consumer.js';
+
+export const readyWithExternal = `${init}:${externalHelper}:${consumed}`;
--- a/integration/fixtures/shared-remote/external-consumer.js
+++ b/integration/fixtures/shared-remote/external-consumer.js
@@ -1,0 +1,3 @@
+import { externalHelper } from 'mock-external-dep';
+
+export const consumed = externalHelper;
--- a/integration/shared-deps.test.ts
+++ b/integration/shared-deps.test.ts
@@ -1,5 +1,5 @@
 import { resolve } from 'path';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import type { ModuleFederationOptions } from '../src/utils/normalizeModuleFederationOptions';
 import { isRollupChunk } from './helpers/assertions';
 import { buildFixture, FIXTURES } from './helpers/build';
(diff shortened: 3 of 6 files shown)


gioboa (commented):
I removed the heuristic in this commit [858d6dd](https://github.com/module-federation/vite/pull/995/commits/858d6ddb65d70fca890c651d421d86598c352764)
Let me know if you are ok with this change.

dmchoi77:
@gioboa Thanks for removing the heuristic — `resolved.external` is the right signal.
But I tested the commit against the project this was found in and the probe doesn't catch it: still 16.4–17.3s, consumer unchanged at 50856 B (expected 2512 B), discard warning on every run.

`resolveId` is first-wins, so the probe only sees resolutions that reach it.
There, a `pre` plugin ordered before ours returns `{ external: true }` for shared ids in the SSR graph, so `parseStart.resolveId` is never called for exactly those. I instrumented it: 136 probe hits, none from an SSR importer,
`externalSet` stays empty. Any plugin returning an external ahead of ours is invisible to an observer hook.

Repeating the resolution instead of observing it works, since `this.resolve` replays the whole chain (`skipSelf` only skips the caller). Probing at seed time in `addPendingIds`:

```ts
void this.resolve(pendingId, id, { skipSelf: true }).then((resolved) => {
  if (!resolved?.external) return;
  controller.externalSet.add(pendingId);
  controller.parseStartSet.delete(pendingId);
  scheduleParseCompletionCheck(controller);
});
```

Still no timer — the id keeps the barrier open until the resolution answers.
Once per pending id instead of wrapping every resolution, and the probe/`custom` machinery goes away: net -6 lines against your commit.

Verified: real project 5.98–6.25s, consumer 2512 B, 4 identical runs, no warnings; unit 994, integration 58 on vite@6/7/8, e2e 23, fmt + typecheck clean.
The two `parseStart.resolveId` calls in your unit tests had to go with the hook.

Pushed as a commit on top — happy to drop it if you'd rather take a different route.

gioboa (approved):
That's perfect @dmchoi77 
Your true test is the most valuable approval.

dmchoi77:
@gioboa Thanks! Challenging the heuristic is what got us to a cleaner fix.🚀

gioboa (commented):

Image

PR #997, opened 2026-07-31
chore: add automatic support for a DEV host consuming a PROD React remote

gioboa:
Adds automatic support for a DEV host consuming a PROD React remote. The Vite dependency optimizer safely guards React’s DEV-only dispatcher.getOwner() call when the production renderer does not provide it. Supports Rolldown and esbuild without modifying React inside node_modules.


gioboa (commented):

Image

PR #1000, opened 2026-08-03
fix: skip preloading dynamic-only remote exposes

### LINKED ISSUE #998: Does `version-first` intentionally preload unused remote expose modules, not just remoteEntry?
## Summary

With `@module-federation/vite@1.20.1` and the default `version-first` strategy, I observed that a remote expose module is requested during host bootstrap even when its dynamic import is never executed.

I would like to confirm whether loading the actual expose module at this point is intentional. I understand that `version-first` may load `remoteEntry.js` early for shared dependency negotiation, but I am not sure whether the same is expected for `remoteApp/App`.

## Reproduction

```ts
// routes.ts
export const loadRemoteApp = () => import('remoteApp/App');
```

```ts
// main.ts
import { loadRemoteApp } from './routes';

// Keep the loader reachable, but do not call it.
window.unusedRemoteRoute = loadRemoteApp;

document.querySelector('#app').textContent = 'unused-page-rendered';
```

Remote configuration:

```ts
federation({
  name: 'host',
  remotes: {
    remoteApp: {
      type: 'module',
      name: 'remoteApp',
      entry: 'http://localhost:3001/remoteEntry.js',
      entryGlobalName: 'remoteApp',
    },
  },
});
```

The page renders successfully, and `loadRemoteApp()` is never called.

## Actual behavior

The browser requests both the remote entry and the actual expose chunk during host bootstrap:

```text
GET /remoteEntry.js
GET /assets/App-*.js
```

The generated host bootstrap contains preload entries similar to:

```js
const __mfRemotePreloads = [
  __mfPreloadRemote('remoteApp', 'remoteApp'),
  __mfPreloadRemote('remoteApp/App', 'remoteApp/App'),
];

await Promise.allSettled(__mfRemotePreloads);
```

With multiple configured remotes, the configured remote entries are also requested during bootstrap, even when the current page does not use them.

## Question

Is it intentional for the expose module to be requested before the loader is executed?

The expose module is requested during host bootstrap, before this loader is called:

```ts
await loadRemoteApp();
```

Is this behavior intentional? If not, could remote entry preloading for `version-first` be separated from expose-module loading so that expose modules remain on-demand?

## Notes

The relevant code path is `src/plugins/pluginAddEntry.ts`, where the generated preload block calls `runtime.loadRemote()` for entries collected from `getUsedRemotesMap()`.

### Version

v1.20.1

### Validations

gioboa:
Close #998 

Prevents `version-first` hosts from preloading dynamic-only remote exposes while preserving remote initialization and static expose preloading.

### DIFF
--- a/integration/host-build.test.ts
+++ b/integration/host-build.test.ts
@@ -65,8 +65,9 @@ describe('host build', () => {
     expect(bootstrapAsset?.source).toContain('const { initHost } = __mfHostInit;');
     expect(bootstrapAsset?.source).toContain('const runtime = await initHost();');
     expect(bootstrapAsset?.source).toMatch(
-      /__mfPreloadRemote\("__mfe_internal__hostApp__mf_owner__\d+__remote1\/Module", "remote1\/Module"\)/
+      /__mfPreloadRemote\("__mfe_internal__hostApp__mf_owner__\d+__remote1", "remote1"\)/
     );
+    expect(bootstrapAsset?.source).not.toContain('remote1/Module');
     expect(bootstrapAsset?.source).toContain('runtime.loadRemote(runtimeRemote)');
     expect(bootstrapAsset?.source).toContain('})().then(() => __mfImport(');
     expect(bootstrapAsset?.source).toContain('globalThis.System.import(src)');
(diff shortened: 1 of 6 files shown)


gioboa (commented):

Screenshot 2026-08-03 at 22 35 19

PR #1001, opened 2026-08-03
fix: preserve public aliases for loadRemote

### LINKED ISSUE #999: loadRemote() fails with RUNTIME-004 in dev mode because the runtime's remote matcher never checks entryGlobalName
### Summary

In dev mode, `@module-federation/vite` rewrites a remote's `name` and `alias` fields to an internal namespaced string (`__mfe_internal__<host>__mf_owner__<n>__<remote>`), but keeps the original, user-configured name in `entryGlobalName`. `@module-federation/runtime-core`'s `matchRemoteWithNameAndExpose()`, the function `loadRemote()` depends on to resolve a request, only ever checks `remote.name` and `remote.alias`, never `entryGlobalName`. The practical effect: calling `loadRemote()` with the remote name exactly as documented fails with `RUNTIME-004: Failed to locate remote`, even though the remote is registered, running, and resolvable via a static `import()` of the same module in the same session.

### Reproduction

Minimal repro repo (host + one remote, verified against a fresh install): https://github.com/rparrish-5542/mf-loadremote-repro

```
cd remote && npm install && npm run dev   # :4174
cd host   && npm install && npm run dev   # :4173
```

Open http://localhost:4173. The page runs both loading strategies against the identical exposed module (`remote/util`):

- `import('remote/util')`, succeeds
- `loadRemote('remote/util')`, fails with `RUNTIME-004: Failed to locate remote`

Versions: `@module-federation/vite@1.20.1`, `@module-federation/runtime@2.8.0`, `vite@8.1.5`.

### What I checked before filing

- Our usage matches the documented API exactly. The docs' own canonical example (runtime-api.html#loadremote) registers a remote as `{ name: 'remote', ... }` and calls `loadRemote('remote/util')`, identical to this repro, so this doesn't look like a config mistake on our end.
- Confirmed in `@module-federation/vite`'s own source (`lib/index.js`) that the generated runtime registration sets `name`/`alias` via `getRuntimeRemoteAlias()` (which produces the `__mfe_internal__...` string), while `entryGlobalName` keeps the original plain name.
- Confirmed at runtime, not just in source: dumping `window.__FEDERATION__.__INSTANCES__[0].options.remotes` in the browser shows exactly this: `name`/`alias` renamed, `entryGlobalName: "remote"`.
- Confirmed `entryGlobalName` is never referenced anywhere in `@module-federation/runtime-core` (grepped the full installed package, zero matches), so there's no path by which the matcher could fall back to it.

### What I'm not certain about

I don't know whether this renaming in dev mode is intentional (`loadRemote()` may simply not be meant to be used against `@module-federation/vite` remotes in dev mode, for reasons not documented) or an oversight. I'm filing this as a mismatch between what one package writes and another reads, without assuming which side is "wrong."

### Related issues

- #311 has the same error code, but reads as several different root causes reported under one title (a build-plugin conflict, a production-only failure, a manual-registration workaround). I couldn't confirm the specific mechanism here was ever identified there, so I'm filing separately rather than asserting a duplicate.
- #675 (`getInstance()` returns null in dev although the instance exists globally) looks like a sibling case of the same underlying class: the internal `__mfe_internal__` rename breaking a public runtime API's ability to find something that demonstrably exists.
- Not the same as #970 (mf_owner generation mismatch on cold start, fixed by PR #973). That fix touches owner-id caching, not the name-matching function this issue is about.

gioboa:
Close #999 

Preserve public remote aliases so loadRemote() works in dev mode while internal names remain owner-scoped.

### DIFF
--- a/e2e/vite-webpack-rspack/tests/index.spec.ts
+++ b/e2e/vite-webpack-rspack/tests/index.spec.ts
@@ -43,6 +43,10 @@ test.describe('Vite Host Tests', () => {
     ]);
   });
 
+  test('loadRemote resolves a plugin-configured public alias', async ({ page }) => {
+    await expect(page.getByTestId('configured-load-remote')).toHaveText('loaded');
+  });
+
   test('test footer - vite host', async ({ page }) => {
     const productsHeading = page.getByRole('heading', {
       level: 3,
--- a/examples/vite-webpack-rspack/host/src/App.tsx
+++ b/examples/vite-webpack-rspack/host/src/App.tsx
@@ -79,6 +79,10 @@ const App = () => {
   const [randomBanner, setRandomBanner] = React.useState<'SignUpBanner' | 'SpecialPromo'>(
     'SpecialPromo'
   );
+  const RuntimeModuleRemoteProduct = useDynamicImport({
+    module: 'Product',
+    scope: 'moduleRemote',
+  });
 
   const Banner = useDynamicImport({
     module: randomBanner,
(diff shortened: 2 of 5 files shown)


gioboa (commented):

Screenshot 2026-08-03 at 23 20 42

PR #1004, opened 2026-08-04
fix: contain side effects of shared export detection

### LINKED ISSUE #1003: [Bug]: shared export detection require()s browser-condition entries in Node, leaking handles so `vite build` never exits
### Describe the bug

`getRequiredNamedExports` enumerates a shared dependency's named exports by **`require()`-ing that dependency inside the Node build process**, after `getPackageNamedExports` resolves its entry with **browser-first conditions** (`['browser', 'import', 'module', 'default']` — note `node` is absent).

Executing browser-targeted code in Node means the module's top-level side effects land in the build process and are never cleaned up. For `react-dom/server` this is fatal: the browser entry's CJS body runs `var channel = new MessageChannel()` at module scope (`react-dom/cjs/react-dom-server.browser.production.js:132`) and never closes it. That one leaked `MessagePort` keeps the libuv event loop permanently non-empty, so **`vite build` prints `✓ built in Xs`, writes a complete and correct `dist/`, and then hangs forever** — CI jobs hang until a timeout kills them.

Chain:

```
writeLoadShareModule()                     src/virtualModules/virtualShared_preBuild.ts:1509
  -> getPackageNamedExports()              :561   resolves with ['browser','import','module','default']
    -> getRequiredNamedExports()           :545   require()s the resolved entry
      -> require('react-dom/server.browser')
        -> react-dom/cjs/react-dom-server.browser.production.js:132
           var channel = new MessageChannel()      // module scope, never closed
```

**Two conditions gate it, and both are common:**

1. The `require()` path runs **only when `shareConfig.import === false`**.
2. `COMMON_SHARED_SUBPATHS` (`src/utils/pathNormalization.ts:5`) auto-expands `react-dom` to `['react-dom/client', 'react-dom/server', 'react-dom/server.browser']`, so a project never has to name `react-dom/server` itself.

So **any React project sharing `react-dom` with `import: false` is affected deterministically.** In our case that is ~19 federated modules.

**Why this is not a Vite or rolldown bug** (I originally misfiled it as one, and have corrected/closed that report): `process.getActiveResourcesInfo()` after `build()` resolves shows exactly one surviving non-stdio handle — `MessagePort` — and an `async_hooks` creation-stack trace names `getRequiredNamedExports` as its origin. It reproduces identically on `vite@8.1.5`/`rolldown@1.1.5` and `vite@8.2.0`/`rolldown@1.2.2`, because neither is the cause.

**Note on the obvious-looking fix:** resolving with `node` conditions for the execution step is *not* correct, and your existing comment at `virtualShared_preBuild.ts:562-564` already anticipates why. I measured the difference:

- `react-dom/server.browser` → `[version, renderToString, renderToStaticMarkup, renderToReadableStream, resume]`
- `react-dom/server.node` → additionally `renderToPipeableStream`, `resumeToPipeableStream`

Switching conditions would emit Node-only names into a browser bundle where the host provides the browser build, so they would resolve to `undefined` at runtime. Any fix needs to keep the browser entry's export shape while containing the side effects.

**Adjacent observation while measuring:** detection appears unmemoized — one of our builds performs **20** `require()` calls for only **6 distinct** modules.

I intend to submit a PR for this and would appreciate a maintainer's steer on which containment approach you'd accept, since it affects the shape of the change:

- **(a)** snapshot active handles around the `require()` and `unref()` newly created ones — ~6 lines, stays synchronous, and I have verified end-to-end that it fixes the hang with byte-identical output. Downside: it needs `process._getActiveHandles()`, an undocumented API, and would be the first internal-API use in this codebase (which today uses only documented builtins).
- **(b)** run the `require()`-based detection in a short-lived `execFileSync` child process — documented API only, and contains *all* side-effect classes rather than just unref-able handles. Costs ~60ms per package, which memoizing (see above) brings down to ~6 spawns.
- **(c)** extend the existing static analysis to follow single-target CJS re-exports so the module never has to be executed — avoids execution entirely, but is the largest and most fragile change.

Happy to implement whichever you prefer. Note `writeLoadShareModule` is synchronous, which rules out worker-thread isolation without an async refactor of `resolveId`.

### Version

@module-federation/vite 1.19.1 (also present in 1.20.0 source); vite 8.1.5 and 8.2.0

### Reproduction

No standalone repo yet (the original is a proprietary app), but the defect reproduces with **no Vite, no rolldown and no plugin involved at all** — this is exactly what `getRequiredNamedExports` does, and it is sufficient to demonstrate the root cause:

```bash
npm i react-dom@19

# what the plugin resolves and executes today -> leaks a MessagePort, hangs forever
node -e "require('react-dom/server.browser'); console.log('exports read, now hanging')"

# the Node-safe entry, for contrast -> exits 0 immediately
node -e "require('react-dom/server.node'); console.log('exports read, exits fine')"
```

To see it via the plugin: any Vite 8 + `@module-federation/vite` remote that shares `react-dom` with `import: false`, then `vite build`. The build completes and writes correct output, then never exits.

I can put together a minimal public repro repo if that would help triage — say the word.

### Relevant log output

```shell
vite v8.1.5 building client environment for production...
transforming...✓ 3381 modules transformed.
rendering chunks...
computing gzip size...
... (full, correct dist/ written) ...
✓ built in 4.55s
# <- process never exits from here; must be killed manually

# handle introspection at this point:
$ node --input-type=module -e "
  import {build} from 'vite';
  await build();
  console.log(process.getActiveResourcesInfo());
"
[ 'PipeWrap', 'PipeWrap', 'MessagePort' ]
#                          ^^^^^^^^^^^ leaked by the introspection require();
#                                      PipeWrap is stdio and predates the build
```

### Validations

Gmendesfonseca:
Close #1003

## Summary

`getRequiredNamedExports` reads a package's export names by requiring it, which runs that package's top-level code inside the build process — and `getPackageNamedExports` deliberately resolves the **browser** entry (`['browser', 'import', 'module', 'default']`, no `node`). A browser entry may open a handle Node never closes: `react-dom/server.browser` holds a module-scope `MessageChannel`. One ref'd handle keeps the event loop alive forever, so `vite build` writes a complete, correct bundle, prints `✓ built in Xs`, and then never exits.

`COMMON_SHARED_SUBPATHS` auto-shares `react-dom/server`, and this path only runs when `shareConfig.import === false`, so any project sharing `react-dom` with `import: false` hits it deterministically. Reproducible with no Vite and no plugin involved:

```
node -e "require('react-dom/server.browser')"   # leaks a MessagePort, hangs
node -e "require('react-dom/server.node')"      # exits 0
```

This unrefs whatever the require created. Safe here because the module is loaded purely to read `Object.keys` off it and is never used afterwards.

**Export names are unchanged.** Switching the execution to the `node` condition would have been the smaller-looking fix, but the export lists genuinely differ — `server.node` adds `renderToPipeableStream` and `resumeToPipeableStream`, which the browser build the host provides does not have, so they would resolve to `undefined` at runtime. The existing comment at `virtualShared_preBuild.ts:562-564` already anticipates this, so the browser-first resolution is preserved as-is.

## Changes

- Bracket the introspection `require()` with a handle snapshot and unref handles created during it. The cleanup is in `finally`, so a package that throws part-way through its side effects is also cleaned up.
- `process._getActiveHandles` is undocumented, so it is called behind a `typeof` guard — if it is ever absent, behaviour degrades to exactly what it is today rather than failing the build.
- Four tests covering: no handle survives a side-effectful require, export names unchanged, a throwing package returns `undefined` and leaves no handle, and the guarded path still reports names.

The test fixture opens a **real** `MessageChannel` and assigns `port1.onmessage`, because a bare `MessageChannel` is not loop-keeping — without the `onmessage` assignment the test would pass vacuously.

## Verification

| Gate | Result |
| --- | --- |
| `pnpm fmt.check` | pass |
| `pnpm typecheck` | pass |
| `pnpm test` | 927 passed / 1 skipped — baseline was 923/1, so +4 and no existing test changed, skipped or removed |
| `pnpm build` | pass |
| `pnpm test:integration` | 51 passed (13 files) |
| `pnpm playwright test` | 4 pre-existing failures in `multi-example`, identical set to an unmodified baseline on this machine across repeated runs — no change from this PR |

Mutation-checked rather than assumed: removing the unref loop fails 2 of the 4 new tests; moving the cleanup out of `finally` onto the success path fails exactly the throwing-package test.

End to end, on a real federated remote (`vite@8.1.5`, `rolldown@1.1.5`) that hangs indefinitely without this change, `vite build` exits 0 in 3.8s with byte-identical `dist/` output.

## Notes for reviewers

- **Known limitation, deliberately not addressed here:** this contains handles only. A dependency whose side effect is not a handle — an `exit` listener, a global mutation — is still not contained. Executing third-party module code in the build process to enumerate exports remains the underlying architectural risk.
- I costed two alternatives before choosing this one and can switch if you prefer either. A throwaway `execFileSync` child process is stricter (documented API, contains every side-effect class) and I verified it returns identical names for all six packages a real build introspects at ~165ms total — but it cannot see `vi.mock`, so the ~38 virtual mock packages in `virtualShared_preBuild.test.ts` become unresolvable and roughly 42 existing assertions fail; fixing that means migrating the harness, which seemed disproportionate for a hang fix. (If you do revisit it: a naive child **also** hangs — it needs `fs.writeSync(1, …)` then `process.exit(0)`, since stdout-to-pipe is async.) `cjs-module-lexer` avoids execution entirely but adds a production dependency to a package that currently has three.
- Unrelated observation while measuring, not touched here: detection looks unmemoized — one build performed 20 `require()` calls for 6 distinct modules. Happy to open a separate PR.


gioboa (commented):

Screenshot 2026-08-04 at 09 08 06

gioboa (approved):
Thanks @Gmendesfonseca 
this is a great fix 👏

smeng9:
@gioboa 

Glad that the issue mentioned in the post I sent to you is fixed here.

> That same confusion produced our most baffling symptom: a build that printed "built successfully" and then never exited. The work was finished; the process simply refused to end. ... We found it because the AI agent harness doing the migration had itself been silently stuck for most of a day

It is also something https://github.com/react/react/pull/36918 we are waiting from react team

PR #1005, opened 2026-08-04
fix: resolve host shared modules in dev SSR

dmchoi77:
## Summary

This change fixes an issue where shared modules in Vite 8 dev SSR can be resolved from the remote environment instead of the host environment.

## Problem

When the host `ModuleRunner` requests modules through the remote `/__mf_runner__` endpoint, shared modules such as `react` may also be resolved using the remote file system.

This can cause the host and remote to use different shared module paths or reuse a runner configured for another host when the same remote is consumed by multiple hosts.

## Changes

- Resolve modules present in `resolvedShared` from the host without forwarding them to the remote runner
- Include `resolvedShared` in the ModuleRunner cache key to isolate runners per host
- Reuse the existing `EXTERNAL_URL_RE` utility
- Add regression coverage to ensure shared module paths are isolated between hosts

## Compatibility

- No public API changes
- Affects only the Vite 8 cross-process dev SSR path
- The existing support scope for dev SSR transport on Vite versions below 8 remains unchanged

## Verification

- `pnpm fmt.check`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `pnpm exec vitest run integration/dev-singleton-react.test.ts`

### DIFF
--- a/src/utils/__tests__/ssrEntryLoader.test.ts
+++ b/src/utils/__tests__/ssrEntryLoader.test.ts
@@ -1199,24 +1199,28 @@ describe('ssrEntryLoaderPlugin — code transformation', () => {
  * freshLoaderWithRunner — like freshLoader but configures Node's createRequire
  * so per-test ModuleRunner behaviour can be injected.
  */
+type RunnerOptions = {
+  transport: {
+    invoke: (payload: unknown) => Promise<unknown>;
+  };
+};
+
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

gioboa (approved):
Thanks @dmchoi77 LGTM 🚀

PR #1006, opened 2026-08-04
chore: pin dep versions

gioboa:


gioboa (commented):

Screenshot 2026-08-04 at 18 08 48

PR #1007, opened 2026-08-04
fix: negotiate shared subpaths in SSR vm strategy

dmchoi77:
### Summary

Fixes shared subpath negotiation in the SSR `vm` strategy when only the parent package is configured as shared.

### Root Cause

`ssrVmStrategy` only checked `instance.options.shared[specifier]` for the exact imported specifier.

With a host configuration such as:

```ts
shared: {
  react: { singleton: true },
}
```

a remote import of `react/jsx-runtime` was not recognized as shared and fell back to `resolvedShared` or a regular host import.

### Changes

- Prefer exact shared key matches.
- Match common shared subpaths such as `react/jsx-runtime` and `react-dom/client` through their parent shared package.
- Support Vue aliases and wildcard shared configurations.
- Pass the original subpath specifier to `loadShare()` so subpath exports are preserved.
- Preserve the existing fallback order when no shared provider is available.
- Leave the `temp-file` strategy and Vite 8 `ModuleRunner` path unchanged.
- No public API changes.

The VM strategy uses a minimal matcher to avoid pulling the entire Vite plugin module into the SSR bundle.

### Verification

- `pnpm fmt.check`
- `pnpm typecheck`
- `pnpm exec vitest run src/utils/__tests__/ssrVmStrategy.test.ts`
- `pnpm exec vitest run src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts`
- `pnpm test`
- `pnpm exec vitest run integration/ssr-shared-exports.test.ts`
- `pnpm build`

### Scope and Risk

- The change only affects SSR environments using `strategy: 'vm'`.
- The existing `temp-file` strategy and CSR behavior are unchanged.


gioboa (approved):
👍 nice fix
Thank you @dmchoi77

PR #1008, opened 2026-08-04
fix: shared export scan should correctly resolve nested template literals

smeng9:
`hasTopLevelDeclaratorComma` treated a backtick as a plain quote, so a template literal nested inside a `${...}` interpolation closed the outer template early. Everything after that point was scanned as code: brace depth drifted, and a `/` sitting in template text was read as the start of a regex, which then never closes and reports a phantom top-level comma.

A phantom comma marks the whole export scan incomplete, so `getSharedNamedExports` returns undefined. Since #918 gated `usesDeferredSingletonFallback` on `hasCompleteExportCoverage`, losing coverage silently downgrades a `singleton: true` share to a locally pre-built copy. The remote then never joins the host's share scope and the library is instantiated twice, which breaks any module-level state or context the singleton was meant to guarantee.

This affects any shared package whose entry transitively re-exports a module that builds markup from nested template literals, and it has been present in every version since 1.17.1.

Track template text and interpolations on an explicit stack so the scanner stays in sync, and ignore commas and semicolons that appear inside an interpolation.

Minimal reproduction:

    const body = 'text';
    export const injectedHtml = `${`<div>`}${body}${`</div>`}`;

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -210,6 +210,9 @@ vi.mock('../../utils/packageUtils', () => ({
     if (pkg === 'mock-package-cjs-star-entry') {
       return '/repo/apps/remote/node_modules/mock-package-cjs-star-entry/index.mjs';
     }
+    if (pkg === 'mock-package-nested-template-literal') {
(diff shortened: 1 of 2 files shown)


smeng9:
getSharedNamedExports needs to report back correct shared exports for the singleton

gioboa (approved):
Thanks for this fix @smeng9 🙏 appreciated

PR #1010, opened 2026-08-05
fix: detect multi-declarator named exports

### LINKED ISSUE #1009: [Bug]: Prion-like propagation of faulty federated module's shared dependencies that infect otherwise healthy modules.
### Describe the bug

As a standard practice for module-federation, a package with react-context is generally shared as a singleton. If there are multiple copies, accessing the context will raise issues like useXXX hook must used within a XXXProvider.

Our application has multiple federated modules, and each is mounted under a /\<path\> using react-router and is supposed to be self contained besides using shared context. One of our shared package publishes version number variables in an inlined manner instead of splitting into multiple exports. This seems cannot be resolved correctly past changes in #918 gets adopted.

```
export const major=5, minor=2, patch=0;
```

We built one of our remote modules using vite plugin 1.20.2 and gated it behind a feature flag for grey testing internally. We noticed not only the issue causing the related app to crash (expected because getSharedNamedExports does not report back correctly so singleton resolution failed), but also other frontend apps built prior to #918 and even apps not built using vite ecosystem start crashing, if we first visit the affected page, then navigate to some other pages (because user can still use app bar to return home).

We dug a little bit more into this prion behavior. You can find a minimal reproduction that a shared counter is not incremented correctly.

Even though we can fix one issue at a time, we probably need the core team to look a bit more into a better architecture to catch a group of issues. Particularly, if host is already having a singleton copy matching the all the dependency and singleton requirements of a remote, it is imperative for the host to write it to the shared cache and force the remote to use it.

### Version

v8.1.5

### Reproduction

https://stackblitz.com/edit/2d2nzmlq?file=README.md&terminal=dev

### Relevant log output

```shell
No log output. This is silent.
```

### Validations

gioboa:
Close #1009

Detects all named exports declared in a single multi-declarator statement, preserving singleton sharing.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -1867,7 +1867,7 @@ describe('writeLoadShareModule', () => {
     expect(generatedCode).not.toContain('export * from "/repo/packages/generic-initializer.ts"');
   });
 
-  it('still detects a real declarator comma after a generic initializer', () => {
+  it('enumerates a later declarator after a generic initializer', () => {
     const pkg = 'mock-package-with-reserved';
     const mockShareItem: ShareItem = {
       name: pkg,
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Screenshot 2026-08-05 at 09 42 16

PR #1013, opened 2026-08-05
fix: dynamically load eager shared map

### LINKED ISSUE #1011: [Bug]: eager shared drops the __mfLocalSharedImportMap namespace import from remoteEntry (ReferenceError at startup, build exits 0)
### Describe the bug

When at least one `shared` entry has `eager: true`, `generateRemoteEntry()` emits a static namespace import and returns it:

```js
// lib/index.js (1.20.2, around L4196 / L4249 / L4284)
const hasEagerShared = Object.values(options.shared ?? {}).some(
  (share) => share?.shareConfig.eager === true && share.shareConfig.import !== false
);
...
${hasEagerShared ? `import * as __mfLocalSharedImportMap from "${getLocalSharedImportMapPath(options)}";` : ""}
...
async function getLocalSharedImportMap() {
  ${hasEagerShared ? "return __mfLocalSharedImportMap;" : ""}
  ...
}
```

If Rollup hoists the Module Federation runtime into its own shared chunk, the emitted `remoteEntry.js` **keeps `return __mfLocalSharedImportMap` but loses the import that binds it**, and stops referencing the `_virtual_mf-localSharedImportMap…` chunk entirely.

`dist/remoteEntry.js` from the reproduction (built with `minify: false`):

```js
import { _ as __vitePreload } from './assets/preload-helper-CWZBUsdZ.js';
import { h as helpers, i as init$1 } from './assets/index-GcC7LmTc.js';
//                                ^ no `_` binding; the _virtual_mf-localSharedImportMap…
//                                  chunk is emitted but never imported here
...
async function getLocalSharedImportMap() {
    return __mfLocalSharedImportMap;   // free, undeclared identifier
}
```

When it works (same config, MF runtime *not* hoisted), all three bindings arrive together from the `localSharedImportMap` chunk:

```js
import { h as helpers, i as init$1, _ as __mfLocalSharedImportMap }
  from './assets/_virtual_mf-localSharedImportMap___mfe_internal__hostApp__mf_owner__1-CiudmNFF.js';
```

**Three conditions are required.** Removing any one keeps the MF runtime inside the `localSharedImportMap` chunk and the binding survives:

1. at least one `shared` entry with `eager: true`
2. two or more build entries (`build.rollupOptions.input`)
3. application source importing `@module-federation/runtime` directly

**Why this is easy to miss:** the build exits `0` with no Rollup warning, and the dev server is unaffected — it only happens in `vite build`. Since `__mfLocalSharedImportMap` becomes an undeclared global, esbuild's minifier cannot rename it, so the identifier survives minification unchanged. CI that only runs a build cannot catch it; the app dies in the browser at bootstrap (`hostInit` → `Module.init`).

We hit this in a production Vue host app; the reproduction linked below is a clean minimal project, not that app.

Affected versions:

| `@module-federation/vite` | result |
| --- | --- |
| 1.16.1 | not affected (no `hasEagerShared` code path yet) |
| 1.17.0 | reproduces |
| 1.19.0 | reproduces |
| 1.20.0 | reproduces |
| 1.20.2 (latest) | reproduces |

`hasEagerShared` was introduced in 1.17.0.

Workaround: set `eager: false` on all `shared` entries.

I have not written a fix, so I am not claiming a PR — but if the intended contract is "the eager path must be self-contained", making `getLocalSharedImportMap()` fall back to the dynamic `import()` form (or keeping the namespace binding alive across chunking) both look viable.

### Version

vite 7.3.3, `@module-federation/vite` 1.20.2, `@module-federation/runtime` 2.8.1, Node 22.18.0, macOS

### Reproduction

https://github.com/oth54477/mf-vite-eager-shared-repro

```bash
npm install
npm run build
npx vite preview --port 5199
# open http://localhost:5199, check the browser console
```

Static check without a browser:

```bash
node -e '
const s = require("fs").readFileSync("dist/remoteEntry.js", "utf8");
console.log("uses the identifier :", s.includes("__mfLocalSharedImportMap"));
console.log("imports it          :", /import\s*\*\s*as\s*__mfLocalSharedImportMap/.test(s));
console.log("references the chunk:", s.includes("localSharedImportMap"));
'
# broken -> true / false / false
# fine   -> true / true  / true
```

### Relevant log output

```shell
# vite build -> exit code 0, no warnings

# browser console on load:
ReferenceError: __mfLocalSharedImportMap is not defined
    at getLocalSharedImportMap (http://localhost:5199/remoteEntry.js:215:5)
    at Module.init (http://localhost:5199/remoteEntry.js:361:45)
    at http://localhost:5199/assets/hostInit-CM6Ia1xK.js:86:45
    at async http://localhost:5199/assets/mf-entry-bootstrap-0-1989a01d.js:32:3
```

### Validations

gioboa:
Close #1011 

Prevent missing eager-share bindings in multi-entry production builds.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1120,7 +1120,7 @@ describe('virtualRemoteEntry', () => {
     expect(code).not.toContain('import {usedShared, usedRemotes} from');
   });
 
-  it('statically includes the local shared map when a local eager share is configured', async () => {
+  it('loads the local shared map dynamically when a local eager share is configured', async () => {
     const mod = await import('../virtualRemoteEntry');
     const eagerShare = {
       name: 'eager-shared',
@@ -1146,10 +1146,9 @@ describe('virtualRemoteEntry', () => {
     );
 
     expect(code).toMatch(
-      /import \* as __mfLocalSharedImportMap from "virtual:mf-localSharedImportMap:__mfe_internal__host__mf_owner__\d+";/
+      /retrySharedInit\(\(\) => import\("virtual:mf-localSharedImportMap:__mfe_internal__host__mf_owner__\d+"\)\)/
     );
-    expect(code).toContain('return __mfLocalSharedImportMap;');
-    expect(code).not.toContain('retrySharedInit(() => import("virtual:mf-localSharedImportMap');
+    expect(code).not.toContain('__mfLocalSharedImportMap');
   });
 
   it('patches server-calc providers from Snapshot and coverage-caches runtime selections', async () => {
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -934,9 +934,6 @@ export function generateRemoteEntry(
   const hasTreeShakingShared = Object.values(options.shared ?? {}).some(
     (share) => !!share?.shareConfig.treeShaking
   );
-  const hasEagerShared = Object.values(options.shared ?? {}).some(
-    (share) => share?.shareConfig.eager === true && share.shareConfig.import !== false
-  );
   const hasMultipleShareScopes = Array.isArray(options.shareScope);
   const runtimeImports = [
     'init as runtimeInit',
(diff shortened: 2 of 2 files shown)


gioboa (commented):

Image

PR #1015, opened 2026-08-05
fix: skip unavailable React DOM client

### LINKED ISSUE #1012: [Bug]: dev server fails with `Failed to resolve import "react-dom/client"` when sharing react-dom on React 17
### Describe the bug

When `react-dom` is declared in `shared` and the installed React version is **17** (which has no
`react-dom/client` entry — it was introduced in React 18), `vite dev` crashes as soon as the page
is loaded:

```
Internal server error: Failed to resolve import "react-dom/client" from
"virtual:mf:__mfe_internal__<name>__mf_owner__1__prebuild__react_mf_2_dom_mf_1_client__prebuild__.js".
Does the file exist?
Plugin: vite:import-analysis
```

`vite build` / `vite preview` are unaffected, because the subpath expansion only runs in serve mode.

### Root cause

In serve mode, sharing `react-dom` automatically expands to the hardcoded `COMMON_SHARED_SUBPATHS`
(`react-dom/client`, `react-dom/server`, `react-dom/server.browser`) — see
`src/utils/pathNormalization.ts:3`.

In the expansion loop in `src/index.ts` (~lines 651–672), `canResolveSharedSubpath(subpath, root)`
is called and correctly returns `false` for `react-dom/client` on React 17 — but the result is only
used to decide `optimizeDeps.include` vs `optimizeDeps.exclude`. `writePreBuildLibPath(subpath,
shareItem, options)` is still called **unconditionally**, so a prebuild virtual module
containing a bare

```js
import * as __mfPrebuildNamespace from "react-dom/client";
```

is emitted, and `vite:import-analysis` fails to resolve it.

There is already a guard for exactly this situation, but it is special-cased to a single subpath
(line 653):

```ts
if (subpath === 'react/compiler-runtime' && !canResolveSubpath) {
  optimizeDeps.exclude.push(subpath);
  continue;
}
```

### Suggested fix

Generalize the `react/compiler-runtime` special case to every unresolvable common subpath:

```ts
if (!canResolveSubpath) {
  optimizeDeps.exclude.push(subpath);
  continue;
}
```

i.e. skip `getLoadShareModulePath` / `writeLoadShareModule` / `writePreBuildLibPath` /
`addUsedShares` for any subpath of `COMMON_SHARED_SUBPATHS` that doesn't resolve in the project.
A subpath that doesn't exist in the installed package version can't be provided as a shared module
anyway, so registering it only produces a broken prebuild wrapper. 

### Workaround (for anyone else hitting this)

A serve-only alias makes the prebuild wrapper resolve into the shared `react-dom` singleton:

vite.config
```ts
resolve: {
  alias: command === 'serve' ? { ...alias, 'react-dom/client': 'react-dom' } : alias,
}
```

## Versions
Vite 5.4.21, @module-federation/vite 1.20.1, React 17.0.2, Node 20.19.1

### Version

Vite 5.4.21 (repro is on 8.2.0 and does not work either)

### Reproduction

https://stackblitz.com/edit/vitejs-vite-kgtmhwxs?file=README.md

### Relevant log output

```shell

```

### Validations

gioboa:
Close #1012

Prevents Vite dev failures on React 17 by skipping the nonexistent react-dom/client prebuild module. React 18 and 19 remain unchanged.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -650,11 +650,13 @@ export default __mfShared.default ?? __mfShared;`,
             }
             for (const subpath of getCommonSharedSubpaths(key)) {
               const canResolveSubpath = canResolveSharedSubpath(subpath, root);
-              if (subpath === 'react/compiler-runtime' && !canResolveSubpath) {
-                // react/compiler-runtime is only exported by newer React
-                // versions. Registering it for React 18 can make a bare import
-                // resolve to an unrelated React version elsewhere in a pnpm
-                // workspace and serve its CommonJS entry directly.
+              if (
+                ['react/compiler-runtime', 'react-dom/client'].includes(subpath) &&
+                !canResolveSubpath
+              ) {
+                // These entry points only exist in newer React versions.
+                // Generating their prebuild wrappers for older versions creates
+                // imports that Vite cannot resolve.
                 optimizeDeps.exclude.push(subpath);
                 continue;
               }


gioboa (commented):

Image

PR #1017, opened 2026-08-05
fix: derive runtime plugin version from dependency

dmchoi77:
## Summary

- Derive the external runtime plugin version from the installed `@module-federation/runtime` package metadata.
- Keep `plugin.version` and `_FEDERATION_RUNTIME_CORE_FROM.version` aligned automatically when the runtime dependency is upgraded.
- Add regression coverage for matching runtime provenance and avoid false multiple-runtime warnings.

## Problem

`injectExternalRuntimeCorePlugin` used a hard-coded `2.8.0` version while the project depended on `@module-federation/runtime@2.8.1`. This could make compatible runtimes appear to have different versions and trigger an unnecessary multiple-runtime warning.

## Changes

- Read `@module-federation/runtime/package.json` directly from the runtime plugin.
- Enable JSON module resolution in TypeScript.
- Replace the stale test expectation with the installed runtime version.
- Verify that a provider reporting the installed version does not emit a warning.

## Validation

- `pnpm test`
- `pnpm test:integration`
- `pnpm typecheck`
- `pnpm build`
- `pnpm fmt.check`

### DIFF
--- a/src/utils/__tests__/injectExternalRuntimeCorePlugin.test.ts
+++ b/src/utils/__tests__/injectExternalRuntimeCorePlugin.test.ts
@@ -1,5 +1,6 @@
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import * as runtimeCore from '@module-federation/runtime/core';
+import { createRequire } from 'node:module';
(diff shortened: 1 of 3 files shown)


gioboa (approved):
Nice catch 👏 Thanks @dmchoi77

PR #1019, opened 2026-08-05
fix: preserve ssrExternals during normalization

dmchoi77:
## Summary

- Preserve the public `ssrExternals` option when normalizing Module Federation options.
- Ensure user-configured SSR-only packages reach `pluginSSRRemoteEntry`.
- Add regression coverage for the full raw-options → normalization → SSR plugin path.

## Problem

`ssrExternals` was declared as a public option but omitted from the normalized options object. Since the SSR plugin receives only normalized options, packages configured by users were not added to the SSR external list.

For example, configuring `ssrExternals: ['sharp']` could still allow `sharp` or a server-only SDK to be bundled into the SSR remote entry, depending on the Vite/Rollup configuration.

## Changes

- Copy `options.ssrExternals` into the normalized options object.
- Add a normalization test for user-provided `ssrExternals`.
- Update the SSR plugin test to exercise the public configuration path.
- Verify configured package subpaths are externalized and unconfigured packages retain the default behavior.

The existing SSR external pattern already matches both `sharp` and `sharp/subpath`, so no regex change was needed.

## Validation

- `pnpm exec vitest run src/utils/__tests__/normalizeModuleFederationOption.test.ts src/plugins/__tests__/pluginSSRRemoteEntry.test.ts`
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- `pnpm fmt.check`

### DIFF
--- a/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
@@ -304,9 +304,29 @@ describe('pluginSSRRemoteEntry', () => {
     });
 
     it('externalises user-provided ssrExternals', () => {
-      const base = makeOptions();
-      const options = { ...base, ssrExternals: ['my-server-only-pkg'] };
-      const plugins = pluginSSRRemoteEntry(options);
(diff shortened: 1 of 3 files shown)


gioboa (approved):
Great catch and clean fix @dmchoi77 💪

PR #1020, opened 2026-08-05
fix: normalize publicPath for dev remote entry URL

dmchoi77:
## Summary

- Normalize path-based `publicPath` values before composing the dev remote entry URL.
- Resolve absolute `http://` and `https://` public paths without prepending the host origin.
- Add regression coverage for slash variants, absolute CDN URLs, and hash-based filenames.

## Problem

When `publicPath` does not end with `/`, the dev host auto-init code concatenates it directly with the remote entry filename. For example, `publicPath: '/cdn'` produces `/cdnremoteEntry.js` instead of `/cdn/remoteEntry.js`.

Absolute public paths have a related problem: `origin` is prepended to values such as `https://cdn.example.com/assets`, producing an invalid import URL.

## Changes

- Keep the original `publicPath` value unchanged for manifest metadata.
- Normalize path-based dev URLs with `ensureTrailingSlash()` before appending the remote entry filename.
- Resolve absolute public paths with the URL API and preserve their origin.
- Cover `/cdn`, `/cdn/`, `/`, absolute CDN URLs, and `remoteEntry-[hash]` in `pluginProxyRemoteEntry` tests.

## Validation

- `pnpm exec vitest run src/plugins/__tests__/pluginProxyRemoteEntry.test.ts src/utils/__tests__/pathNormalization.test.ts`
- `pnpm test`
- `pnpm typecheck`
- `pnpm exec oxfmt --check src/plugins/pluginProxyRemoteEntry.ts src/plugins/__tests__/pluginProxyRemoteEntry.test.ts`
- `pnpm build`

### DIFF
--- a/src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
@@ -150,7 +150,7 @@ describe('pluginProxyRemoteEntry', () => {
   it('uses a dev-safe filename for hash-pattern host init remote entry imports', async () => {
     normalizeModuleFederationOptions({ name: 'test' });
     const plugin = pluginProxyRemoteEntry({
-      options: getDefaultMockOptions({ filename: 'remoteEntry-[hash]' }),
+      options: getDefaultMockOptions({ filename: 'remoteEntry-[hash]', publicPath: '/cdn' }),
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks @dmchoi77 
Clean, focused fix with solid test coverage. 🚀

PR #1021, opened 2026-08-05
fix: reuse compatible non-singleton shared fallbacks

### LINKED ISSUE #1018: [Bug]: Non-singleton shared deps with a bundled fallback are downloaded once per remote, even when every container registers the same version
### Describe the bug

A workspace package is shared by the host and every remote as `singleton: false` with a bundled fallback (no `import: false`). All containers register the identical version:

```
// host + each remote
shared: {
  'my-ui-lib': { singleton: false, version: '1.2.3', requiredVersion: '^1.2.3' },
}
```

We expected version negotiation to resolve every container to one provider, so the implementation downloads once.

What actually happens is every remote that initializes downloads its own full copy of the fallback. In our app the shared UI library is about 1 MB, and a session that renders 1 remote downloads copies of the library for all 7 remotes.

From reading `lib/index.js` and the generated bundles, three places combine to force the download:

1. The deferred fallback template (`generateLazyWorkspaceSingletonExports`), which keeps the implementation in its own chunk and only `import()`s it after init on a cache miss, is gated to `singleton === true` (`usesDeferredSingletonFallback`). Non-singleton fallbacks get the eager template instead, so the full implementation is compiled into the loadShare virtual chunk that every consumer statically imports. The module graph decides the download before any negotiation runs.
2. `__mfSeedLocalShared` runs `await share.get()` for every non-`import:false` share at container init, and `__mfNeedsPreInitSeedBarrier` returns false for non-singletons. Remotes initializing concurrently all miss the cache check and each fetch their own copy.
3. The post-init bridge (`__mfBridgeExternalSharedProvider`) that materializes the negotiated winner returns early for non-singletons. `__mfBridgeMaterializedProvider` does handle non-singletons, but only when the external provider is already lib, loading, or loaded, which is rarely true at remote-init time.

### Version

1.20.2

### Reproduction

n/a

### Relevant log output

```shell

```

### Validations

gioboa:
Close #1018

Defers workspace fallbacks until federation initialisation, reusing an existing exact-version provider instead of downloading duplicate implementations. Preserves singleton behaviour.

### DIFF
--- a/e2e/vite-vite/tests/host-preview.spec.ts
+++ b/e2e/vite-vite/tests/host-preview.spec.ts
@@ -14,6 +14,26 @@ test.describe('vite-vite host preview', () => {
     await expect(heading).toBeVisible();
   });
 
+  test('downloads a compatible non-singleton fallback once', async ({ page }) => {
+    const scriptResponses: Promise<{ url: string; body: string }>[] = [];
+    page.on('response', (response) => {
+      if (response.request().resourceType() === 'script') {
+        scriptResponses.push(
+          response.text().then((body) => ({ url: response.url(), body }))
+        );
+      }
+    });
+
+    await page.goto('/');
+    await expect(page.getByTestId('shared-counter-[shared-lib] Host')).toBeVisible();
+    await expect(page.getByTestId('shared-counter-[shared-lib] Remote')).toBeVisible();
+
+    const sharedLibResponses = (await Promise.all(scriptResponses)).filter(({ body }) =>
+      body.includes('[Shared Lib] Initialized')
+    );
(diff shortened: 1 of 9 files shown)


dmchoi77:
@gioboa I have a question about whether this fix also covers the singleton case.

For a workspace shared package configured as:

{ singleton: true }

(without import: false), the host provides and initializes the exact same version, but the remote still requests its generated fallback chunk.

Does this PR also prevent the fallback chunk from being requested in this scenario?

gioboa (commented):

Image

gioboa:
> Does this PR also prevent the fallback chunk from being requested in this scenario?

Yes, singleton workspace fallbacks already check the cache through PRs #883 and #918; this PR extends it to exact-version non-singletons.

PR #1023, opened 2026-08-05
feat: add `disableSsrRemoteEntry` to skip the node SSR remote entry

### LINKED ISSUE #1022: Node SSR remote entry with a bare `@module-federation/runtime` import is emitted into the client output for browser-only remotes, with no opt-out
## Summary

For a browser-only remote, `pluginSSRRemoteEntry` emits `<filename>.ssr.js` into the **client** output directory whenever `exposes` is non-empty. That file is Node-targeted — it imports `@module-federation/runtime` as a bare specifier — and there is no way to opt out of emitting it.

When the client output is published to a CDN (the normal deployment for a remote), the result is a publicly fetchable JavaScript module that no browser can evaluate:

```
TypeError: Failed to resolve module specifier "@module-federation/runtime".
Relative references must start with either "/", "./", or "../".
```

## Reproduction

Five files, no framework, no host — `vite` + `@module-federation/vite` only.

**package.json**
```json
{
  "name": "mf-ssr-entry-repro",
  "private": true,
  "type": "module",
  "scripts": { "build": "vite build" },
  "devDependencies": {
    "@module-federation/vite": "1.20.3",
    "vite": "^8.0.1"
  }
}
```

**vite.config.mjs** — browser-only remote: no SSR, no `remotes`, no `build.ssr`, no `environments.ssr`
```js

export default {
  plugins: [
    federation({
      name: 'repro',
      filename: 'remoteEntry.js',
      exposes: { './thing': './src/expose.js' },
      shared: {},
    }),
  ],
  build: { target: 'es2022' },
};
```

**index.html**
```html
<!doctype html><html><body><script type="module" src="/src/main.js"></script></body></html>
```

**src/expose.js**
```js
export const hello = () => 'hello from the exposed module';
```

**src/main.js**
```js
import { hello } from './expose.js';
document.body.textContent = hello();
```

Then:

```console
$ npm install && npm run build
$ ls dist
assets  index.html  remoteEntry.js  remoteEntry.ssr.js

$ cat dist/remoteEntry.ssr.js
import{t as e}from"./assets/vite-preload-helper-Czpn1I53.js";import{init as t}from"@module-federation/runtime";…
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                              bare specifier, unresolvable in a browser
```

Environment: `@module-federation/vite` 1.20.3, `vite` 8.2.0, node 24.13.0. Also present in 1.19.1 — the emit logic is unchanged between them.

`dist/remoteEntry.js` and everything under `dist/assets/` are clean; the bare specifier exists only in `remoteEntry.ssr.js`.

## Why there is no way to avoid it

`pluginSSRRemoteEntry`'s `buildStart` gate is `isServe` → environment name → empty `exposes`. For a plain client build (`environmentName === 'client'`, no `environments.ssr`, no `build.ssr`) it falls through and emits.

Notably `options.target` is **not** consulted anywhere in `pluginSSRRemoteEntry` — only `options.ssrExternals` is. So declaring the documented `target: 'web'`, which is already the default for a non-SSR build, does not suppress the SSR entry. I could not find any other option that does.

Both bundler paths emit the file; only the emit type differs (`type: 'asset'` on Rollup, `type: 'chunk'` on Rolldown), per the branch in `buildStart` and the existing tests around it.

## The file demonstrably produces that error in a browser

Not a theoretical concern. Serving the emitted `remoteEntry.ssr.js` and importing it the way a host would:

```js
import('./department.ssr.js')
  .then(m => console.log('loaded:', Object.keys(m)))
  .catch(e => console.error('name=' + e.constructor.name + ' :: message=' + e.message));
```

Headless Chrome (via Playwright, `channel: 'chrome'`):

```
[console.error] name=TypeError :: message=Failed to resolve module specifier
                "@module-federation/runtime". Relative references must start with either "/", "./", or "../".
```

Worth noting for anyone debugging this: the browser fetched `remoteEntry.ssr.js` and then **never requested its relative dependency chunk**, because specifier resolution precedes dependency fetching. So a session that really evaluated this file shows the `.ssr.js` request with none of its siblings.

## What I am *not* claiming

We have an intermittent production failure with exactly that error signature, on a browser-only remote whose host loads the browser `remoteEntry.js` directly by URL. 21 of our 33 remotes publish a `.ssr.js` today (HTTP 200, `content-type: text/javascript`).

**I cannot attribute that failure to this issue, and I am not asking you to treat it as evidence.** I tried to and failed. Every mechanism I could identify by which our app would evaluate the file is eliminated:

| Mechanism | Check | Result |
| --- | --- | --- |
| Runtime selects `ssrRemoteEntry` via `getRemoteEntryInfoFromSnapshot` | Read the condition in our shipped host bundle | Impossible for us — `ENV_TARGET` is a free identifier there (our host has no MF plugin), so `isBrowserEnvValue` resolves `true` in a browser and the `remoteEntry` branch always wins |
| `@module-federation/vite/ssrEntryLoader` deriving `${base}.ssr.js` | grep deployed host + remote chunks for `__mf_server__` / `__mf_ssr__` / `.ssr.js` | Not shipped — zero occurrences (expected: our remote has no `remotes`, so `injectSsrEntryLoader` is off) |
| The `import.meta.env.SSR` branch in the *browser* remote entry surviving DCE and executing | Rebuilt with `define: { 'import.meta.env.SSR': 'true' }` | Refuted — 317 asset chunks, 0 bare specifiers; the specifier stays confined to `remoteEntry.ssr.js` |
| Host bundle shipping its own bare specifier | AST scan of every host entry chunk | 0 — and a static one would stop the app booting entirely |

So the cause of our production error is still unknown and may well be unrelated to this plugin (our own deploy pipeline has a separate lockfile-drift problem that fits the symptom better). **Please read the production error only as what motivated the investigation.**

The request stands without it: **a remote that is only ever consumed by a browser host should not be forced to publish a Node-targeted module it cannot execute, into the directory that gets served to browsers.**

## Proposed fix

Add an opt-out, defaulting to current behaviour so nothing changes for existing SSR users. PR follows this issue: it adds `disableSsrRemoteEntry?: boolean`, named after the existing `disableRemote` / `disableShared` / `disableSnapshot` options, and returns early from the `buildStart` emit.

Alternatives considered and rejected:

- **Skip when `target === 'web'`.** Reads well, but `target` defaults to `'web'` for non-SSR builds and Nuxt emits the SSR entry from the *client* pass, so this would silently stop emitting for existing Nuxt/SSR users. Breaking.
- **Skip when the project has no SSR environment.** Depends on `environments.ssr` detection whose semantics vary across Vite majors and Nuxt; risks false negatives that break working SSR setups.
- **Emit as an asset on Rolldown too**, mirroring the Rollup path. Does not help — the asset is still written to the client `outDir` and still published. It also appears to exist deliberately on Rolldown so exposes are bundled into the Node SSR graph.

Gmendesfonseca:
Closes #1022

## Problem

For a browser-only remote, `pluginSSRRemoteEntry` emits `<filename>.ssr.js` into the **client** output directory whenever `exposes` is non-empty. That file is Node-targeted — it imports `@module-federation/runtime` as a bare specifier — and there is currently no way to opt out.

Since a remote's client output is normally published to a CDN, the result is a publicly fetchable module that no browser can evaluate (`Failed to resolve module specifier "@module-federation/runtime"`). `options.target` does not help: `pluginSSRRemoteEntry` never reads it, so the documented `target: 'web'` has no effect here.

Full reproduction (five files, no framework) is in the issue.

## Change

Adds `disableSsrRemoteEntry?: boolean`, named after the existing `disableRemote` / `disableShared` / `disableSnapshot` options, and returns early from the `buildStart` emit when it is set.

```js
federation({
  name: 'repro',
  filename: 'remoteEntry.js',
  exposes: { './thing': './src/expose.js' },
  disableSsrRemoteEntry: true,   // browser-only remote
})
```

**Default is unchanged** — the option is `undefined` unless set, so every existing SSR and Nuxt setup emits exactly as before. Three files, +66 lines, no behaviour change without the flag.

## Why an opt-out rather than inferring it

Recorded so the alternatives aren't re-derived:

- **Skip when `target === 'web'`.** Reads better, but `target` defaults to `'web'` for non-SSR builds and Nuxt emits the SSR entry from the *client* pass — this would silently stop emitting for existing Nuxt/SSR users. Breaking.
- **Skip when the project has no SSR environment.** Depends on `environments.ssr` detection whose semantics differ across Vite majors and under Nuxt; risks false negatives that break working SSR setups.
- **Emit as an asset on Rolldown too**, mirroring the Rollup branch. Doesn't solve it — the asset still lands in the client `outDir` and is still published. The chunk path also looks deliberate on Rolldown, so exposes bundle into the Node SSR graph.

Gating the emit in `buildStart` leaves the dev `/__mf_ssr__/` resolve and middleware paths untouched, since those don't produce deploy artifacts.

## Tests

Two cases added to `src/plugins/__tests__/pluginSSRRemoteEntry.test.ts`, one per bundler path (Rolldown/chunk, Rollup/asset).

Each pairs a **control** (flag absent → the file really is emitted) with the **treatment** (flag set → not emitted), so neither can pass for an unrelated reason such as the hook bailing out early. Each also asserts the option survived normalization, rather than inferring that from the absence of an emit.

Both were verified to fail without the source change, using two separate mutations:

| Mutation | Failure |
| --- | --- |
| Option removed from the normalizer | `expected undefined to be true` — the normalization guard |
| Only the `buildStart` gate deleted (option still normalizes) | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |

The second is the one that matters: it proves the early return is behaviour, not tidiness, so a later refactor that moves it turns the suite red.

## Verification

Run via package scripts, on `origin/main` @ `a5a59d4`.

| Gate | Baseline (pristine main) | With this PR |
| --- | --- | --- |
| `pnpm typecheck` | exit 0 | exit 0 |
| `pnpm test` | 1023 passed / **1 skipped** / 3 failed (1027) | 1025 passed / **1 skipped** / 3 failed (1029) |
| `pnpm fmt.check` | 1 file: `src/utils/codePositionMap.ts` | same 1 file |

**+2 tests, skip count unchanged, and the 3 failures are byte-identical to baseline** — they reproduce on pristine `main` across two consecutive runs, so they are pre-existing and unrelated to this change:

- `src/__tests__/index.test.ts > vite:module-federation-early-init > includes shared react in dev optimizeDeps when react-redux is installed`
- `src/__tests__/index.test.ts > vite:module-federation-early-init > includes shared singleton react in dev optimizeDeps without react-redux`
- `src/plugins/hmr/__tests__/react.test.ts > reactAdapter > serves the local React refresh runtime under the configured base path`

`src/utils/codePositionMap.ts` also fails `oxfmt --check` on pristine `main`. I left it alone rather than widen the diff by reformatting a file this PR doesn't touch — I confirmed it with the repo's pinned `oxfmt` (via `pnpm exec`), not a version `npx` happened to fetch, so it is a genuine pre-existing violation and not a formatter-version artifact.

Because of it, `.husky/pre-commit` (`npx oxfmt --check src` + `pnpm typecheck`) fails on this branch for a reason unrelated to the change, so **the commit was made with `--no-verify`**. Disclosing that rather than leaving it to be discovered: I re-ran both of the hook's gates against the committed state — `oxfmt --check` reports only the same pre-existing file, and `pnpm typecheck` passes. Happy to fold a `codePositionMap.ts` reformat in, or split it to its own PR, whichever you prefer.

End-to-end against the reproduction, using this branch's build:

```console
# control — flag absent
$ npm run build && ls dist/*.js
dist/remoteEntry.js  dist/remoteEntry.ssr.js
$ grep -rl '@module-federation/runtime' dist/ | wc -l
1

# treatment — disableSsrRemoteEntry: true
$ npm run build && ls dist/*.js
dist/remoteEntry.js
$ grep -rl '@module-federation/runtime' dist/ | wc -l
0

# the browser entry is unaffected
$ cat dist/remoteEntry.js
import{n as e,t}from"./assets/virtual_mf-REMOTE_ENTRY_ID___mfe_internal__repro__remoteEntry_js-wglAFbns.js";export{t as get,e as init};
```

## Known limitations

- **This does not explain the production error that led me here.** We see `Failed to resolve module specifier "@module-federation/runtime"` intermittently on a browser-only remote whose host loads the browser `remoteEntry.js` directly by URL. `remoteEntry.ssr.js` is the only artifact in our deploy whose contents can produce that string, but I could not prove how a browser comes to evaluate it, and two mechanisms I suspected were measured and refuted. So this PR is scoped to the defect I *can* demonstrate — a browser-only remote being forced to publish a Node module — and makes no claim to fix that error.
- **The opt-out is manual.** A remote that should have set it and didn't still publishes the file. Inferring it automatically is the better end state, but every inference I could find risks breaking existing SSR/Nuxt users (above).
- **Dev-mode SSR routes are untouched.** With the flag set, `/__mf_ssr__/…` still resolves in `vite serve`. That produces no deploy artifact, so I left it out to keep the change to one concern — happy to gate those too if you'd prefer the flag to mean "no SSR entry anywhere".
- **Verification is author-run**, not independently reproduced.

### DIFF
--- a/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
@@ -836,6 +836,53 @@ describe('pluginSSRRemoteEntry', () => {
       );
     });
 
+    // A browser-only remote must not publish `<filename>.ssr.js`: it imports
+    // `@module-federation/runtime` as a bare specifier, which no browser can resolve,
+    // and it lands in the client output directory that gets served to browsers.
+    // Each test pairs a control (flag absent -> emitted) with the treatment
+    // (flag set -> not emitted), so it cannot pass for an unrelated reason such as
+    // the hook bailing out early.
+    it.each([
+      ['Rolldown', true, 'chunk'],
+      ['Rollup', false, 'asset'],
+    ])(
+      'does not emit the node SSR remote entry when disableSsrRemoteEntry is set (%s)',
+      (_label, rolldown, expectedType) => {
+        getIsRolldownMock.mockReturnValue(rolldown as boolean);
+
+        const control = makeEmitFile();
+        callHook(
+          pluginSSRRemoteEntry(makeOptions())[1].buildStart,
+          {
+            meta: makePluginMeta(rolldown as boolean),
+            emitFile: control,
+          } as unknown as Rollup.PluginContext,
+          {} as Rollup.NormalizedInputOptions
+        );
+        // Control: without the flag this configuration really does emit, so the
+        // treatment assertion below is meaningful.
+        expect(control).toHaveBeenCalledWith(
+          expect.objectContaining({ type: expectedType, fileName: 'remoteEntry.ssr.js' })
+        );
+
+        const options = makeOptions({ disableSsrRemoteEntry: true });
+        // Guard: prove the option survived normalization rather than being dropped.
+        expect(options.disableSsrRemoteEntry).toBe(true);
+
+        const emitFile = makeEmitFile();
+        callHook(
+          pluginSSRRemoteEntry(options)[1].buildStart,
+          {
+            meta: makePluginMeta(rolldown as boolean),
+            emitFile,
+          } as unknown as Rollup.PluginContext,
+          {} as Rollup.NormalizedInputOptions
+        );
+
+        expect(emitFile).not.toHaveBeenCalled();
+      }
+    );
+
     it('skips emit when no exposes are configured', () => {
       const emitFile = makeEmitFile();
       const plugins = pluginSSRRemoteEntry(makeOptions({ exposes: {} }));
--- a/src/plugins/pluginSSRRemoteEntry.ts
+++ b/src/plugins/pluginSSRRemoteEntry.ts
@@ -496,6 +496,10 @@ export function pluginSSRRemoteEntry(options: NormalizedModuleFederationOptions)
       buildStart() {
         // Only emit the SSR entry chunk during vite build — not vite serve.
         if (isServe) return;
+        // Opt-out for remotes that are only ever consumed by a browser host. The SSR
+        // entry imports `@module-federation/runtime` as a bare specifier, so shipping
+        // it to a browser-only deploy publishes a module no browser can evaluate.
+        if (options.disableSsrRemoteEntry) return;
         // `this.meta` is available in Rollup/Rolldown hooks — use it to detect
         // whether we're running under Rolldown (Vite 8+) so we can choose the
         // right output format and file extension.
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -567,6 +567,20 @@ export type ModuleFederationOptions = {
    */
   ssrExternals?: string[];
   /**
+   * Skips emitting the Node SSR remote entry (`<filename>.ssr.js`) entirely.
+   *
+   * The SSR remote entry is Node-targeted: it imports `@module-federation/runtime`
+   * as a bare specifier, which is correct for Node but unresolvable in a browser.
+   * It is otherwise emitted into the client output directory whenever `exposes` is
+   * non-empty, so a browser-only remote publishes a file it can never evaluate —
+   * and, when that output is served by a CDN, a file that is publicly fetchable.
+   *
+   * Enable this for remotes that are only ever consumed by a browser host.
+   *
+   * @default false
+   */
+  disableSsrRemoteEntry?: boolean;
+  /**
    * Experimental Module Federation capabilities.
    *
    * @see https://module-federation.io/configure/experiments
@@ -806,6 +820,7 @@ export function normalizeModuleFederationOptions(
     varFilename: options.varFilename,
     target: options.target,
     ssrExternals: options.ssrExternals,
+    disableSsrRemoteEntry: options.disableSsrRemoteEntry,
     disableRemote: options.disableRemote,
     disableShared: options.disableShared,
     disableSnapshot: options.disableSnapshot,


gioboa (commented):
Thanks, @Gmendesfonseca.
This behaviour wasn’t intended.
#1024 fixes it by avoiding SSR entry generation during client-only builds.

PR #1024, opened 2026-08-05
fix: emit SSR entry during SSR builds

### LINKED ISSUE #1022: Node SSR remote entry with a bare `@module-federation/runtime` import is emitted into the client output for browser-only remotes, with no opt-out

gioboa:
Close #1022 #1023

Fix issue with browser builds.
Stop emitting the Node-targeted SSR remote entry, aligning behaviour with Rspack and Webpack.

### DIFF
--- a/src/plugins/pluginSSRRemoteEntry.ts
+++ b/src/plugins/pluginSSRRemoteEntry.ts
@@ -505,26 +505,17 @@ export function pluginSSRRemoteEntry(options: NormalizedModuleFederationOptions)
         const environmentName = (this as { environment?: { name?: string } }).environment?.name;
         const hasSsrEnvironment = Boolean(viteConfig?.environments?.ssr);
         const isLegacySsrBuild = Boolean(
-          (this as { environment?: { config?: ResolvedConfig } }).environment?.config?.build?.ssr
+          (this as { environment?: { config?: ResolvedConfig } }).environment?.config?.build?.ssr ??
+          viteConfig?.build?.ssr
         );
 
         // Environment API (client + ssr): emit only in the ssr environment so exposes
         // are bundled with the Node SSR graph (nested loadRemote stays on the server).
-        // Nuxt only runs the client Vite graph for federation output (dist/client);
-        // skipping the client pass leaves no remoteEntry.ssr.js in .output/public and
-        // the host falls back to the browser remoteEntry (SourceTextModule errors).
-        if (hasSsrEnvironment) {
-          // Non-Nuxt: SSR entry belongs in the dedicated `ssr` environment build.
-          // Nuxt: federation assets land in `dist/client` only — emit there (or the
-          // unnamed client pass), never in the Nitro `ssr` pass (wrong output dir).
-          if (isNuxtProject) {
-            if (environmentName === 'ssr') return;
-          } else if (environmentName !== 'ssr') {
-            return;
-          }
-        } else if (isLegacySsrBuild) {
-          // Legacy `vite build --ssr` pass (e.g. vue-ssr dual build:server).
-        } else if (environmentName && environmentName !== 'client') {
+        // Emit only for an SSR environment or legacy `vite build --ssr`.
+        if (
+          (hasSsrEnvironment && environmentName !== 'ssr') ||
+          (!hasSsrEnvironment && !isLegacySsrBuild)
+        ) {
           return;
         }
 
--- a/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
@@ -3,7 +3,7 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { pathToFileURL } from 'node:url';
-import type { Rollup, ResolvedConfig } from 'vite';
+import type { Plugin, Rollup, ResolvedConfig } from 'vite';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
 import { normalizeModuleFederationOptions } from '../../utils/normalizeModuleFederationOptions';
@@ -789,11 +789,23 @@ describe('pluginSSRRemoteEntry', () => {
   });
 
   describe('main plugin — buildStart', () => {
+    function configureLegacySsrBuild(plugin: Plugin) {
+      callHook(
+        plugin.configResolved,
+        {} as Rollup.PluginContext,
+        {
+          command: 'build',
+          build: { ssr: true },
+        } as ResolvedConfig
+      );
+    }
+
     it('emits SSR entry as a pre-generated ESM asset for Rollup (avoids code-splitting side effects)', () => {
       getIsRolldownMock.mockReturnValue(false);
       const emitFile = makeEmitFile();
       const plugins = pluginSSRRemoteEntry(makeOptions());
       const mainPlugin = plugins[1];
+      configureLegacySsrBuild(mainPlugin);
 
       callHook(
         mainPlugin.buildStart,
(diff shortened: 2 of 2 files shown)


gioboa (commented):

Image

src/plugins/pluginSSRRemoteEntry.ts
```diff
@@ -505,26 +505,17 @@ export function pluginSSRRemoteEntry(options: NormalizedModuleFederationOptions)
...
-          // unnamed client pass), never in the Nitro `ssr` pass (wrong output dir).
-          if (isNuxtProject) {
-            if (environmentName === 'ssr') return;
-          } else if (environmentName !== 'ssr') {
-            return;
-          }
-        } else if (isLegacySsrBuild) {
-          // Legacy `vite build --ssr` pass (e.g. vue-ssr dual build:server).
-        } else if (environmentName && environmentName !== 'client') {
+        // Emit only for an SSR environment or legacy `vite build --ssr`.
```
dmchoi77, line 514: Hi! @gioboa I have a question about the SSR remote entry behavior after PR #1024.

For Vite 5–7, `@module-federation/vite` emits `remoteEntry.ssr.js` as a Rollup asset. Since Vite’s default `ssrEmitAssets` is `false`, the asset is not written to the final SSR output unless users configure:

```ts
build: {
  ssrEmitAssets: true,
}
```

In Vite 8, the plugin emits the SSR entry as a chunk, so it works even with ssrEmitAssets: false.

Is the intended contract that Vite 5–7 users must set ssrEmitAssets: true manually, or should the plugin enable it automatically for the generated SSR remote entry?

Also, should this requirement be documented and covered by an integration test that checks the final remoteEntry.ssr.js file?

gioboa: That's a valid concern, let me fix it.

PR #1025, opened 2026-08-06
fix: guard shared React mixed-mode dispatchers

gioboa:
Follow-up of -> chore: add automatic support for a DEV host consuming a PROD React remote (#997)

Prevents getOwner crashes when @module-federation/runtime loads production remotes into development hosts.


gioboa (commented):

Image

PR #1026, opened 2026-08-06
fix: defer npm non-singleton shared fallbacks

gioboa:
Defers production fallback loading until federation initialization, reusing an existing exact-version provider when available.


gioboa (commented):

Image

PR #1028, opened 2026-08-06
fix: honor eager workspace singleton exports

### LINKED ISSUE #1027: Workspace singleton read synchronously at module-eval time gets async export shape when not "consumed by a peer singleton" → binding is undefined
## Environment

- `@module-federation/vite@1.20.0`
- Vite: rolldown-vite
- Framework: Angular (standalone components), Module Federation v2, host + remotes
- Repro shape: a **first-party (npm-workspace) package** shared as `singleton` and imported for a **value** read that happens synchronously at module-evaluation time (e.g. inside an Angular component's static class metadata `ɵcmp`)

## Summary

For **workspace** (first-party, in-repo) singleton shares, the generated `loadShare` module can take one of two export shapes:

- **eager / synchronous** — `generateEagerWorkspaceSingletonExports`: emits a static `import * as __mfLocalShare from "<local>"` and synchronously assigns the exported bindings, so they hold real values at module-eval time.
- **lazy / asynchronous** — `generateLazyWorkspaceSingletonExports`: the exported bindings are assigned later inside `initPromise.then(...)`, so at module-eval time they are `undefined`.

The choice is made by `usesEagerWorkspaceFallback` in `writeLoadShareModule` (`lib/index.js`, ~L2630):

```js
const usesEagerWorkspaceFallback = hasCompleteExportCoverage && isWorkspaceSingleton && isConsumedByPeerSingleton;
```

`isConsumedByPeerSingleton` comes from `isSharedSingletonConsumedByPeer(pkg, options)` (~L2147). It returns `true` only when **another shared singleton** imports `pkg` — either:

1. a direct subpath check: another shared singleton key that ``startsWith(`${pkg}/`)``, or
2. a recursive walk of the shared-singleton **dependency graph** (`dependencies`/`peerDependencies`/`optionalDependencies` from each shared package's `package.json`).

A **leaf** workspace singleton — one that no _other shared singleton_ imports, but that the **application code** reads synchronously at eval time — therefore gets the **async** shape, and its exported binding is `undefined` when read.

## Impact

In Angular, a component's static class metadata (`ɵcmp`) is constructed during class definition — i.e. at module-evaluation time — and can synchronously reference a shared value export (a component class, a factory function, a token, an enum, etc.). With the async export shape, that reference is `undefined`, producing a runtime crash inside the static initializer, e.g.:

```
Uncaught (in promise) TypeError: <X> is not a function
    at <static_initializer> (…)
```

In our case `<X>` was a component class exported from a shared workspace **subpath** entry (`@scope/ui/icon`) and referenced in another component's `ɵcmp` `features` array. The binding was `undefined` at static-init because the generated `loadShare` module for that subpath used the async shape.

Note: **`shareConfig.eager: true` does not fix this.** `eager` only affects _chunking_ (`generateLocalSharedImportMap` — whether the share's bytes go into the initial chunk vs. a dynamic `import()`); it is **not** an input to `usesEagerWorkspaceFallback`, which decides the _export shape_. So a share can be `eager: true` yet still receive the async binding shape.

## Root cause — granularity mismatch

The deeper issue is that `isSharedSingletonConsumedByPeer` reasons at **package-name** granularity (it maps each shared key through `getPackageName(key)` and walks `package.json` dependencies, which are package names), while shares are frequently **declared at subpath granularity** (`@scope/pkg/subA`, `@scope/pkg/subB`, …) for tree-shaking.

Consequences:

- A consumer's `package.json` dependency on `@scope/pkg` resolves (via the `sharedKeyByPackageName` map) to a **single canonical key** for that package name — never to a specific subpath key. So the walk's `sharedDependency === pkg` check can never match a _subpath_ `pkg` like `@scope/pkg/subA`.
- Adding the **barrel** `@scope/pkg` as a shared singleton does **not** help a subpath leaf: the direct subpath check (``key.startsWith(`${pkg}/`)``) looks for keys _under_ the subpath, and the graph walk collapses the package to the barrel key.
- **Subpath → sibling-subpath** eval-time reads (`@scope/pkg/subA` statically `extends`/references `@scope/pkg/subB`) are likewise invisible: same package name, collapsed to one canonical key.

So there is **no share-list arrangement** that makes a subpath leaf (or a subpath→sibling edge) satisfy `isConsumedByPeerSingleton`. The heuristic structurally under-covers this legitimate consumption pattern.

## Minimal repro

1. Workspace package `@scope/leaf` with `export const Thing = () => {};`.
2. Share it: `shared: { '@scope/leaf': { singleton: true, eager: true } }` — and ensure **no other shared singleton** depends on `@scope/leaf`.
3. In host application code, read it **synchronously at eval time**, e.g. a module-level capture or a static class field:
   ```ts
   export class Foo {
   	static readonly captured = Thing(); // runs at class-definition (module-eval) time
   }
   ```
4. Build federated + serve.

**Observed:** `Thing` (the exported binding) is `undefined` at the eval-time read → `TypeError: Thing is not a function`. In the generated `loadShare` module for `@scope/leaf`, `export { __mf_0 as Thing }` where `__mf_0` is only assigned inside `pendingShareLoads.push(initPromise.then(...))`, never synchronously.

**Expected:** a supported way to force the **synchronous** export shape for a workspace singleton that is read at module-eval time, independent of whether a _peer singleton_ imports it — since eval-time value reads (Angular decorator metadata, module-level captures, `class X extends SharedBase`) are a legitimate consumption pattern.

## Suggested fix

Add a per-share opt-in that also satisfies `usesEagerWorkspaceFallback`. The smallest, backward-compatible change is to honor an explicit per-share flag (e.g. `forceEager`, or reuse `eager`) as a second trigger:

```js
const usesEagerWorkspaceFallback =
	hasCompleteExportCoverage &&
	isWorkspaceSingleton &&
	(isConsumedByPeerSingleton || shareItem.shareConfig.forceEager === true);
```

with the flag plumbed through `normalizeShareItem`:

```js
forceEager: !!shareItem.forceEager,
```

## Current workaround

We ship exactly the two-line change above as a local `patch-package` patch against `1.20.0` (adds a `forceEager` share flag in `normalizeShareItem` and uses it as an extra trigger in `usesEagerWorkspaceFallback`). It resolves all our eval-time-read crashes with no observed regressions.

Happy to open a PR with this change (plus a test covering the subpath-leaf case) if the maintainers agree with the approach — or a different API name if `forceEager` isn't preferred.

gioboa:
Close #1027

Makes eager: true generate synchronous exports for workspace singletons, preventing undefined bindings during module evaluation.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -3324,6 +3324,39 @@ describe('writeLoadShareModule', () => {
     expect(generatedCode).toContain('Promise.resolve().then');
   });
 
+  it('emits eager local exports for an explicitly eager leaf workspace singleton', () => {
+    normalizeModuleFederationOptions({
+      name: 'host',
+      shared: {
+        'workspace-producer': { singleton: true, eager: true },
+      },
+    });
+    const pkg = 'workspace-producer';
+    const mockShareItem: ShareItem = {
+      name: pkg,
+      from: '',
+      version: '1.0.0',
+      shareConfig: {
+        singleton: true,
+        eager: true,
+        strictVersion: false,
+        requiredVersion: '^1.0.0',
+      },
+      scope: 'default',
+    };
+
+    writeLoadShareModule(pkg, mockShareItem, 'build', false);
+
+    const generatedCode = writeSyncSpy.mock.calls.at(-1)?.[0] as string;
+
+    expect(generatedCode).toContain(
+      'import * as __mfLocalShare from "/repo/packages/workspace-producer/src/index.ts";'
+    );
+    expect(generatedCode).toContain('const __mfApplyEagerShareExports = (mod) => {');
+    expect(generatedCode).toContain('__mfApplyEagerShareExports(exportModule);');
+    expect(generatedCode).not.toContain('initPromise.then');
+  });
+
   it('does not treat parent package.json name mismatches as workspace package matches', () => {
     const pkg = 'workspace-name-mismatch';
     const mockShareItem: ShareItem = {
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -1939,7 +1939,9 @@ export function writeLoadShareModule(
     resolvedOptions.hostInitInjectLocation === 'entry' &&
     isConsumedByPeerSingleton;
   const usesEagerWorkspaceFallback =
-    hasCompleteExportCoverage && isWorkspaceSingleton && isConsumedByPeerSingleton;
+    hasCompleteExportCoverage &&
+    isWorkspaceSingleton &&
+    (isConsumedByPeerSingleton || shareItem.shareConfig.eager === true);
   const usesDeferredTreeShakingFallback = hasCompleteExportCoverage && Boolean(treeShakingConsumer);
   const reactMixedModeGuard = pkg === 'react' ? createReactMixedModeRuntimeGuard() : '';
   let exportLine: string;


gioboa:
@vsternbach I'm merging this fix. this should solve the problem. Can you confirm?

gioboa (commented):

Screenshot 2026-08-06 at 13 51 24

PR #1032, opened 2026-08-06
fix: namespace Vue template HMR rerender ids

### LINKED ISSUE #1031: [Bug]: Vue template HMR in federated remotes uses an unprefixed component ID
### Describe the bug

With Vue and `dev.remoteHmr: true`, a template-only edit in a federated remote is received by
Vite, but the mounted remote component is not rerendered.

The Vue adapter in `@module-federation/vite@1.20.1` prefixes the HMR ID assigned by the main SFC
module so host and remote components cannot collide in the shared Vue HMR runtime:

```js
_sfc_main.__hmrId = "remote-<sfc-id>"
```

However, `@vitejs/plugin-vue@6` emits template hot updates in a separate module with a direct
literal call:

```js
import.meta.hot.accept(({ render }) => {
  __VUE_HMR_RUNTIME__.rerender("<sfc-id>", render)
})
```

The remote adapter currently returns early unless the transformed code contains
`__VUE_HMR_RUNTIME__.createRecord(`. It therefore never processes this template submodule. The
live component is registered as `remote-<sfc-id>`, while the hot update tries to rerender
`<sfc-id>`. The lookup misses, so the DOM remains unchanged even though Vite reports a successful
hot update.

I can submit a PR if the proposed approach below is acceptable.

### Version

Vite `8.2.0`  
Also reproduced with:  - `@module-federation/vite@1.20.1` - `@vitejs/plugin-vue@6.0.7` - `vue@3.5.40`

### Reproduction

 Repository: https://github.com/stalker3343/vue-hmr-reproduction 

```bash pnpm install pnpm dev ```  1. Open `http://localhost:5173/`. 2. Increment the remote counter so its local state is non-zero. 3. Change `Template version: A` to `Template version: B` in    `remote/src/RemoteCounter.vue`. 4. Save the file.  Actual result: Vite reports a template hot update, but the DOM still shows `Template version: A`.  Expected result: the DOM changes to `Template version: B` without navigation or a document reload, and the counter retains its local state.

### Relevant log output

```shell
[vite] hot updated: /src/RemoteCounter.vue?vue&type=template&...


The generated modules show the ID mismatch:


// /src/RemoteCounter.vue after the federation transform
_sfc_main.__hmrId = "remote-feb1273b"

// /src/RemoteCounter.vue?vue&type=template&lang.js
__VUE_HMR_RUNTIME__.rerender("feb1273b", render)


There is no runtime exception. Calling the same rerender manually with the federation-prefixed
ID updates the component immediately:


__VUE_HMR_RUNTIME__.rerender("remote-feb1273b", render)



**Proposed fix**

Extend the Vue remote HMR transform so it also prefixes literal IDs in direct template-submodule
calls shaped like:


__VUE_HMR_RUNTIME__.rerender("<sfc-id>", render)


The existing `.__hmrId = "..."` rewrite and its collision protection should remain unchanged.
The new rewrite should be idempotent when the federation prefix is already present. Unit tests
should cover both the main SFC module and the Vue template submodule forms.
```

### Validations

stalker3343:
Prefix HMR IDs used by template submodules with the federation name. This prevents host and remote Vue components from colliding in the shared HMR runtime.

Fix https://github.com/module-federation/vite/issues/1031

### DIFF
--- a/src/plugins/hmr/__tests__/vue.test.ts
+++ b/src/plugins/hmr/__tests__/vue.test.ts
@@ -83,6 +83,14 @@ describe('vueAdapter', () => {
         '});',
       ].join('\n');
 
+    const templateModule = (hmrIdLiteral: string) =>
+      [
+        'export function render() {}',
+        'import.meta.hot.accept(({ render }) => {',
+        `  __VUE_HMR_RUNTIME__.rerender("${hmrIdLiteral}", render);`,
+        '});',
+      ].join('\n');
+
     it('prefixes _sfc_main.__hmrId with the federation name', () => {
       const result = vueAdapter.remote?.transform?.(
         sfcModule('abc123'),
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Excellent contribution, clean and well-tested.
Thanks @stalker3343 👏

PR #1035, opened 2026-08-06
fix: preserve SSR entry during cleanup

gioboa:
Ensures Vite 5–7 emits remoteEntry.ssr.js even when ssrEmitAssets is disabled, without enabling unrelated SSR assets.


gioboa (commented):
@dmchoi77 is this solving your concern?

gioboa (commented):

Image

dmchoi77:
@gioboa Yes, this perfectly addresses my concern. Thanks for the quick fix!🚀

PR #1037, opened 2026-08-06
fix: include peerDependencies in automatic shared detection

dmchoi77:
## Summary

Component remotes commonly declare framework packages such as React or Vue in `peerDependencies` and `devDependencies`, while omitting an explicit `shared` configuration. Those framework packages could be missed by automatic shared detection. This change includes installed `peerDependencies` in the automatic shared candidates so the remote and host can use the same singleton instance.

## Root Cause

When `shared` was omitted, automatic detection only iterated over `packageJson.dependencies` and ignored `peerDependencies`.

As a result:

- The remote bundled its own React instance.
- The host provided a different React singleton.
- The two React instances conflicted at runtime.
- The application could fail with `Invalid hook call`, `useState` errors, or broken context identity.

## Changes

- Include `peerDependencies` in automatic shared detection.
- Register only peers that are installed and resolvable.
- Preserve the peer range as `shareConfig.requiredVersion`.
- Continue using the installed package version as `version`.
- When a package is declared in both `dependencies` and `peerDependencies`, use the peer range as the required version while retaining the installed version.
- Preserve the existing behavior where explicit `shared: {}` disables automatic detection.

The shape passed to runtime-core remains unchanged:

```ts
{
  version: "19.2.4",
  shareConfig: {
    singleton: true,
    requiredVersion: "^19.0.0",
  },
}
```

## Tests

Added regression coverage for:

- Automatically sharing React declared only in `peerDependencies + devDependencies`.
- Ignoring an uninstalled peer without failing the build.
- Handling a package declared in both `dependencies` and `peerDependencies`.
- Preserving the explicit `shared: {}` opt-out.

Verification:

- `pnpm test`
- `pnpm test:integration`
- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm build`

## Scope and Compatibility

- Explicit shared configurations are unchanged.
- Existing `dependencies`-based automatic sharing is preserved.
- `optionalDependencies` are intentionally outside the scope of this change.

### DIFF
--- a/src/utils/__tests__/normalizeModuleFederationOption.test.ts
+++ b/src/utils/__tests__/normalizeModuleFederationOption.test.ts
@@ -591,6 +591,116 @@ describe('normalizeModuleFederationOption', () => {
       }
     });
 
+    it('auto-shares installed peer dependencies from a component remote', () => {
+      const path = require('node:path');
+      const fs = require('node:fs');
+      const fixtureRoot = path.join(require('node:os').tmpdir(), 'mf-vite-peer-auto-share');
+      const reactDir = path.join(fixtureRoot, 'node_modules/react');
+      fs.rmSync(fixtureRoot, { force: true, recursive: true });
+      fs.mkdirSync(reactDir, { recursive: true });
+      fs.writeFileSync(
+        path.join(fixtureRoot, 'package.json'),
+        JSON.stringify({
+          name: 'component-remote',
+          peerDependencies: {
+            react: '^19.0.0',
+            'missing-peer': '^1.0.0',
+          },
+          devDependencies: {
+            react: '19.2.4',
+          },
+        })
+      );
+      fs.writeFileSync(
+        path.join(reactDir, 'package.json'),
+        JSON.stringify({ name: 'react', version: '19.2.4', exports: './index.js' })
+      );
+      fs.writeFileSync(path.join(reactDir, 'index.js'), '');
+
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks for the contribution and for bringing this behaviour to my attention. 👀 
Automatic dependency sharing was introduced `by mistake` during the large #638 refactor. Rspack/webpack share only explicitly configured dependencies. Extending this accidental behaviour to peerDependencies would increase the divergence and implicitly mark more packages as singletons.

We should remove the automatic detection block from normalizeShared() in src/utils/normalizeModuleFederationOptions.ts and release the fix in a minor version.

gioboa (changes requested):

dmchoi77:
@gioboa Thanks for the clarification. I agree that extending automatic detection to `peerDependencies` would further diverge from webpack/Rspack’s explicit `shared` configuration.

I’ll revise this PR to remove the automatic dependency detection block from `normalizeShared()`, restore the behavior where an omitted `shared` option normalizes to `{}`, and update the regression tests accordingly.

I’ll also update the PR for a minor-version release and force-push the revised branch after verification.

dmchoi77:
@gioboa I see that you’ve already opened #1038 to revert the accidental automatic dependency sharing.
Since that PR contains the intended fix, would you prefer that I close this PR to avoid duplication?

gioboa:
Thanks for your patience, @dmchoi77. I’ll close this PR in favor of #1038.

PR #1038, opened 2026-08-06
fix: revert accidental automatic dependency sharing

gioboa:
Removes automatic dependency sharing, introduced accidentally during a large refactor. Restores explicit configuration and alignment with Rspack (https://rspack.dev/plugins/webpack/module-federation-plugin#shared) and webpack (https://webpack.js.org/plugins/module-federation-plugin/#sharing-libraries).

```
Correct behavior:

// Shares nothing
federation({ name: 'remote' });

// Shares React explicitly
federation({
  name: 'remote',
  shared: {
    react: { singleton: true },
  },
});
```

My main concern is keeping the Vite, Rspack, and webpack implementations aligned as much as possible. The Module Federation specification is the result of extensive research and was carefully designed for predictable, optimal behaviour.

Since Rspack and webpack require explicit shared configuration, Vite should follow the same semantics and avoid implicit dependency detection. This reduces surprises and makes migration between bundlers easier.

### DIFF
--- a/src/utils/__tests__/normalizeModuleFederationOption.test.ts
+++ b/src/utils/__tests__/normalizeModuleFederationOption.test.ts
@@ -734,92 +734,29 @@ describe('normalizeModuleFederationOption', () => {
       setPackageDetectionCwd(process.cwd());
     });
 
-    it('skips Nuxt module packages from implicit shared deps', () => {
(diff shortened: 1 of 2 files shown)


Nsttt (approved):

gioboa (commented):

Image

PR #1039, opened 2026-08-06
fix: isolate SSR entry emission by environment

tonoizer:
## Summary

- Remove shared per-build SSR emission state from `pluginSSRRemoteEntry`.
- Re-evaluate SSR eligibility and Rolldown detection for each hook invocation.
- Keep SSR entry emission scoped to the current SSR environment during parallel client/SSR builds.
- Add regression coverage for both client/SSR build-start orderings.

## Root cause

The plugin instance is shared across Vite environments, but `shouldEmitSsrEntry` and `isRolldown` were closure-global. A client build could reset or observe state belonging to the SSR build, causing the SSR entry to be omitted or emitted into the client output.

## Validation

- `git diff --check`
- Added deterministic unit coverage for both interleavings.
- Vitest, typecheck, and formatting could not run in this checkout because `node_modules` is unavailable.
- A real Vite 6/7 parallel Environment API integration test remains a follow-up validation opportunity.

No security or performance behavior is changed by this PR.

### DIFF
--- a/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
@@ -958,6 +958,63 @@ describe('pluginSSRRemoteEntry', () => {
       expect(emitFile).not.toHaveBeenCalled();
     });
 
+    it.each([
+      ['SSR starts before client', 'ssr-first'],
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks, this fix looks solid @tonoizer 🚀

PR #1040, opened 2026-08-06
perf: cache generated SSR remote entry source

tonoizer:
## Summary

- Cache the generated SSR remote-entry source within each plugin instance.
- Reuse the stable base source across multiple Rollup outputs.
- Preserve independent Nuxt expose-chunk rewriting for every output.
- Add multi-output coverage proving one generation and two correctly rewritten assets.

## Motivation

Rollup can invoke output-generation hooks for multiple outputs. The generated SSR entry depends only on normalized federation options, so regenerating the same base source for every output is unnecessary.

## Scope

This is intentionally limited to the SSR source-generation path. It does not change environment gating, publication, dev-server behavior, or generated output semantics.

## Validation

- `git diff --check`
- Added focused multi-output Nuxt unit coverage.
- Vitest, typecheck, and formatting could not run in this checkout because `node_modules` is unavailable.

This is a small optimization; the benefit is proportional to the number of Rollup outputs and may be most useful for multi-output builds.

### DIFF
--- a/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginSSRRemoteEntry.test.ts
@@ -1152,6 +1152,68 @@ describe('pluginSSRRemoteEntry', () => {
         })
       );
     });
+
+    it('caches the base SSR entry across outputs while rewriting Nuxt imports per output', () => {
+      getIsRolldownMock.mockReturnValue(false);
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks @tonoizer 
Nice one 👏

gioboa (approved):

Issue #1041, opened 2026-08-07
[Bug]: Too long initial startup time


Sddoo:
### Describe the bug

Starting with version 1.17.1, the initial page loading time has increased significantly. I guess it happened because of this PR: https://github.com/module-federation/vite/pull/918. 
I do not know how to create a reproducible example, because the project in which the long load reproduced is corporate. If the problem is incorrect or not where it should be, please tell me how and where to do it correctly. 
P.S. it is noticeable that your plugin is running slower than a similar one from originjs, are any optimizations planned?

Example with 1.17.0 version:
<img width="1007" height="331" alt="Image" src="https://github.com/user-attachments/assets/7b921880-f693-4e9d-98ae-d338c7495440" />

Example with 1.17.1 version: 
<img width="1011" height="490" alt="Image" src="https://github.com/user-attachments/assets/8266952b-5d7a-4a35-85a9-eebeb8647690" />

Federation config:
```
plugins: [
  federation({
    name: 'host',
    dts: false,
    remotes: {
      mf1: {
        type: 'module',
        name: 'mf1',
        entryGlobalName: 'mf1',
        entry: 'mf1-url'
      },
      mf2: {
        type: 'module',
        name: 'mf2',
        entryGlobalName: 'mf2',
        entry: 'mf2-url'
      },
    },
    dev: { remoteHmr: isLocalRemote, disableDynamicRemoteTypeHints: isLocalRemote },
    shared: {
      react: {
        singleton: true,
        requiredVersion: dependencies.react,
      },
      'react-dom': {
        singleton: true,
        requiredVersion: dependencies['react-dom'],
      },
      'react-redux': {
        singleton: true,
        requiredVersion: dependencies['react-redux'],
      },
      '@reduxjs/toolkit': {
        singleton: true,
        requiredVersion: dependencies['@reduxjs/toolkit'],
      },
      'uikit1': {
        singleton: true,
        requiredVersion: dependencies['uikit1'],
      },
      'uikit2': {
        singleton: true,
        requiredVersion: dependencies['uikit2'],
      },
    },
  }),
  react(),
].filter(Boolean)
```

Server config:
```
{
    host: env.HOST,
    port: 3000,
    open: true,
    allowedHosts: true,
    cors: true,
    watch: {
      usePolling: true,
      interval: 300,
      binaryInterval: 300,
      awaitWriteFinish: {
        stabilityThreshold: 300,
        pollInterval: 150,
      },
    },
    warmup: {
[...]

PR #609, opened 2026-04-10
fix: expose imports run one-at-a-time

gioboa:
Close #604

PR #1043, opened 2026-08-07
fix: allow independent expose loads without reintroducing Safari TLA races

dmchoi77:
## Summary

This PR removes the global one-at-a-time expose import queue introduced in #609. That queue serialized unrelated exposes and created a waterfall when a remote contained many large exposed modules.

The replacement keeps only an in-flight promise per expose key. The same expose is deduplicated, failed loads can retry, and independent exposes can start in parallel.

## Merge note

If this implementation direction is accepted, merge the functional fix from `877a89e` but exclude the benchmark commit `922323d` from the final merge. The benchmark is included as independently revertible evidence only; it can be excluded by merging only the first commit or reverted with:

```sh
git revert 922323d
```

## Why this is safe

The expose loader is not responsible for shared-module initialization. That readiness is represented by explicit Promise barriers in the current federation graph:

- `loadShare` tracks deferred shared exports in `pendingShareLoads`.
- `remoteEntry.get()` waits for pending shared loads before invoking an expose.
- remote named-export proxies publish `__mf_remote_dependency_pending` instead of using generated top-level `await`.
- host bootstrap awaits the optional `__tla` promise before reading `initHost`.

This preserves the no-TLA direction from [PR #638](https://github.com/module-federation/vite/pull/638) while removing the leftover global mutex.

## Changes

- Replace the global expose promise queue with a per-expose in-flight map.
- Remove the test expectation that unrelated exposes are serialized.
- Add coverage for same-expose deduplication and retry after failure.
- Add a build regression test that targets `safari14` and parses generated federation chunks for top-level `await`.
- Add an opt-in real-browser waterfall benchmark for issue #1041.

## Benchmark

Workload: 21 exposes, approximately 1.5 MB each, with a 120 ms response delay.

| Metric | Global queue | Per-expose map |
| --- | ---: | ---: |
| `maxInFlight` | 1 | 6 |
| `startSpanMs` | ~2,567 ms | ~420 ms |
| `totalMs` | ~2,739 ms | ~581 ms |
| Expose bytes | 31,510,543 | 31,510,543 |

The browser limited concurrent HTTP requests to six in this run. The detailed procedure is in `docs/EXPOSE_QUEUE_BENCHMARK.md`, and the successful benchmark command is listed in Validation below.

## Validation

- `pnpm test` — 49 test files, 1,033 passed, 1 skipped
- `pnpm test:integration` — 15 test files, 61 passed, 1 skipped
- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm benchmark:expose-queue`
- `git diff --check`

## Compatibility and risk

- No public API or configuration change.
- CSS de-duplication behavior is unchanged.
- The benchmark uses Chromium; a Safari/WebKit runtime is not available in the current environment. Safari coverage is based on the generated-code AST regression and the existing Promise readiness barriers.
- User-authored top-level await inside exposed modules is unchanged; this PR prevents federation-generated control code from introducing TLA races.

## Review focus

- Confirm that the in-flight map is scoped to one expose key and removes failed promises so a later request can retry.
- Confirm that shared/remote readiness remains in the existing Promise barriers, rather than in the expose scheduler.
- Confirm that the benchmark remains isolated in its own revertible commit.

## Related

- Issue: [module-federation/vite#1041](https://github.com/module-federation/vite/issues/1041)
- Previous queue change: [#609](https://github.com/module-federation/vite/pull/609)
- No-TLA refactor: [#638](https://github.com/module-federation/vite/pull/638)


gioboa (commented):
Could you double check if #604 is still fixed?

dmchoi77:
@gioboa I double-checked this directly in Safari Technology Preview 249 through the Safari MCP server. I built a remote with three exposes sharing the same dependency and invoked all expose factories concurrently. 
All three resolved successfully, including the generated __loadShare__ chunk, with zero Safari console errors and no ReferenceError: Cannot access ... before initialization. This confirms that the per-expose queue change does not regress #604. This was verified on current Safari Technology Preview/WebKit rather than Safari 18 specifically.

gioboa (approved):
Thanks for this fix and Safari verification @dmchoi77 
Great work 👏

gioboa (commented):

Image

PR #1044, opened 2026-08-08
perf: reuse SSR remote entry source in dev paths

tonoizer:
## Summary

- Reuse the cached SSR remote-entry source in the dev-server SSR entry middleware.
- Reuse the same source for virtual SSR entry `load()` calls.
- Preserve Nuxt-specific URL rewriting on a local string copy.
- Add coverage proving repeated load and middleware requests generate the base source once.

## Motivation

The existing SSR source cache covered multi-output build generation, but the dev-server middleware and virtual-module `load()` path still regenerated the same deterministic source. This follow-up applies the cache consistently across all SSR remote-entry generation paths.

## Testing

- `git diff --check`
- `pnpm exec vitest run src/plugins/__tests__/pluginSSRRemoteEntry.test.ts` *(not run: `node_modules` is unavailable in this checkout)*
- `pnpm exec tsc -p tsconfig.json --noEmit` *(not run: `node_modules` is unavailable in this checkout)*

## Risks

- Low-risk performance-only change; generated source and Nuxt rewrites remain unchanged.
- The cache is scoped to the plugin instance and uses normalized federation options.

## Related

- Follow-up to #1040, which added the build-output cache.


gioboa (approved):
Thank you @tonoizer 👏

PR #1045, opened 2026-08-08
refactor: clarify SSR entry cleanup paths

tonoizer:
## Summary

- Remove the no-op Rolldown verification block from the SSR entry generator.
- Clarify that the shared plugin instance can span build environments and outputs.
- Rename the path-containment helpers to distinguish canonical and already-resolved paths.

## Motivation

These are behavior-preserving maintenance cleanups in the SSR entry plugin. The previous Rolldown block checked a chunk but had no observable success path, and the two containment helpers had similar names despite different symlink and boundary semantics.

## Testing

- `git diff --check`
- `pnpm exec vitest run src/plugins/__tests__/pluginSSRRemoteEntry.test.ts` *(not run: `node_modules` is unavailable in this checkout)*
- `pnpm exec tsc -p tsconfig.json --noEmit` *(not run: `node_modules` is unavailable in this checkout)*

## Risks

- Low-risk, behavior-preserving refactor.
- Path helper implementations are unchanged; only their names and call sites are clarified.
- No security behavior or generated output semantics are intentionally changed.

## Related

- Independent companion to #1044.


gioboa (approved):
Thank you @tonoizer  👏

PR #1047, opened 2026-08-08
fix: match zero-whitespace side-effect imports in SSR entries

### LINKED ISSUE #1046: [Bug]: SSR entry loader's import-discovery regex misses zero-whitespace side-effect imports (ERR_UNSUPPORTED_ESM_URL_SCHEME)
### Describe the bug

The SSR entry loader's default `'temp-file'` strategy fetches a remote's SSR entry, rewrites its relative ESM specifiers to absolute HTTP URLs, then **separately** scans the same source to decide which of those rewritten imports must be fetched and inlined into a local temp file (Node cannot import an `http:` specifier directly).

Those two steps use two different regexes, and they disagree about a side-effect import with **zero whitespace** between `import` and the string literal — exactly what a minifier emits.

- **Rewrite** — [`ssrEntryLoader.ts:713`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/utils/ssrEntryLoader.ts#L713) uses `import\s*` (zero-or-more), so it **matches** `import"./chunk.js"` and rewrites it to an absolute URL.
- **Discovery** — [`ssrEntryLoader.ts:800`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/utils/ssrEntryLoader.ts#L800) uses `import\s*(?:\(|\s)`, requiring either a literal `(` or **at least one** whitespace character. Zero whitespace with no paren satisfies neither, so this shape is **invisible** to discovery.

Net effect: the specifier is rewritten to an absolute `http://…` URL but never queued for inlining. The generated temp file then contains a bare `import"http://…"`, which Node's ESM loader rejects with `ERR_UNSUPPORTED_ESM_URL_SCHEME`.

### This looks like an accidental regression, not a deliberate asymmetry

The discovery pattern originally handled the zero-whitespace form. `git log -L` on that line shows it changed exactly twice:

- introduced in `b1f2e4d` (*"feat: improve SSR support"*, #715) as `import\s*\(?\s*` — paren optional, so `import"./x.js"` matched;
- narrowed in `d1ee72d` (*"fix: change htmlFileNames logic"*, #716) to `import\s*(?:\(|\s)`.

The substantive change in #716 was broadening the **capture group** from relative-only to any specifier (which is why the `startsWith('./')` filter appears in the loop). The boundary narrowing looks like collateral, unrelated to that commit's stated purpose — the old pattern already handled the nested-`import()` case its comment cites as justification, and no test in the suite exercises or depends on the exclusion.

### Version

Vite `8.2.0` (host). `@module-federation/vite@1.20.5` (current npm `latest`), `@module-federation/runtime@2.8.2`.

Also verified on `origin/main` at `107223d72d0b14cd63c6e494330b7ed94196f324` — the two regex patterns are unchanged there; only their line numbers differ (`709`/`796` at the `1.20.5` tag vs `713`/`800` on `main`, accounted for by an unrelated dev-mode hunk earlier in the file). **Reproduces on latest published and on current `main`.**

### Reproduction

https://gist.github.com/Gmendesfonseca/221b398403c93856d8bba1ceca5e4dc3

Dependency-free — no `npm install`, no network, no build step. `node repro.mjs`.

It extracts both regexes from the pinned source, asserts they **agree** on a normally-spaced import first (so a broken instrument can't be mistaken for a passing check), then shows the disagreement on the zero-whitespace form and feeds Node the exact statement the mismatch produces.

### Relevant log output

```shell
[control] both regexes agree on `import "./chunk.js";`  -> OK

[classify] input: import"./chunk.js";
  rewrite regex   (ssrEntryLoader.ts:713) matches -> true
  discovery regex (ssrEntryLoader.ts:800) matches -> false
  -> disagreement confirmed: rewritten but never queued for fetch-and-inline.

[rewrite] "import"./chunk.js";" -> "import"http://127.0.0.1:5999/assets/chunk.js";"
[node]    import("/tmp/mf-vite-r2-7NFu0g/ssr-entry.mjs") ...
  -> threw ERR_UNSUPPORTED_ESM_URL_SCHEME: Only URLs with a scheme in: file and data are supported by the default ESM loader. Received protocol 'http:'

Reproduced: ERR_UNSUPPORTED_ESM_URL_SCHEME.
```

### Expected vs actual

- **Expected:** every specifier the rewrite step turns into an absolute URL is also queued by the discovery step for local fetch-and-inline.
- **Actual:** `import"./x.js"` is rewritten but never queued, so the temp file contains a bare `http:` import specifier that Node refuses.

### I intend to submit a PR

I have a fix prepared that restores the original boundary (`import\s*\(?\s*`) while keeping #716's broadened capture group, plus a regression test. All four gates pass locally (`pnpm test` 1033 → 1034 passed, skip count unchanged at 1; `test:integration`, `typecheck`, `fmt.check` unchanged and green), and the new test was verified red before the fix. Two additional mutations confirm existing tests still guard the `import (` and `import(` forms, so the change can't silently regress them.

That said — I'd rather follow your preference than push my own shape. There are ~5 other files doing regex-based import/export scanning (`pluginRemoteNamedExports.ts`, `pluginSSRRemoteEntry.ts`, `bundleHelpers.ts`, `controlChunkSanitizer.ts`, `htmlEntryUtils.ts`); if you'd prefer a shared pattern across them rather than a local fix, I'm happy to rework it that way, or to leave the fix to you entirely.

### Known limitations

- The reproduction exercises the regex disagreement and Node's rejection of the resulting output directly against the pinned source; it does not drive `ssrEntryLoaderPlugin()` end-to-end against a live remote.
- Whether a given project hits this depends on its built output shape — a build that emits no zero-whitespace bare relative import won't trigger it, so reachability is topology-dependent even though the regex disagreement itself is version-independent across `1.20.5` and `main`.
- Verified on Node v24.13.0 / Vite 8.2.0 only, not across the full `^5`–`^8` peer range.

### Validations

- [x] Checked that there isn't already an issue reporting the same bug (searched `ssrEntryLoader`, `ERR_UNSUPPORTED_ESM_URL_SCHEME`, `temp-file`, "minified import" — nearest hit #753 is an unrelated resolve-path error).

Gmendesfonseca:
Closes #1046

## Problem

A zero-whitespace relative side-effect import — `import"./chunk.js";`, which minified Vite/Rolldown output commonly emits — gets rewritten to an absolute HTTP URL during SSR entry loading, but is never queued for fetch-and-inline into the temp file. Node's ESM loader then rejects the resulting file with `ERR_UNSUPPORTED_ESM_URL_SCHEME`.

## Mechanism

Two regexes in `ssrEntryLoader.ts` classify the same specifiers and disagree:

- **Rewrite** ([`ssrEntryLoader.ts#L713`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/utils/ssrEntryLoader.ts#L713)) — `import\s*` allows zero or more whitespace, so it matches and rewrites `import"./chunk.js"` to an absolute HTTP URL.
- **Discovery** ([`ssrEntryLoader.ts#L800`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/utils/ssrEntryLoader.ts#L800)) — `import\s*(?:\(|\s)` requires either `(` or at least one whitespace character after `import`, so `import"./chunk.js"` never matches and its specifier is never queued for fetch-and-inline.

Net effect: the rewrite runs, the fetch doesn't, and the temp file ends up with a bare `http:` import specifier Node can't load.

The discovery pattern's `(?:\(|\s)` requirement was introduced in a refactor that broadened the specifier capture group to support nested `import()` call-sites ([d1ee72df](https://github.com/module-federation/vite/commit/d1ee72df7438b36edc8d35f8e8a7a1ba987f8e11)). The pattern it replaced (`import\s*\(?\s*` — both boundary characters optional) already handled that case, and the zero-whitespace case. Nothing in that commit's comment or diff, or in any test since, depends on excluding zero-whitespace — as far as I can tell this was an unintended side effect of the broadening, not a deliberate constraint.

## Fix

Restore the discovery regex's `import` alternative to `import\s*\(?\s*`, matching the rewrite regex's leniency. The loop's existing `m[1].startsWith('./') || m[1].startsWith('../')` filter and `isVitePreloadHelperSpecifier` exclusion are untouched, so this cannot admit bare (non-relative) specifiers — it only changes which spellings of a relative import are recognized.

```diff
     // Collect relative HTTP imports before transforming. Keep this pattern
-    // broad enough for nested import() call-sites generated by Vite/Rolldown.
+    // broad enough for nested import() call-sites generated by Vite/Rolldown,
+    // and for the zero-whitespace `import"./x.js"` side-effect form that
+    // minified Vite/Rolldown output commonly emits — mirror transformSsrCode's
+    // rewrite pattern above so discovery and rewriting agree on what counts
+    // as a relative import.
     const relImports: string[] = [];
-    const relRegex = /(?:from|export\s*\*\s*from|import\s*(?:\(|\s))\s*["'`]([^"'`\s]+)["'`]/g;
+    const relRegex = /(?:from|export\s*\*\s*from|import\s*\(?\s*)\s*["'`]([^"'`\s]+)["'`]/g;
```

## Tests

New test: `fetches a zero-whitespace relative side-effect import` in `src/utils/__tests__/ssrEntryLoader.test.ts`. It fetches an SSR entry whose body is `import"./assets/chunk.js";export async function init() {}` and asserts the chunk was actually fetched, and that the written temp-file source contains no bare `http:` import specifier.

RED, before the fix (`pnpm test -- -t zero-whitespace`):

```
 ❯ src/utils/__tests__/ssrEntryLoader.test.ts (51 tests | 1 failed) 465ms
     × fetches a zero-whitespace relative side-effect import 10ms

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/utils/__tests__/ssrEntryLoader.test.ts > ssrEntryLoaderPlugin — code transformation > fetches a zero-whitespace relative side-effect import
AssertionError: expected false to be true // Object.is equality

- Expected
+ Received

- true
+ false

 ❯ src/utils/__tests__/ssrEntryLoader.test.ts:994:92
    992|       remoteInfo: { name: 'r', entry: 'http://localhost:5001/remoteEnt…
    993|     });
    994|     expect(fetch.mock.calls.some((c) => c[0] === 'http://localhost:500…
       |                                                                                            ^
    995|       true
    996|     );

 Test Files  1 failed | 48 passed (49)
      Tests  1 failed | 1033 passed | 1 skipped (1035)
```

(Trimmed: vitest's pre-existing `vi.mock` hoisting deprecation warnings, unrelated to this change, are omitted from the excerpt above.)

Mutation check: stashing just the source change (test left in place) reproduces this exact failure; restoring it goes green again. Two further mutations confirm the widening isn't a vacuous pass-everything change — requiring a mandatory `(` breaks the pre-existing `finishes fetching circular relative import graphs` test (plain `import "..."` with a space, no parens); requiring mandatory whitespace with no paren option breaks the pre-existing `fetches loader-wrapped template literal dynamic imports` test (`import(...)`). Both pre-existing forms stay independently guarded.

Gates, run via `corepack pnpm …` on `origin/main` @ `107223d`:

| Gate | Before | After |
| --- | --- | --- |
| `pnpm test` | 1033 passed / 1 skipped / 1034 total | 1034 passed / 1 skipped / 1035 total |
| `pnpm test:integration` | 61 passed / 0 skipped / 61 total | 61 passed / 0 skipped / 61 total |
| `pnpm typecheck` | clean | clean |
| `pnpm fmt.check` | clean | clean |

Skip count is unchanged; the only new result is the new test moving from failing to passing.

## Known limitations

- Verification is author-run, not independently reproduced.
- This widens discovery to also recognize `import"x"` wherever it appears in the fetched SSR module source, not anchored to a statement boundary — but none of the existing alternatives (`from`, `export * from`, `import(`, `import `) are anchored either, so this doesn't introduce a new class of false positive, just extends an existing one.
- I didn't audit the other files in this repo doing regex-based import/export scanning (`pluginRemoteNamedExports.ts`, `pluginSSRRemoteEntry.ts`, `bundleHelpers.ts`, `controlChunkSanitizer.ts`, `htmlEntryUtils.ts`) for the same zero-whitespace gap. Kept this PR to the one file/one concern that has a reproducible failure; happy to open follow-up issues for the others, or fold the pattern into a shared helper, if that's useful.

---

Happy to reshape this however you'd prefer — different layer, a shared helper instead of a local regex tweak, or something else entirely. Flagging that up front because the last PR I opened against this repo proposed a config opt-out and you preferred fixing the underlying condition directly instead; I don't want to assume a one-line regex change is the right shape here either without checking.


gioboa (approved):
Nice catch 👏 
Thanks @Gmendesfonseca

PR #1050, opened 2026-08-09
fix: prefer Vite SSR, fallback Node

### LINKED ISSUE #1048: [Bug]: isNodeServer() false-positives in browser realms that shim process.versions.node, inconsistent with SERVER_ENV_GUARD
### Describe the bug

This package answers "am I running on the server?" three different ways, in three places, and they don't agree:

1. **`isNodeServer()`** — `typeof process.versions.node === 'string'` ([`ssrEntryLoader.ts#L46-L50`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/utils/ssrEntryLoader.ts#L46-L50))
2. **`SERVER_ENV_GUARD`** — `import.meta.env.SSR` ([`ssrCapabilities.ts#L4-L8`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/utils/ssrCapabilities.ts#L4-L8))
3. **Generated remote wrappers** — `typeof globalThis.window === 'undefined'` ([`virtualRemoteEntry.ts#L1245`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/virtualModules/virtualRemoteEntry.ts#L1245))

Both failure directions this package cares about are real, and each was fixed in only one of the two places sharing the underlying signal:

- **Direction A** — a genuine Node SSR realm where a DOM shim defines `window`. A `typeof window === 'undefined'` guard breaks this. #859 (`907c08d`) fixed it for `isNodeServer()` by switching to `process.versions.node`, and renamed the regression test to `'still intercepts on Node when a DOM shim defines window'`. That test is live and passing today.
- **Direction B** — a browser-like realm where a framework shims `process.versions.node` onto `globalThis.process` (Nuxt's client shim is the example cited in this repo's own comment). `process.versions.node` alone false-positives here. #922 (`701ee34`) fixed exactly this for `SERVER_ENV_GUARD` by switching *away* from `process.versions.node` to `import.meta.env.SSR`.

`isNodeServer()` was never updated for direction B. It still uses precisely the signal that `SERVER_ENV_GUARD`'s own comment — one file over — identifies as unsafe for this purpose:

> Use Vite's environment flag rather than `process.versions.node`. Frameworks such as Nuxt can expose a browser `process` shim containing `versions.node`, which made client remote wrappers enter the SSR bootstrap and silently resolve through its fallback instead of fetching the browser remote entry.

**Observable cost:** in a realm with `window` defined **and** a `process.versions.node` shim, `isNodeServer()` returns `true` and `ssrEntryLoaderPlugin().loadEntry()` makes **4 real `fetch()` calls** trying to discover an SSR entry — network work a browser should never do. The failures are caught internally, so `loadEntry()` resolves `undefined` either way; the return value never reveals it. Only the `fetch` calls do.

### Version

Vite `8.2.0` (host). `@module-federation/vite@1.20.5` (current npm `latest`), `@module-federation/runtime@2.8.2`.

Also verified on `origin/main` at `107223d72d0b14cd63c6e494330b7ed94196f324` — `isNodeServer()` is byte-identical there (`ssrEntryLoader.ts:48`). **Reproduces on latest published and on current `main`.**

### Reproduction

https://gist.github.com/Gmendesfonseca/6da1f0beb5f5d9681396044bf26c2b2b

`npm install && node repro.mjs` — one script, one dependency (this package), no build step.

It defines `window`/`document` on a realm that also has Node's real `process.versions.node`, spies on `fetch`, and calls `ssrEntryLoaderPlugin().loadEntry()`. The spy is validated with a known-positive control first — one direct `fetch()` asserted to record exactly one call — so an inert spy can't be mistaken for "no calls made".

### Relevant log output

```shell
[control] fetch spy recorded 1 call after 1 direct fetch() -> OK

realm: typeof window = object, typeof document = object, process.versions.node = 24.13.0

loadEntry() resolved: undefined
fetch() call count: 4
  [1] HEAD http://127.0.0.1:5999/__mf_server__/remoteEntry.ssr.js
  [2] GET http://127.0.0.1:5999/mf-manifest.json
  [3] HEAD http://127.0.0.1:5999/remoteEntry.ssr.js
  [4] HEAD http://127.0.0.1:5999/__mf_ssr__/remoteEntry.ssr.js
```

### Expected vs actual

- **Expected:** a browser-like realm makes no SSR-entry-discovery requests, regardless of whether a framework has shimmed `process`.
- **Actual:** four requests are attempted, then silently swallowed.

### I'm deliberately not proposing a patch

The obvious change — adding `&& typeof window === 'undefined'` to `isNodeServer()` — is wrong, and I want to say so rather than have you point it out. It would reintroduce exactly what #859 fixed and break the live `'still intercepts on Node when a DOM shim defines window'` test. Direction A and direction B cannot both be satisfied by adding a single condition to the current signal.

The direction that looks promising to me is adopting `import.meta.env.SSR` / `SERVER_ENV_GUARD` at this call site too — the same migration #922 already made for the sibling check — but I haven't verified that against this call site's constraints, and `isNodeServer()` runs in a context where a build-time-substituted flag may not be available. So I'm filing the inconsistency and the reproduction, and leaving the choice to you. Happy to do the work on whichever approach you prefer.

### Known limitations

- The reproduction shims `window`/`document` onto a Node realm rather than running inside a real browser with a framework's `process` shim. That's the same shape the maintainers' own comment describes, but it is a simulation of that environment, not the environment itself.
- I have not measured whether this causes user-visible breakage beyond the wasted requests — the fallback resolves, so the cost may be limited to latency and noise rather than incorrect behaviour. I'd rather state that than overclaim severity.
- Verified on Node v24.13.0 / Vite 8.2.0 only.

### Validations

- [x] Checked that there isn't already an issue reporting the same bug (searched `isNodeServer`, `process.versions.node`, `SERVER_ENV_GUARD`, "Nuxt shim", "SSR detection" — no match for this mechanism).

gioboa:
Close #1048

Use Vite’s SSR flag first, with Node detection as fallback.

### DIFF
--- a/src/utils/ssrEntryLoader.ts
+++ b/src/utils/ssrEntryLoader.ts
@@ -43,11 +43,16 @@ async function nodeImport(id: string): Promise<unknown> {
   return importCache.get(id);
 }
 
-// DOM shims can define window/document on the server. Node's version marker is
-// not affected by those shims, and is absent in real browser environments.
-const isNodeServer = (): boolean =>
-  typeof (globalThis as { process?: { versions?: { node?: string } } }).process?.versions?.node ===
-  'string';
+// Detect whether loadEntry should intercept browser remote loading for SSR.
+// Prefer Vite's realm-specific flag; fall back for direct, untransformed Node usage.
+const isNodeServer = (): boolean => {
+  const viteSsr = (import.meta as ImportMeta & { env?: { SSR?: boolean } }).env?.SSR;
+  return (
+    viteSsr ??
+    typeof (globalThis as { process?: { versions?: { node?: string } } }).process?.versions
+      ?.node === 'string'
+  );
+};
 
 // ---------------------------------------------------------------------------
 // Vite 8+ ModuleRunner path (dev mode only)


gioboa (commented):

Image

PR #1051, opened 2026-08-09
fix: fix SSR shared singleton resolution

### LINKED ISSUE #1049: [Bug]: server-rendered remote's shared singleton resolves to its own inlined copy instead of the host's instance, silently
### Describe the bug

A remote exposes a module that imports a package declared as a shared singleton. The remote is built as a **proper `vite build --ssr` remote** (two passes: client, then `--ssr`), and consumed server-side by a host through the standard runtime API — `registerShared` → `registerRemotes` → `loadRemote`.

The exposed module ends up bound to the remote's **own inlined copy** of the shared package rather than the instance the host registered. Nothing throws, nothing hangs, and the exposed function returns a normal-looking value. Sharing is simply lost, silently.

The reproduction detects this with a **per-evaluation identity marker** rather than a static value — a static export can't reveal this, since a duplicate copy exports the same literal and any assertion on it passes whether sharing worked or not.

### Version

Vite `8.2.0`. `@module-federation/vite@1.20.5` (current npm `latest`), `@module-federation/runtime@2.8.2`.

Also verified against a from-source build of `origin/main` at `107223d72d0b14cd63c6e494330b7ed94196f324` — same result, three consecutive runs, no flake. **Reproduces on latest published and on current `main`.**

I checked the two most recent commits touching shared initialisation and neither changes this path: `e9da956` (#1042, lazy shared initialisation) adds export-detection memoisation and touches an unrelated `ModuleRunner` transport path; `107223d` (#1043) only changes per-expose load queuing in `virtualExposes.ts`. Confirmed by diff **and** by rebuilding from source and re-running.

### Reproduction

https://github.com/Gmendesfonseca/mf-vite-ssr-shared-binding-repro

```
npm install
npm run build     # client pass, then --ssr pass
node host.mjs     # exits non-zero on mismatch
```

A known-positive control runs first: it seeds the exact cache key the generated proxy reads from with a sentinel and confirms the remote reports that sentinel back. Without it, "wrong instance" could just mean "we misread the cache shape."

### Relevant log output

```shell
[control] remote reported the manually-seeded cache value ("control-sentinel-value") -> OK

host's own shared-lib instance id:      mskwlgan-zbevz7qc2tt
instance id the remote's Widget saw:    mskwlge6-fshsy9qag1d

MISMATCH -- the remote used a different shared-lib instance than the one
the host registered as the singleton. Sharing was lost silently: no error,
no hang, renderWidget() returned a normal-looking value either way.
```

### Expected vs actual

- **Expected:** a host that registers a shared singleton and loads a remote through the standard container protocol gets that instance used by the remote's exposed module.
- **Actual:** the exposed module resolves the package to its own bundled copy. The container's `init(shareScope)` — which `loadRemote()` calls — never writes to the cache the exposed module's shared-import proxy actually reads from.

### Suspected root cause

`generateLazyWorkspaceSingletonExports` ([`virtualShared_preBuild.ts#L1643-L1699`](https://github.com/module-federation/vite/blob/107223d72d0b14cd63c6e494330b7ed94196f324/src/virtualModules/virtualShared_preBuild.ts#L1643-L1699)) emits one template containing both an `if (import.meta.env.SSR)` branch and an `else` branch, relying on Vite's standard per-target inlining of `import.meta.env.SSR` to select which survives in each compiled output.

That inlining isn't the issue. The question is what the SSR branch does once selected: it treats *"no shared instance cached yet"* and *"the local copy is an acceptable substitute"* as equivalent, and it makes that decision **synchronously during the SSR chunk's own module evaluation** — before `loadRemote()` has returned control to any calling code. So a host has no point at which it could intervene.

The non-SSR branch at least creates a window (a promise pushed onto `pendingShareLoads`) during which something could resolve it. In the SSR-targeted output that `else` branch is eliminated, so the window doesn't exist.

**This may well be deliberate** — an SSR render path generally doesn't want to await late-arriving async shared state, and preferring a synchronous local copy is a reasonable way to guarantee that. I'm inferring the mechanism from reading the generated output, not from any statement of intent, so I can't tell whether this is a bug or a documented-nowhere design constraint.

If it's intended, I think it's worth documenting explicitly: a hand-built SSR host cannot participate in shared-singleton resolution for a remote's exposed modules, and any measurement of "is my singleton actually shared" taken server-side will silently read a duplicate. That's a surprising property to discover empirically, and it's invisible — there's no warning, no error, and the render output looks correct.

### I'm not proposing a fix

The mechanism spans several generated virtual modules plus a scoping decision whose rationale isn't established from the outside, so any patch I wrote would be speculation. I'd rather bring you the reproduction and let you decide whether this is a defect, a documentation gap, or working as designed. Happy to do the implementation work on whichever you prefer.

### Known limitations

- Verified on Vite `8.2.0` only, not across the `^5`–`^8` peer range.
- Verified for `singleton: true` with workspace-package resolution only — not `eager`, not a non-default `scope`, not the `node_modules`-resolved variant of the same code path.
- Host and remote run in one process (loopback HTTP, in-process `import()`); genuinely separate OS processes weren't tested.
- The host is a plain script consuming `@module-federation/runtime` directly rather than a second Vite build. That's deliberate — it isolates the remote's SSR output as the variable — but it means this doesn't cover a Vite-built SSR host consuming the same remote.
- The `import.meta.env.SSR` elimination explanation comes from comparing the two build outputs, not from tracing Rolldown internals.

### Validations

- [x] Checked that there isn't already an issue reporting the same bug (searched `pendingShareLoads`, `loadRemote shared`, `shared singleton SSR`, `init shareScope` — nearest hits #895/#849/#866/#874/#892 are distinct, already-fixed shared-loading races).

gioboa:
Close #1049

Seed SSR’s shared cache from runtime-selected host singletons during container initialization.
Prevents remotes silently using local copies.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntrySSR.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntrySSR.test.ts
@@ -40,4 +40,34 @@ describe('virtualRemoteEntrySSR', () => {
     expect(code).toContain('initRes.initShareScopeMap(scopeName, scopeShare)');
     expect(code).toContain('initRes.initializeSharing(scopeName');
   });
+
+  it('caches host-provided SSR singletons during container init', () => {
+    const options = getDefaultMockOptions({
+      name: 'remote',
+      shared: {
+        react: {
+          name: 'react',
+          version: '19.0.0',
+          scope: 'default',
+          from: '',
+          shareConfig: { singleton: true, requiredVersion: '^19.0.0' },
+        },
+        lodash: {
+          name: 'lodash',
+          version: '4.17.21',
+          scope: 'default',
+          from: '',
+          shareConfig: { singleton: false, requiredVersion: '^4.17.21' },
+        },
+      },
+    } as any);
+
+    const code = generateRemoteEntrySSR(options);
+
+    expect(code).toContain('const sharedSingletons = {"react"');
+    expect(code).not.toContain('"lodash"');
+    expect(code).toContain('const factory = await initRes.loadShare(pkg');
+    expect(code).toContain("cache[scopeName + ':' + pkg] ??= module");
+    expect(code.slice(0, code.indexOf('async function'))).not.toContain('await ');
+  });
 });
--- a/src/virtualModules/virtualRemoteEntrySSR.ts
+++ b/src/virtualModules/virtualRemoteEntrySSR.ts
@@ -28,10 +28,14 @@ export function getSsrRemoteEntryFileName(browserFilename: string): string {
  */
 export function generateRemoteEntrySSR(options: NormalizedModuleFederationOptions): string {
   const virtualExposesSSRId = getVirtualExposesSSRId(options);
+  const sharedSingletons = Object.fromEntries(
+    Object.entries(options.shared).filter(([, share]) => share.shareConfig.singleton)
+  );
 
   return `
   import { init as runtimeInit } from "@module-federation/runtime";
 
+  const sharedSingletons = ${JSON.stringify(sharedSingletons)};
   let exposesMapPromise;
 
   async function getExposesMap() {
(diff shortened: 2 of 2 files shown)


gioboa (commented):

Image

PR #1057, opened 2026-08-10
fix: fix up React Router v8 wrapping

### LINKED ISSUE #1055: [Bug]: Incompatible with React Router v8 Framework Mode
### Describe the bug

Related to https://github.com/module-federation/vite/issues/976, when using React Router v8, the client entry input is not detected correctly. `entry.client.tsx` is not wrapped, and therefore module federation doesn't run at all in the client.

In another, more complex project, the wrong file is wrapped and breaks the application.

### Version

v8.1.3

### Reproduction

https://github.com/WoodyWoodsta/react-router-mod-fed-entrypoint-repro-custom-server/tree/react-router-v8

### Relevant log output

```shell

```

### Validations

gioboa:
Close #1055

Wrap React Router v8’s hidden dev client entry so Module Federation initialises correctly.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -510,6 +510,35 @@ describe('pluginAddEntry', () => {
     expect(result?.code).not.toMatch(/^import "\/virtual\/hostInit\.js";\nimport/m);
   });
 
+  it('wraps React Router v8 default client entry during dev serve', async () => {
+    const [servePlugin] = addEntry({
+      entryName: 'hostInit',
+      entryPath: '/virtual/hostInit.js',
+      inject: 'entry',
+    });
+
+    runConfig(
+      servePlugin,
+      {} as ConfigPluginContext,
+      {},
+      { command: 'serve', mode: 'development' }
+    );
+    runConfigResolved(servePlugin, {
+      root: '/repo',
+      base: '/',
+      build: { rollupOptions: {} },
+    } as unknown as ResolvedConfig);
+    const code = await runTransform(
+      servePlugin,
+      'import { hydrateRoot } from "react-dom/client";\nimport { HydratedRouter } from "react-router/dom";\nhydrateRoot(document, <HydratedRouter />);',
+      '/repo/node_modules/@react-router/dev/dist/config/defaults/entry.client.tsx?v=123'
+    );
+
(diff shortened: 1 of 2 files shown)


gioboa (commented):
This is solving the issue 💪  @WoodyWoodsta

gioboa (commented):

Image

src/plugins/pluginAddEntry.ts
```diff
@@ -578,6 +578,19 @@ const __mfCurrentScript = document.currentScript;
         return getBootstrapSource(initSrc, entrySrc);
       },
       transform(code, id) {
+        // React Router v8 hides its default client entry from rollupOptions.input in dev.
+        // Match the framework entry by its stable hydration signature instead of its version/path.
+        const isReactRouterClientEntry =
+          code.includes('hydrateRoot') && code.includes('HydratedRouter');
```
WoodyWoodsta, line 584: @gioboa What happens if you abstract the entry file?

In our enterprise scale application, we provide a custom entry implementation which we import into `entry.route.tsx` - my understanding is that this logic will wrap the wrong code file.

In react router, the entry file is _always_ called `entry.route.tsx`, so I don't understand why we cannot match on the filename.

src/plugins/__tests__/pluginAddEntry.test.ts
```diff
@@ -510,6 +510,35 @@ describe('pluginAddEntry', () => {
     expect(result?.code).not.toMatch(/^import "\/virtual\/hostInit\.js";\nimport/m);
   });
 
+  it('wraps React Router v8 default client entry during dev serve', async () => {
```
WoodyWoodsta, line 513: The issue is not present only in dev, it also exist in build

src/plugins/pluginAddEntry.ts
gioboa: Thanks for your check, I created #1065 to address your suggestions.

src/plugins/__tests__/pluginAddEntry.test.ts
gioboa: Thanks for your check, I created #1065 to address your suggestions.

WoodyWoodsta: Skimming that diff - looks like we're on track. Will test it when released. Thank you for sorting that out!

PR #1058, opened 2026-08-10
fix: a failing runtime-only share provider must not reject remote `init()`

dmchoi77:
## Summary

While a remote resolves its runtime-only (`import: false`) shares during `init()`, a provider whose chunk fails to load propagates the rejection out of `init()`. The host surfaces this as `RUNTIME-015 Remote container initialization failed`, so **the whole container becomes unusable — every expose fails, including those that never import the failing share.**

This PR contains the failure inside the loop that already knows how to stop: the rejection is logged and seeding halts, exactly as the loop's own comment prescribes for an unresolved provider.

## Root cause

`src/virtualModules/virtualRemoteEntry.ts`, emitted into the remote's `init()`:

```js
// Resolve runtime-only dependencies and seed local fallbacks in dependency
// order. Stop at an unresolved provider so its consumers cannot capture an
// undefined or provisional singleton.
for (const pkg of __mfDeferredSeedKeys) {
  ...
  await __mfResolveImportFalseShared(pkg, share);   // rejects → escapes init()
  ...
  if (__mfIsRuntimeOnlySharePending(pkg)) break;
}
await __mfSeedLocalShared(__mfReadyDeferredSeedKeys);
initResolve(initRes)                                // never reached
```

`__mfResolveImportFalseShared` awaits `__mfLoadPinnedRuntimeShare` with no `try`/`catch`.

Two details shaped the fix below.

**Control flow** comes from this loop itself. It already ends at a provider left *pending* (`if (__mfIsRuntimeOnlySharePending(pkg)) break;`), and its order is dependency-derived — `__mfSeedOrder` is `getShareBatches(...).flat()`, which levels shares by their `dependencies` / `peerDependencies`. Continuing past a provider that did not resolve is what the comment warns against, so a rejection is routed into that same `break`.

**Logging** follows `__mfBridgeExternalSharedProvider` a few lines above, which reports the same class of failure rather than throwing:

```js
catch (e) {
  console.error('[Module Federation] Failed to bridge external shared module "' + pkg + '"', e)
}
```

## The fix

```diff
       if (__mfIsRuntimeOnlySharePending(pkg)) {
-        if (share.treeShaking) {
-          await __mfResolveTreeShakingShared(pkg, share);
-        } else if (share.shareConfig?.import === false) {
-          await __mfResolveImportFalseShared(pkg, share);
+        try {
+          if (share.treeShaking) {
+            await __mfResolveTreeShakingShared(pkg, share);
+          } else if (share.shareConfig?.import === false) {
+            await __mfResolveImportFalseShared(pkg, share);
+          }
+        } catch (err) {
+          // A rejected provider is an unresolved provider: stop here as the comment above
+          // prescribes, instead of escalating to a container-wide init() failure.
+          console.error(
+            `[Module Federation] Failed to resolve runtime-only shared module "${pkg}"`,
+            err
+          );
+          break;
         }
       }
       if (__mfIsRuntimeOnlySharePending(pkg)) break;
```

## Reproduction and evidence

A standalone harness — no browser needed — loads a built `remoteEntry.js`, injects a share scope built from the remote's `mf-manifest.json` where exactly one provider rejects, and calls `init()`:

```js
import * as core from '@module-federation/runtime-core';
globalThis._FEDERATION_RUNTIME_CORE = core;

const provider = (name) => ({
  version: '1.0.0',
  get: () => name === FAILING
    ? Promise.reject(Object.assign(new Error('Loading chunk 51741 failed.'), { name: 'ChunkLoadError' }))
    : Promise.resolve(() => ({ __esModule: true, default: {} })),
  shareConfig: { singleton: true, requiredVersion: '*' },
  scope: ['default'], useIn: [], from: 'host', deps: [], loaded: true, strategy: 'version-first',
});

const { init } = await import('./remoteEntry.js');
await init(shareScope, []);
```

On a remote declaring 33 shares, with a single workspace singleton set to reject:

| | before | after |
|---|---|---|
| `init()` | **rejected** — `Loading chunk 51741 failed.` | **resolved** |
| provider `get()` invocations | 34 | 34 |

The unchanged call count matters: this PR does not alter *what* is resolved or *when*, only whether one failure aborts the container.

## Behavior change

A share left unresolved now surfaces at the point a module imports it, via the existing deferred host-provided export path (`Shared module "<pkg>" was imported before federation bootstrap finished`), rather than at `init()`.

Hosts that relied on `container.init()` rejecting as the signal that a share provider failed will no longer receive it there; the failure now reaches them when a consuming module is imported.

## Tests

Adds `does not reject init when a runtime-only provider fails to load` to `src/virtualModules/__tests__/virtualRemoteEntry.test.ts`, executing the generated deferred resolution code with a rejecting `__mfResolveImportFalseShared`.

- Fails on `main`, passes with this change.
- Full suite: 49 files / 1039 tests passing.

## Out of scope

Resolution is still driven by the remote's materialized share graph rather than the requested expose's dependency closure, so shares an expose never imports are still fetched during `init()`. That needs expose-level shared dependency metadata and is better discussed separately. This PR only stops a single provider failure from taking down the whole container.


gioboa (approved):
Thanks @dmchoi77 great fix 🚀

PR #1061, opened 2026-08-10
fix: reject remote init when share negotiation fails

dmchoi77:
### Summary

This fixes an SSR remote container that logged `initializeSharing()` and singleton `loadShare()` failures while still resolving `init()`.

Shared initialization now rejects with scope/package context and writes global module-cache entries only after all share initialization succeeds. This prevents exposes from being evaluated after share negotiation fails and avoids partial cache state, unintended fallbacks, or duplicate singleton instances.

### Changes

- Collects scope-initialization and singleton-loading failures and reports them as an `AggregateError`.
- Treats a missing host provider or incompatible provider selection as an initialization failure.
- Writes cache entries only after all share initialization succeeds.
- Adds regression tests that execute the generated SSR entry.

### Impact and Compatibility

- Changes are limited to the SSR remote entry and its test file.
- The browser remote entry, public configuration types, and existing configuration APIs are unchanged.
- SSR is fail-closed by default. Consumers relying on a local/Node fallback without a host singleton provider must provide a compatible host provider.
- No separate fail-open option was added.

### Verification

- SSR targeted tests:

  `pnpm exec vitest run src/virtualModules/__tests__/virtualRemoteEntrySSR.test.ts src/virtualModules/__tests__/virtualExposesSSR.test.ts src/plugins/__tests__/pluginSSRRemoteEntry.test.ts src/virtualModules/__tests__/virtualRuntimeInitStatus.test.ts integration/ssr-shared-exports.test.ts integration/build.test.ts`

  6 files, 88 tests passed

- `pnpm test`: 49 files, 1044 tests passed, 1 skipped
- `pnpm test:integration`: 15 files, 61 tests passed
- `pnpm typecheck`, `pnpm fmt.check`, `pnpm build`, and `git diff --check`: all passed
- All commands exited with code 0. The output contained only pre-existing Vitest, TypeScript, and tsdown warnings unrelated to this change; no command failed.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntrySSR.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntrySSR.test.ts
@@ -158,14 +158,12 @@ describe('virtualRemoteEntrySSR', () => {
     expect(dynamicImport).not.toHaveBeenCalled();
   });
 
-  it('rejects init when a required singleton is absent from the host share scope', async () => {
+  it('allows the SSR local fallback when a singleton is absent from the host share scope', async () => {
     const loadShare = vi.fn();
     const runtimeInit = createRuntime(undefined, loadShare);
     const entry = createEntry(runtimeInit, { shared: { react: singleton('react') } });
 
-    await expect(entry.init({})).rejects.toThrow(
-      'scope "default" package "react": No compatible host provider was registered in the share scope'
-    );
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks @dmchoi77 it looks great to me, I did a small change to preserve fallback

gioboa (approved):

PR #1062, opened 2026-08-11
fix: resolve directory imports in the entry shared-import scan

smeng9:
`registerEntrySharedImports` crawls the entry graph to decide which shared packages to materialize. Its candidate list starts with the bare request path and is filtered with `existsSync`, which is also true for directories, so a directory request wins before the `index.*` candidates below it are tried. The crawl then reads that candidate and dies:

    Error: EISDIR: illegal operation on a directory, read
      at registerEntrySharedImports

This takes down `vite dev` startup entirely, and a bare directory import is ordinary code — `import { AppContextProvider } from './provider'` where `src/provider/` holds an `index.ts` is enough to trigger it.

Accept a candidate only when it is a file, so a directory request falls through to the `index.*` candidates that were already there for exactly this case.

Verified against a real host that could not start on 1.20.5: dev server now boots and serves, and the regression test reproduces the original EISDIR without the fix.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1151,6 +1151,49 @@ describe('vite:module-federation-early-init', () => {
     }
   });
 
+  it('resolves directory imports in the entry graph to their index file', () => {
+    const root = mkdtempSync(path.join(tmpdir(), 'mf-entry-dir-import-'));
+    mkdirSync(path.join(root, 'src'));
+    mkdirSync(path.join(root, 'src/provider'));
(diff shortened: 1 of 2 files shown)


smeng9:
Hi @gioboa 

This issue seems is happening only in unreleased main (beyond 1.20.5 on pkg.pr.new).

We are trying to import directly from a directory instead of its index.js/ts

gioboa (approved):
Thanks @smeng9 Nice catch 👏

PR #1063, opened 2026-08-11
perf(shared): cache concrete shared import source resolution

smeng9:
`getConcreteSharedImportSource` walks from the project root up to the filesystem root, probing module resolution at every level. The shared-module resolver injected into Vite's dependency optimizer asks for the same handful of packages tens of thousands of times on a cold dev start — measured on one host: 26,279 calls returning the same answers.

Cache the resolution. The package detection cwd is part of the key because `setPackageDetectionCwd` can move it between config hooks, so a plain per-package cache would return another root's answer.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -91,6 +91,7 @@ import {
   getPreBuildLibImportId,
   materializeCachedLoadShareModule,
   prependWorkspaceSingletonSsrImport,
+  resetConcreteSharedImportSourceCache,
   writeLoadShareModule,
   writePreBuildLibPath,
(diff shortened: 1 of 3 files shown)


smeng9:
We also encountered super long initial startup time for dev server similar to https://github.com/module-federation/vite/issues/1041 . Our project takes ~1min to start dev server.... 

After https://github.com/module-federation/vite/pull/1062 is merged hopefully we can use pkg.pr.new with faster speed

gioboa (approved):
Thanks for this great improvement @smeng9 
I added a small commit to reset the cache

PR #1065, opened 2026-08-11
fix: fix up React Router wrapping

gioboa:
Wraps the correct React Router client entry in development and production without affecting imported hydration helpers.


gioboa (commented):

Image

PR #1067, opened 2026-08-11
fix: avoid Webpack provider init deadlock

### LINKED ISSUE #1064: [Bug]: Vite remote inside a webpack host: same-version shared registration poisons the host's share scope (never-resolving get()); auto-added subpath shares inherit import:false and crash
### Describe the bug

A Vite-built remote consumed by a **webpack host** (`@module-federation/enhanced`) registers share-scope providers whose `get()` never resolves outside a Vite host. Two user-visible failures come out of this, both reproduced in the linked repo:

**1. Auto-added subpath shares inherit `import: false` and are unresolvable**

With `shared: { react: { singleton: true, requiredVersion: '^18.3.1', import: false }, 'react-dom': { ... } }` (the documented "host provides these" setup), the plugin automatically adds `react/jsx-runtime` / `react-dom/client` as their own share keys (`COMMON_SHARED_SUBPATHS` in `src/utils/pathNormalization.ts`), inheriting `import: false`. A typical webpack host only shares the bare packages — no host realistically provides those subpath keys — so loading the remote throws:

```
[Module Federation] Shared module 'react/jsx-runtime' must be provided by host
```

There is no way to opt out from the remote side, since the subpath keys are added implicitly. Providing the subpath from the host doesn't reliably help because of failure 2:

**2. Same-version registration overwrites the host's provider with a never-resolving `get()` — the host itself hangs**

When the remote registers a share at the **same version** the host provides (e.g. both sides pin `react-dom@18.3.1` — the common production case), the remote's entry takes the version slot in the host's share scope:

```
webpack_host | react-dom@18.3.1 from=remote_app loaded=false lib=false   <-- host's own provider is gone
webpack_host | react@18.3.1     from=webpack_host loaded=true lib=true   <-- survived only because it was consumed before the remote registered
```

That entry's `get()` never settles (it appears to wait on the plugin's host-auto-init handshake, which only a Vite host performs — probed via `Promise.race`, still pending after 3s). The host's own `import('./bootstrap')` then consumes `react-dom` through the poisoned slot and hangs forever: **blank page, zero console output**. Which package gets poisoned is a race between host consumption and remote registration.

This reproduces identically with and without `import: false` (with fallback assets present, `lib` is still `false` on the registered entry), so carrying fallbacks is not a workaround.

The remote works fine standalone and under Vite hosts; everything breaks only when embedded in a non-Vite host.

**Relation to #952:** that issue was the remote picking its *own* throwing stub during *its* consumption (fixed; commit `361cab5` is contained in 1.20.5 per the compare API — `status: behind`). This report is about the registration side breaking the **host's** resolution: webpack's consume logic reads the scope slot directly, so the consume-side fix can't help.

Happy to test patches / pre-releases.

### Version

vite 7.3.6, @module-federation/vite 1.20.5 (also reproduces on 1.20.3); host: webpack 5.109.2 + @module-federation/enhanced 2.8.1; react/react-dom 18.3.1 exact on both sides

### Reproduction

https://github.com/h11g/mf-vite-webpack-host-repro

### Relevant log output

```shell
# failure mode 1 (remote built with import:false, default `npm run build`)
[ Federation Runtime ] Version 0 from 0 of shared singleton module react/jsx-runtime does not satisfy the requirement of remote_app which needs ^18.3.1)
[Module Federation] Failed to bridge external shared module "react/jsx-runtime"
Error: [Module Federation] Shared module 'react/jsx-runtime' must be provided by host

# failure mode 2 (remote built with fallbacks, `npm run build:with-fallback`)
# -> no output at all; host page stays blank. Share scope dump:
webpack_host | react-dom@18.3.1        from=remote_app   loaded=false lib=false
webpack_host | react@18.3.1            from=webpack_host loaded=true  lib=true
webpack_host | react/jsx-runtime@18.3.1  from=remote_app loaded=false lib=false
webpack_host | react-dom/client@18.3.1   from=remote_app loaded=false lib=false
# probing the poisoned entry:
await Promise.race([e.get().then(()=>'resolved',()=>'rejected'), new Promise(r=>setTimeout(()=>r('still pending'),3000))])
# -> 'still pending'
```

dmchoi77:
## Summary

- Fix remote initialization deadlock with plain Webpack hosts.
- Defer only unresolved Webpack-compatible shared providers.
- Preserve normal materialization behavior for non-Webpack providers.
- Bridge the host shared modules after runtime initialization and before expose evaluation.
- Clarify that shared package subpaths require trailing-slash prefix matching, such as `react-dom/`.
  See the [Module Federation shared dependency documentation](https://module-federation.io/configure/shared#prefix-matching-with-trailing-slash).

Closes #1064

## Root Cause

A plain Webpack host can register a shared provider whose `get()` function waits for Webpack's own consume initialization. Calling that provider during the remote's `init()` creates a circular wait.

The workaround is now limited to providers whose getter contains `__webpack_require__`, instead of treating every unresolved foreign provider the same way.

## Validation

- `pnpm test`
- `pnpm test:integration`
- `pnpm typecheck`
- `pnpm build`
- `pnpm fmt.check`
- Webpack host reproduction:
  - Remote build succeeded
  - Webpack host build succeeded
  - Remote widget rendered
  - Clicking the widget changed `0` to `1`

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -506,7 +506,7 @@ describe('module parse wiring', () => {
 
     callHook(parseEnd.moduleParsed, ctx, { id: '/src/main.ts' } as never);
 
-    expect(await resolvesQuickly(Promise.resolve(pendingRemoteEntry))).toBe(true);
+    expect(await Promise.resolve(pendingRemoteEntry)).toEqual(expect.any(String));
   });
 
   it('keeps the complete import:false export surface when parse analysis is incomplete', async () => {
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -355,6 +355,22 @@ describe('pluginProxySharedModule_preBuild', () => {
       expect(findSharedKey('missing-dep/subpath', shared)).toBeUndefined();
       expect(findSharedKey('missing-dep/subpath', shared)).toBeUndefined();
     });
+
+    it('does not implicitly share common subpaths when the package uses import:false', () => {
+      const shared = makeShared();
+      shared.react.shareConfig.import = false;
+
+      expect(findSharedKey('react', shared)).toBe('react');
+      expect(findSharedKey('react/jsx-runtime', shared)).toBeUndefined();
+
+      const explicitlyShared = makeShared();
+      explicitlyShared.react.shareConfig.import = false;
+      explicitlyShared['react/jsx-runtime'] = {
+        ...explicitlyShared.react,
+        name: 'react/jsx-runtime',
+      };
+      expect(findSharedKey('react/jsx-runtime', explicitlyShared)).toBe('react/jsx-runtime');
+    });
   });
 
   beforeEach(() => {
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -135,12 +135,18 @@ function getSharedKeyMatcher(shared: NormalizedShared | undefined): SharedKeyMat
 
   for (const key of keys) {
     const keyBase = key.endsWith('/') ? key.slice(0, -1) : key;
+    const shareItem = shared[key];
 
     if (!vueKey && keyBase === 'vue') vueKey = key;
     if (key.endsWith('/')) wildcardKeys.push({ key, base: keyBase });
 
-    for (const subpath of getCommonSharedSubpaths(keyBase)) {
-      if (!commonSubpathKeys.has(subpath)) commonSubpathKeys.set(subpath, key);
+    // `import: false` applies to the configured key only. Treating common
+    // subpaths as implicit shares creates unfulfillable runtime-only entries
+    // for hosts which provide the bare package but not every package export.
+    if (shareItem.shareConfig?.import !== false) {
+      for (const subpath of getCommonSharedSubpaths(keyBase)) {
+        if (!commonSubpathKeys.has(subpath)) commonSubpathKeys.set(subpath, key);
+      }
     }
   }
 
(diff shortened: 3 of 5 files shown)


gioboa (commented):
Thanks @dmchoi77 
Did you verify the fix with the [reproduction repo](https://github.com/h11g/mf-vite-webpack-host-repro) ?

dmchoi77:
@gioboa I verified the fix with my local example, but I haven't checked it against the reproduction repo yet. Let me take a look and verify it there as well. I'll check it now and get back to you.🙏

gioboa:
We can use this check to behave differently
```
function isWebpackProvider(provider) {
  if (typeof provider?.get !== 'function') return false;
  const source = Function.prototype.toString.call(provider.get);
  return source.includes('__webpack_require__');
}
```

dmchoi77:
@gioboa Thanks, that distinction was the missing piece.

I updated the fix to detect Webpack-compatible providers based on their getter source:

```js
function isWebpackProvider(provider) {
  if (typeof provider?.get !== 'function') return false;
  const source = Function.prototype.toString.call(provider.get);
  return source.includes('__webpack_require__');
}
```

The workaround now defers only unresolved Webpack providers during remote.init(). Other providers, including Vite providers, continue through the normal materialization path.

The late bridge still runs after sharing initialization and before expose evaluation, so the Webpack host’s React instance is selected once it becomes available.

Verified with the Webpack host repro and the full test suite.

gioboa (commented):
Thanks! Tested the latest diff against #1064:

- Fallback build works.
- `import:false` still hangs; `react-dom@18.3.1` remains owned by `remote_app`.

Likely because Webpack guards require `!scopeRoot`, but enhanced Webpack provides a `scopeRoot`. Provider detection should apply regardless of `scopeRoot`.
Please test both repro build modes.

dmchoi77:
@gioboa Thanks, this was the missing case.

I updated the Webpack provider guard to apply regardless of `scopeRoot`, and added the same check to the `import:false` deferred-resolution path using the initial host provider snapshot. This prevents `remote.init()` from calling an unresolved Webpack provider getter.

I verified both repro modes with a fresh package:

- `pnpm run build` (`import:false`)
- `pnpm run build:with-fallback`

Both modes rendered successfully with the Webpack host providing the exact subpath shares (`react/jsx-runtime` and `react-dom/client`).

gioboa (approved):
Thanks @dmchoi77 
Detecting Webpack and handling this scenario differently is definitely the right approach. Great PR! 🚀

gioboa (commented):
This fix matches Webpack/Rspack semantics:
  - react: { import:false } → only react must come from host.
  - react/jsx-runtime bundles locally.
  - Explicit react/ or react/jsx-runtime with import:false → subpaths must come from host.

gioboa:

Screenshot 2026-08-11 at 23 01 22

PR #1042, opened 2026-08-07
perf: lazy shared initialization improves startup performance

gioboa:
Close #1041

## Performance results

| Metric | 1.17.0 | Fixed | Improvement |
| :-- | --: | --: | --: |
| Cold FCP | 2,704 ms | 2,264 ms | **16.3% faster** |
| Cold app ready | 2,906 ms | 2,355 ms | **19.0% faster** |
| Cold last resource | 2,608 ms | 2,180 ms | **16.4% faster** |
| Cold network idle | 3,194 ms | 2,755 ms | **13.7% faster** |
| Median FCP | 1,304 ms | 1,112 ms | **14.7% faster** |
| Median app ready | 1,349 ms | 1,364 ms | 1.1% variance |
| Median last resource | 1,251 ms | 1,061 ms | **15.2% faster** |
| Median network idle | 1,798 ms | 1,603 ms | **10.8% faster** |
| Resource requests | 148 | 137 | **7.4% fewer** |

## Changes

- Registers every configured provider, but materializes only imports found in the entry/expose graph.
- Preserves dependency ordering while batching independent providers.
- Deduplicates provider initialization.
- Stabilizes optimizer inputs and caches export inspection.
- Excludes nested shared dependencies before discovery.
- Removes the unrelated remote-root preload experiment, which caused React ownership races.
- Keeps the `vite-vite` configuration unchanged.
- Uses no top-level `await`.
- Downloads no Vue package payload; only the existing `vue.svg` asset loads.

## Validation

- Unit tests: **1030 passed, 1 skipped**
- E2E tests: **25/25 passed**, including cold multi-framework React sharing
- Typecheck, build, and formatting: **passed**
- Added entry-graph and lazy-share regression coverage

Issue #419, opened 2026-02-19
[Bug]: ENOENT: no such file or directory, open Error for __mf__virtual On CI/CD pipeline first time it build


yashnextgenit1-design:
### Describe the bug

Im getting error while creating build in CI environment first time only if i retry the pipeline for creating build then its working fine. im attaching my vite.config.js file can someone help me to solve this issue.

**Error:**
[commonjs--resolver] ENOENT: no such file or directory, open '/app/node_modules/__mf__virtual/analytics__loadShare___mf_0_mui_mf_1_styles__loadShare__.js'
file: /app/node_modules/@nv2/nv2-pkg-js-shared-components/lib/Tabs/index.js
    at Object.writeFileSync (node:fs:2426:20)
    at VirtualModule.writeSync (/app/node_modules/@module-federation/vite/lib/index.cjs:1167:8)
    at writeLoadShareModule (/app/node_modules/@module-federation/vite/lib/index.cjs:1240:26)
    at Object.customResolver (/app/node_modules/@module-federation/vite/lib/index.cjs:2018:13)
    at Object.resolveId (file:///app/node_modules/vite/dist/node/chunks/config.js:8145:76)
    at Object.handler (file:///app/node_modules/vite/dist/node/chunks/config.js:33705:13)
    at file:///app/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:22494:40
    at async PluginDriver.hookFirstAndGetPlugin (file:///app/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:22376:28)
    at async resolveId (file:///app/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:20852:26)
    at async ModuleLoader.resolveId (file:///app/node_modules/vite/node_modules/rollup/dist/es/shared/node-entry.js:21289:15)

**Vite.config.js**
import fs from 'node:fs';
import path from 'node:path';
import { defineConfig } from 'vite';
import react from '@vitejs/plugin-react';
import { federation } from '@module-federation/vite';
import svgr from 'vite-plugin-svgr';
import cssInjectedByJsPlugin from 'vite-plugin-css-injected-by-js';
import { dependencies, version as packageVersion } from './package.json';
import { REMOTE_NAMES } from './src/runtime/remotes.js';

const resolveBuildVersion = () => {
  const buildId = Date.now().toString(36);

  const commitHash = process.env.VITE_COMMIT_HASH || null;

  if (commitHash) {
[...]

PR #1072, opened 2026-08-11
fix: isolate react-server share cache from client/ssr environments

ryanmagoon:
## Problem

In a vinext (vite-rsc) app, SSR of federated remote components silently fails with React error #419 ("The server could not finish this Suspense boundary"). The server ships only the Suspense fallbacks plus an error digest, and the client re-renders from scratch — the page looks correct, but nothing federated is actually server-rendered, the SSR HTML is missing content, and the browser console shows uncaught #419 errors.

Reproduced with the `module-federation/vinext` example apps (`pnpm preview`, host on 4173).

## Root cause

vite-rsc runs the RSC and SSR server environments in **one Node process**, but the RSC environment resolves React with the `react-server` condition — a build with no state hooks. The runtime share cache is keyed on a single process-global (`__mf_module_cache__`), so whichever environment initializes first publishes its React for everyone:

- On 1.20.1: the RSC environment's hostAutoInit published react-server React into `default:react`; the SSR environment's remote wrappers consumed it → `TypeError: __mf_38 is not a function` (`useState` missing) → #419.
- After #1042 excluded react from materialization for vinext: the host never publishes *any* React, so remote wrappers fall back to their own bundled copy — a different instance from the one `react-dom-server` renders with → `Cannot read properties of null (reading 'useState')` (null hooks dispatcher) → same #419 symptom.

## Fix (three layers)

1. **Per-environment cache bucket**: generated code for environments resolving with the `react-server` condition reads/writes `__mf_module_cache_react_server__`; client/ssr keep the legacy key, so already-deployed remotes stay compatible. hostAutoInit, loadShare/prebuild wrappers, and remoteEntry are regenerated per environment (reusing the existing refresh-on-load mechanism).
2. **Flavor guard**: `__mfWriteSharedCache` refuses react-family modules whose flavor (server vs client internals marker) doesn't match the bucket the code was generated for.
3. **Local-provider fallback**: the MF runtime registry is also process-global, so share negotiation can hand the SSR environment the RSC environment's react-server factory. When hostAutoInit detects the flavor mismatch it republishes the environment's **local** share provider — the same React instance the local renderer uses.

With flavor isolation in place, the vinext react-materialization exclusion from #1042 is no longer needed (it traded the poisoning for the null-dispatcher crash), so it's reverted — vinext hosts materialize react like every other app.

## Verification

- 9 new unit tests (`virtualSharedReactServerIsolation.test.ts`), full suite green.
- vinext example apps: SSR HTML now contains the remote components with no `<!--$!--><template data-dgst>` markers, browser console is clean, and both e2e tests pass (a new raw-HTML SSR assertion — PR to vinext coming — plus the existing hydration test, all counters interactive).

## Notes for reviewers

- The bucket key is derived from export conditions at codegen time (`react-server` in the requesting environment's resolve conditions), not from environment names.
- `.remote` module cache and dev-mode (`vite dev`) RSC hosts are unchanged; the dev path may deserve a follow-up.


gioboa (commented):
Hi @ryanmagoon thanks for your help.

gioboa (changes requested):
The flavor check only matters server-side. could you skip injecting it into client bundles and gate it to SSR?

ryanmagoon:
> The flavor check only matters server-side. could you skip injecting it into client bundles and gate it to SSR?

Done. while I was in there I also tightened up the react-dom flavor detection.

while validating this against the vinext apps I hit a separate regression from #1024: client-only remotes stopped emitting remoteEntry.ssr.js, so SSR hosts crash trying to import the browser entry over http. fix is in [#1073](https://github.com/module-federation/vite/pull/1073) with an e2e example covering the scenario. the vinext demo needs both PRs but they're independent code-wise, merge order doesn't matter.

gioboa (changes requested):
The bug looks valid, but the fix feels broad: ~12 files/700 LOC and changes cache generation across all environments.
Could we narrow this to the React publication boundary? Specifically, detect react-server React and prevent it from entering the normal SSR/client cache, with a local-provider fallback where needed.
If a guard alone is insufficient, could you add an integration test demonstrating why per-environment cache keys and every modified generation path are required? That would make the broader scope easier to justify.

gioboa (approved):
Thanks for your help @ryanmagoon 🙏 
Based on your work, I narrowed the fix to minimize regression risk across heterogeneous framework configurations while resolving the verified RSC failure.

gioboa:

Screenshot 2026-08-12 at 14 01 03

PR #1073, opened 2026-08-12
fix: emit SSR entry for client-only remotes again

ryanmagoon:
## Problem

Since #1024 (1.20.2), client-only remotes — apps with `exposes` but no `ssr` Vite environment of their own — no longer emit `remoteEntry.ssr.js`. Their SSR entry is consumed by the *host's* server over HTTP, so it must ship with the client assets. Consuming SSR hosts (e.g. vinext) now fall back to importing the browser remote entry with Node's ESM loader and crash:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file and data are supported by the default ESM loader. Received protocol 'http:'
```

Found while validating #1072 against the `module-federation/vinext` example apps: bisected to 09c08e5 (#1024).

## Why the regression happened

#1024's model — "the SSR entry belongs to SSR builds" — is correct for apps that have their own `environments.ssr`, but for a client-only remote the SSR entry is not an internal build artifact: it's a cross-app deliverable that a *different* app's server fetches over HTTP. The only comment documenting that invariant was attached to the Nuxt carve-out (also removed in #1024), so the general case read as accidental behavior that was safe to clean up, and the tests asserting it were inverted to match the new model rather than treated as a contract.

## Fix

`isSsrRemoteEntryBuild` regains the pre-#1024 client-pass emission for apps with no ssr environment and no legacy `--ssr` build — **Rolldown builds only (Vite 8+ / rolldown-vite)**. Unchanged from #1024: `environments.ssr` setups still emit only in the ssr environment, and the `?? viteConfig?.build?.ssr` legacy-detection fallback is kept.

**Why Rolldown-only:** review of this PR surfaced that the Rollup asset path has *never* produced a working client-only SSR entry — its `virtual:mf-exposes-ssr` import is only rewritten for Nuxt, so the emitted file throws `ERR_UNSUPPORTED_ESM_URL_SCHEME` at first expose load. We reproduced this byte-identically on published **1.20.1** with a minimal vite@7 remote (runtime `get('./Widget')` fails). Emitting nothing on Rollup keeps Vite ≤7 at exact parity with current main — a missing file fails fast in the loader, where a present-but-broken one returns 200 and masks the fallbacks. Bundling the SSR exposes graph into Rollup client builds risks the code-splitting side effects the original asset-path comment warns about, so that remains out of scope.

## Test coverage

- **Unit contract**: client-only emission pinned for Rolldown, non-emission pinned for Rollup (with the 1.20.1 evidence in the test comment), plus the pre-existing environment/legacy cases.
- **e2e (new `vite-ssr` example + Playwright project)**: this repo's e2e previously had no coverage of the SSR-consumption architecture — the suite passed while this regression was live on main. The new example (dual-build SSR host + client-only SPA remote) closes that: the spec asserts `remoteEntry.ssr.js` is served as actual JavaScript with the container's `init`/`get` exports, the federated widget appears in raw server-rendered HTML, and it hydrates with **zero hydration-mismatch console errors**. Verified red on main and green with the fix. Readiness probes a `/healthz` endpoint so regressions surface as assertion failures, not webServer timeouts.

The e2e also guards the vinext scenario's server-loading half without any vinext dependency; the React share-cache half is covered by #1072.

## Verification

- Full unit suite green (1055); full Playwright matrix green including the new project.
- vinext example apps rebuilt with this fix + #1072: both e2e tests pass end to end.

Note: the vinext demo needs **both** this PR and #1072; they are independent at the code level (zero file overlap, both based on main) and can merge in either order.

### DIFF
--- a/e2e/vite-ssr/ssr.spec.ts
+++ b/e2e/vite-ssr/ssr.spec.ts
@@ -3,14 +3,20 @@ import { expect, test } from '@playwright/test';
 // Guards the SSR-consumption architecture: a client-only remote publishes
 // remoteEntry.ssr.js with its client assets, and an SSR host's Node server
 // fetches it over HTTP to render federated components before hydration.
-// Regressed in #1024 (no SSR entry emitted → hosts crashed with
-// ERR_UNSUPPORTED_ESM_URL_SCHEME and served no federated content).
+// Regression (#1024): no SSR entry emitted → hosts crashed with
+// ERR_UNSUPPORTED_ESM_URL_SCHEME and served no federated content.
 
 test('client-only remote publishes its SSR entry as JavaScript', async ({ request }) => {
   const response = await request.get('http://localhost:5177/remoteEntry.ssr.js');
   expect(response.status()).toBe(200);
   // Guard against an SPA index.html fallback masquerading as a 200.
   expect(response.headers()['content-type']).toContain('javascript');
+  const body = await response.text();
+  expect(body.trimStart()).not.toMatch(/^</);
+  // The container contract: named `init` and `get` exports. Local identifiers
+  // are minified, but the exported names survive (`export{s as get,o as init}`).
+  expect(body).toMatch(/export\s*\{[^}]*\binit\b/);
+  expect(body).toMatch(/export\s*\{[^}]*\bget\b/);
 });
 
 test('host serves the federated widget as server-rendered HTML', async ({ request }) => {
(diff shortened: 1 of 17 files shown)


gioboa (changes requested):
We maintain runnable examples in https://github.com/module-federation/module-federation-examples. Could you move the new vite-ssr example there?

gioboa:
The behavior from #1024 is intentional and aligns with Rspack and Webpack: browser builds must not emit a Node-targeted SSR remote entry. A client-only remote should publish only the browser entry. 

For SSR, configure a separate server build.

[Here](https://github.com/gioboa/module-federation-vite-examples/blob/main/vinext/island/vite.config.ts#L20) you can find a working example of this 👇 

```
export default defineConfig({
  builder: {},
  plugins: [
    federation({
      name: 'remote',
      filename: 'remoteEntry.js',
      exposes: { './Widget': './src/Widget.tsx' },
    }),
  ],
  environments: {
    client: {
      build: { outDir: 'dist/client' },
    },
    ssr: {
      build: {
        ssr: 'src/entry-server.ts',
        outDir: 'dist/server',
      },
    },
  },
});
```

PR #1079, opened 2026-08-12
fix: isolate remote caches across environments

### LINKED ISSUE #1074: [Bug]: remote-module cache and runtime-init moduleCache are shared across Vite server environments, leaking modules between react-server and ssr graphs
### Describe the bug

#1072 partitions the **share** cache per environment so vite-rsc's react-server environment can't poison the React instance the ssr environment consumes. The **remote-module** cache has no equivalent partition: a federated module evaluated under one server environment's module graph is cached and returned to the other environment in the same Node process.

Three cooperating mechanisms, none environment-aware:

1. The loadRemote wrapper codegen never receives export conditions — the dev bootstrap ([virtualRemotes.ts#L382](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRemotes.ts#L382)) and the build cache bootstrap ([#L392](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRemotes.ts#L392)) both bind the default bucket, and the load hooks refresh `LOAD_SHARE`/`PREBUILD`/`__H_A_I__` ids per environment but never `LOAD_REMOTE`.
2. `remoteCacheKey` has no environment component ([virtualRemotes.ts#L397](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRemotes.ts#L397)) — the same remote id maps to one slot for every environment.
3. The runtime-init global pins `moduleCache` first-writer-wins: created at [virtualRuntimeInitStatus.ts#L170](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRuntimeInitStatus.ts#L170) and later only `||=` ([#L182](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRuntimeInitStatus.ts#L182)), while the `globalKey` itself is per-owner, not per-environment — so whichever environment bootstraps first decides which bucket every `globalThis[globalKey].moduleCache` consumer sees. `writeRuntimeInitStatus` ([#L267](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRuntimeInitStatus.ts#L267)) also generates the status module without conditions.

### Version

Applies to `main` @ 6d91862ea730250b0ecb3321e4a1a43d4b9eee78 and all published 1.20.x. Found during review of #1072 by code reading; mechanisms verified by quoting the generated output.

### Reproduction status — latent, honestly

We could **not** demonstrate this live in the `module-federation/vinext` example: the only remote loaded from both server environments there (the React-18 island) deliberately bundles its own React, so the cross-environment cache hit returns a flavor-neutral module. The hazard becomes live when a **server component and a client component both import the same exposed key** from a remote that shares React — then the first environment's evaluation (bound to its React flavor via share negotiation) is served to the other. That configuration is plausible for users but not covered by any example or test.

### Expected vs actual

- **Expected:** each Vite server environment resolves remote modules against its own module graph, as #1072 established for shares.
- **Actual:** `.remote` entries and the pinned `moduleCache` alias are process-global; results are initialization-order-dependent.

### Suggested direction

Thread export conditions through `generateRemotes`' two bootstrap calls, refresh `LOAD_REMOTE` ids per environment like their siblings, and either key `remoteCacheKey` by bucket or make the line-182 `||=` an always-assign from the requesting environment's bucket.

gioboa:
Close #1074

Completes environment isolation for remote-module caches, extending the shared-module isolation.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -78,6 +78,7 @@ import {
   LOAD_REMOTE_TAG,
   LOAD_SHARE_TAG,
   PREBUILD_TAG,
+  refreshRemoteModuleForEnvironment,
   TREE_SHAKING_GRAPH_QUERY,
   TREE_SHAKING_PROVIDER_TAG,
   writeLocalSharedImportMap,
@@ -1014,6 +1015,17 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
     });
   };
 
+  const refreshLoadRemoteModuleForEnvironment = (
+    id: string,
+    context: LoadHookContext,
+    loadOptions?: LoadHookOptions
+  ) =>
+    refreshRemoteModuleForEnvironment(
+      id,
+      options,
+      getLoadHookExportConditions(context, loadOptions)
+    );
+
   const refreshPreBuildModuleForEnvironment = (
     id: string,
     context: LoadHookContext,
@@ -1114,6 +1126,12 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
         return virtualModule.getResolvedId();
       },
       load(id: string, loadOptions?: LoadHookOptions) {
+        if (
+          id.includes(LOAD_REMOTE_TAG) &&
+          !refreshLoadRemoteModuleForEnvironment(id, this as LoadHookContext, loadOptions)
+        ) {
+          return;
+        }
         if (
           command !== 'build' &&
           id.includes(LOAD_SHARE_TAG) &&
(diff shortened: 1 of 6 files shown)


gioboa (commented):

Image

PR #1080, opened 2026-08-12
fix: scope shared caches per SSR environment

### LINKED ISSUE #1075: [Bug]: generated SSR remote entry writes negotiated shares into the default share cache unguarded, and the poison is sticky
### Describe the bug

The generated SSR remote entry writes host-negotiated shares into a **hardcoded** `globalThis.__mf_module_cache__` via direct `??=` assignment — bypassing both the per-environment bucket selection from #1072 and its flavor guard:

[virtualRemoteEntrySSR.ts#L116](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/virtualModules/virtualRemoteEntrySSR.ts#L116) — the entry's `init()` loads each host share via `initRes.loadShare(pkg)` and publishes it under `"<scope>:<pkg>"` plus the bare alias, straight into the default bucket, no `__mfWriteSharedCache`, no conditions.

Because the MF runtime registry is process-global, `loadShare` inside a remote's init can hand back whatever flavor another environment registered first. And the poison is **sticky**: hostAutoInit's preload loop skips any key that's already cached, so the host's own guarded publish never repairs it.

Evidence from a real artifact — any remote built by this repo's `examples/vite-ssr` emits it (minified):

```
$ grep -o 'globalThis.__mf_module_cache__[^;]*' examples/vite-ssr/vite-remote/dist/remoteEntry.ssr.js
globalThis.__mf_module_cache__||={share:{},remote:{}},t=e.share||={}
```

### Version

Applies to `main` @ 6d91862ea730250b0ecb3321e4a1a43d4b9eee78 and all published 1.20.x. Found during review of #1072; write path and reachability verified by code reading (ssrEntryLoader evaluates the entry in the host's process — temp-file strategy imports via `file://`, vm strategy in-context — so `globalThis` is shared).

### Scenario

vite-rsc host: the react-server environment's `loadRemote` evaluates a remote's SSR entry before the ssr environment has seeded `default:react`; the entry publishes a react-server-flavored React into the empty default-bucket slot; ssr consumers read it → React error #419. The `??=` means it only fills empty slots — but RSC renders before SSR, so the empty-slot case is the common one.

### Deployment caveat

This code ships in **remote build artifacts**, so a codegen fix only protects rebuilt remotes. A complete fix likely needs a host-side component too (e.g. flavor validation on read, or hostAutoInit not treating a cached entry as authoritative), which is why we left it out of #1072 rather than half-fix it.

### Expected vs actual

- **Expected:** every share-cache write path goes through the guarded, bucket-aware helper.
- **Actual:** this one writes directly to the default bucket and its values suppress the host's guarded publishes.

gioboa:
Close #1075

SSR remotes now write negotiated singletons into the requesting environment’s cache.
Hosts repair stale unscoped entries from older remotes.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -2468,6 +2468,9 @@ describe('virtualRemoteEntry', () => {
 
     expect(code).toContain('const __mfGetSharedCacheDescriptor =');
     expect(code).toContain('__mfReadSharedCache(__mfModuleCache.share, cacheDescriptor)');
+    expect(code).toContain(
+      '__mfReadSharedCacheOwner(__mfModuleCache.share, cacheDescriptor) !== undefined'
+    );
     expect(code).toMatch(
       /__mfWriteSharedCache\(\s*__mfModuleCache\.share,\s*cacheDescriptor,\s*resolved,\s*"host"\s*\)/
     );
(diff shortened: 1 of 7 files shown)


gioboa (commented):

Image

PR #1081, opened 2026-08-12
fix: preserve tree-shaking environment cache isolation

### LINKED ISSUE #1076: [Bug]: tree-shaking shares bypass the react flavor isolation (unguarded partial-provider writers + conditionless dev refresh)
### Describe the bug

**Follow-up to #1072 — applies once it merges.** React-family packages shared with `shareConfig.treeShaking` bypass the flavor-isolation mechanism entirely, through two cooperating gaps:

1. **Unguarded writers.** The flavor guard lives only in the generated `__mfWriteSharedCache`; the tree-shaking writers (`__mfWriteTreeShakingSharedCache`, `__mfWriteTreeShakingSharedSelection`, [branch permalink](https://github.com/ryanmagoon/vite/blob/50cf688/src/utils/packageUtils.ts)) perform no flavor check, and generated hostAutoInit deliberately skips tree-shaking shares in its preload loop (`if (share.treeShaking) continue;`) — so the republish fallback never applies either. In `generateRemoteEntry`'s `__mfResolveTreeShakingShared`, the full-module branch writes through the guarded helper, but the **partial-provider branch** writes through the unguarded pair, and `__mfReadTreeShakingSharedSelection` serves those values to consumers with no check anywhere on the path.
2. **Conditionless refresh.** `refreshTreeShakingModules` rewrites tree-shaking prebuild and loadShare wrappers with no `exportConditions` and a hardcoded `'build'` command, hard-resetting environment-stamped cache keys and guards to the default bucket in dev — unlike `writeHostAutoInit`, these writers have no conditions memory, so a watch rebuild that skips the per-environment load refresh serves default-bucket code to a react-server environment.

### Version

Mechanisms verified by code reading against PR #1072's branch (`50cf688`) during its review; the refresh call sites exist on `main` @ 6d91862ea730250b0ecb3321e4a1a43d4b9eee78.

### Reality check, honestly

Nothing in this repo's docs, tests, or examples tree-shakes react — the documented use case is component libraries (`antd`). The config is expressible (`normalizeModuleFederationOptions` only forbids eager+treeShaking) but semantically dubious for React's hook namespace. We classified this as a real defense-in-depth hole with a theoretical trigger, which is why it was documented out of #1072 rather than fixed there.

### Expected vs actual

- **Expected:** every share-cache write path enforces the same flavor/bucket invariants.
- **Actual:** the tree-shaking partial-provider path and its dev refresh both operate outside them.

gioboa:
Close #1076

SSR/RSC-only bug environments. Client-only builds are unaffected.
Keeps refreshed shared wrappers bound to their original export conditions, command, and bundler mode. Prevents React server shares from falling into the default cache.

### DIFF
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -282,6 +282,16 @@ export function proxySharedModule(options: {
   const hasAnalyzableShares = Object.values(shared).some((share) =>
     shouldAnalyzeSharedExports(share)
   );
+  const getEnvironmentConditions = (context: unknown): string[] | undefined =>
+    (context as { environment?: { config?: { resolve?: { conditions?: string[] } } } }).environment
+      ?.config?.resolve?.conditions;
+  const refreshTreeShakingForEnvironment = (context: unknown) =>
+    refreshTreeShakingModules(
+      federationOptions,
+      _command,
+      getIsRolldown(context),
+      getEnvironmentConditions(context)
+    );
 
   const normalizeTreeShakingOutputPath = (value: string) => {
     const normalized = normalizePathForImport(value);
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Image

PR #1082, opened 2026-08-12
fix: preserve React isolation for worker SSR

### LINKED ISSUE #1077: [Bug]: webworker-SSR environments lose the react flavor guard — 'browser' condition conflated with browser process
### Describe the bug

**Follow-up to #1072 — applies once it merges.** `shouldIncludeReactFlavorGuard` strips the guard from any bundle whose resolve conditions include `'browser'` without `'react-server'` ([branch permalink](https://github.com/ryanmagoon/vite/blob/50cf688/src/utils/packageUtils.ts#L268)). That heuristic conflates the *browser export condition* with an *actual browser process*:

`DEFAULT_WEBWORKER_SSR_EXPORT_CONDITIONS` is `['worker','browser','import','module','default']` ([sharedExportConditions.ts#L3](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/utils/sharedExportConditions.ts#L3)), and `getSharedExportConditions` also appends these to user-configured ssr conditions for `ssrTarget: 'webworker'` — so a webworker-SSR environment always loses the guard while still using the **default** cache bucket. workerd-style environments commonly resolve `'browser'` too.

The guard defends against a vector the bucket partition does not: the victim environment itself calling `runtime.loadShare` against the process-global registry and receiving a sibling react-server environment's provider. A webworker-SSR isolate co-resident with an RSC environment would accept that wrong-flavor publish unguarded.

### Version

Verified by code reading against PR #1072's branch (`50cf688`) during its review.

### Reality check, honestly

No supported deployment shape hits this today: the repo has no workerd/edge examples, the SSR machinery targets Node, and every RSC setup we know of resolves node conditions for its ssr sibling. This is hardening for a future workerd/RSC scenario, not a live regression — which is why #1072 shipped the simple heuristic (requested in its review to keep client bundles lean) with this filed as the follow-up.

### Suggested direction

The right predicate is not "does this bundle target the browser" but "**can this process host more than one cache bucket**" — answerable at config time from `config.environments` (guard only when a sibling react-server environment exists). That also removes the guard bytes from every plain Node-SSR app with no RSC sibling, a strict improvement over the current heuristic in both directions.

gioboa:
Close #1077

Treat worker environments as SSR even when their export conditions include browser, preserving local React selection.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -643,8 +643,13 @@ describe('virtualRemoteEntry', () => {
     const clientInit = mod.generateHostAutoInitCode('"virtual:remoteEntry"', 'build', undefined, [
       'browser',
     ]);
+    const workerInit = mod.generateHostAutoInitCode('"virtual:remoteEntry"', 'build', undefined, [
+      'worker',
+      'browser',
+    ]);
     expect(serverInit).toContain('const localFactory = await share.get()');
     expect(clientInit).not.toContain('const localFactory = await share.get()');
+    expect(workerInit).toContain('const localFactory = await share.get()');
   });
 
   it('uses configured share import path in localSharedImportMap', async () => {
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1083, opened 2026-08-12
fix: await both SSR share-load queues

### LINKED ISSUE #1078: [Bug]: entry bootstrap's pendingShareLoads barrier always awaits the default cache bucket in react-server environments
### Describe the bug

**Follow-up to #1072 — applies once it merges.** The injected entry bootstrap builds its cache reference with `getRuntimeModuleCacheBootstrapCode()` — no export conditions — and then awaits `__mfModuleCache.pendingShareLoads` ([pluginAddEntry.ts#L375-L377](https://github.com/module-federation/vite/blob/6d91862ea730250b0ecb3321e4a1a43d4b9eee78/src/plugins/pluginAddEntry.ts#L375-L377)). After #1072, loadShare wrappers stamped for a react-server environment queue their pending loads in `__mf_module_cache_react_server__` — so an entry belonging to that environment awaits the **default** bucket's (empty) list and resolves immediately, re-opening the deferred-singleton-exports race the barrier exists to close.

### Version

Applies to `main` @ 6d91862ea730250b0ecb3321e4a1a43d4b9eee78 (the conditionless call site) in combination with #1072's bucket partition; found during #1072's review by code reading.

### Reachability, honestly

Medium confidence. It requires entry injection (`hostInitInjectLocation: 'entry'` with the pending-loads barrier emitted) targeting a **server** environment; the vinext path skips the html-entry fallback, and we did not construct a live reproduction. Verifiers traced the producers (`virtualShared_preBuild`'s deferred singleton assignments push into the per-conditions bucket) and the consumer (this barrier) reading different buckets by construction, so the mismatch is mechanical even though the triggering config is uncommon.

### Suggested direction

The entry bootstrap is generated per app entry; thread the environment's conditions through `getBootstrapSource` the same way #1072 threads them through hostAutoInit — or have the barrier await both buckets, which is two lines and order-safe.

gioboa:
Close #1078

Entry bootstrap waits for pending React Server shared-module loads before starting the application.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -416,6 +416,9 @@ describe('pluginAddEntry', () => {
 
     expect(result?.code).toContain('__mfModuleCache.pendingShareLoads');
     expect(result?.code).toContain('await Promise.all(__mfModuleCache.pendingShareLoads)');
+    expect(result?.code).toContain(
+      'await Promise.all(__mfReactServerModuleCache.pendingShareLoads)'
+    );
 
     const bootstrap = result?.code ?? '';
     expect(bootstrap.indexOf('await initHost();')).toBeLessThan(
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1086, opened 2026-08-13
fix: ignore export member method calls

### LINKED ISSUE #1084: [Bug]: `export` used as a method name in a shared package fails named-export detection, silently disabling sharing (`canLiveRebind: false`)
## Describe the bug

Named-export detection fails closed when a shared package's ESM entry uses `export` as a **method name**. The bare `\bexport\b` sweep in `getNamedExportsViaRegex` cannot distinguish `key.export(...)` from the `export` keyword, so it marks the scan incomplete and `getSharedNamedExports` returns `undefined`.

That verdict then silently disables sharing for the package:

1. `getPackageNamedExports` → `undefined`
2. `const canLiveRebind = shareItem.shareConfig.import === false || detectedNamedExports !== undefined` → **`false`**
3. The local provider is routed to a `__prebuild__` chunk instead of a `__loadShare__` one
4. The two `if (usedShare.canLiveRebind === false) return;` guards skip consumer rewriting entirely, so `import { useAuth0 } from '@auth0/auth0-react'` stays a plain static import of the remote's **own bundled copy**

There is no warning and the build succeeds. The share is registered in `shareScopeMap` and shows `loaded: true`, but `useIn` stays `[]` forever because `loadShare` is never called.

**Real-world impact:** `@auth0/auth0-react` is affected. Its ESM bundle contains `e.export()` and `e.export({format:"jwk"})` (Node `KeyObject` calls in the bundled DPoP code). A Vite remote that declares it as a `singleton` silently gets a second copy of `Auth0Context`, so every `useAuth0()` in the remote throws:

> You forgot to wrap your component in `<Auth0Provider>`

even though the host renders `Auth0Provider` and provides the singleton. React, `@apollo/client`, `@emotion/styled` etc. all share correctly in the same app — only packages tripping this scanner break, which makes it very hard to attribute.

This is the same class of bug as #986 (`t.exports` misclassification), #990/#991 (trailing commas) and #1008 (nested template literals), all in this scanner.

### Root cause

`src/virtualModules/virtualShared_preBuild.ts`:

```ts
// Regex extraction must fail closed. Valid syntax can omit whitespace or put
// comments between tokens, and silently treating an unmatched declaration as
// default-only would mix a cache-backed default with local named exports.
const exportKeywordRegex = /\bexport\b/g;
while ((match = exportKeywordRegex.exec(source)) !== null) {
  if (!codePositions[match.index]) continue;
  if (!recognizedExportStarts.has(match.index)) {
    scanState.complete = false;
    break;
  }
}
```

Failing closed is right, but a member property can never be a declaration. `hasCommonJsExports` in the same file already handles exactly this for the `exports` marker:

```ts
// `exports` is a CommonJS marker only when it is a standalone identifier.
// Member properties such as `wrapper.exports` describe arbitrary objects
// and do not determine the containing file's module format.
let previousCodeIndex = match.index - 1;
while (previousCodeIndex >= 0 && (/\s/.test(source[previousCodeIndex]) || !codePositions[previousCodeIndex])) {
  previousCodeIndex--;
}
if (source[previousCodeIndex] === '.') continue;
```

The keyword sweep is missing the same guard.

### Introduced by

Both `canLiveRebind` and this sweep came from 76e222d33949ee85ee7f28134717eca60c5d77b0 (#918, "fix: resolve lazy shared providers"), first released in **1.17.1**. Neither has been modified since — `git log -S'exportKeywordRegex' 76e222d..HEAD -- src/` and `git log -S'const canLiveRebind' 76e222d..HEAD -- src/` are both empty — so 1.20.6 is still affected.

## Version

Vite 6.3 / 7.x, `@module-federation/vite` 1.17.1 through 1.20.6 (reproduced on `main` @ 2a192a9)

## Reproduction

**Minimal, as a failing test in this repo's own suite** — the whole bug is in one pure function. Add a fixture whose entry uses `export` as a method name and assert detection:

```ts
// src/virtualModules/__tests__/virtualShared_preBuild.test.ts
if (filePath.endsWith('node_modules/mock-package-export-method-call/index.js')) {
  return [
    "const key = globalThis.crypto.subtle;",
    'export function toJwk() {',
    "  return key.export({ format: 'jwk' });",
    '}',
  ].join('\n');
}

it('detects named exports when `export` appears as a method name', () => {
  expect(getSharedNamedExports('mock-package-export-method-call')).toEqual(['toJwk']);
});
```

**Real-world:** declare `'@auth0/auth0-react': { singleton: true, requiredVersion: '2.22.0' }` as `shared` in a Vite remote, build, and inspect the generated `localSharedImportMap` chunk — the entry reads `canLiveRebind:!1` and its provider points at a `__prebuild__` chunk, while every other share reads `canLiveRebind:!0` with a `__loadShare__` chunk.

## Relevant log output

```shell
# No build error or warning is emitted. The failing unit test:

FAIL  src/virtualModules/__tests__/virtualShared_preBuild.test.ts > writeLoadShareModule
      > detects named exports when `export` appears as a method name

AssertionError: expected undefined to deeply equal [ 'toJwk' ]

- Expected:
[
  "toJwk",
]

+ Received:
undefined


# Generated localSharedImportMap in an affected build:

"@auth0/auth0-react": { ..., from:"appA", canLiveRebind:!1 }   <-- only share with false
"@apollo/client":     { ..., from:"appA", canLiveRebind:!0 }

"@auth0/auth0-react": async()=>await a(()=>import("./..._prebuild___mf_0_auth0_mf_1_auth0_mf_2_react__prebuild__~U52lox5U.js"))
"@apollo/client":     async()=>await a(()=>import("./..._loadShare___mf_0_apollo_mf_1_client__loadShare__.js~CBkDRhVm.js").then(r=>r.q))


# Runtime consequence — host provides the singleton, remote never consumes it:

shareScopeMap.default["@auth0/auth0-react"]["2.22.0"] = {
  from: "host", loaded: true, useIn: []        <-- never consumed
}
shareScopeMap.default["@apollo/client"]["3.8.4"] = {
  from: "host", loaded: true, useIn: ["appA"]
}
```

## Suggested fix

Reuse the member-access guard already used by `hasCommonJsExports` in the same file:

```diff
   const exportKeywordRegex = /\bexport\b/g;
   while ((match = exportKeywordRegex.exec(source)) !== null) {
     if (!codePositions[match.index]) continue;
-    if (!recognizedExportStarts.has(match.index)) {
-      scanState.complete = false;
-      break;
-    }
+    if (recognizedExportStarts.has(match.index)) continue;
+
+    // `export` is a keyword only as a standalone identifier. A member property
+    // such as `key.export({ format: 'jwk' })` names an arbitrary method and
+    // cannot be a declaration, so it must not fail the scan closed.
+    let previousCodeIndex = match.index - 1;
+    while (
+      previousCodeIndex >= 0 &&
+      (/\s/.test(source[previousCodeIndex]) || !codePositions[previousCodeIndex])
+    ) {
+      previousCodeIndex--;
+    }
+    if (source[previousCodeIndex] === '.') continue;
+
+    scanState.complete = false;
+    break;
   }
```

With this applied, `virtualShared_preBuild.test.ts` passes 118/118 and the full unit suite passes 1120/1122 (the one failure, `integration/dev-singleton-react.test.ts > boots on first load with a nested static remote import (#1056)`, also fails on unmodified `main` here, and 1 test is skipped).

Happy to open a PR with the patch and the test above if that's useful.

## Workaround

Point `import` at a build whose exports the plugin can read via `require()`, which restores `canLiveRebind: true`:

```ts
'@auth0/auth0-react': {
  singleton: true,
  requiredVersion: auth0ReactVersion,
  import: '@auth0/auth0-react/dist/auth0-react.cjs.js',
},
```

gioboa:
Close #1084

Prevents .export() from breaking shared-package named-export detection.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -218,6 +218,9 @@ vi.mock('../../utils/packageUtils', () => ({
     if (pkg === 'mock-package-nested-template-literal') {
       return '/repo/apps/remote/node_modules/mock-package-nested-template-literal/index.js';
     }
+    if (pkg === 'mock-package-export-method-call') {
+      return '/repo/apps/remote/node_modules/mock-package-export-method-call/index.js';
+    }
     if (pkg === 'mock-package-esm-only/stores' || pkg === 'mock-package-esm-only') {
       return '/repo/apps/remote/node_modules/mock-package-esm-only/dist/stores.js';
     }
@@ -690,6 +693,14 @@ export const buttonBase = 1;`;
       return [
         "const body = 'text';",
         'export const injectedHtml = `${`<div>`}${body}${`</div>`}`;',
+      ].join('\n');
+    }
+    if (filePath.endsWith('node_modules/mock-package-export-method-call/index.js')) {
+      return [
+        'const key = globalThis.crypto.subtle;',
+        'export function toJwk() {',
+        "  return key.export({ format: 'jwk' });",
+        '}',
       ].join('\n');
     }
     if (
@@ -2708,6 +2719,10 @@ describe('writeLoadShareModule', () => {
         String(filePath).endsWith('/mock-package-nested-template-literal/index.js')
       )
     ).toHaveLength(1);
+  });
+
+  it('detects named exports when `export` appears as a method name', () => {
+    expect(getSharedNamedExports('mock-package-export-method-call')).toEqual(['toJwk']);
   });
 
   it('uses worker conditional exports for webworker SSR', () => {
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -752,10 +752,20 @@ function getNamedExportsViaRegex(
   const exportKeywordRegex = /\bexport\b/g;
   while ((match = exportKeywordRegex.exec(source)) !== null) {
     if (!codePositions[match.index]) continue;
-    if (!recognizedExportStarts.has(match.index)) {
-      scanState.complete = false;
-      break;
+    if (recognizedExportStarts.has(match.index)) continue;
+
+    // A member named `export` is not an export declaration.
+    let previousCodeIndex = match.index - 1;
+    while (
+      previousCodeIndex >= 0 &&
+      (/\s/.test(source[previousCodeIndex]) || !codePositions[previousCodeIndex])
+    ) {
+      previousCodeIndex--;
     }
+    if (source[previousCodeIndex] === '.') continue;
+
+    scanState.complete = false;
+    break;
   }
 
   return Array.from(names);


gioboa (commented):

Image

PR #1087, opened 2026-08-13
fix: defer remote singleton fallback imports

### LINKED ISSUE #1085: Shared singleton module evaluated twice across host/remote despite singleton:true and matching version (no mismatch warning)
### Describe the bug

A dependency shared with `{ singleton: true }` on both a host and a remote, with both resolving to the identical version (`1.0.0`), is evaluated twice instead of once when the host loads the remote. No version-mismatch warning is logged.

Expected: with a satisfied singleton version requirement, the shared module should be evaluated exactly once across the federated runtime.

Actual: the shared dependency in this repro is `shared-lib`, whose module body logs `[shared-lib] evaluated` when it runs. That log fires twice, from two different origins — the host's own bundle and the remote's own bundle:
```
[shared-lib] evaluated  (http://localhost:5174/@fs/.../packages/shared-lib/index.js:1)   <- host, port 5174
[shared-lib] evaluated  (http://localhost:5173/@fs/.../packages/shared-lib/index.js:1)   <- remote, port 5173
```

The same result occurs whether or not `remote` also sets `eager: true` for `shared-lib` (both variants tested; see repro README for the eager/eager case).

### Version

@module-federation/vite: 1.20.6 (vite: 8.2.1)

### Reproduction

https://github.com/arshabh-agarwal/module-federation-vite-singleton-repro

Setup/run/config details are in the repo's README.

### Relevant log output

```
[vite] connecting...
[vite] connected.
[shared-lib] evaluated
[vite] connecting...
[vite] connected.
[shared-lib] evaluated
[remote] importing shared-lib
[remote] evaluated
[host] importing shared-lib
[host] evaluated
[host] remote loaded
```

### Validations

- [x] Read the docs.
- [x] Checked that there isn't already an issue that reports the same bug to avoid creating a duplicate.

gioboa:
Close #1085

Remote dev wrappers now check the shared cache before loading their local fallback, preventing duplicate singleton evaluation.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -3560,7 +3560,7 @@ describe('writeLoadShareModule', () => {
 
     const generatedCode = writeSyncSpy.mock.calls.at(-1)?.[0] as string;
 
-    expect(generatedCode).toContain('import * as __mfLocalShare from "lit";');
+    expect(generatedCode).not.toContain('import * as __mfLocalShare from "lit";');
     expect(generatedCode).toContain('if (import.meta.env.SSR');
     expect(generatedCode).toContain('import("lit").then((mod) => {');
     expect(generatedCode).toContain('export { __mf_default as default };');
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

Issue #591, opened 2026-04-02
[Bug]: federation-fix-preload regex incorrectly matches Stencil's getScopeId function, causing runtime crash


chirag3092:
## Bug Report
**Package version:** @module-federation/vite@1.13.2 (also affects 1.12.2, 1.12.3, 1.13.x)

## Description
The module-federation-fix-preload plugin introduced in 1.12.2 contains a generateBundle post-processing step that uses a regex to find and replace Vite's preload helper function. The regex is too broad and incorrectly matches **Stencil's minified getScopeId function**, corrupting the bundle and causing a runtime crash.

## Root Cause
In `src/index.ts`, the arrow function regex is intended to match Vite's preload helper like `=(e)=>"/"+e` but it also matches Stencil's minified `getScopeId function: va=(e,t)=>"sc-"+e.$tagName$`

 Because:
 - `(\w+)` captures `e`
 - `(?:,\w+)?` matches ,`t`
 - the string literal matches `"sc-"`
 - `\+\s*\1` matches `+e` (false positive - `e` is the backreference `\1`)
 The same issue exists in the function expression regex variant.

## What happens
 After the replacement, the bundle contains:
`va=function(e){return new URL("../"+e,import.meta.url).href}.$tagName$`

 Which makes `va` undefined at runtime, causing:
`Uncaught TypeError: va is not a function`

 This breaks **all Stencil web components** on the page.

### Version

v1.12.2, v1.12.3, v1.13.x

### Reproduction

https://github.com/module-federation/vite/pull/589/changes/dadb3d4fa7c3117c1724b222fe16c65b2584392e

### Relevant log output

```shell

```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite).
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues).
- [x] Check that there isn't [already an issue](https://github.com/module-federation/vite/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1091, opened 2026-08-17
fix: treat `$`-prefixed minified names as real identifiers

Ryan-Haines:
`\b` cannot test whether an identifier is referenced. A word boundary sits between a `\w` and a non-`\w` character, and `$` is not a word character, so `\b` never matches at the start of a name beginning with `$` — `\b$8\b` matches nothing at all, in any input.

Minifiers produce those names routinely. esbuild and Oxc both draw single-character names from a 54-character alphabet whose last two entries are `_` and `$`, so any chunk with more than about 54 module-scope bindings has `$`-prefixed names, and a bundled application chunk has dozens of them.

Verified against esbuild 0.27.4, counting distinct module-scope bindings in a chunk:

| bindings | names starting with `$` |
| --- | --- |
| 60 | `$` |
| 500 | `$ $t $e $l $v $1 $2 $3 $4` |
| 3500 | 64 of them, including `$8` |

The regex behaviour on its own:

```js
new RegExp(`\\b$8\\b`).test('var e=$8,r=PS;')   // false — should be true
new RegExp(`\\b\\$8\\b`).test('var e=$8,r=PS;') // false — escaping alone does not help
```

Affects 1.13.0 through 1.20.7; the line is byte-identical across all of them. Vite 8 / Rolldown is
equally affected: the CJS proxy rewrite has no `isRolldown` guard, and Oxc's mangler uses the same
alphabet.

When such a check reports "unused", a still-referenced import gets rebound or deleted and its references are orphaned. The chunk builds and minifies cleanly and throws `ReferenceError: <name> is not defined` as soon as a browser evaluates it. For a remote that surfaces as the whole expose failing to load.

## What changed

Adds `isIdentifierReferenced(name, code)` — escape, then lookarounds instead of `\b` — and uses it for all three usage checks:

| Site | Effect of the bug |
| --- | --- |
| `resolveProxyAlias` (`bundleHelpers`) | Import rebound, references orphaned. Observed in production. |
| Re-export mapping in `collectSystemProxyInfos` (`bundleHelpers`) | Export mis-mapped |
| Helper import elision in `stripEmptyPreloadCalls` (`controlChunkSanitizer`) | Import statement deleted |

The latter two also interpolated the name **unescaped**, where a leading `$` is a regex end-anchor rather than a literal — a second, independent reason the same names went unmatched.

`stripEmptyPreloadCalls` additionally captured its alias with `\w+`, which skipped `$`-prefixed imports instead of processing them. That made site 3 latent rather than live. The capture and the check are widened together, because widening either alone is worse than neither: widening only the capture makes the bug reachable, and fixing only the check leaves the import unprocessed.

Also escapes the remaining raw identifier interpolations in `function <name>(` patterns, and switches those `String.replace` calls to callbacks so a `$` in the replacement is not read as a substitution pattern — the same hazard #461 fixed elsewhere.

No behaviour change for names that do not involve `$`.

## Verification

**Tests fail without the fix.** 8 in `integration/proxy-alias-deconflict.test.ts` and 1 in `src/utils/__tests__/controlChunkSanitizer.test.ts`. I confirmed this by reverting the fix inside the new helper and re-running, so the tests are pinned to the behaviour rather than merely passing alongside it.

The existing `resolveProxyAlias` tests from #461 cover `$` only in non-leading positions (`require$$0`, `commonjsGlobal$1`), which all pass under `\b`. That is why a leading `$` slipped through, so the new cases target that position specifically, plus the "genuinely unused" direction so the fix cannot be trivially satisfied by always returning `true`.

**Full suite green:** `pnpm typecheck`, `pnpm test` (49 files, 1061 passed / 1 skipped), `pnpm test:integration` (15 files, 78 passed, including real Vite builds), `pnpm fmt.check`.

**Verified end-to-end against the real failing build,** not only fixtures. Rebuilding the exact application tree that broke, with the fix reversed and then applied:

```
# fix reversed
build/assets/vendor-BQm3xKp1.js
  $8 — referenced 2x, first at 17:133828, bound nowhere in this chunk
bundle check: 1 chunk references an identifier that is never bound.

# fix applied
bundle check: 45 chunks clean
```

The two artifacts differ by one byte, in the import the plugin rewrites:

```js
// fix reversed — plugin decided $8 was unused
import{r as l,c as Zg}from"./..._commonjs-proxy-hash.js";
// fix applied — binding preserved
import{r as $8,c as Zg}from"./..._commonjs-proxy-hash.js";
```

react-dom's binding in the same chunk was `tI` and was handled correctly either way, which is the whole difference between the two.

Scope note: site 1 is verified end-to-end on that build; sites 2 and 3 and the escaping changes are covered by unit tests with verified negative controls.

Refs #460, #461, #591, #1014.

### DIFF
--- a/integration/proxy-alias-deconflict.test.ts
+++ b/integration/proxy-alias-deconflict.test.ts
@@ -1,7 +1,59 @@
 import { describe, expect, it } from 'vitest';
-import { resolveProxyAlias } from '../src/utils/bundleHelpers';
+import { isIdentifierReferenced, resolveProxyAlias } from '../src/utils/bundleHelpers';
+
+describe('isIdentifierReferenced', () => {
+  it.each(['$8', '$', '$1', '$$0'])('finds `%s` at the start of an identifier', (name) => {
+    expect(isIdentifierReferenced(name, `var n=j3,e=${name},r=PS;`)).toBe(true);
+  });
+
+  it.each(['An', 'd6', 'N4', 'require$$0', 'commonjsGlobal$1'])(
+    'still finds the ordinary name `%s`',
+    (name) => {
+      expect(isIdentifierReferenced(name, `const o=${name}.createElement("div");`)).toBe(true);
+    }
+  );
+
+  it('does not match a longer identifier that merely contains the name', () => {
+    expect(isIdentifierReferenced('$8', 'const x=$80;')).toBe(false);
+    expect(isIdentifierReferenced('$', 'const x=$foo;')).toBe(false);
+    expect(isIdentifierReferenced('r', 'const x=react;')).toBe(false);
+  });
+
+  it('does not match when the name is only a suffix of another identifier', () => {
+    expect(isIdentifierReferenced('$8', 'const x=a$8;')).toBe(false);
+    expect(isIdentifierReferenced('o', 'const x=foo;')).toBe(false);
+  });
+
+  it('reports an absent name as unreferenced', () => {
+    expect(isIdentifierReferenced('$8', 'console.log(other);')).toBe(false);
+  });
+});
 
 describe('resolveProxyAlias', () => {
+  // The `\b` check reported a minified `$8` binding unused, so the import was
+  // rebound to the proxy's export name and its references were left dangling —
+  // a clean build that threw `ReferenceError: $8 is not defined` in the browser.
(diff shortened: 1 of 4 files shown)


gioboa (approved):
Thanks @Ryan-Haines nice fix 👍

PR #1092, opened 2026-08-18
fix: include expose static closures in federation manifests

dmchoi77:
## Summary

Include the complete static JavaScript closure required to evaluate exposed modules in federation manifest assets.

The manifest previously recorded only the expose root chunk, omitting its static dependencies and the container bootstrap chunks required by `init`/`get`.

## Changes

- reuse the existing cycle-safe static chunk traversal
- add container bootstrap dependencies and expose static closures to `assets.js.sync`
- expand async roots through static imports without crossing nested dynamic boundaries, with `sync` taking precedence
- support Vite 5–7 direct containers and the Vite 8 static-wrapper topology using scoped virtual module IDs

Shared/CSS analysis, development, SSR, and `disableAssetsAnalyze` behavior remain unchanged.

## Verification

- `pnpm test`
- `pnpm test:integration`
- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm build`
- emitted bundle graphs verified with Vite 5.4.21, 6.4.3, 7.3.6, and 8.2.0

### DIFF
--- a/integration/build.test.ts
+++ b/integration/build.test.ts
@@ -54,6 +54,16 @@ describe('build', () => {
 
       const parsed = JSON.parse(manifest!.source as string);
       expect(parsed).toHaveProperty('exposes');
+
+      const syncAssets = parsed.exposes[0].assets.js.sync;
+      const chunks = manifestOutput.output.filter((item) => item.type === 'chunk');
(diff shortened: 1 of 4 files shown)


gioboa (approved):
Thank you @dmchoi77 good improvement 💪

PR #1094, opened 2026-08-18
fix: only patch svelte in svelte application

### LINKED ISSUE #1093: [Bug]: patchSvelteKitStaticHtml executed on React application, adding ~25sec delaying process in my application
### Describe the bug

Hello 

First of all, thanks for the library :) 

On a **React** (non-SvelteKit) client-side build, `pluginAddEntry`'s `closeBundle` hook
unconditionally runs [`patchSvelteKitStaticHtml()`](https://github.com/module-federation/vite/blob/49b84452f509606915d0d73d7c59850e75d066b4/src/plugins/pluginAddEntry.ts#L252) and retries it **20 times**, because the
function can never report success on a non-SvelteKit project. 

In my application, this adds **~25s**.

---

I think an early return is missing in the [`closeBundle`](https://github.com/module-federation/vite/blob/49b84452f509606915d0d73d7c59850e75d066b4/src/plugins/pluginAddEntry.ts#L808) : 
```ts
if (!hasPackageDependency('@sveltejs/kit')) {
  return;
}
```

Adding it in my project fixes the problem.

### Version

1.20.7

### Reproduction

https://github.com/romain-trotard/vite/tree/repro -> I have changed the README.md with the STR

### Relevant log output

With `node --cpu-prof` : 

Image

Image

### Validations

romain-trotard:
Close #1093

Hello,

This is a PR to fix my issue https://github.com/module-federation/vite/issues/1093

It allows not to `patchSvelteKitStaticHtml` when we are not in `svelte` application.

Do you think I should try to write a test? I don't really know how to write one for the moment ^^

Thanks

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -15,8 +15,10 @@ import type {
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { toViteEncodedId, VITE_ID_PREFIX } from '../../utils/VirtualModule';
 
+const mockHasPackageDependency = vi.hoisted(() => vi.fn((_pkg: string) => true));
+
 vi.mock('../../utils/packageUtils', () => ({
-  hasPackageDependency: () => true,
+  hasPackageDependency: mockHasPackageDependency,
 }));
 
 const mockMfOptions = vi.hoisted(() => ({
@@ -98,6 +100,11 @@ function runBuildStart(
   callHook(plugin.buildStart, ctx, options);
 }
 
+function runCloseBundle(plugin: AddEntryPlugin): void {
+  if (!plugin.closeBundle) throw new Error(`${plugin.name} closeBundle hook not found`);
+  callHook(plugin.closeBundle, {} as Rollup.PluginContext);
+}
+
 function runGenerateBundle(
   plugin: AddEntryPlugin,
   ctx: Rollup.PluginContext,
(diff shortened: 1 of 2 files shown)


dmchoi77:
If you decide to add a test, a focused unit test around closeBundle should be enough: mock hasPackageDependency('@sveltejs/kit') to return false and verify that no retry timer is scheduled. A small SvelteKit fixture could also verify that the existing static HTML patching behavior is preserved. The fix looks correct otherwise!

romain-trotard:
Thanks @dmchoi77 I add some tests with your recommandation soon :)

src/plugins/__tests__/pluginAddEntry.test.ts
```diff
@@ -1949,4 +1960,52 @@ describe('pluginAddEntry', () => {
...
+      inject: 'html',
+    });
+
+    runCloseBundle(buildPlugin);
+
+    expect(setTimeoutSpy).not.toHaveBeenCalled();
+  });
+
+  it('closeBundle patches SvelteKit static HTML on disk after build', () => {
+    vi.useFakeTimers();
```
romain-trotard, line 1981: I am using fake timers, to keep a fast execution. 
Is it okay for you? or do you prefer that I use `vi.waitFor` ?

src/plugins/__tests__/pluginAddEntry.test.ts
```diff
@@ -1949,4 +1960,52 @@ describe('pluginAddEntry', () => {
...
+    } as unknown as ResolvedConfig);
+
+    runCloseBundle(buildPlugin);
+    vi.runAllTimers();
+
+    const patched = fs.readFileSync(path.join(buildDir, '200.html'), 'utf-8');
+
+    expect(patched).toContain('const __mfCurrentScript = document.currentScript;');
+    expect(patched).toContain('.then(({ initHost }) => initHost());');
+    expect(patched).toContain('kit.start(app, element);');
```
romain-trotard, line 2009: I am checking some content, but not sure if that's enough. Do not hesitate to tell me if I should check other content

gioboa (approved):
Thanks @romain-trotard 
That's great 👏

src/plugins/__tests__/pluginAddEntry.test.ts
gioboa: That's fine 💪

PR #1098, opened 2026-08-19
fix: detect minified Webpack shared providers

### LINKED ISSUE #1097: [Bug]: `isWebpackProvider` incorrectly returns false
### Describe the bug

Hi!

I'm encountering a startup failure on my half-legacy/half-modern module federation. I've step-debugged the code to find the error:

- It's a loading error, a Vite consumer fails to load `i18next` exposed by a webpack producer
- The webpack producer is minified

Image

```js
Function.prototype.toString.call(provider.get)
// "()=>i.e(796).then((()=>()=>i(796)))"
```

(`i` is probably `__webpack_require__` before minification if I had to take a guess)

Image

`isWebpackProvider` returns false, causing `if (provider && isWebpackProvider(provider) && !provider.lib && !provider.loaded) return provider;` not to return `provider` and then my lib to be undefined at runtime.

---

Bingo! Here is the code before minification

```js
register("i18next", "19.9.2", () => (__webpack_require__.e("vendors-node_modules_i18next_dist_esm_i18next_js").then(() => (() => (__webpack_require__(/*! ./node_modules/i18next/dist/esm/i18next.js */ "./node_modules/i18next/dist/esm/i18next.js"))))));
```

### Version

8.2.1 / Vite-module-federation 1.20.7

### Reproduction

none yet

### Relevant log output

```shell

```

### Validations

dmchoi77:
## Summary

Fixes an issue where `isWebpackProvider` incorrectly classified minified Webpack shared providers as non-Webpack providers.

## Related Issue

Fixes #1097 

## Background

Webpack shared providers can expose a lazy `get()` function. In production builds, Webpack may minify the internal `__webpack_require__` identifier. For example:

```js
() => i.e(796).then(() => () => i(796));
```

The previous implementation only looked for the `__webpack_require__` string. As a result, this minified provider was not deferred by the Webpack-specific provider guards. The Vite consumer could then continue with an unresolved provider and observe an undefined shared library at runtime.

## Root Cause

`isWebpackProvider` inferred the provider type from an unminified function identifier:

```js
source.includes("__webpack_require__");
```

That identifier is not preserved by production minification, while Webpack's lazy chunk-loading shape remains recognizable as `.e(...).then(...)`.

## Changes

- Recognize the minified Webpack lazy chunk-loading pattern `.e(...).then(...)`.
- Add a regression test using the exact minified getter reported in the issue.

## Verification

The following checks pass:

- `pnpm test` — 49 test files, 1065 tests passed
- `pnpm test:integration` — 15 test files, 78 tests passed
- `pnpm typecheck`
- `pnpm build`
- `pnpm fmt.check`

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -2881,8 +2881,10 @@ describe('virtualRemoteEntry', () => {
     const webpackGet = new Function(
       'return function webpackGet() { return __webpack_require__("react"); };'
     )();
+    const minifiedWebpackGet = new Function('return ()=>i.e(796).then((()=>()=>i(796)))')();
 
     expect(isWebpackProvider({ get: webpackGet })).toBe(true);
+    expect(isWebpackProvider({ get: minifiedWebpackGet })).toBe(true);
     expect(isWebpackProvider({ get: () => Promise.resolve(() => ({})) })).toBe(false);
     expect(isWebpackProvider({})).toBe(false);
     expect(code).toContain('const isWebpackScope = !scopeRoot &&');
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks for your help @dmchoi77 🙏

PR #1099, opened 2026-08-19
fix: `loaded-first` static remote evaluation

### LINKED ISSUE #1096: [Bug]: loaded-first deferred remote proxy throws its pending Promise out of module evaluation, app never boots
### Version

`@module-federation/vite@1.20.7` (also 1.20.6 — `applyRewrites` is byte-identical in both), `vite@7.3.6`, production build, browser.

### Summary

With `shareStrategy: 'loaded-first'`, any consumer that touches a remote namespace at **module-evaluation** time throws the deferred proxy's pending Promise out of module evaluation. That rejects the entry's `import()`, so the app never boots. The only console output is a bare `Uncaught (in promise)` — no stack, because the rejection *reason* is a `Promise`, not an `Error`.

This looks like the same family as #787 / #740 / #824, fixed for Vue static imports in #789 / #742, but the framework-agnostic case is still broken: plain JS module-scope reads.

### Minimal repro

Runnable repo: <https://github.com/h11g/mf-vite-loaded-first-repro>

Inlined below as well — no React, no Vue, 7 files.

<details>
<summary><code>remote/vite.config.js</code>, <code>remote/src/util.js</code>, <code>remote/index.html</code></summary>

```js
// remote/vite.config.js
import {federation} from '@module-federation/vite';
import {defineConfig} from 'vite';

export default defineConfig({
	base: 'http://localhost:5174/',
	plugins: [
		federation({
			name: 'remoteApp',
			filename: 'remoteEntry.js',
			exposes: {'./util': './src/util.js'},
			shared: {},
		}),
	],
	build: {target: 'chrome91'},
});
```

```js
// remote/src/util.js
export const REMOTE_NAME = 'remoteApp';

export function greet() {
	return `hello from ${REMOTE_NAME}`;
}
```

```html
<!-- remote/index.html -->
<!doctype html>
<html><body>remote</body></html>
```

</details>

<details>
<summary><code>host/vite.config.js</code>, <code>host/index.html</code>, <code>host/src/*</code></summary>

```js
// host/vite.config.js
import {federation} from '@module-federation/vite';
import {defineConfig} from 'vite';

export default defineConfig({
	plugins: [
		federation({
			name: 'host',
			filename: 'remoteEntry.js',
			remotes: {
				remoteApp: {
					type: 'module',
					name: 'remoteApp',
					entry: 'http://localhost:5174/remoteEntry.js',
				},
			},
			shared: {},
			shareStrategy: 'loaded-first', // drop this line and the repro passes
		}),
	],
	build: {target: 'chrome91'},
});
```

```html
<!-- host/index.html -->
<!doctype html>
<html>
	<body>
		<div id="root">boot did not finish</div>
		<script type="module" src="/src/main.js"></script>
	</body>
</html>
```

```js
// host/src/main.js
import {label} from './consumer-named.js'; // or './consumer-namespace.js'

document.getElementById('root').textContent = label();
```

```js
// host/src/consumer-named.js  — variant A
import {REMOTE_NAME} from 'remoteApp/util';

export function label() {
	return `host sees ${REMOTE_NAME}`;
}
```

```js
// host/src/consumer-namespace.js  — variant B
import * as remoteUtil from 'remoteApp/util';

export const {REMOTE_NAME} = remoteUtil;

export function label() {
	return `host sees ${REMOTE_NAME}`;
}
```

</details>

```bash
cd remote && npm i && npx vite build && npx vite preview --port 5174 --strictPort
cd host   && npm i && npx vite build && npx vite preview --port 5173 --strictPort
```

Open <http://localhost:5173/>.

**Expected:** `host sees remoteApp`
**Actual:** the page still reads `boot did not finish`, console shows `Uncaught (in promise)`.

The rejection reason is the remote's own load Promise:

```js
window.addEventListener('unhandledrejection', async (e) => {
	typeof e.reason.then;            // 'function'  -> the reason is a Promise
	Object.keys(await e.reason);     // ['REMOTE_NAME', 'greet']
});
```

Both variants reproduce. Reproduces on build only.

### Why

`shareStrategy: 'loaded-first'` makes `shouldDeferRemoteLoad()` true, so a browser build resolves the remote to `__mfCreateDeferredRemoteProxy()`. Its `get` trap ends with:

```js
if (prop === "default") return proxyTarget;
throw ensurePending();
```

That throw is a React-Suspense signal and only works inside a render pass. But `applyRewrites()` emits bindings that read the namespace during module evaluation. Minified variant-A host chunk:

```js
let f = cache.remote[key];
f === void 0 && (f = createDeferredRemoteProxy());
const g = startRemoteLoad().then(assign);
const { REMOTE_NAME: p } = f;   // <-- throws the pending Promise
Promise.all([g]);               // __mf_remote_dependency_pending, nothing awaits it
```

Three affected branches in `applyRewrites()`:

| import form | generated read |
| --- | --- |
| `import { x } from 'remote'` | `const { x: x } = nsId;` |
| `export { x } from 'remote'` | `let tmp = nsId["x"];` |
| `import * as ns from 'remote'` | none generated — but user source like `export const { x } = ns` reads it just the same |

`__mf_remote_dependency_pending` is awaited for exposes (`generateExposes`), for dynamic imports (`wrapDynamicImport`) and for remote-to-remote loads, but nothing gates it for the host's own entry graph — which is where these consumers live.

The default strategy is unaffected because static remote entries are resolved during host init, so `__mf_module_cache__.remote[key]` is already populated when the entry graph evaluates and the proxy branch never runs.

### Impact

On a real app (~4,600 TS files, 11 remotes) this is every route white-screening with no usable stack. 11 distinct eval-time read sites; evaluation aborts at the first one, so only one remote ever appears in `__mf_module_cache__.remote`, which makes it look like a single-remote problem.

### Candidate fix

Awaiting the pending barrier in all three branches before the namespace is read fixes both repro variants and the production app:

```diff
 			if (imp.namespaceLocal && !imp.defaultLocal && imp.named.length === 0) {
 				const pendingId = `${imp.namespaceLocal}__mf_pending`;
 				dependencyPendingIds.push(pendingId);
-				ms.overwrite(imp.start, imp.end, `import { __moduleExports as ${imp.namespaceLocal}, __mf_remote_pending as ${pendingId} } from ${src};`);
+				ms.overwrite(imp.start, imp.end, `import { __moduleExports as ${imp.namespaceLocal}, __mf_remote_pending as ${pendingId} } from ${src};\nawait Promise.resolve(${pendingId}).catch(() => {});`);
 			} else {
@@
-				let rewrite = `import { ${importParts.join(", ")} } from ${src};`;
+				let rewrite = `import { ${importParts.join(", ")} } from ${src};\nawait Promise.resolve(${pendingId}).catch(() => {});`;
@@ reexport branch
-			const varLines = vars.map((v) => `let ${v.tmp} = ${nsId}[${JSON.stringify(v.local)}];`).join("\n");
+			const varLines = `await Promise.resolve(${pendingId}).catch(() => {});\n` + vars.map((v) => `let ${v.tmp} = ${nsId}[${JSON.stringify(v.local)}];`).join("\n");
```

I'd rather get your call on two points before opening a PR, since both are policy decisions rather than mechanics:

1. **This makes every remote consumer a TLA module.** Fine for a `chrome91`-style target, but it will break builds whose `build.target` predates top-level await, and CJS/SSR output where TLA is illegal. Should it be gated on target capability and skipped for server/CJS consumers, or put behind an option?
2. **`.catch(() => {})` keeps a failing remote local.** Without it, one unreachable remote entry takes down the whole boot instead of only its own consumers, which is the current behaviour. The cost is that bindings stay `undefined` — related to #824. Propagating instead may be preferable.

A non-TLA alternative would be emitting lazy getters for the generated bindings so nothing is read until first use. That covers the two generated flavours but not `export const { x } = ns` in user source, so it is not a complete fix on its own.

Happy to send the PR with an e2e case in whichever shape you prefer.

dmchoi77:
## Summary

Fixes #1096 without introducing top-level await (TLA).

When a host uses a static remote import with `shareStrategy: 'loaded-first'`, the host entry must not be evaluated until that remote namespace is ready. The previous bootstrap deliberately skipped all remote preloads for `loaded-first`, which could cause named or namespace imports to observe an incomplete remote module and fail during application startup.

This change preloads only remotes that are statically imported by the client entry graph. Dynamic-only remotes remain on-demand.

## Root cause

- `pluginRemoteNamedExports` rewrites remote named and namespace imports while preserving a TLA-free module wrapper.
- The `loaded-first` host bootstrap previously skipped remote preloading entirely.
- As a result, the application entry could be evaluated before a statically imported remote namespace was available.
- The failure was commonly surfaced as a bare `Uncaught (in promise)` during production startup.

## Implementation

- Add a shared import descriptor scanner that distinguishes:
  - runtime static imports;
  - dynamic `import()` calls;
  - CommonJS `require()` calls;
  - type-only imports and exports.
- Track the actual static remote module IDs discovered from the configured client entry graph.
- For client production builds using `loaded-first`:
  - register each static remote with the runtime;
  - preload only those static remotes;
  - await them with `Promise.all()` before importing the application entry.
- Preserve the existing behavior for:
  - dynamic-only remotes;
  - `version-first` and its `Promise.allSettled()` error policy;
  - SSR and server consumer wrappers;
  - CommonJS handling;
  - `skipRemotePreload` paths;
  - remote wrapper dependency-pending exports.
- Keep generated wrappers and bootstrap code free of TLA.

## Error behavior

Static remote preload failures now reject the bootstrap with the original remote loading error. The application entry is not imported, so entry side effects do not run after a static remote failure.

Dynamic-only remote failures remain deferred until the remote is requested.

## Compatibility

- No public `ModuleFederationOptions` or runtime API changes.
- No new TLA or `await` at module top level.
- The change is limited to internal static-import tracking, host bootstrap sequencing, and regression coverage.

## Tests

Passed on the fix branch:

- `pnpm test` — 49 files, 1,073 passed, 1 skipped
- `pnpm typecheck`
- `pnpm build`
- `pnpm test:integration` — 16 files, 84 passed
- `pnpm fmt.check`
- `git diff --check`

The browser regression coverage verifies:

- named static imports;
- namespace static imports;
- remote request ordering before host entry evaluation;
- absence of page errors and unhandled rejections on success;
- no host entry side effects after a static remote load failure;
- propagation of the original runtime loading error.

## Review guide

Recommended review order:

1. Entry import descriptors and static remote registry. (https://github.com/module-federation/vite/commit/5b8af3a7ecb15f0a779cfc54966bdd01633becaa) 
2. `loaded-first` bootstrap preload and runtime registration. (https://github.com/module-federation/vite/commit/5abdb67a6aea8a87420285b234a9bf5ad522d3fc)
3. Production/browser regression fixtures and failure-path assertions. (https://github.com/module-federation/vite/commit/194399633f0e76a7e309123a4a181041e264e8e6)

### DIFF
--- a/src/utils/htmlEntryUtils.ts
+++ b/src/utils/htmlEntryUtils.ts
@@ -66,7 +66,7 @@ export function findModuleImportSources(code: string): string[] {
   return Array.from(
     new Set(
       findModuleImportDescriptors(code)
-        .filter(({ syntax }) => syntax === 'import')
+        .filter(({ syntax, typeOnly }) => syntax === 'import' && !typeOnly)
         .map(({ source }) => source)
     )
   );
--- a/src/utils/htmlEntryUtils.ts
+++ b/src/utils/htmlEntryUtils.ts
@@ -1,20 +1,75 @@
 import { createCodePositionMap } from './codePositionMap';
 
-export function findModuleImportSources(code: string): string[] {
-  const codePositions = createCodePositionMap(code);
-  const sources = new Set<string>();
-  const patterns = [
-    /\b(?:import|export)\s+[^;]*?\bfrom\s*["']([^"']+)["']/g,
-    /\bimport\s*\(\s*["']([^"']+)["']/g,
-    /\bimport\s*["']([^"']+)["']/g,
-  ];
+export interface ModuleImportDescriptor {
+  kind: 'static' | 'dynamic';
+  syntax: 'import' | 'require';
+  source: string;
+  typeOnly: boolean;
+}
 
-  for (const pattern of patterns) {
-    for (const match of code.matchAll(pattern)) {
-      if (codePositions[match.index!]) sources.add(match[1]);
-    }
+function isTypeOnlyClause(clause: string): boolean {
+  const normalized = clause.trim();
+  if (/^type\b/.test(normalized)) return true;
+
+  const namedSpecifiers = normalized.match(/^\{([\s\S]*)\}$/)?.[1];
+  if (!namedSpecifiers) return false;
+
+  const specifiers = namedSpecifiers
+    .split(',')
+    .map((specifier) => specifier.trim())
+    .filter(Boolean);
+  return specifiers.length > 0 && specifiers.every((specifier) => /^type\b/.test(specifier));
+}
+
+/**
+ * Finds module imports while ignoring comments, strings, and regular expressions.
+ * The descriptor keeps enough information for callers to distinguish runtime
+ * static imports from type-only imports without introducing a parser dependency.
+ */
+export function findModuleImportDescriptors(code: string): ModuleImportDescriptor[] {
+  const codePositions = createCodePositionMap(code);
+  const descriptors: ModuleImportDescriptor[] = [];
+  const staticFromPattern = /\b(?:import|export)\s+([\s\S]*?)\s+from\s*(["'])([^"']+)\2/g;
+  const dynamicPattern = /\bimport\s*\(\s*(?:\/\*[\s\S]*?\*\/\s*)?(["'])([^"']+)\1\s*\)/g;
+  const requirePattern = /\brequire\s*\(\s*(["'])([^"']+)\1\s*\)/g;
+  const sideEffectPattern = /\bimport\s*(["'])([^"']+)\1/g;
+
+  for (const match of code.matchAll(staticFromPattern)) {
+    if (!codePositions[match.index!]) continue;
+    descriptors.push({
+      kind: 'static',
+      syntax: 'import',
+      source: match[3],
+      typeOnly: isTypeOnlyClause(match[1]),
+    });
   }
-  return Array.from(sources);
+
+  for (const match of code.matchAll(dynamicPattern)) {
+    if (!codePositions[match.index!]) continue;
+    descriptors.push({ kind: 'dynamic', syntax: 'import', source: match[2], typeOnly: false });
+  }
+
+  for (const match of code.matchAll(requirePattern)) {
+    if (!codePositions[match.index!]) continue;
+    descriptors.push({ kind: 'dynamic', syntax: 'require', source: match[2], typeOnly: false });
+  }
+
+  for (const match of code.matchAll(sideEffectPattern)) {
+    if (!codePositions[match.index!]) continue;
+    descriptors.push({ kind: 'static', syntax: 'import', source: match[2], typeOnly: false });
+  }
+
+  return descriptors;
+}
+
+export function findModuleImportSources(code: string): string[] {
+  return Array.from(
+    new Set(
+      findModuleImportDescriptors(code)
+        .filter(({ syntax }) => syntax === 'import')
+        .map(({ source }) => source)
+    )
+  );
 }
 
 export function sanitizeDevEntryPath(devEntryPath: string): string {
--- a/integration/fixtures/basic-remote/exposed-module.js
+++ b/integration/fixtures/basic-remote/exposed-module.js
@@ -1,3 +1,5 @@
+export const REMOTE_NAME = 'remoteApp';
+
 export function greet(name) {
   return `Hello, ${name}!`;
 }
--- a/integration/fixtures/loaded-first-namespace-host/entry.js
+++ b/integration/fixtures/loaded-first-namespace-host/entry.js
@@ -1,0 +1,5 @@
+import * as remoteUtil from 'remote1/Module';
+
+export const name = remoteUtil.REMOTE_NAME;
+document.querySelector('#app').textContent = name;
+console.log('__mf_host_entry_evaluated__');
--- a/integration/fixtures/loaded-first-namespace-host/index.html
+++ b/integration/fixtures/loaded-first-namespace-host/index.html
@@ -1,0 +1,7 @@
+<!doctype html>
+<html>
+  <body>
+    <div id="app"></div>
+    <script type="module" src="./entry.js"></script>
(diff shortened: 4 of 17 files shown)


src/utils/htmlEntryUtils.ts
```diff
@@ -1,20 +1,75 @@
...
-    /\bimport\s*\(\s*["']([^"']+)["']/g,
-    /\bimport\s*["']([^"']+)["']/g,
-  ];
-
-  for (const pattern of patterns) {
-    for (const match of code.matchAll(pattern)) {
-      if (codePositions[match.index!]) sources.add(match[1]);
-    }
+  const descriptors: ModuleImportDescriptor[] = [];
+  const staticFromPattern = /\b(?:import|export)\s+([\s\S]*?)\s+from\s*(["'])([^"']+)\2/g;
```
gioboa, line 32: This code misclassifies the type-only import as a runtime static remote, so the host bootstrap fails preloading remote/type:

```ts
import 'polyfill';
import type { T } from 'remote/type';
```

gioboa (changes requested):
The PR is great, I added a small comment though

src/utils/htmlEntryUtils.ts
dmchoi77: Thanks for catching this. Type-only imports and exports are now excluded from remote tracking, with regression coverage added.

https://github.com/module-federation/vite/pull/1099/commits/f5a106a27c54318a0c26d823b5511df464e7cb75

gioboa (approved):
Thank you @dmchoi77 🙏

PR #1102, opened 2026-08-20
fix: exclude workspace sources from hydration fallback

### LINKED ISSUE #1100: Workspace source packages are treated as entry candidates in dev (only node_modules is excluded)
## Summary

In a pnpm workspace where the host app consumes a **workspace source package** (not a built dist), `useI18n`-style named exports from that package become `undefined` at dev runtime, while `build` and the composed production output are fine.

The root cause appears to be that the hydration-entry fallback excludes only `node_modules` paths. Workspace sources arrive through Vite's `/@fs/` prefix, so they are still treated as entry candidates and get the federation entry injected — which reorders module initialization and leaves the shared package's exports unresolved.

## Environment

```
@module-federation/vite   1.20.0  (also reproduced on 1.20.7)
vite                      7.3.6
pnpm workspace, host + remotes
```

## Where

`lib/index.js:5115` (v1.20.0):

```js
const isHydrationEntryFallback =
  inject === "entry" &&
  entryFiles.length === 0 &&
  (!htmlFilePath || !fs.existsSync(htmlFilePath)) &&
  !clientInjected &&
  !isFederationInternalVirtualId(id) &&
  !id.includes("node_modules") &&    // <-- only node_modules is excluded
  (...)
```

A workspace package imported as source resolves to something like
`/@fs/<repo>/packages/i18n/src/index.ts`, which contains no `node_modules` segment.

## Why the usual workarounds don't apply

We ruled these out by measurement before filing:

| Attempt | Result |
| --- | --- |
| Build the shared packages first | No change — `useI18n` appears 6× in the built dist |
| Check the source export | Present (`export { useI18n } from './useI18n'`) |
| Check the shared list | Present in the federation config |
| Clear `node_modules/.vite` | No change |
| Upgrade 1.20.0 → 1.20.7 | Condition is byte-identical in both |
| Set `build.rollupOptions.input` | The failing branch requires `entryFiles.length === 0`, so it never consults it |
| Pass `skipTransformFor` directly | Not a user-facing option — it is only populated from `exposes` |

Both internal locks (`skipTransformFor`, `forceClientInjected`) are derived from `exposes`, and a **host** has no `exposes` by definition, so there is no user-side lever.

## Suggested direction

Treat `/@fs/`-resolved workspace sources the same way as `node_modules` for entry-candidate purposes — or expose `skipTransformFor` as a plugin option so hosts can opt a path out.

We are happy to test a patch against our workspace if that helps.

dmchoi77:
## Summary

Fixes #1100

When `hostInitInjectLocation: 'entry'` is used without an explicit HTML or Rollup entry in dev, the hydration fallback identifies a module that calls `hydrateRoot`, `createRoot`, or Vue `mount()` and injects the host-init bootstrap into it.

The fallback previously excluded `node_modules` but not workspace source modules. Vite can represent those modules as `/@fs/` IDs or as absolute filesystem paths outside `viteConfig.root`. A workspace package entry containing `createRoot` could therefore be mistaken for the application entry and wrapped with `?mf-entry-bootstrap`, breaking its original exports.

## Changes

- Decode Vite IDs and remove query/hash suffixes before filesystem classification.
- Exclude `/@fs/` modules and existing filesystem modules outside `viteConfig.root` from the hydration fallback.
- Keep Module Federation virtual IDs and framework virtual entries eligible for the existing fallback.
- Preserve explicit `entryFiles`, `node_modules` filtering, and `forceClientInjected` behavior.
- Add regression coverage for workspace source IDs, external absolute source paths, app-root hydration entries, and virtual hydration entries.

## Validation

- `pnpm exec vitest run src/plugins/__tests__/pluginAddEntry.test.ts`
- `pnpm test`
- `pnpm run typecheck`
- `pnpm run fmt.check`
- `pnpm build`

## Scope note

Existing non-`node_modules` filesystem entries outside the application root are no longer eligible for this implicit hydration fallback. Applications using such entries should use an explicit Rollup input or a virtual entry instead.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -573,6 +573,119 @@ describe('pluginAddEntry', () => {
     expect(other).toBeUndefined();
   });
 
+  it('does not consume hydration fallback for Vite /@fs workspace source IDs', async () => {
+    const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mf-add-entry-fs-workspace-'));
+    const plugins = addEntry({
+      entryName: 'hostInit',
+      entryPath: '/virtual/hostInit.js',
+      inject: 'entry',
+    });
+    const buildPlugin = plugins[1];
+
+    runConfig(
+      buildPlugin,
+      {} as ConfigPluginContext,
+      { build: { rollupOptions: {} } },
+      { command: 'serve', mode: 'development' }
+    );
+    runConfigResolved(buildPlugin, {
+      root: tempDir,
+      base: '/',
+      command: 'serve',
+      build: { rollupOptions: {} },
+    } as unknown as ResolvedConfig);
+
+    const result = await runTransform(
+      buildPlugin,
+      'export const useI18n = () => undefined;\ncreateRoot(document);',
+      '/@fs//workspace/packages/i18n/src/index.ts'
+    );
+
+    expect(result).toBeUndefined();
+  });
+
+  it('does not consume hydration fallback for an existing source outside the app root', async () => {
+    const appRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'mf-add-entry-app-root-'));
+    const workspaceRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'mf-add-entry-workspace-root-'));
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Nice fix 👍 thank you @dmchoi77

PR #1106, opened 2026-08-21
fix: mirror remote snapshots under their real container name

### LINKED ISSUE #1105: [Bug]: remotes are registered under a synthetic per-owner name, so rspack/webpack containers resolving the same remote by its real name fail with RUNTIME-011
### Describe the bug

`getRuntimeRemoteAlias` rewrites every remote's **`name`** into a synthetic per-owner string before handing it to the runtime, while `alias` already carries the local specifier. Global snapshot keys are derived from that name, so the manifest snapshot is cached under the synthetic key and the container's real name never gets one.

Any rspack/webpack container in the same page that declares the same remote — which it does by the real container name — misses that key, falls back to the remote's own bare-name entry (written by `SnapshotHandler.loadRemoteSnapshotInfo`, whose `remoteEntry` is `""` by design), and the runtime raises `RUNTIME-011` instead of fetching the manifest.

Repro output, with `provider` loaded by the Vite host through the alias `aliasProvider` and then by an rspack container under its real name:

```
__FEDERATION__.moduleInfo keys:
  host_vite                                                                                       remoteEntry: ""
  __mfe_internal__host_vite__mf_owner__1__aliasProvider:http://localhost:3002/mf-manifest.json     remoteEntry: "remoteEntry.js"
  provider                                                                                        remoteEntry: ""
  leaf:http://localhost:3003/mf-manifest.json                                                     remoteEntry: "remoteEntry.js"
```

`provider:<manifestUrl>` — the key an rspack consumer asks for — does not exist. `leaf:<manifestUrl>` shows what a correct key looks like: `leaf` was loaded by an rspack container, so it landed under its real name.

**Where it comes from** (`lib/index.js`, 1.20.6):

```js
// :586
const MF_OWNER_INFIX = "__mf_owner__";

// :1117-1120
function getRuntimeRemoteAlias(alias, options) {
	if (!options) return alias;
	return `${options.internalName}${MF_OWNER_INFIX}${getRuntimeInitOwnerId(options)}__${alias}`;
}
```

used as the runtime `name`:

```js
// :4929-4936
runtime.registerRemotes([{
	entryGlobalName: ...,
	name: options ? runtimeRemoteAlias : remote.name,   // :4931
	alias: remoteAlias,
	entry: ...,
	...
}]);
```

The real container name is right there — `options.remotes[alias].name`, preserved by `normalizeRemoteItem` (`:380-389`) — and is dropped. Same for `getRuntimeRemoteId` (`:4751-4754`) and the `usedRemotes` list (`:3072`). Snapshot keys come from that name only (`runtime-core/dist/global.js:107-111` + `utils/tool.js:9-13`).

Three things make this hard to work around:

- The `if (!options) return alias` escape at `:1118` is unreachable from config: `getRemoteVirtualModule` defaults `options` to `getNormalizeModuleFederationOptions()` (`:4691`) and every call site passes it. The prefix is applied even with a single `federation()` instance and owner id `1`.
- There is no opt-out: `normalizeModuleFederationOptions` (`:530-572`) whitelists the accepted options and none of them affects remote naming. Passing `name` explicitly in the remote object does not help — `registerRemotes` overwrites it.
- `1.1.5` has no `__mf_owner__` / `getRuntimeRemoteAlias` at all, so this arrived somewhere between 1.1.5 and 1.20.6. Apps that host webpack/rspack remotes break on upgrade.

Owner scoping presumably exists to keep multiple `federation()` instances from colliding, but `__FEDERATION__.moduleInfo` is a page-global namespace keyed by container name and shared with every other bundler's containers. Using the identity field for local disambiguation, when `alias` is already doing that job, is what breaks cross-bundler resolution. Would a fix along the lines of "prefer `options.remotes[alias].name` for the runtime `name`, keep the synthetic string for internal ids only" be acceptable? Happy to send a PR if so.

### Secondary defect (in `runtime-core`, filing here since it is reached through this path)

`getTargetSnapshotInfoByModuleInfo` (`runtime-core@2.8.2/dist/global.js:95-105`) accepts the bare entry when only its `version` matches, even though its `remoteEntry` is `""`:

```js
if ("version" in moduleInfo && moduleInfo["version"]) {
	const { version, ...resModuleInfo } = moduleInfo;
	const getModuleInfoWithoutVersion = getInfoWithoutType(
		nativeGlobal.__FEDERATION__.moduleInfo,
		getFMId(resModuleInfo)
	).value;
	if (getModuleInfoWithoutVersion?.version === version) return getModuleInfoWithoutVersion;
}
```

That snapshot is truthy, so `isManifestProvider` is false, the manifest fetch at `SnapshotHandler.js:109-116` is never reached, and `assignRemoteInfo` throws RUNTIME-011 (`plugins/snapshot/index.js:11-12`). A snapshot with no entry should arguably never satisfy a lookup whose purpose is to find one — falling through to the manifest fetch would make this failure self-healing regardless of the naming issue.

(The bare entry's `version` is the manifest URL because the parent host propagates its resolved version into the child container's own options: `plugins/snapshot/index.js:18` -> `module/index.js:23-24` -> `webpack-bundler-runtime/dist/initContainerEntry.js:7-11`. That is why the two keys collide.)

### Consequence in a real app

The failure surfaces far from its cause. In the app where we hit it, the failing require sat in an rspack container's exposed entry module, at module-evaluation time. The runtime installs a throwing module factory on remote-load failure (`webpack-bundler-runtime/dist/remotes.js:19-26`) and swallows the rejection into the ensure-chunk chain (`:31-32`), so the exposed module aborts halfway through its body, webpack 5 neither evicts nor retries the module cache, **all ten of that container's exports read as `undefined` for the rest of the page**, and the only visible error is a `TypeError: Cannot read properties of undefined` inside an unrelated export getter. RUNTIME-011 never reaches the console.

It is also intermittent: if the consuming container resolves the remote before the bare entry exists, the manifest is fetched, the correct real-name key is cached, and everything works.

### Version

`@module-federation/vite` 1.20.6, `vite` 7.3.6 (also `@module-federation/enhanced` 2.8.2, `runtime-core` 2.8.2 resolved, node 24.4.1)

### Reproduction

https://github.com/h11g/mf-vite-real-name-snapshot-repro

`pnpm install && pnpm start`, then open http://localhost:3000/. Four containers exposing plain functions, no framework:

```
host-vite  :3000   vite + @module-federation/vite    remotes: aliasProvider -> provider, aliasConsumer -> consumer
consumer   :3001   rspack                           remotes: provider (real container name); exposes ./widget
provider   :3002   rspack                           remotes: leaf;                           exposes ./thing
leaf       :3003   rspack                           exposes ./noop
```

The page loads `aliasProvider/thing` (works), dumps the snapshot table, then loads `aliasConsumer/widget` (fails). `provider` consumes `leaf` for one reason only: loading any remote is what makes the runtime write the loading container's own bare-name entry.

The host sets `shareStrategy: 'loaded-first'` to make the ordering deterministic. With the default `'version-first'` every remote initialises at boot and `consumer` happens to fetch `provider`'s manifest itself, repairing the table by accident; the bug still occurs under `'version-first'` whenever the consuming container is loaded lazily, after the provider has already run.

### Relevant log output

```shell
step 1  provider/thing -> provider(leaf)

__FEDERATION__.moduleInfo keys after step 1:
  host_vite
      remoteEntry: ""
  __mfe_internal__host_vite__mf_owner__1__aliasProvider:http://localhost:3002/mf-manifest.json
      remoteEntry: "remoteEntry.js"
  provider
      remoteEntry: ""
  leaf:http://localhost:3003/mf-manifest.json
      remoteEntry: "remoteEntry.js"

step 2  consumer/widget threw:
Error: [ Federation Runtime ]: Remote container initialization failed. #RUNTIME-015
args: {"hostName":"host_vite","remoteName":"__mfe_internal__host_vite__mf_owner__1__aliasConsumer","remoteEntryUrl":"http://localhost:3001/remoteEntry.js","remoteEntryKey":"consumer","shareScope":"default"}
Original Error Message:
 Error: [ Federation Runtime ]: The remoteEntry URL is missing from the remote snapshot. #RUNTIME-011
args: {"remoteName":"provider"}
    at logAndReport (http://localhost:3000/assets/index-BuDsyMy7.js:479:10)
    at error (http://localhost:3000/assets/index-BuDsyMy7.js:487:34)
    at async Module$1.init (http://localhost:3000/assets/index-BuDsyMy7.js:1832:7)
    at async Module$1.get (http://localhost:3000/assets/index-BuDsyMy7.js:1858:32)
    at async RemoteHandler.loadRemote (http://localhost:3000/assets/index-BuDsyMy7.js:3028:31)
```

### Workaround

A host runtime plugin that mirrors each resolved manifest snapshot under its real name, so the primary lookup hits before the empty fallback is consulted:

```js
const realNameSnapshotPlugin = () => ({
	name: 'real-name-snapshot',
	afterLoadSnapshot(args) {
		const {globalName, version, remoteEntry} = args.remoteSnapshot ?? {};
		if (!globalName || !version || !remoteEntry) return args;
		const {moduleInfo} = globalThis.__FEDERATION__ ?? {};
		const key = `${globalName}:${version}`;
		if (moduleInfo && !moduleInfo[key]) moduleInfo[key] = args.remoteSnapshot;
		return args;
	},
});
```

It only repairs lookups that carry a version; one arriving without a version still reads the bare key directly (`SnapshotHandler.js:29-36` -> `global.js:96`).

### Validations

- [x] Read the docs.
- [x] Checked that there isn't already an issue that reports the same bug.

h11g:
Fixes #1105.

## Problem

Remotes are registered with the runtime under an owner-scoped name — `getRuntimeRemoteAlias` (`src/virtualModules/virtualRuntimeInitStatus.ts:53-56`) — and global snapshot keys derive from the remote *name* (`runtime-core/dist/global.js:107-111` + `utils/tool.js:9-13`). So `__FEDERATION__.moduleInfo` ends up with

```
__mfe_internal__host__mf_owner__1__aliasProvider:https://…/provider/mf-manifest.json   remoteEntry: "remoteEntry.js"
```

and never an entry for `provider:https://…/provider/mf-manifest.json`.

A webpack/rspack container in the same page declares the same remote by its **real container name**. Its lookup misses that key, then `getTargetSnapshotInfoByModuleInfo` (`global.js:95-105`) strips the version and accepts the bare-name entry the provider wrote to describe itself, whose `remoteEntry` is `""` (`SnapshotHandler.loadRemoteSnapshotInfo`, `:59-70`). That snapshot is truthy, so `isManifestProvider` is false, the manifest fetch at `SnapshotHandler.js:109-116` is never reached, and `assignRemoteInfo` throws:

```
[ Federation Runtime ]: The remoteEntry URL is missing from the remote snapshot. #RUNTIME-011
args: {"remoteName":"provider"}
```

It is intermittent — resolving before the bare entry exists fetches the manifest and works — and it surfaces far from its cause: if the failing require sits in an exposed entry module at module-evaluation time, the runtime installs a throwing module factory (`webpack-bundler-runtime/dist/remotes.js:19-26`), swallows the rejection into the ensure-chunk chain (`:31-32`), webpack 5 never evicts the module cache, and **every export of that container reads as `undefined` for the rest of the page**. The only visible error is a `TypeError: Cannot read properties of undefined` in an unrelated export getter.

## Approach

The obvious fix — put the real container name in the runtime `name` and move the synthetic string to `alias` — is deliberately **not** what this PR does. #1001 ("preserve public aliases for loadRemote") moved the alias slot the other way on purpose, and `virtualRemotes.test.ts` ("does not reuse pending or loaded remotes across federation instances") pins the property that two owners pointing at *different entries for the same container name* stay separate. Registering the real name would make the second `registerRemote` a silent no-op (`RemoteHandler.registerRemote` skips a duplicate `name` unless `force`), so the second owner would silently load the first owner's entry.

So this PR leaves naming and registration untouched and repairs the lookup table instead: a generated runtime plugin mirrors every resolved manifest snapshot under `<globalName>:<version>` in `afterLoadSnapshot`. `globalName` is the real container name and `version` is the manifest URL (`sdk/generateSnapshotFromManifest.js:50,53`), so the mirrored key is exactly what another bundler's container asks for, and its primary lookup hits before the empty fallback is consulted.

Properties:

- **Additive.** Owner scoping, public aliases and `loadRemote` ids are unchanged; nothing that reads a remote name back is touched.
- **Multi-owner safe.** Different entries for the same container name produce different `version`s, hence different mirrored keys — no collision.
- **Self-limiting.** Only snapshots that carry `globalName`, `version` *and* a non-empty `remoteEntry` are mirrored; mirroring an entry-less snapshot would spread the very failure this fixes. An existing key is never clobbered.
- **Ordering-proof.** Once mirrored, the provider's own `loadRemoteSnapshotInfo` finds `provider:<url>` and no longer writes the bare stub at all, so the fallback is never reachable.

It follows the existing generated-plugin pattern (`__mfSharePinLifecyclePlugin`, `__mfTreeShakingSnapshotPlugin`) and is registered in the same `runtimeInit({ plugins: [...] })` list.

## Testing

- New unit test in `src/virtualModules/__tests__/virtualRemoteEntry.test.ts` extracts the generated `__mfRealNameSnapshotPlugin` and drives `afterLoadSnapshot` over three cases: mirrors a manifest snapshot under its real name, skips a snapshot with an empty `remoteEntry`, and keeps an existing real-name entry.
- `pnpm test` — 49 files, 1078 passed.
- `pnpm build && pnpm test:integration` — 16 files, 84 passed.
- `pnpm typecheck`, `pnpm fmt.check` clean.
- End to end against a minimal four-container repro (vite host + three rspack containers, plain functions, no framework): https://github.com/h11g/mf-vite-real-name-snapshot-repro

  Before — `provider:<manifestUrl>` absent, bare `provider` present with `remoteEntry: ""`:
  ```
  step 2  consumer/widget threw:
  RUNTIME-015 → Original: RUNTIME-011  The remoteEntry URL is missing from the remote snapshot.
  args: {"remoteName":"provider"}
  ```
  After, with this branch linked into the host:
  ```
  provider:http://localhost:3002/mf-manifest.json      remoteEntry: "remoteEntry.js"
  step 2  consumer/widget -> widget(provider(leaf))
  ```

I did not add an e2e spec: expressing the four-container graph means reshaping apps under `examples/vite-webpack-rspack/` that other specs share. Happy to add one there if you'd prefer it in-repo.

## Note on the secondary defect

`runtime-core`'s `getTargetSnapshotInfoByModuleInfo` accepting a snapshot whose `remoteEntry` is `""` (`global.js:99-104`) is the other half of this: a snapshot with no entry should arguably never satisfy a lookup whose purpose is to find one, and falling through to the manifest fetch would make this class of failure self-healing. That belongs in `module-federation/core`; happy to file it there too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1011,6 +1011,67 @@ describe('virtualRemoteEntry', () => {
     expect(code).toContain('initRes.initializeSharing(shareScopeName');
   });
 
+  it('mirrors resolved manifest snapshots under the real container name', async () => {
+    const mod = await import('../virtualRemoteEntry');
+
+    const code = mod.generateRemoteEntry(
+      {
+        internalName: '__mfe_internal__host',
+        name: 'host',
+        filename: 'remoteEntry.js',
+        remotes: {},
+        runtimePlugins: [],
+        shareScope: 'default',
+        shareStrategy: 'loaded-first',
+      } as any,
+      'virtual:exposes',
+      'build'
+    );
+
+    const pluginCode = code.slice(
+      code.indexOf('function __mfRealNameSnapshotPlugin()'),
+      code.indexOf('const runtimeResolveShareHook =')
+    );
+    const plugin = new Function(`${pluginCode}; return __mfRealNameSnapshotPlugin();`)() as {
+      afterLoadSnapshot: (args: { remoteSnapshot: Record<string, string> }) => unknown;
+    };
+
+    const manifestUrl = 'https://cdn.invalid/provider/mf-manifest.json';
+    const moduleInfo: Record<string, unknown> = {
+      [`__mfe_internal__host__mf_owner__1__alias:${manifestUrl}`]: {},
+      // The provider's own entry, which shadows a real-name lookup.
+      provider: { version: manifestUrl, remoteEntry: '' },
+    };
+    const federationGlobal = globalThis as { __FEDERATION__?: unknown };
+    const hadFederationGlobal = '__FEDERATION__' in federationGlobal;
+    const previous = federationGlobal.__FEDERATION__;
+    federationGlobal.__FEDERATION__ = { moduleInfo };
+
+    try {
+      const remoteSnapshot = {
+        globalName: 'provider',
+        version: manifestUrl,
+        remoteEntry: 'remoteEntry.js',
+      };
+      plugin.afterLoadSnapshot({ remoteSnapshot });
+      expect(moduleInfo[`provider:${manifestUrl}`]).toBe(remoteSnapshot);
+
+      plugin.afterLoadSnapshot({
+        remoteSnapshot: { globalName: 'other', version: manifestUrl, remoteEntry: '' },
+      });
+      expect(`other:${manifestUrl}` in moduleInfo).toBe(false);
+
+      plugin.afterLoadSnapshot({ remoteSnapshot: { ...remoteSnapshot } });
+      expect(moduleInfo[`provider:${manifestUrl}`]).toBe(remoteSnapshot);
+    } finally {
+      if (hadFederationGlobal) {
+        federationGlobal.__FEDERATION__ = previous;
+      } else {
+        delete federationGlobal.__FEDERATION__;
+      }
+    }
+  });
+
   it('inlines a dedicated build-only initResolve bootstrap into remoteEntry', async () => {
     const mod = await import('../virtualRemoteEntry');
 
@@ -1264,7 +1325,7 @@ describe('virtualRemoteEntry', () => {
       'if (share.treeShaking) continue;'
     );
     expect(code).toContain(
-      'plugins: [__mfSharePinLifecyclePlugin(), __mfTreeShakingSnapshotPlugin(),'
+      'plugins: [__mfSharePinLifecyclePlugin(), __mfRealNameSnapshotPlugin(), __mfTreeShakingSnapshotPlugin(),'
     );
   });
 
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -1339,7 +1339,7 @@ export function generateRemoteEntry(
       name: mfName,
       remotes: ${options.shareStrategy === 'loaded-first' ? '[]' : 'usedRemotes'},
       shared: usedShared,
-      plugins: [__mfSharePinLifecyclePlugin(), ${hasTreeShakingShared ? '__mfTreeShakingSnapshotPlugin(),' : ''} ...__browserPlugins, ...__ssrPlugins],
+      plugins: [__mfSharePinLifecyclePlugin(), __mfRealNameSnapshotPlugin(), ${hasTreeShakingShared ? '__mfTreeShakingSnapshotPlugin(),' : ''} ...__browserPlugins, ...__ssrPlugins],
       ${options.shareStrategy ? `shareStrategy: '${options.shareStrategy}'` : ''}
     });
     ${
@@ -1371,6 +1371,29 @@ export function generateRemoteEntry(
             return defaultResolver(...resolverArgs);
           };
           lifecycle.pinned.reveal();
+          return args;
+        }
+      };
+    }
+    function __mfRealNameSnapshotPlugin() {
+      return {
+        name: "vite-real-name-snapshot-plugin",
+        afterLoadSnapshot(args) {
+          // Remotes are registered under an owner-scoped runtime name, so the
+          // global snapshot only ever holds that name. Containers built by other
+          // bundlers ask for the same remote by its real container name, miss, and
+          // fall back to that container's own entry, whose remoteEntry is empty
+          // (RUNTIME-011). Mirroring the resolved snapshot under the real name
+          // keeps the primary lookup off that fallback.
+          const snapshot = args && args.remoteSnapshot;
+          const globalName = snapshot && snapshot.globalName;
+          const version = snapshot && snapshot.version;
+          if (!globalName || !version || !snapshot.remoteEntry) return args;
+          const moduleInfo = globalThis.__FEDERATION__ && globalThis.__FEDERATION__.moduleInfo;
+          const realNameKey = globalName + ":" + version;
+          if (moduleInfo && !moduleInfo[realNameKey]) {
+            moduleInfo[realNameKey] = snapshot;
+          }
           return args;
         }
       };


gioboa (approved):
Thank you for this fix @h11g

PR #1107, opened 2026-08-21
fix: preload remote-only dev singleton wrappers before entry import

### LINKED ISSUE #1104: [Bug]: Blank page on first cold dev start in 1.20.7 — `_jsxDEV is not a function` (regression from 1.20.6)
### Describe the bug

On the first load after a cold `node_modules/.vite` cache, the dev server renders a
blank page and the console shows:

    Uncaught (in promise) TypeError: _jsxDEV is not a function

A manual refresh recovers, and it does not reproduce again until the Vite dep cache
is cleared. Bisecting the plugin version isolates the regression to 1.20.7.

### Reproduction
1. Vite + React app with `react` declared as a shared singleton
2. `rm -rf node_modules/.vite`
3. `pnpm run dev`
4. Open the app in a fresh browser context (no warm cache)

### Bisect

| @module-federation/vite | first cold load |
|---|---|
| 1.19.1 | OK |
| 1.20.0 | OK |
| 1.20.4 | OK |
| 1.20.6 | OK |
| **1.20.7** | **blank page — `_jsxDEV is not a function` (3/3 runs)** |

Same project, same config, only the plugin version changed.

### Environment
- vite 8.2.1
- react / react-dom 19.2.8
- react-router 8.x
- pnpm 10.x
- Reproduced on Windows 11 and on Linux (Node 22)

### Minimal config
```ts
federation({
  name: 'overviewRemote',
  filename: 'remoteEntry.js',
  dts: false,
  exposes: { './Overview': './src/pages/Overview.tsx' },
  shared: {
    react: { singleton: true },
    'react/jsx-runtime': { singleton: true },
    'react/jsx-dev-runtime': { singleton: true },
    'react-router': { singleton: true },
  },
})
```

### Things that did NOT help
- Async boundary (`main.tsx` containing only `import('./bootstrap')`)
- Explicitly awaiting `loadShare('react')` / `loadShare('react/jsx-dev-runtime')`
  before importing the app
- `react({ jsxRuntime: 'classic' })` — the error simply becomes
  `React is not defined`, i.e. the same race with a different binding
- `optimizeDeps.noDiscovery: true` with a complete `include` list
- Adding / removing the jsx runtimes from `shared`

The `jsxRuntime: 'classic'` result suggests the issue is that the shared scope is not
populated when the first module evaluates, rather than anything specific to the JSX
runtime entry point.

### Version

8.2.1

### Reproduction

1. Vite + React app with `react` declared as a shared singleton 2. `rm -rf node_modules/.vite` 3. `pnpm run dev` 4. Open the app in a fresh browser context (no warm cache)

### Relevant log output

```shell

```

### Validations

dmchoi77:
Fixes #1104 

## Summary

Fixes a cold-start failure in remote-only development builds where a shared singleton wrapper is evaluated after the application entry, leaving `react/jsx-dev-runtime`'s `_jsxDEV` undefined on the first page load.

## Root cause

- #1087 removed the static local fallback import from remote-only dev singleton wrappers to prevent the duplicate singleton evaluation described in #1085.
- When the wrapper is first evaluated while the application entry is loading, its shared exports are assigned asynchronously.
- The existing bootstrap awaited `pendingShareLoads`, but the wrapper had not been evaluated yet, so no pending promise had been registered.

## Changes

- Preload remote-only dev singleton wrappers from the `hostInit` bootstrap.
- Await the resulting `pendingShareLoads` before importing the application entry.
- Limit preloading to `singleton: true` shares, excluding `import: false`, tree-shaken shares, and unresolvable generated subpaths.

### DIFF
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -330,6 +330,51 @@ describe(`singleton React dev fallback (Vite ${viteVersion})`, () => {
     expect(pageErrors).toEqual([]);
   }, 60_000);
 
+  it('boots on first load when a remote-only singleton imports jsx-dev-runtime (#1104)', async () => {
+    const exposedEntry = path.join(fixtureRoot, 'src/Issue1104App.js');
+    await Promise.all([
+      writeFile(
+        path.join(fixtureRoot, 'src/main.js'),
+        `import { jsxDEV } from 'react/jsx-dev-runtime';
+
+const element = jsxDEV('div', { children: 'ready' }, undefined, false, undefined, undefined);
+window.__issue1104Result = element.type === 'div' ? 'ready' : 'invalid';
+document.body.textContent = window.__issue1104Result;
+`
+      ),
+      writeFile(exposedEntry, 'export default function Issue1104App() { return null; }\n'),
+    ]);
+
+    const host = await createDevServer(
+      'issue-1104-host',
+      {
+        name: 'issue1104Host',
+        exposes: { './App': exposedEntry },
+        dts: false,
+        shared: {
+          react: { singleton: true },
+        },
+      },
+      fixtureRoot
+    );
+    const browser = await chromium.launch({ channel: 'chrome', headless: true });
+    cleanupTasks.push(() => browser.close());
+    const page = await browser.newPage();
+    const pageErrors: string[] = [];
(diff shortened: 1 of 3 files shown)


gioboa (approved):
That's great 👏 thank you @dmchoi77

PR #1109, opened 2026-08-22
fix: fix dev cold-start import:false shares

### LINKED ISSUE #1090: [Bug]: dev cold start fails for import:false host-provided shares (bootstrap race / undefined loadShare export)
### Describe the bug

**Dev only:** After restarting the remote Vite dev server, the **first** hard refresh from a cross-origin host fails to load an exposed module (e.g. `./lifecycle`). A **second** refresh usually succeeds. **Preview and production builds are unaffected.**

**Setup:**
- Remote: Vite + `@module-federation/vite`, with multiple **`import: false`** shared dependencies (host-provided)
- An exposed module **statically imports** one of those shared packages at module top level
- Host: separate Module Federation host (webpack), loads `remoteEntry.js` cross-origin, then calls `remote.init(hostShared)` and `remote.get('./lifecycle')`

**Errors** (same underlying timing issue):
1. `[Module Federation] Shared module <package> was imported before federation bootstrap finished.`
2. `TypeError: Cannot destructure property 'X' of '<pkg>' as it is undefined.`

**What we observed:**
- `remote.init()` can resolve while a critical `import: false` share is **still missing** from the generated share cache; other shares may already be present
- The generated `__loadShare__` virtual module for host-provided shares can **export `undefined` synchronously** during module evaluation, before async bootstrap completes

This points to a **bootstrap ordering race in dev** between `init()`, share cache population, and expose module evaluation — not application business logic.

### Version

v1.20.7

### Reproduction

None at the moment.

### Relevant log output

```shell
Minimal reproduction: We could not publish a reliably failing minimal repo. The bug reproduces in our production-like remote + runtime host setup under the steps above. We attempted a simplified public repro but it always succeeds due to shallow module graph / race timing. We can provide recordings, logs, and a verified patch on 1.20.7.
```

### Validations

Image

gioboa:
Close #1090

Fixes a dev cold-start race where import:false shares were skipped before the usage scanner marked them materialized, leaving exports undefined.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1865,6 +1865,89 @@ describe('virtualRemoteEntry', () => {
     expect(order).toEqual(['@repro/shared-lib', '@repro/core']);
   });
 
+  it('resolves an import:false share during init even before the dev scanner has materialized it (#1090)', async () => {
+    // Dev-only cold start: a cross-origin host requests remoteEntry.js before Vite's
+    // resolveId has ever processed the exposed module's own import of the share, so
+    // the share is configured but not yet materialized when this remoteEntry.js is
+    // generated. Because import:false shares have no local fallback, init() must
+    // still attempt to resolve them — otherwise the share is never seeded, and the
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1111, opened 2026-08-22
fix: dev shared-key cache invalidation

### LINKED ISSUE #1101: [Bug]: Sharing a package that is a nested dependency of another shared package crashes the dev server with "Cannot read properties of undefined (reading 'shareConfig')"
## Describe the bug

When a `shared` entry is **both** (a) declared for module federation and (b) a dependency of *another* shared package, the plugin's dev-mode logic `excludeSharedSubDependencies()` **deletes** that entry from the `shared` map (and logs `Auto-excluding "<pkg>" from shared modules for dev mode`). After the deletion, the shared resolver still resolves imports of the deleted package, and `writeLoadShareModule()` is called with an **`undefined` `shareItem`**. `getSharedCacheDescriptor()` then dereferences `shareItem.shareConfig.singleton` and throws.

The dev server does not hard-crash (Vite keeps running and still serves `remoteEntry.js`), but the console is flooded with the `TypeError` — one occurrence per import site of the excluded package — which is alarming and hides real errors. `vite build` is unaffected (the exclusion only runs for `command === "serve"`).

**Trigger condition (important):** the exclusion — and therefore the crash — only happens when the nested dependency is shared in a form that is **neither `singleton: true` nor `import: false`** (e.g. the plain version-string / bare form `'pkg': '^1.0.0'`, or `'pkg': {}`). Declaring it `{ singleton: true }` or `{ import: false }` makes `excludeSharedSubDependencies()` skip it (guard below), and the crash disappears — which is the current workaround.

### Root cause (TypeScript source, tag [`1.20.6`](https://github.com/module-federation/vite/tree/1.20.6))

1. [`excludeSharedSubDependencies()`](https://github.com/module-federation/vite/blob/1.20.6/src/plugins/pluginProxySharedModule_preBuild.ts#L229) deletes a shared entry that is also a nested dependency of another shared package — **unless** that entry is `singleton: true` / `import: false` (the guard). It logs the `Auto-excluding` warning and `delete shared[depKey]`:
   ```ts
   // src/plugins/pluginProxySharedModule_preBuild.ts#L239-L253
   if (
     shared[depKey]?.shareConfig.singleton === true ||   // guard (L241-242)
     shared[depKey]?.shareConfig.import === false
   ) {
     continue;
   }
   mfWarn(`"${dep}" is a dependency of shared package "${parentKey}" ... Auto-excluding ...`); // L247
   delete shared[depKey];   // L253  <-- entry removed from `shared`
   ```
   It runs only for `command === 'serve'`, called at [`src/index.ts#L571`](https://github.com/module-federation/vite/blob/1.20.6/src/index.ts#L571) and [`pluginProxySharedModule_preBuild.ts#L399`](https://github.com/module-federation/vite/blob/1.20.6/src/plugins/pluginProxySharedModule_preBuild.ts#L399).

2. The dev optimizeDeps plugin [`module-federation:optimize-shared-resolver`](https://github.com/module-federation/vite/blob/1.20.6/src/index.ts#L582) still resolves imports of the (now-deleted) package and forwards an **`undefined` `shareItem`**:
   ```ts
   // src/index.ts#L606-L633 (resolveId)
   const key = findSharedKey(source, shared);   // L611 — still matches the deleted key
   if (!key) return;
   ...
   const shareItem = shared[key];               // L614 — undefined (key was deleted)
   ...
   writeLoadShareModule(source, shareItem, _command, isRolldown, options);   // L633
   ```
   [`findSharedKey`](https://github.com/module-federation/vite/blob/1.20.6/src/plugins/pluginProxySharedModule_preBuild.ts#L105) goes through [`getSharedKeyMatcher`](https://github.com/module-federation/vite/blob/1.20.6/src/plugins/pluginProxySharedModule_preBuild.ts#L122), which memoizes the matcher in a `WeakMap` keyed by the `shared` object (decl [L120](https://github.com/module-federation/vite/blob/1.20.6/src/plugins/pluginProxySharedModule_preBuild.ts#L120), read L125, set L183). `excludeSharedSubDependencies` mutates `shared` with `delete` but never invalidates that cache — so a matcher built before the deletion keeps matching the removed key. (I haven't instrumented the exact cache timing; maintainers can confirm. The observable contradiction is clear regardless: the key is gone from `shared`, yet `resolveId` still resolves it and passes `undefined` as `shareItem`.)

3. [`getSharedCacheDescriptor()`](https://github.com/module-federation/vite/blob/1.20.6/src/utils/packageUtils.ts#L247) dereferences `shareItem.shareConfig.singleton` with no undefined guard, throwing the `TypeError`:
   ```ts
   // src/utils/packageUtils.ts#L247-L253
   export function getSharedCacheDescriptor(pkg: string, shareItem: ShareItem): SharedCacheDescriptor {
     const parts = getSharedCacheKeyParts({
       pkg,
       singleton: shareItem.shareConfig.singleton,   // L250 — shareItem is undefined -> TypeError
       version: shareItem.version,
       scope: shareItem.scope,
     });
   ```

(The stack trace in the log below points at the published `lib/*.js` bundle; the line numbers above are the corresponding TypeScript source.)

### Suggested fix (either or both)
- In the `resolveId` at [`src/index.ts#L614`](https://github.com/module-federation/vite/blob/1.20.6/src/index.ts#L614), bail out when the matched entry no longer exists: `const shareItem = shared[key]; if (!shareItem) return;` (treat an excluded key as not-shared). The esbuild-optimizer variant of the same resolver has the identical unguarded `const shareItem = shared[key]` at [`src/index.ts#L677`](https://github.com/module-federation/vite/blob/1.20.6/src/index.ts#L677).
- And/or invalidate the [`getSharedKeyMatcher`](https://github.com/module-federation/vite/blob/1.20.6/src/plugins/pluginProxySharedModule_preBuild.ts#L122) `WeakMap` after `excludeSharedSubDependencies` mutates `shared` (e.g. `sharedKeyMatcherCache.delete(shared)`), so `findSharedKey` stops matching deleted keys.

## Version
```
vite: 8.2.0
@module-federation/vite: 1.20.6
rolldown: 1.2.3
@vitejs/plugin-react: 6.0.5 (classic runtime)
node: 22.18.0
OS: macOS (Darwin 25.6.0)
```

## Reproduction

Minimal repro: https://github.com/TDliuwang/mf-vite-shareconfig-repro

```bash
npm install
npm run dev      # command === "serve"  -> reproduces
# npm run build  -> succeeds (exclusion only runs for serve)
```

A single remote is enough (no host needed — the error fires at dev startup). It uses two trivial local workspace packages where **`pkg-a` depends on `pkg-b`**, so no external/registry packages are involved. The relevant config:

```ts
// vite.config.ts
federation({
  name: 'remote',
  filename: 'remoteEntry.js',
  exposes: { './x': './src/x.ts' },
  shared: {
    'pkg-a': { singleton: true },   // parent — kept
    'pkg-b': '^1.0.0'               // nested dep of pkg-a, BARE form -> excluded -> crash
  }
})
```
`pkg-a` (`package.json` lists `pkg-b` as a dependency) re-exports from `pkg-b`; `src/x.ts` imports both.

**Expected:** dev server starts without errors; `pkg-b` is either shared or excluded silently.

**Actual (verified on the versions above):** startup logs the `Auto-excluding` warning immediately followed by the `TypeError` (stack below). `npm run build` succeeds.

## Relevant log output

```shell
[Module Federation] "pkg-b" is a dependency of shared package "pkg-a" and is also shared separately. This may cause initialization order issues in dev mode. Consider sharing only "pkg-a".
  Auto-excluding "pkg-b" from shared modules for dev mode.

[plugin module-federation:optimize-shared-resolver]
TypeError: Cannot read properties of undefined (reading 'shareConfig')
    at getSharedCacheDescriptor (…/node_modules/@module-federation/vite/lib/dtsConstants-DyJrx8ah.js:187:24)
    at getSharedCacheDescriptorLiteral (…/node_modules/@module-federation/vite/lib/index.js:1642:24)
    at writeLoadShareModule (…/node_modules/@module-federation/vite/lib/index.js:2756:26)
    at PluginContextImpl.resolveId (…/node_modules/@module-federation/vite/lib/index.js:9239:9)
    at plugin (…/node_modules/rolldown/dist/shared/bindingify-input-options-CMab5gbt.mjs:1457:30)
    at …/node_modules/rolldown/dist/shared/bindingify-input-options-CMab5gbt.mjs:1928:21
    at plugin.<computed> (…/node_modules/rolldown/dist/shared/bindingify-input-options-CMab5gbt.mjs:2125:18)
```

## Screenshot

Image

## Validations
- [x] Read the docs.
- [x] Checked there isn't already an issue reporting the same bug.

gioboa:
Close #1101

Clears the cached share-key matcher when a nested dependency is auto-excluded in dev, so it stops resolving deleted keys and crashing.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -161,7 +161,11 @@ vi.mock('../../utils/VirtualModule', () => ({
   }),
 }));
 
-import { findSharedKey, proxySharedModule } from '../pluginProxySharedModule_preBuild';
+import {
+  excludeSharedSubDependencies,
+  findSharedKey,
+  proxySharedModule,
+} from '../pluginProxySharedModule_preBuild';
 import {
   getResolvedLocalSharedImportMapId,
   getUsedShares,
@@ -370,6 +374,57 @@ describe('pluginProxySharedModule_preBuild', () => {
         name: 'react/jsx-runtime',
       };
       expect(findSharedKey('react/jsx-runtime', explicitlyShared)).toBe('react/jsx-runtime');
+    });
+
+    // findSharedKey memoizes its matcher per `shared` object reference. If
+    // something (e.g. the entry-import scanner) resolves a source through the
+    // matcher before excludeSharedSubDependencies deletes that key from the
+    // same (in-place mutated) `shared` object, a later lookup must not keep
+    // resolving the now-deleted key from the stale cached matcher — otherwise
+    // callers downstream (e.g. writeLoadShareModule) dereference an undefined
+    // shareItem and crash.
+    it('stops resolving a key excludeSharedSubDependencies deletes, even if already cached', () => {
+      hasPackageDependencyMock.mockReturnValue(false);
+      existsSyncMock.mockImplementation(
+        (p: string) => p === '/repo/apps/remote/node_modules/pkg-a/package.json'
+      );
+      readFileSyncMock.mockImplementation((p: string) => {
+        if (p === '/repo/apps/remote/node_modules/pkg-a/package.json') {
+          return JSON.stringify({ name: 'pkg-a', dependencies: { 'pkg-b': '^1.0.0' } });
+        }
+        return '{}';
+      });
+
+      const shared: NormalizedShared = {
+        'pkg-a': {
+          name: 'pkg-a',
+          from: '',
+          version: '1.0.0',
+          scope: 'default',
+          shareConfig: { singleton: true, requiredVersion: '^1.0.0', strictVersion: false },
+        },
+        'pkg-b': {
+          name: 'pkg-b',
+          from: '',
+          version: '1.0.0',
+          scope: 'default',
+          shareConfig: { singleton: false, requiredVersion: '^1.0.0', strictVersion: false },
+        },
+      };
+
+      // Prime findSharedKey's cached matcher for this `shared` object, as the
+      // entry-import scanner does before excludeSharedSubDependencies runs.
+      expect(findSharedKey('pkg-b', shared)).toBe('pkg-b');
+
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+      excludeSharedSubDependencies(shared);
+      warnSpy.mockRestore();
+
+      expect(shared).not.toHaveProperty('pkg-b');
+      expect(findSharedKey('pkg-b', shared)).toBeUndefined();
+
+      existsSyncMock.mockReset().mockReturnValue(false);
+      readFileSyncMock.mockReset().mockReturnValue('{}');
     });
   });
 
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -253,6 +253,9 @@ export function excludeSharedSubDependencies(shared: NormalizedShared): void {
         delete shared[depKey];
         sharedKeys.delete(depKey);
         sharedKeyByBase.delete(dep);
+        // A matcher cached before this deletion would keep resolving the
+        // removed key and hand callers an undefined shareItem.
+        sharedKeyMatcherCache.delete(shared);
       }
     }
   }


gioboa (commented):

Image

PR #1114, opened 2026-08-24
fix: parallelize federation network waterfall

### LINKED ISSUE #1112: Staircase/serial network requests when loading Module Federation remote — high initial load follow up
### Describe the bug

Image

i tested with fix. still i observe  staircase network request whenever remote Resources are loading.  since we have 5 remotes. literally around 4 to 5 seconds spend on here . 

appreciate your feedback and improvements. 

### Version

1.20.7

### Reproduction

https://github.com/Arun-2095/mf-sample

### Relevant log output

```shell

```

### Validations

gioboa:
Close #1112

Preload host init chunks with their static imports and remote entries; remote entry warms its own init chunks when evaluated.

image


gioboa (commented):

Image

PR #1115, opened 2026-08-24
fix: fix up version-first cross-bundler singleton sharing

gioboa:
Refreshes Vite’s shared cache from the runtime-selected provider after initialization, while preserving pre-init deadlock protections.
Fixes Vite↔Webpack/Rspack singleton duplication.


longngo0617:
Will this bug, which I reported in core https://github.com/module-federation/core/issues/5002, be resolved?

gioboa (commented):
Hi @longngo0617 yes, that problem should be solved.

PR #1116, opened 2026-08-24
fix: heal stale loadShare IDs directly

gioboa:
Handles Vite module-graph reuse that bypasses resolveId, preventing stale virtual IDs from falling through to filesystem loading.


gioboa (commented):
@imtaotao here is the change. Could you test the pre-release?

imtaotao:
Tested the PR package locally with our fallback disabled. I verified the stale owner ID path that goes directly through the load hook without running through resolveId again, and it now correctly returns the current loadShare module. Looks good on our side, thanks!

gioboa (commented):

Image

PR #1117, opened 2026-08-24
fix: preserve host-owned singleton cache

gioboa:
Preserves local Vite singleton identity while still refreshing runtime-selected Webpack/Rspack shares.


gioboa (commented):

Image

PR #1120, opened 2026-08-24
fix: control missing import false warnings

### LINKED ISSUE #1118: @module-federation/vite repeatedly logs import: false shared dependency warnings when starting the Vite dev server
## Environment

  - @module-federation/vite: 1.20.7
  - Vite: 8.2.1
  - Node.js: 24.19.0
  - pnpm: 10.27.0

### Description

Some shared dependencies in the Vite configuration use import: false and are intentionally not installed locally in the remote  application:
```ts
  shared: {
    '@pos-dashboard/host': {
      singleton: true,
      requiredVersion: '1.0.0',
      import: false,
    },
    '@pos-dashboard/components': {
      singleton: true,
      requiredVersion: '1.0.0',
      import: false,
    },
  }
```

This is intentional: these modules are provided by the host at runtime, and the remote should not bundle a local fallback.

When starting the Vite dev server, the same warning is logged multiple times for each dependency:

```
  [Module Federation] Shared dependency "@pos-dashboard/host" has import: false but is not installed locally.
    Named imports (e.g. import { ... } from '@pos-dashboard/host') will not work in production builds.
```

The same warning is also repeatedly logged for @pos-dashboard/components.

### Additional context

The application only uses default imports at runtime:

```ts
  import host from '@pos-dashboard/host';
  import components from '@pos-dashboard/components';
```

All named imports are TypeScript type-only imports:

```ts
  import type { HostTaxCode } from '@pos-dashboard/host';
```

These type-only imports are removed from the production output, so the runtime named-export risk described by the warning does not apply in this case.

### Expected behavior

Please consider one or more of the following:

  1. Log the warning at most once per shared dependency during a single Vite invocation.
  2. If possible, only emit the warning when runtime named imports are actually used.
  3. Provide an option to suppress this warning for intentionally host-only shared modules.
  4. Avoid requiring host-only virtual modules to be installed as local devDependencies solely for named-export detection.

### Actual behavior

The same warning is printed multiple times during a single dev-server startup. This creates noisy output and can make it appear as though multiple distinct configuration problems exist.

### Minimal reproduction

```ts

  export default defineConfig({
    plugins: [
      federation({
        name: 'remote',
        shared: {
          '@example/host': {
            singleton: true,
            requiredVersion: '1.0.0',
            import: false,
          },
        },
      }),
    ],
  });
```

Do not install @example/host locally, then run: `vite dev`

Observe whether the same warning is logged multiple times during a single startup.

Image

### Version

@module-federation/vite: 1.20.7，Vite: 8.2.1

### Reproduction

None at the moment.

### Relevant log output

```shell

```

### Validations

gioboa:
Close #1118

 Deduplicates warnings and adds opt-in suppression for intentional host-only dependencies. Default remains false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,22 @@ export default defineConfig({
 In this remote app configuration, we define a remoteEntry.js file that will expose the App component.
 The shared property ensures that both host and remote applications use the same vue library.
 
+### Host-only shared dependencies
+
+Use `import: false` when a dependency must be provided by the host and should not be bundled as a local fallback. If it is intentionally not installed locally, the named-export detection warning can be suppressed:
+
+```ts
+shared: {
+  "@pos-dashboard/host": {
+    singleton: true,
+    import: false,
+    suppressMissingImportWarning: true,
+  },
+},
+```
+
+`suppressMissingImportWarning` defaults to `false`. Enable it only when the host is guaranteed to provide the dependency and local named-export detection is unnecessary.
+
 ## The Host Application configuration
 
 file **host/vite.config.ts**
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -142,6 +142,7 @@ export interface ShareItem {
   shareConfig: SharedConfig &
     moduleFederationPlugin.SharedConfig & {
       treeShaking?: TreeShakingConfig;
+      suppressMissingImportWarning?: boolean;
     };
 }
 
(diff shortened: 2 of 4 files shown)


gioboa (commented):
@imtaotao Could you check this fix please?
Is it solving the issue?

imtaotao:
> @imtaotao Could you check this fix please? Is it solving the issue?

Thanks for the quick fix! I tested the preview package, and it works as expected. Each missing shared dependency now emits the warning only once during dev-server startup.

imtaotao:
Perhaps we could add an option to suppress this warning, since the dependency is intentionally host-only and not installed locally by design—for example, `suppressMissingImportWarning`?

```ts
 shared: {
    '@pos-dashboard/host': {
      singleton: true,
      import: false,
      suppressMissingImportWarning: true,
    },
  }
```

gioboa:
> Perhaps we could add an option to suppress this warning, since the dependency is intentionally host-only and not installed locally by design—for example, `suppressMissingImportWarning`?

There we go, can you test if it's good for you?

imtaotao:
Perfect, thank you!

image

gioboa (commented):

Image

PR #1123, opened 2026-08-24
fix: honor hashed filenames across bundlers

### LINKED ISSUE #1122: [Bug]: [hash] is ignored in `filename`
### Describe the bug

When a remote is built through Vite's Environment API, using `[hash]` in the `filename` option is ignored.

With `filename: 'remote-entry-[hash].js'` and `assetsDir: 'static'`, I expect `dist/remote-entry-<hash>.js`. The build instead emits `dist/static/remoteEntry-<hash>.js`.

`mf-manifest.json` correctly references the emitted fallback filename, so the remote remains loadable. The configured filename does not control the file name or path.

The gap appears to be in `patchHashEntryFileName`: it wraps `output.entryFileNames`, while Rolldown applies `chunkFileNames` to the emitted remote entry in this environment. Applying the same handling to both output options should cover this case.

I can follow up with a small PR if there is agreement about the behavior.

### Version

Vite 8.0.10 with `@module-federation/vite` 1.20.7. Also reproduced against `main` at `b67d1fc` (package version 1.20.8).

### Reproduction

https://gist.github.com/gu-stav/4d82589e632cbc9906348d8978b7e101

### Relevant log output

```shell
Configured filename: remote-entry-[hash].js
Emitted file: dist/static/remoteEntry-DyqXMO2l.js
Manifest reference: static/remoteEntry-DyqXMO2l.js
```

### Validations

- [x] Read the [docs](https://github.com/module-federation/vite)
- [x] Read the [common issues list](https://github.com/module-federation/vite/issues)
- [x] Checked that there isn't already an issue reporting the same bug
- [x] The provided reproduction is a minimal reproducible example

gioboa:
Close #1122

Apply configured hash filenames to Rollup and Rolldown entries/chunks, including Environment API builds, while preserving asset directories.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -233,6 +233,53 @@ describe('pluginAddEntry', () => {
     expect(next).toHaveBeenCalledOnce();
   });
 
+  it('applies hash-pattern remote filenames to Rolldown chunks', () => {
+    const plugins = addEntry({
+      entryName: 'remoteEntry',
+      entryPath: 'virtual:mf-remote-entry',
+      fileName: 'remote-entry-[hash].js',
+    });
+    const config = {
+      environments: {
+        federation: {
+          build: {
+            assetsDir: 'static',
+            rolldownOptions: { input: './index.ts' },
+          },
+        },
+      },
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1124, opened 2026-08-24
fix: share application-local modules by adopting the resolved file as…

mshima:
… import source

A shared key that is not an installed package (an application module such as 'app/store' resolved by a user resolver plugin) has no package entry to inspect for named exports, so the generated loadShare module silently fell back to the local copy and every consumer got its own instance of a singleton.

Resolve such keys through Vite in the loadShare resolveId hook and adopt the resolved project file as shareConfig.import, which is what shared: { 'app/store': { import: './src/store' } } configures explicitly.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -6,6 +6,9 @@ import type {
   UserConfig,
 } from 'vite';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mkdtempSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import * as path from 'node:path';
 import { callHook } from '../../utils/__tests__/viteHookHelpers';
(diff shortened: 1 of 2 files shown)


gioboa (changes requested):
Thanks for your help @mshima 
I added tests to demonstrate concurrent resolutions can consume stale shared wrappers, while query-bearing resolver IDs lose semantics when adopted as imports.
We need to fix these regressions before merging.

gioboa (changes requested):
Should application-local shares require an explicit import instead?
Vite resolution is importer/environment-dependent, while shared configuration is global. Adopting the first observed resolution makes behavior order-dependent and risks mishandling queries, SSR/client variants, and concurrent resolutions.
A warning directing users to configure import: './src/store' seems safer and deterministic.

eg. Two modules concurrently import an aliased store; the second consumes stale wrapper code before asynchronous resolution updates its provider path.

gioboa (commented):
I'm closing this for now because it's stale, and there are a few open architectural questions to resolve (see comments) before creating a solution.
Thank you for your time and ideas @mshima

PR #1134, opened 2026-08-26
fix: handle shared named import prebundling

### LINKED ISSUE #1129: [Bug]: dev prebundle `mf-shared:` proxy re-exports through `export *` from an external module, so named imports between two shared packages fail with "No matching export"
## Describe the bug

When two packages are both declared in `shared` and one of them statically imports a **named** export from the other, the dev server dies during dependency optimization:

```
✘ [ERROR] No matching export in "mf-shared:pkg-a" for import "hello"

    node_modules/pkg-b/index.js:3:9:
      3 │ import {hello} from 'pkg-a';
        ╵         ~~~~~

Error: Error during dependency optimization:
    at failureErrorWithLog (node_modules/esbuild/lib/main.js:1748:15)
```

This is not the same as #1101. There, the nested shared entry was *dropped* from `shared` and `writeLoadShareModule()` got an `undefined` `shareItem`. Here the entry is deliberately **kept** (both packages are `singleton: true`, so `excludeSharedSubDependencies()` hits its guard and skips them), both end up in `optimizeDeps.include`, and the failure happens later, inside the esbuild prebundle.

`vite build` is unaffected — the `optimizeDeps.include` push is `command === 'serve'` only.

### Trigger conditions (each verified independently against the repro below)

| variant | result |
| --- | --- |
| `pkg-b` statically imports a named export from `pkg-a` | dev server starts fine |
| … **and** something in the prebundled graph also does `import('pkg-a')` | ✘ **fails** |
| `pkg-a` shared without `singleton: true` / `import: false` | fine — `excludeSharedSubDependencies()` drops it, so no proxy is generated |
| `vite build` | fine |
| vite `8.2.2` (rolldown optimizer) | fine — see below |

The dynamic `import()` does not have to be in your own code — a third-party shared package doing it is enough. In the app where I hit this, three packages were shared as singletons: a UI package that statically imports a named export from an analytics package, and an auth package that dynamically imports the same analytics package (`Promise.all([import('<analytics>')])`). Only the analytics package failed. The auth package, statically imported by the UI package with named imports too, did **not** fail — nothing dynamically imports it in that reduced config. With the app's full shared set it failed as well, and grepping turned up a fourth shared package doing `import('<auth>')`. That asymmetry is what isolated the condition above.

## Root cause

`module-federation:optimize-shared-proxy` (in `src/index.ts`, pushed onto `optimizeDeps.esbuildOptions.plugins`) routes every non-entry import of a shared package into the `mf-shared` namespace, and its `onLoad` returns:

```js
import * as __mfShared from "virtual:mf:__mfe_internal__host__mf_owner__1__loadShare__pkg_mf_2_a__loadShare__.js";
export * from "virtual:mf:__mfe_internal__host__mf_owner__1__loadShare__pkg_mf_2_a__loadShare__.js";
export default __mfShared.default ?? __mfShared;
```

The sibling `onResolve` in the same plugin marks every `virtual:mf:` id `external: true` (confirmed by instrumenting the plugin: `-> {"path":"virtual:mf:…loadShare__pkg_mf_2_a__loadShare__.js","external":true}`).

So the proxy's entire named-export surface is an `export *` from a module esbuild never sees. As long as esbuild can keep the star export dynamic it emits a runtime `__reExport` and everything works — which is why the plain static-import case passes. But once the proxy module has to be materialized across a chunk boundary, esbuild needs its export list statically, cannot expand the star export, and reports every named import of that shared package as missing.

Reduced to plain esbuild, no Vite and no MF — same 3-line proxy contents, same `external: true` on the virtual id:

```js
// entry.js
import {capture} from 'pkg-a';           // static named import
export const go = () => capture('x');
export const lazy = () => import('pkg-a'); // dynamic import of the same package
```

* `splitting: true` → `No matching export in "mf-shared:pkg-a" for import "capture"`
* `splitting: false` → builds fine

Vite's dep optimizer always runs with `splitting: true`, so in practice any dynamic `import()` of a shared package anywhere in the prebundled graph turns the static named imports of that package into hard errors.

### Why the rolldown path (vite 8) is not affected

The rolldown sibling plugin `module-federation:optimize-shared-resolver` returns `{id: loadSharePath, external: true}` directly from `resolveId` instead of generating an intermediate `export *` proxy. The prebundled output on vite `8.2.2` is:

```js
import { hello } from "virtual:mf:__mfe_internal__host__mf_owner__1__loadShare__pkg_mf_2_a__loadShare__.js";
…
return import("virtual:mf:__mfe_internal__host__mf_owner__1__loadShare__pkg_mf_2_a__loadShare__.js").then((m) => m.hello(name));
```

No proxy module, no star export, no error.

## Suggested fix (either)

1. **Emit explicit named re-exports in the `mf-shared` proxy** instead of `export *`. The plugin already has the machinery: `writeLoadShareModule()` uses `getPackageNamedExports()` + `generateDeferredHostProvidedExports()` for the `shareConfig.import === false` path, and the generated Vite-side loadShare virtual module already emits a full `export { __mf_0 as ASPixel, __mf_2 as capture, … }` list. Reusing that list here would make the proxy's exports statically known:
   ```js
   import * as __mfShared from "<virtual:mf:…>";
   export { a, b, c } from "<virtual:mf:…>";
   export default __mfShared.default ?? __mfShared;
   ```
2. **Or mirror the rolldown branch** in the esbuild branch: resolve the import straight to the external loadShare id and skip the `mf-shared` namespace entirely.

A cheaper mitigation, if neither is desirable: when a shared package is statically imported by another package that is itself in `optimizeDeps.include`, keep the *importing* package out of the prebundle (see workaround).

## Workaround

```js
// vite.config.js — after federation()
{
  name: 'mf-skip-prebundle-nested-shared',
  config(config) {
    const nested = ['pkg-b']; // the packages that import other shared packages
    const od = (config.optimizeDeps ??= {});
    od.include = (od.include ?? []).filter((d) => !nested.includes(d));
    od.exclude = [...(od.exclude ?? []), ...nested];
  },
}
```

Two notes on this:

* Adding to `exclude` alone is **not** enough. `normalizeOptimizeDeps` (`src/utils/normalizeOptimizeDeps.ts`) has a `configResolved` hook that filters out any `exclude` entry which is also present in `include`, and the early-init plugin has already pushed the package into `include`. The package has to be removed from `include` in the same pass.
* The plugin must not be `enforce: 'pre'`, so that its `config` hook runs after the federation plugin's.

Sharing is preserved with this workaround: the excluded package goes through Vite's normal ESM pipeline, where its imports of the other shared packages are still rewritten to the `virtual:mf:…loadShare…` modules, so the singletons stay single.

## Version

```
vite: 7.3.0 (also reproduced on 7.3.6)
@module-federation/vite: 1.20.8
esbuild: 0.27.2 / 0.27.7
node: 24.4.1
OS: macOS (Darwin 25.6.0)
```

Not reproducible on vite `8.2.2` (rolldown optimizer).

## Reproduction

https://github.com/h11g/mf-vite-shared-dynamic-import-repro

```bash
npm install --install-links
npm run dev     # dies with: No matching export in "mf-shared:pkg-a" for import "hello"
npm run build   # succeeds
```

Ten files, no private packages. The two that matter:

```js
// vite.config.js
federation({
  name: 'host',
  filename: 'remoteEntry.js',
  dts: false,
  exposes: {},
  remotes: {},
  shared: {
    // singleton: true is what makes excludeSharedSubDependencies()
    // keep pkg-a in `shared` for dev instead of dropping it.
    'pkg-a': {singleton: true, requiredVersion: '^1.0.0'},
    'pkg-b': {singleton: true, requiredVersion: '^1.0.0'},
  },
})
```

```js
// packages/pkg-b/index.js — pkg-a is imported both ways; both are needed
import {hello} from 'pkg-a';

export function greet(name) {
  return hello(name).toUpperCase();
}

export function greetLazy(name) {
  return import('pkg-a').then((m) => m.hello(name));
}
```

`--install-links` matters: with a plain symlinked `file:` dependency Vite treats the package as linked and never prebundles it, so the bug does not show. (For the same reason, note that npm reuses the copied package when only its file contents change — bump the version or copy the file into `node_modules/` by hand when editing `pkg-b` to test variants.)

Deleting `greetLazy` from `packages/pkg-b/index.js` — leaving only the static named import — makes the dev server start cleanly, which isolates the dynamic `import()` as the second half of the trigger.

gioboa:
Close #1129

Resolve ESM shared imports directly during esbuild optimization, preserving named exports with dynamic imports.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -1669,6 +1669,18 @@ describe('vite:module-federation-early-init', () => {
     });
     expect(result.contents).toContain(JSON.stringify(loadSharePath));
     expect(result.contents).not.toContain(JSON.stringify(encodedLoadSharePath));
+    for (const kind of ['import-statement', 'dynamic-import']) {
+      expect(
+        onResolveHandlers[1]({
+          path: 'pkg-foo/a',
+          importer: '/repo/node_modules/pkg-b/index.js',
+          kind,
+        })
+      ).toEqual({
+        path: expect.stringContaining(LOAD_SHARE_TAG),
+        external: true,
+      });
+    }
   });
 
   it('does not proxy shared deps when esbuild resolves optimizeDeps entry points', () => {
@@ -1872,7 +1884,10 @@ describe('vite:module-federation-early-init', () => {
         importer: '/repo/node_modules/.vite/deps/pkg.js',
         kind: 'import-statement',
       })
-    ).toEqual({ path: '@ui-lib/button', namespace: 'mf-shared' });
+    ).toEqual({
+      path: expect.stringContaining(LOAD_SHARE_TAG),
+      external: true,
+    });
   });
 
   it('redirects System.register commonjs-proxy consumers to loadShare chunks', () => {
--- a/src/index.ts
+++ b/src/index.ts
@@ -726,6 +726,15 @@ function createEarlyVirtualModulesPlugin(options: NormalizedModuleFederationOpti
                   )
                     return;
                   addUsedShares(args.path, options);
+                  if (args.kind === 'import-statement' || args.kind === 'dynamic-import') {
+                    const shareItem = shared[key];
+                    const loadSharePath = getLoadShareModulePath(args.path, isRolldown, options);
+                    writeLoadShareModule(args.path, shareItem, _command, isRolldown, options);
+                    if (shareItem.shareConfig?.import !== false) {
+                      writePreBuildLibPath(args.path, shareItem, options);
+                    }
+                    return { path: loadSharePath, external: true };
+                  }
                   return { path: args.path, namespace: 'mf-shared' };
                 });
                 build.onLoad({ filter: /.*/, namespace: 'mf-shared' }, (args: any) => {


gioboa (commented):

Image

PR #1136, opened 2026-08-26
feat: preserve user chunking config alongside federation chunks

joshuaellis:
The plugin currently discards any user-provided chunking config: `output.codeSplitting.groups` (Rolldown) is deleted and `output.manualChunks` (Rollup) is replaced. This makes it impossible for a consumer to emit a module as its own chunk with a stable file name in a federation build — for example to rewrite one small file at deploy time.

The original reason for discarding user config still holds: federation's runtime-init and load-share wrappers must stay isolated in their own chunks, or a remote can deadlock on startup. This change keeps that guarantee while letting user config through:

- Rolldown: user `codeSplitting.groups` are kept, but federation groups are installed at a higher priority and user priorities are clamped below it. Rolldown assigns each module to the highest-priority matching group, so a user group can never capture a federation module — it only gets what federation leaves behind.
- Rollup: a user `manualChunks` function is composed in as a fallback behind the plugin's own, which claims federation modules first. The object form is still rejected, with a warning pointing at the function form.

Warnings now only fire when the plugin actually mutates user config (clamping a priority, dropping object-form `manualChunks`, removing `manualChunks` on a Rolldown output).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -127,6 +127,13 @@ type CodeSplittingGroup = {
   priority?: number;
 };
 
+// Federation groups must always outrank user-provided codeSplitting groups so a
+// user group can never capture a runtimeInit/loadShare wrapper or the preload
+// helper (Rolldown assigns each module to the highest-priority matching group).
+// User group priorities are clamped below this value.
+const MF_GROUP_PRIORITY = 1_000_000;
+const USER_GROUP_MAX_PRIORITY = MF_GROUP_PRIORITY - 1;
+
 type ViteWatchOptions = NonNullable<NonNullable<UserConfig['server']>['watch']>;
 type ViteWatchConfig = ViteWatchOptions | boolean | null | undefined;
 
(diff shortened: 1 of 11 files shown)


dmchoi77:
Looks good! Could we update README.md in this PR as well?

The current documentation says that codeSplitting.groups and manualChunks are always ignored. With this change, Rolldown user groups are preserved below federation groups, and the function form of Rollup manualChunks is composed as a fallback.

Updating the docs would prevent users from following outdated guidance and clarify the differences between Vite 5–7 and Vite 8.

src/index.ts
```diff
@@ -1414,49 +1421,46 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
...
-          ) {
-            return;
-          }
+        // Groups installed by applyManualChunks: the dynamic name() group (tracked
+        // in patchedManualChunks) and the preload-helper test group.
+        const isFederationGroup = (group: unknown): boolean => {
+          const candidate = group as Partial<CodeSplittingGroup> | undefined;
+          return (
+            (typeof candidate?.name === 'function' && patchedManualChunks.has(candidate.name)) ||
+            candidate?.name === PRELOAD_HELPER_CHUNK
```
gioboa: Could we identify plugin-owned groups by object identity instead of name? A user group named vite-preload-helper is currently filtered out silently. Please add a regression test.

Please align the readme file with the new behaviour.

gioboa (changes requested):

src/index.ts
joshuaellis: sounds good. change pushed.

gioboa (commented):
Could we cover this in vite-vite by keeping the host on Vite 8 and downgrading the remote to Vite 7?
Then test codeSplitting.groups on the host and manualChunks on the remote, asserting both custom chunks are emitted and loaded.
Since this suite uses production preview, it validates real bundler behavior and federation startup.

gioboa (approved):
I updated the readme, just a couple of words.
Thanks @joshuaellis this is a nice improvement 👏

joshuaellis:
Thanks @gioboa for the quick merge.

PR #1138, opened 2026-08-26
chore: update pnpm action for Node 24

dmchoi77:
## Summary

Update `pnpm/action-setup` references in all workflows from the Node.js 20 action commit to the official `v4.4.0` commit that targets Node.js 24.

## Why

The current action declares `using: node20`, but GitHub-hosted runners already execute it with Node.js 24. 
GitHub will remove Node.js 20 from Actions runners on September 23, 2026. After that, the compatibility path for actions targeting Node.js 20 will no longer be available, and the pnpm setup step may fail before install, test, or publish steps run.

See GitHub's [Node.js 20 deprecation notice](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

## Change

- Replace `pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1` with `pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320`.
- The new SHA is the direct upstream successor of the old SHA and the `v4.4.0` release commit. It changes the action runtime from Node.js 20 to Node.js 24 without changing the workflow configuration.

Official references: [commit](https://github.com/pnpm/action-setup/commit/fc06bc1257f339d1d5d8b3a19a8cae5388b55320) · [v4.4.0 release](https://github.com/pnpm/action-setup/releases/tag/v4.4.0)

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
 
       - name: PNPM Install
-        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
+        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v4.4.0
         with:
           version: 10.28.2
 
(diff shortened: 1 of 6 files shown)


gioboa (approved):
Thanks @dmchoi77 🍒

PR #1143, opened 2026-08-27
fix: preserve named exports in CommonJS

### LINKED ISSUE #1142: [Bug]: sharing a CommonJS package whose `default` export is a function drops all its named exports (rollup build)
### Describe the bug

Sharing a CommonJS package whose `default` export is a **function** (alongside named
exports) makes every CommonJS consumer of that package lose all of its named exports in
a rollup production build.

Any consumer compiled by `tsc` with `esModuleInterop` — i.e. `import * as ns from 'pkg'`
becoming `__importStar(require('pkg'))` — receives the `default` **function** instead of
the module object, so `ns.<namedExport>` is `undefined`:

```
TypeError: Cannot read properties of undefined (reading 'A')
```

Not shared → works. Dev → works. Vite 8 / rolldown → works. Only the Vite 7 (rollup)
production build is affected.

This is **not** the "shared module used before bootstrap finished" timing problem: the
repro waits for the share cache to be populated and prints it, so you can see the cache
holds the full module object (`{ SOME_ENUM, createThing, default }`) while `require()`
still receives only the function.

Named-export *detection* works correctly here (the generated code does read
`mod["SOME_ENUM"]`); the problem is purely which value is handed to `require()`.

### Version

`@module-federation/vite` 1.21.0, vite 7.3.6 (also present in `main` / 1.20.7), node 24.4.1

### Reproduction

https://github.com/h11g/mf-vite-cjs-fn-default-repro

```bash
git clone https://github.com/h11g/mf-vite-cjs-fn-default-repro
cd mf-vite-cjs-fn-default-repro
npm install && npm run build && npm run preview   # http://localhost:4180
```

5 files, ~60 significant lines, no third-party packages involved. Inlined below as well.

<details>
<summary><code>packages/cjs-fn-default/index.js</code> — the shared CommonJS package</summary>

The only thing that matters is the shape: `__esModule: true`, named exports, and a
`default` that is a function.

```js
'use strict';
Object.defineProperty(exports, '__esModule', { value: true });

function createThing(config) {
  return { config: config };
}
exports.createThing = createThing;

var SOME_ENUM;
(function (SOME_ENUM) {
  SOME_ENUM.A = 'A';
  SOME_ENUM.B = 'B';
})((SOME_ENUM = exports.SOME_ENUM || (exports.SOME_ENUM = {})));

// `default` is a FUNCTION, not an object. This is the trigger.
Object.defineProperty(exports, 'default', {
  enumerable: true,
  get: function () {
    return createThing;
  },
});
```

`package.json`: `{ "name": "cjs-fn-default", "version": "1.0.0", "main": "index.js" }`

</details>

<details>
<summary><code>packages/cjs-fn-default-consumer/index.js</code> — a CommonJS consumer</summary>

Verbatim `tsc` output for `import * as lib from 'cjs-fn-default'` with
`module: commonjs` + `esModuleInterop: true`. The `__importStar` / `__createBinding` /
`__setModuleDefault` helpers are tsc's, unmodified.

```js
'use strict';
var __createBinding =
  (this && this.__createBinding) ||
  (Object.create
    ? function (o, m, k, k2) {
        if (k2 === undefined) k2 = k;
        var desc = Object.getOwnPropertyDescriptor(m, k);
        if (!desc || ('get' in desc ? !m.__esModule : desc.writable || desc.configurable)) {
          desc = { enumerable: true, get: function () { return m[k]; } };
        }
        Object.defineProperty(o, k2, desc);
      }
    : function (o, m, k, k2) {
        if (k2 === undefined) k2 = k;
        o[k2] = m[k];
      });
var __setModuleDefault =
  (this && this.__setModuleDefault) ||
  (Object.create
    ? function (o, v) { Object.defineProperty(o, 'default', { enumerable: true, value: v }); }
    : function (o, v) { o['default'] = v; });
var __importStar =
  (this && this.__importStar) ||
  function (mod) {
    if (mod && mod.__esModule) return mod;
    var result = {};
    if (mod != null)
      for (var k in mod)
        if (k !== 'default' && Object.prototype.hasOwnProperty.call(mod, k))
          __createBinding(result, mod, k);
    __setModuleDefault(result, mod);
    return result;
  };
Object.defineProperty(exports, '__esModule', { value: true });
exports.probe = void 0;

var lib = __importStar(require('cjs-fn-default'));

function probe() {
  return {
    defaultIsFunction: typeof lib.default === 'function',
    enumA: lib.SOME_ENUM.A, // <- throws
  };
}
exports.probe = probe;
```

`package.json`: `{ "name": "cjs-fn-default-consumer", "version": "1.0.0", "main": "index.js", "dependencies": { "cjs-fn-default": "file:../cjs-fn-default" } }`

</details>

<details>
<summary><code>vite.config.ts</code> — one <code>shared</code> entry is the whole repro</summary>

```ts

export default defineConfig({
  plugins: [
    federation({
      name: 'app',
      filename: 'remoteEntry.js',
      exposes: { '.': './src/expose.ts' },
      remotes: {},
      shared: {
        // Comment this out and the page renders.
        'cjs-fn-default': { singleton: false },
      },
    }),
  ],
  build: { target: 'chrome89', minify: false },
});
```

`src/expose.ts` is just `export const name = 'app';` so the build is a real container.

</details>

<details>
<summary><code>src/main.ts</code> — waits for the share cache, then prints both facts</summary>

```ts
async function main() {
  const out = document.getElementById('out')!;

  // Wait until the federation runtime has finished seeding the share cache.
  // Without this the repro would be ambiguous: a CommonJS module that evaluates
  // before bootstrap also sees `undefined`, which throws the same TypeError.
  await new Promise((resolve) => setTimeout(resolve, 500));

  const cache = (globalThis as Record<string, any>).__mf_module_cache__;
  const share = cache?.share ?? {};
  const key = Object.keys(share).find((k) => k.startsWith('default:cjs-fn-default'));
  const entry = key ? share[key] : undefined;

  const lines = [
    `share cache key    : ${key ?? '<missing>'}`,
    `share cache value  : ${typeof entry}` +
      (entry && typeof entry === 'object' ? ` { ${Object.keys(entry).join(', ')} }` : ''),
    `  .default         : ${entry ? typeof entry.default : '-'}`,
    `  .SOME_ENUM       : ${entry ? typeof entry.SOME_ENUM : '-'}`,
  ];

  // Imported dynamically so the CommonJS consumer evaluates *after* the wait.
  const { probe } = await import('cjs-fn-default-consumer');

  try {
    lines.push(`probe()            : OK ${JSON.stringify(probe())}`);
  } catch (error) {
    lines.push(`probe()            : THREW ${error}`);
  }

  out.textContent = lines.join('\n');
  console.log(lines.join('\n'));
}

void main();
```

`index.html` only needs `<pre id="out">loading…</pre>` and
`<script type="module" src="/src/main.ts"></script>`.

</details>

Steps: `npm install && npm run build && npm run preview`, then open the page.

One setup note: the two local packages are installed with `install-links=true` in
`.npmrc` so they land as real copies under `node_modules` instead of symlinks — Vite's
default `build.commonjsOptions.include` is `[/node_modules/]`, and with a symlink the
realpath falls outside it and the CommonJS transform never runs. In a real project these
are ordinary registry dependencies and no setting is needed.

### Relevant log output

```shell
# actual
share cache key    : default:cjs-fn-default@1.0.0
share cache value  : object { SOME_ENUM, createThing, default }
  .default         : function
  .SOME_ENUM       : object
probe()            : THREW TypeError: Cannot read properties of undefined (reading 'A')

# expected (what every other configuration prints)
probe()            : OK {"defaultIsFunction":true,"enumA":"A"}
```

### Result matrix

Every row verified with this repro; only `vite.config.ts` / the `vite` version changed.

| Vite | Mode | `shared` entry | Result |
|---|---|---|---|
| 7.3.6 (rollup) | build | `'cjs-fn-default': { singleton: false }` | ❌ `TypeError` |
| 7.3.6 (rollup) | build | `'cjs-fn-default': { singleton: true }` | ❌ `TypeError` |
| 7.3.6 (rollup) | build | not shared | ✅ `OK {"defaultIsFunction":true,"enumA":"A"}` |
| 7.3.6 (rollup) | dev | `'cjs-fn-default': { singleton: false }` | ✅ |
| 8.2.2 (rolldown) | build | `'cjs-fn-default': { singleton: false }` | ✅ |

Vite 8 / rolldown is fine because the consumer gets
`__toCommonJS(<loadShare namespace>)`, which keeps the named exports.

### Root cause

In the rollup build, the CommonJS consumer's `require()` is routed through a
`?commonjs-proxy` copy of the generated `loadShare` module, and that copy exports only
the **ESM default** of the shared module:

```js
// dist/assets/__virtual_mf___..._loadShare__cjs_mf_2_fn_mf_2_default__loadShare__.js_commonjs-proxy-<hash>.js
let __mf_default;
const __mfApplyLazyShareExports = (mod) => {
  mod["createThing"];
  mod["SOME_ENUM"];
  __mf_default = mod.default ?? mod;   // <- mod.default is `createThing`, a function
};
export { __mf_default as _ };
```

The consumer chunk binds `require()` to that default:

```js
// dist/assets/index-<hash>.js
import { _ as __mf_default } from "./__virtual_mf___..._commonjs-proxy-<hash>.js";
...
var lib = __importStar(__mf_default);   // __mf_default === createThing
```

`__importStar` passes a value through untouched only when it has `__esModule: true`. A
function does not, so tsc's helper builds `{ default: createThing }` — a function has no
enumerable own properties to copy — and every named export is dropped.

`__mf_default = mod.default ?? mod` is the right value for the module's *ESM default
export*, but the wrong value to hand a `require()`. CommonJS interop needs the module
object itself (`__esModule: true` + `default` + the named exports), which is exactly what
the share cache already holds.

With `singleton: true` the generated code differs but has the same defect — an unwrap
loop that returns `default` as soon as it is not an object:

```js
const defaultExport = current == null ? void 0 : current.default;
if (!defaultExport || typeof defaultExport !== "object") return defaultExport ?? current;
current = defaultExport;
```

For a CommonJS module whose `default` is a function this always returns the function, so
the loop can never yield a value carrying both `default` and the named exports.

Both spots live in `src/virtualModules/virtualShared_preBuild.ts`:
`generateShareModuleUnwrapCode()` called with
`stopWithReturn: 'defaultExport ?? current'`, and the
`__mf_default = mod.default ?? mod` assignments in the lazy-share export generators.

### Suggested direction

Leave the ESM `default` export as it is, but give the `?commonjs-proxy` variant the
module object rather than the unwrapped default, preserving `__esModule` so
`__importStar` passes it through.

The shape is common in packages compiled by `tsc` with `esModuleInterop` that export a
factory as `default` plus enums/constants as named exports — `@shopify/app-bridge` is one
published example (`default` is `createApp`, with `LifecycleHook`, `MessageType`,
`Context`, … beside it), and sharing it breaks `@shopify/app-bridge-react`'s `Provider`
on `AppBridge.LifecycleHook.DispatchAction`.

Happy to send a PR if you agree with the direction.

### Validations

- [x] Checked that there isn't already an issue reporting the same bug (searched
  `commonjs-proxy`, `named exports`, `importStar`, `esModuleInterop`; #986 / #1084 /
  #1129 are about named-export *detection*, which works fine here).

gioboa:
Closes #1142

Ensure Rollup CommonJS proxies return the complete shared module when its default export is a function.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -805,6 +805,14 @@ describe('module-federation-esm-shims', () => {
     expect(result).toBeNull();
   });
 
+  it('returns the full shared module to CommonJS consumers', () => {
+    const plugin = getEsmShimsPlugin();
+    const target = `\0virtual:mf:host${LOAD_SHARE_TAG}cjs-package${LOAD_SHARE_TAG}.js`;
+    const result = callHook(plugin.load, {} as any, `\0${target}?commonjs-proxy`);
+
+    expect(result).toBe(`export { __moduleExports as default } from ${JSON.stringify(target)};`);
+  });
+
   it('filters federation control chunks from js dynamic modulepreload deps', () => {
     const plugin = getEsmShimsPlugin();
     const config: any = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1642,6 +1642,12 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
         }
       },
       load(id: string, loadOptions?: LoadHookOptions) {
+        const commonJsProxySuffix = '?commonjs-proxy';
+        if (id.includes(LOAD_SHARE_TAG) && id.endsWith(commonJsProxySuffix)) {
+          const target = id.slice(id.startsWith('\0') ? 1 : 0, -commonJsProxySuffix.length);
+          return `export { __moduleExports as default } from ${JSON.stringify(target)};`;
+        }
+
         const loadVirtualModule = (
           importFalseExportUsage?: ReturnType<typeof getSharedExportUsage>
         ) => {


gioboa (commented):

Image

PR #1147, opened 2026-08-28
fix: nested container loses host scope

### LINKED ISSUE #1146: [Bug]: container with both exposes and remotes loses the host's share scope in build — generated hostInit calls init() with no arguments
### Describe the bug

A container that declares **both `exposes` and `remotes`** — an app that is a remote of an outer host *and* a host of its own sub-remotes — silently stops sharing singletons with the outer host after `vite build`. It ends up with its own share scope and loads a second copy of every shared dependency.

In our production app that second copy is React, and the embedded remote dies with `Cannot read properties of null (reading 'useMemo')`, because the second React's `ReactCurrentDispatcher.current` is `null` while the outer host's React is the one rendering.

`dev` is not affected.

#### Cause

In build mode every generated `loadRemote` module statically imports a self-executing `hostInit` chunk — `generateRemotes`, `lib/index.js:5321` in the published package:

```js
const importLine = command === "build"
  ? `... import { hostInitPromise as __mfHostInitPromise } from "<hostAutoInit>";`
  : `<devRuntimeBootstrap>`;                     // dev: globalThis[key].initPromise handshake
const remoteLoadRuntimePromise = command === "build" ? "__mfHostInitPromise" : "initPromise";
```

`hostInit` has a top-level side effect and calls the container's own entry with no arguments — `generateHostAutoInitCode`, `lib/index.js:4974`:

```js
const remoteEntry = await import(<remoteEntry>);
const runtime = await remoteEntry.init();     // no shareScope, no initScope
```

So the generated `init(shareScope = {}, initScope = [])` runs a **second** time with an empty scope and calls `initShareScopeMap("default", {})`, replacing the scope the outer host handed over moments earlier via `loadRemote` → `initContainer`. The container then resolves its shared deps from its own scope.

Instrumenting the exported `init` on our production build shows exactly two calls:

```
#1 args=3 scope=<host scope, 29 pkgs>   stack: host loadRemote -> Module.get -> Module.init
#2 args=0 scope=undefined               stack: <container>/assets/hostInit-*.js
```

The container's own re-entrancy guard cannot catch call #2, because it keys off `initScope` and `hostInit` passes none, so `a` falls back to a fresh `[]`:

```js
var G = je[Pe]; if (G || (G = je[Pe] = {from: k}), a.indexOf(G) >= 0) return; a.push(G)
```

#### Why it is easy to miss

If the outer host is *also* a `@module-federation/vite` build, the bug is masked: the plugin's private `globalThis.__mf_module_cache__` hands the already-loaded singleton back to the nested container even though its share scope was replaced. I hit this while writing the reproduction — my first version had a vite-plugin host and passed. Hosts that are not vite-plugin builds (webpack, rspack, or a hand-written `init()`) have no such cache and get duplicates, so the reproduction's host is deliberately a plain `@module-federation/runtime` host.

#### No opt-out today

- `hostInitInjectLocation` only chooses where the standalone bootstrap is injected (`html` vs `entry`); the `import` inside the generated `loadRemote` module is unconditional in build mode.
- `isRemoteOnlyContainer()` requires `remotes` to be empty, so a container that is both a host and a remote can never take that path.

#### Not a duplicate of #895 / #1049

- **#895** (1.16.14) was the `__loadShare__<pkg>` virtual module evaluating before the container's `init()` bridge filled `__mf_module_cache__`, plus `pendingShareLoads` not being awaited by a pure runtime host. That path is healthy in 1.20.9 — see the `leaf` container in the reproduction: same plugin, same version, same page, pure runtime host, and it binds the host's singleton correctly. It differs from `mid` only in having no `remotes` of its own.
- **#1049** is SSR-specific. This one is plain client-side `vite build`.

#### Possible fixes

1. Use the `initPromise` handshake in build mode too, as dev already does, instead of hard-binding `__mfHostInitPromise`.
2. Have `hostInit` pass an `initScope`, so the entry's existing re-entrancy guard fires on the second call.
3. Make the generated `init` ignore a scope-less call once the container has already been initialized.

I am happy to send a PR for whichever direction you prefer.

### Version

vite 7.3.6 / `@module-federation/vite` 1.20.9 (bundles `@module-federation/runtime` 2.9.0) / node 25.9.0

### Reproduction

https://github.com/h11g/mf-vite-hostinit-repro

```bash
npm install
npm run build      # leaf + mid + host
npm run serve      # http://localhost:5599
open http://localhost:5599/host/index.html
```

| app | role | `exposes` | `remotes` |
| --- | --- | --- | --- |
| `apps/host` | outer host, plain `@module-federation/runtime`, no vite plugin | – | `mid` |
| `apps/mid` | remote **and** host — the failing combination | `./probe` | `leaf` |
| `apps/leaf` | plain remote — built-in control | `./x` | – |

`packages/shared-lib` is a `singleton: true` package whose module body generates a random id per evaluation, so two ids on the page means two copies.

**Expected:** `mid` resolves `shared-lib` from the host, exactly like `leaf` does.

Two extra builds confirm the cause:

```bash
npm run build:mid:no-remotes && npm run build:host   # drop mid's sub-remote
npm run build:mid:guard      && npm run build:host   # keep it, guard the init
```

| build of `mid` | `hostInit` in the `remoteEntry` graph | `mid` scope === host scope | `shared-lib` copies |
| --- | --- | --- | --- |
| default | yes | **false** | **2** |
| `NO_REMOTES=1` | no (only in `mid`'s own `index.html`) | true | 1 |
| `MF_GUARD=1` | yes | true | 1 |

`MF_GUARD=1` enables a ~10-line userland `transform` that wraps the generated entry's `init` so a scope-less call cannot clobber an initialization that already happened. It string-matches generated virtual-module output, so it is a stopgap, not a proposed fix.

### Relevant log output

```shell
# http://localhost:5599/host/index.html — default build
host shared-lib : shared-lib#32772o
mid  shared-lib : shared-lib#a0uj9g
FAIL duplicate instance
instances       : host, mid, leaf
scope host : shared-lib@1.0.0 from=host
scope mid  : shared-lib@1.0.0 from=mid      <-- its own copy
scope leaf : shared-lib@1.0.0 from=host     <-- control, adopted the host's
mid scope === host scope : false

# the generated loadRemote chunk, dist/mid/assets/_virtual_mf___..._loadRemote__leaf_...js
import{hostInitPromise as d}from"./hostInit-BcCFQiQc.js";

# dist/mid/assets/hostInit-*.js
(await import("../../remoteEntry.js")).init()
```

### Validations

gioboa:
Closes #1146

A dual-role container’s automatic initialization replaces its outer host’s shared dependency scope.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1461,6 +1461,48 @@ describe('virtualRemoteEntry', () => {
     expect(code).not.toContain('runtime.loadRemote("remote/remote-app")');
   });
 
+  it.each(['version-first', 'loaded-first'] as const)(
+    'keeps a scoped %s build init when host auto init re-enters a dual-role container',
+    async (shareStrategy) => {
+      const mod = await import('../virtualRemoteEntry');
+      const code = mod.generateRemoteEntry(
+        {
+          internalName: '__mfe_internal__mid',
+          name: 'mid',
+          filename: 'remoteEntry.js',
+          exposes: { './probe': './probe.js' },
+          remotes: {
+            leaf: {
+              name: 'leaf',
+              entry: 'http://localhost:4173/remoteEntry.js',
+              type: 'module',
+            },
+          },
+          shared: {},
+          runtimePlugins: [],
+          shareScope: 'default',
+          shareStrategy,
+        } as any,
+        'virtual:exposes',
+        'build'
+      );
+
+      expect(code).toContain('if (shared === undefined && __mfInitPromise)');
+      expect(code).toContain('export { __mfGuardedInit as init, getExposes as get }');
+      const guardStart = code.indexOf('let __mfInitPromise;');
+      const guardEnd = code.indexOf('export { __mfGuardedInit', guardStart);
+      const guardCode = code.slice(guardStart, guardEnd);
+      const scopedInit = Promise.resolve({ shareScopeMap: { default: {} } });
+      const init = vi.fn(() => scopedInit);
+      const guardedInit = new Function('init', `${guardCode}; return __mfGuardedInit;`)(init);
+
+      expect(guardedInit({ react: {} }, [])).toBe(scopedInit);
+      expect(guardedInit()).toBe(scopedInit);
+      expect(init).toHaveBeenCalledOnce();
+      expect(guardCode).not.toContain('await ');
+    }
+  );
+
   it('does not preload generated subpath shares from a root shared package', async () => {
     const mod = await import('../virtualRemoteEntry');
 
--- a/src/virtualModules/virtualRemoteEntry.ts
+++ b/src/virtualModules/virtualRemoteEntry.ts
@@ -1042,6 +1042,10 @@ export function generateRemoteEntry(
     (share) => !!share?.shareConfig.treeShaking
   );
   const hasMultipleShareScopes = Array.isArray(options.shareScope);
+  const guardHostAutoInit =
+    command === 'build' &&
+    Object.keys(options.exposes ?? {}).length > 0 &&
+    Object.keys(options.remotes ?? {}).length > 0;
   const materializedShareBatches = toSafeJsLiteral(getShareBatches(options, false));
   const runtimeImports = [
     'init as runtimeInit',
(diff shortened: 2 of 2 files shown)


gioboa (commented):

Screenshot 2026-08-28 at 10 37 40

PR #1148, opened 2026-08-28
fix: preserve react/ as consumer-only namespace prefix

tonoizer:
## Summary

The Vite plugin already has generic prefix matching, but it special-cases React: `react/` is normalized to `react`, and known subpaths (`jsx-runtime`, etc.) are only auto-mapped when a local provider exists. With `import: false`, that mapping is skipped, so the generic prefix matcher never receives `react/`.

This PR preserves `react/` as a namespace prefix so consumer-only shares cover actually-imported subpaths (namespace coverage, not a hardcoded React export list).

`react-dom/` still collapses. A browser-wide prefix would also capture `react-dom/server.*`, so `react-dom/client` stays an explicit browser-safe entry unless the plugin gains environment-aware filtering.

## Test plan

- [ ] `normalizeSharedKey('react/')` stays `react/` (does not collapse to `react`)
- [ ] `normalizeSharedKey('react-dom/')` still collapses so `react-dom/server*` is not captured by a browser prefix
- [ ] Consumer-only `import: false` + `react/` materializes `react/jsx-runtime` (and siblings) when actually imported
- [ ] Existing local-provider React auto-mapping tests still pass

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -364,6 +364,9 @@ describe('pluginProxySharedModule_preBuild', () => {
       const shared = makeShared();
       shared.react.shareConfig.import = false;
 
+      // Bare `react` + import:false shares only the package root. Known JSX
+      // subpaths are not auto-mapped (that would invent unfulfillable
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Screenshot 2026-08-28 at 10 51 29

gioboa (approved):
Thanks @tonoizer nice catch

PR #1151, opened 2026-08-28
fix: honor vite-ignore in HTML bootstrap

### LINKED ISSUE #1149: [Bug]: Absolute path of external script converted to relative path and vite-ignore has no effect
### Describe the bug

In our application we reference an external JS file from `index.html`:

```html
<script type="module" src="/external/external.js" vite-ignore></script>
```

The external JS is served from our own domain, which is why we use an absolute (root-relative) path. After upgrading `@module-federation/vite` from **1.15.5** to the current **1.21.1**, the file no longer loads: the plugin turns the absolute path into a relative one.

Also the `vite-ignore` attribute, which Vite documents as the opt-out for HTML processing, has no effect.

#### When the regression was introduced

The absolute path to relative path conversion was introduced with 1.16.5.

#### Proposed fix

Either of these would resolve it:

a) Honour `vite-ignore`. Skip elements carrying the attribute. This matches [[Vite's documented behaviour](https://vite.dev/guide/features#html)](https://vite.dev/guide/features#html) and gives users an explicit escape hatch.

b) Do not rebase root-relative paths or leave paths alone that do not correspond to a file in the bundle. 

### Version

Vite v8.2.2. and module-federation/vite v1.21.0

### Reproduction

Add <script type="module" src="/external/external.js" vite-ignore></script> into index.html. Run the build and have a look at the generated mf-generated-bootstrap-xxx.js. I can provide a full repro if needed. 

### Relevant log output

```shell
Output in the mf-entry-bootstrap-xxx.js:

(async () => {
  const { initHost } = await __mfImport("./hostInit-DI1mVXas.js");
  await initHost();
})().then(() => __mfImport("../external/external.js"));
```

### Validations

gioboa:
Closes #1149

Preserves ignored module scripts during HTML rewriting, preventing root-relative external paths from being rebased.
vite-ignore is an official Vite HTML attribute. It tells Vite to skip processing that element, especially useful for external/CDN assets. Vite documentation

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -1958,6 +1958,61 @@ describe('pluginAddEntry', () => {
     expect(bootstrapFile!.source as string).toContain('__mfImport("../../src/main.tsx")');
   });
 
+  it('leaves vite-ignore scripts unchanged', () => {
+    const plugins = addEntry({
+      entryName: 'hostInit',
+      entryPath: '/virtual/hostInit.js',
+      inject: 'html',
+    });
+    const buildPlugin = plugins[1];
+    const emitted: Rollup.EmittedFile[] = [];
+    const bundle: any = {
+      'index.html': {
+        type: 'asset',
+        source:
+          '<script type="module" src="/src/main.tsx"></script><script type="module" src="/external/external.js" vite-ignore></script>',
+      },
+    };
+
+    runConfigResolved(buildPlugin, {
+      root: '/repo/host',
+      base: '/',
+      command: 'build',
+      build: { rollupOptions: {} },
+    } as unknown as ResolvedConfig);
+    runBuildStart(
+      buildPlugin,
+      {
+        emitFile: (file: Rollup.EmittedFile) => {
+          emitted.push(file);
+          return file.type === 'chunk' ? 'host-init-ref' : `bootstrap-${emitted.length}`;
+        },
+      } as unknown as Rollup.PluginContext,
+      {} as Rollup.NormalizedInputOptions
+    );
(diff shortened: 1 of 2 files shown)


src/plugins/__tests__/pluginAddEntry.test.ts
```diff
@@ -1958,6 +1958,61 @@ describe('pluginAddEntry', () => {
...
+      entryName: 'hostInit',
+      entryPath: '/virtual/hostInit.js',
+      inject: 'html',
+    });
+    const buildPlugin = plugins[1];
+    const emitted: Rollup.EmittedFile[] = [];
+    const bundle: any = {
+      'index.html': {
+        type: 'asset',
+        source:
```
dmchoi77, line 1972: Since the issue is specifically that the root-relative external script is converted to a relative path and `vite-ignore` has no effect, could we also test the actual Vite build pipeline?

Vite's HTML build transform removes `vite-ignore` from ignored script tags before Rollup's output-generation hooks run: [Vite's HTML transform](https://github.com/vitejs/vite/blob/v8.2.2/packages/vite/src/node/plugins/html.ts#L543-L548)

The current test manually constructs the bundle HTML with `vite-ignore` still present, so the guard is not tested against the actual post-transform input. In that case, `/external/external.js` may still be processed and rewritten by MF.

It would be useful to have an integration test asserting that the final build output preserves `/external/external.js` and does not route it through the MF bootstrap.

gioboa: Nice catch, I'm fixing it.

gioboa: Thanks 🙏 @dmchoi77

PR #1152, opened 2026-08-28
fix: cache repeated shared export inspections

### LINKED ISSUE #1150: [Bug]: inspectSharedExportsFromFile re-lexes a shared package's entire re-export graph on every import site
### Describe the bug

`inspectSharedExportsFromFile` recursively `readFileSync`s and regex-lexes a shared package’s whole `export *` / `export { } from` graph, and it is invoked once per **import site** of that package. Nothing on that path is memoized for workspace / source-barrel entries.

---

**Where**

`src/virtualModules/virtualShared_preBuild.ts` (~line 156):

```
writeLoadShareModule → getSharedNamedExports → inspectSharedExportsFromFile → getNamedExportsViaRegex
```

`getNamedExportsViaRegex` dedupes **within** one call (`visited`), not **across** calls. `writeLoadShareModule` has many call sites, so a shared package imported from *N* modules is fully re-walked *N* times.

`getPackageNamedExports` has `packageNamedExportsCache`, but it is disabled for workspace file paths (`!isWorkspaceFilePath`). The configured-`import` path in `getSharedNamedExports` also has no cache. Both hit this function.

---

**Cost**

Workspace UI lib, source barrel, 29 nested `index.ts`, **467** files, **~1,490** import sites → **~700k** reads.

CPU profile: **96.8% (309,445ms)** in the plugin hook; ~53% self time in `RegExp` (`ID_Start` / `ID_Continue`, `export` / `module.exports`). A second host with **36** import sites of the same lib optimized in **~3s** — time tracks import-site count.

On a Vue 3 + Vite 8 host whose shared UI library is a workspace package with `exports.default` → `./index.ts`, that turned a cold `vite` optimize into **~309s** in in my project. 

---

**Repro**

```json
// packages/ui/package.json
{ "name": "ui", "exports": { ".": { "default": "./index.ts" } } }
```

```ts
// packages/ui/index.ts — N≈400 modules is enough
export * from './a';
export * from './b';
```

```ts
shared: { ui: { singleton: true } }
// src/views/View001.ts … View150.ts
import { something } from 'ui';
```

```bash
vite --force   # Vite 8, plugin 1.21.0
```

**Actual:** optimize time grows with **M × N** (ours: **309s**).

Then add `export const NEW = 1` to an already-barreled leaf and import it **without restarting**. **Actual (memo only):** `does not provide an export named 'NEW'`. **Expected:** next transform sees `NEW`.

---

**Proposed fix**

**Patch that fixed this for my project =**
[@module-federation__vite@1.21.0.patch](https://github.com/user-attachments/files/31558165/%40module-federation__vite%401.21.0.patch)

1. Memoize `inspectSharedExportsFromFile` on `` `${entryPath}\0${exportConditions}` ``. Cache negatives. Keep `!entryPath` and thrown errors as `undefined`; early-return `!entryPath` **outside** the cache.

2. On `vite:module-federation-virtual-modules`, clear the memo from `configureServer` on watcher `change` / `add` / `unlink` when the path does not include `node_modules`.

Memoizing `(entryPath, exportConditions)` dropped it to **~1s** for me. Same 59 optimized deps; nothing skipped. That said, a process-lifetime memo is not enough when barrels use `export *`. Adding or removing a named export inside a leaf file changes the reachable name set without touching the entry. Left un-invalidated, the shared `loadShare` proxy keeps the old names and consumers get `does not provide an export named 'X'` until restart, so invalidation fixes that.

### Version

vite 8.1.5, @module-federation/vite 1.21.0

### Reproduction

Example above

### Relevant log output

```shell

```

### Validations

gioboa:
Closes #1150

Memoizes shared export graph scans and invalidates cached results when workspace files change, improving startup performance without stale exports.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -363,6 +363,18 @@ describe('federation in test environment', () => {
 });
 
 describe('virtual module resolution', () => {
+  it('invalidates shared export inspections for workspace watcher events', () => {
+    const on = vi.fn();
+
+    callHook(
+      getVirtualModulesPlugin().configureServer,
+      {} as MinimalPluginContextWithoutEnvironment,
+      { watcher: { on } } as any
+    );
+
+    expect(on.mock.calls.map(([event]) => event)).toEqual(['change', 'add', 'unlink']);
+  });
+
   it('resolves the SSR entry loader from the plugin package', () => {
     const loaderPath = '/plugin/lib/utils/ssrEntryLoader.js';
     const resolverMock = vi.mocked(resolveImportPath);
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,14 @@ import { existsSync, readFileSync, statSync, writeFileSync } from 'fs';
 import { createRequire } from 'module';
 import * as path from 'node:path';
 import { pathToFileURL, fileURLToPath } from 'url';
-import type { ConfigEnv, EnvironmentOptions, Plugin, ResolvedConfig, UserConfig } from 'vite';
+import type {
+  ConfigEnv,
+  EnvironmentOptions,
+  Plugin,
+  ResolvedConfig,
+  UserConfig,
+  ViteDevServer,
+} from 'vite';
 import { version as viteVersion } from 'vite';
 import addEntry, { getBuildInput } from './plugins/pluginAddEntry';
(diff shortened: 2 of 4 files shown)


gioboa (commented):
Could you try this pre-release @dfagan2 ?

PR #1153, opened 2026-08-28
chore: improve preserve ignored scripts through Vite

gioboa:
Tracks ignored script sources before Vite removes their marker, then excludes them from federation bootstrap rewriting.

### DIFF
--- a/integration/host-build.test.ts
+++ b/integration/host-build.test.ts
@@ -1,5 +1,6 @@
 import { mkdir, rm, symlink } from 'fs/promises';
 import { dirname, resolve } from 'path';
+import { version as viteVersion } from 'vite';
 import { describe, expect, it } from 'vitest';
 import type { ModuleFederationOptions } from '../src/utils/normalizeModuleFederationOptions';
 import { buildFixture, FIXTURES } from './helpers/build';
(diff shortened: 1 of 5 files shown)


integration/host-build.test.ts
```diff
@@ -146,6 +146,22 @@ describe('host build', () => {
     expect(bootstrapAsset?.source).toContain('hostInit');
   });
 
+  it('leaves vite-ignore scripts unchanged in a Vite build', async () => {
```
dmchoi77: I checked Vite 5's HTML plugin and found that it does not handle `vite-ignore`; module scripts are always converted into imports: [Vite 5 HTML plugin](https://github.com/vitejs/vite/blob/v5.4.21/packages/vite/src/node/plugins/html.ts#L435-L457)

This is why `Integration (vite@5)` fails: Vite/Rollup tries to resolve `/external/external.js` before MF reaches its `generateBundle` hook.

Could this test be gated to Vite versions that support `vite-ignore` in HTML, while keeping the existing Vite 5 integration coverage unchanged?

gioboa: Yeah, thanks for your help. I did that 👏

PR #1155, opened 2026-08-30
fix: warn when remote type omitted

tonoizer:
Warns when a remote omits type while preserving the existing "var" default.
Adds documentation and regression tests.


gioboa (approved):
Thanks 👏 @tonoizer

PR #1156, opened 2026-08-30
fix: expand react/ prefix in SSR loadShare and eager seed

tonoizer:
## Summary

Follow-up to #1148 (`react/` stays a namespace prefix). Two holes in `virtualRemoteEntry`:

1. **SSR `loadShare`:** `generateRemoteEntrySSR` iterated `options.shared` keys. After #1148, `react/` stays `react/`. The host share scope has `react` / `react/jsx-runtime`, not `react/`, so `if (!scopeShare[pkg]) continue` skipped the prefix and never seeded `jsx-runtime` into `__mf_module_cache__`.
2. **Eager seed:** `if (!pkg.endsWith('/') && share.shareConfig?.eager)` meant `eager: true` on `react/` was a no-op.

Fix: skip keys ending in `/`; expand via the same matcher/usedShares as the browser entry. Do not seed the string `react/`. Does not change `normalizeSharedKey`.

## Test plan

- [ ] `shared: { 'react/': { singleton: true, import: false } }` → SSR init `loadShare`s `react` + `react/jsx-runtime`, never the string `react/`
- [ ] `eager: true` on `react/` seeds `react` + imported subpaths
- [ ] Existing #1148 `react/` prefix tests still pass

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -723,6 +723,75 @@ describe('virtualRemoteEntry', () => {
     expect(code).toMatch(/loaded: false,\s+materialize: true,\s+eager: true,\s+from: "host"/);
   });
 
+  it('expands eager react/ prefix into concrete shared modules, never the prefix string', async () => {
+    const mod = await import('../virtualRemoteEntry');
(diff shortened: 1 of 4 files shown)


gioboa (approved):
Thanks @tonoizer 👏 
Nice fix

PR #1157, opened 2026-08-30
fix: hashed SSR remoteEntry filename matches browser entry

tonoizer:
## Summary

P0 follow-up to #1148 / the hashed `filename` default from #1123. Browser `remoteEntry-[hash].js` is hashed; SSR `pluginSSRRemoteEntry` still wrote a stable `remoteEntry.js`. SSR hosts then miss the hashed asset.

This aligns the SSR emit filename with the browser entry so hashed and custom `filename` values resolve on both sides.

Does not reopen #1148.

## Test plan

- [ ] Default hashed `filename` → SSR and browser emit the same `remoteEntry-[hash].js` basename
- [ ] Custom `filename` (no hash) still matches on both sides
- [ ] Existing SSR host fixtures still resolve the entry


gioboa (commented):
Thanks @tonoizer nice fix 🚀

gioboa (approved):

PR #1158, opened 2026-08-30
fix: Keep React DOM Sharing Client-Safe

tonoizer:
## Summary

  Prevents `react-dom/server*` from being automatically added to client share sets through `COMMON_SHARED_SUBPATHS`.

  Browser-safe subpaths remain automatically shared:

  - `react-dom/client`
  - `react-dom/profiling`

  During SSR, server subpaths resolve normally or can be configured as explicit shared keys.

  No public API changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ export default defineConfig({
       name: "host",
       remotes: {
         remote: {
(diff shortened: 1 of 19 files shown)


gioboa (approved):
That's great 👏 @tonoizer

PR #1160, opened 2026-08-31
fix: ignore package-manager protocol requiredVersion

tonoizer:
Prevents package-manager protocol specifiers from reaching runtime semver checks by falling back to the installed package version. Supports catalog, workspace, Yarn, npm, Git, and URL protocols.

### DIFF
--- a/src/utils/normalizeModuleFederationOptions.ts
+++ b/src/utils/normalizeModuleFederationOptions.ts
@@ -1,5 +1,5 @@
 import { SharedConfig, ShareStrategy } from '@module-federation/runtime/types';
-import { isRequiredVersion, type moduleFederationPlugin } from '@module-federation/sdk';
+import type { moduleFederationPlugin } from '@module-federation/sdk';
 
 export type RemoteEntryType =
   | 'var'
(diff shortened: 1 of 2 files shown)


src/utils/normalizeModuleFederationOptions.ts
```diff
@@ -278,6 +278,14 @@ function normalizeShareItem(
       },
     };
   }
+  // Package-manager protocols (catalog:, workspace:*, npm:pkg@range, patch:, …)
+  // are not semver ranges. Passing them through breaks runtime satisfy().
+  // Treat non-isRequiredVersion strings as unset and fall back like auto-path.
+  const userRequiredVersion = shareItem.requiredVersion;
+  const hasUsableRequiredVersion =
+    userRequiredVersion === false ||
+    (typeof userRequiredVersion === 'string' && isRequiredVersion(userRequiredVersion));
```
dmchoi77: isRequiredVersion() in the SDK is not a complete semver range validator; it is only a regex-based prefix check.

As a result, valid ranges such as the following may be classified as unusable, causing the user-provided value to be replaced with a fallback based on the installed version (^installedVersion) or *.
```
requiredVersion: ' >= 8.0.0'
requiredVersion: '* || ^8.0.0'
```
This can change shared provider selection and may lead to a runtime error when strictVersion: true.

Instead of relying on isRequiredVersion(), it would be safer to explicitly exclude only known package-manager protocols such as catalog:, workspace:, npm:, and patch:, or use a complete semver range validator. Regression tests for the ranges above should also be added.

tonoizer: Good catch. isRequiredVersion is only a prefix check. I will stop using it and only rewrite known package-manager protocols. Adding regression tests.

gioboa: I added a small commit  to do this.
Detect any URI-style package specifier and replace it with an installed-version fallback, preventing invalid semver comparisons at runtime.

gioboa (commented):

Image

gioboa (approved):
Thanks @tonoizer 🍍

PR #1162, opened 2026-09-01
fix: keep Lazy Remotes On Demand

### LINKED ISSUE #1161: [Bug]: `loaded-first` preloads code-split-only remotes at bootstrap and one offline remote blocks the whole app (since 1.20.8)
### Describe the bug

Since **1.20.8** a client build with `shareStrategy: 'loaded-first'` emits a bootstrap that calls `loadRemote()` for every remote it considers "static" and `await Promise.all(...)` on them before importing the application entry (PR #1099, fixing #1096). Two things go wrong with that:

**1. The preload set is not the entry graph.** PR #1099 says:

> This change preloads only remotes that are statically imported by the client entry graph. Dynamic-only remotes remain on-demand.

In practice a remote that is only referenced from a code-split chunk — a module reachable exclusively through `import()` — ends up in the preload list. In the reproduction, `host/src/main.js` never mentions `remote_b`; only `host/src/lazy-route.js` does, and that module is loaded by `await import('./lazy-route.js')` on a button click. The generated bootstrap still preloads it:

```js
const __mfRemotePreloads = [
  __mfPreloadRemote("...__remote_a/widget", "remote_a/widget", {...}),
  __mfPreloadRemote("...__remote_b/widget", "remote_b/widget", {...})   // code-split-only remote
];
await Promise.all(__mfRemotePreloads);
// ...
})().then(() => __mfImport("./index-CNpXhWPo.js"));                     // app entry, only on success
```

On first paint the browser downloads all of `remote_b` (`remoteEntry.js`, its runtime chunk, `virtualExposes`, and the exposed `widget-*.js`) while the host's own `lazy-route-*.js` chunk is correctly *not* requested — so the remote half of a split route is no longer split.

**2. One unreachable remote stops the host from booting.** The bootstrap has no `catch`, no `errorLoadRemote`, no `failOver`, and the app entry is imported only in `.then()`. With `remote_b` offline the host entry never runs and the page stays blank — even though the lazy route it belongs to was never opened. This is the failure mode that #732 reported and PR #735 fixed for `loaded-first`; it is back for any remote that the new code classifies as static.

The docs still describe `loaded-first` the other way round (https://module-federation.io/configure/shareStrategy.html):

> Remotes are loaded on-demand rather than during initialization.
>
> The 'loaded-first' strategy defers remote loading until modules are actually requested, which means offline remotes only cause failures when those specific remotes are accessed, not during application startup.

The new path is also build-only (`isLoadedFirstClientBuild` requires `command === 'build'`), so `vite dev` never exercises it and the difference only shows up in production.

**Behaviour matrix** from the reproduction (`remote_a` statically imported by the entry, `remote_b` only by the code-split chunk):

| | 1.20.7 | 1.20.8 / 1.21.1 |
| --- | --- | --- |
| all remotes up | host boots; `widgetA.label` reads `""` (this is #1096) | host boots; `widgetA.label` correct |
| `remote_b` (code-split-only) down | host boots; only the lazy route would fail | **host never boots**, blank page |

So #1099 does fix #1096 for the entry-graph case; the problem is the extra remotes it pulls in and the fail-fast on all of them.

**Where it comes from** (line numbers from 1.20.8 `lib/index.js`, unchanged in 1.21.1):

* `5642`–`5645`: when `isLoadedFirstClientBuild` is set, `remoteSources` is `Array.from(getStaticRemotes(normalizedOptions))`, and the `isDynamicOnlyRemote(...)` filter that the other branch applies is skipped.
* `5687`: `await Promise.all(__mfRemotePreloads)` on this path (`Promise.allSettled` otherwise).
* `9495`–`9510`: the static crawl over entry + exposes threads a `preloadRemotes` flag and stops descending at dynamic imports — this part does implement entry-graph semantics.
* `8635`: `pluginRemoteNamedExports.transform` calls `markStaticRemote(source, options)` for **any** module in the graph containing a static import of a remote, with no entry-graph gating. That is what puts `remote_b` into `getStaticRemotes()`.
* `skipRemotePreload` is only passed internally for the Nuxt serve case (`6019`), so there is no user-facing way to opt out.

**Suggested fixes**

1. Preload only what the entry-graph crawl marked, not what the `transform` hook saw — e.g. keep transform-discovered static imports in a separate set, or mark static only when the importer is itself reachable through static imports from the entry.
2. Keep the app entry reachable on failure: `Promise.allSettled`, or `Promise.all` with a `catch` that routes through `errorLoadRemote` and still imports the entry. Rejecting the bootstrap turns one broken remote into a blank page for the whole application.
3. Either way, make `skipRemotePreload` (or an equivalent) a public option and align the `shareStrategy` docs with the final behaviour.

### Version

vite 7.3.6, `@module-federation/vite` 1.21.1 (also verified on 1.20.8; 1.20.7 is unaffected)

### Reproduction

https://github.com/h11g/mf-vite-loaded-first-eager-preload

```bash
pnpm install
pnpm -r build
pnpm bootstrap   # shows the preload block in host/dist/assets/mf-entry-bootstrap-*.js
pnpm compare     # same host built with 1.20.7 / 1.20.8 / 1.21.1, preload block diffed

# runtime: start remote_a and the host, leave remote_b down
pnpm serve:remote-a
pnpm serve:host   # http://localhost:5000 stays blank on 1.20.8+, boots on 1.20.7
```

### Relevant log output

```shell
# pnpm compare
----- @module-federation/vite 1.20.7 -----
(no remote preload in bootstrap)

----- @module-federation/vite 1.20.8 -----
const runtime = await initHost();
__mfRemotePreloads = [__mfPreloadRemote("__mfe_internal__host__mf_owner__1__remote_a/widget", "remote_a/widget", ),__mfPreloadRemote("__mfe_internal__host__mf_owner__1__remote_b/widget", "remote_b/widget", )]
await Promise.all(__mfRemotePreloads)

----- @module-federation/vite 1.21.1 -----
const runtime = await initHost();
__mfRemotePreloads = [__mfPreloadRemote("__mfe_internal__host__mf_owner__1__remote_a/widget", "remote_a/widget", ),__mfPreloadRemote("__mfe_internal__host__mf_owner__1__remote_b/widget", "remote_b/widget", )]
await Promise.all(__mfRemotePreloads)

# browser, remote_b offline, 1.21.1: host entry never runs
Failed to load resource: net::ERR_CONNECTION_REFUSED   http://localhost:5002/remoteEntry.js
Uncaught (in promise)
  at loadRemote (index-*.js)
  at __mfPreloadRemote (mf-entry-bootstrap-0-*.js:46:52)
  at (mf-entry-bootstrap-0-*.js:59:312)
  --- await ---
  at (mf-entry-bootstrap-0-*.js:68:3)
```

### Validations

gioboa:
Closes #1161

Preload only remotes reachable through static entry imports, preserving code splitting and offline resilience.

### DIFF
--- a/integration/fixtures/loaded-first-lazy-remote-host/entry.js
+++ b/integration/fixtures/loaded-first-lazy-remote-host/entry.js
@@ -1,0 +1,4 @@
+import { REMOTE_NAME } from 'remote1/Module';
+
+document.querySelector('#app').textContent = REMOTE_NAME;
+export const loadLazy = () => import('./lazy.js');
--- a/integration/fixtures/loaded-first-lazy-remote-host/index.html
+++ b/integration/fixtures/loaded-first-lazy-remote-host/index.html
@@ -1,0 +1,2 @@
+<div id="app"></div>
+<script type="module" src="./entry.js"></script>
--- a/integration/fixtures/loaded-first-lazy-remote-host/lazy.js
+++ b/integration/fixtures/loaded-first-lazy-remote-host/lazy.js
@@ -1,0 +1,3 @@
+import { REMOTE_NAME } from 'remote2/Module';
+
+export const lazyName = REMOTE_NAME;
--- a/integration/host-build.test.ts
+++ b/integration/host-build.test.ts
@@ -84,6 +84,32 @@ describe('host build', () => {
     expect(bootstrapCode).toContain('await initHost();');
   });
 
+  it('does not preload static remote imports behind a dynamic chunk', async () => {
+    const output = await buildFixture({
+      fixture: 'loaded-first-lazy-remote-host',
+      mfOptions: {
+        ...LOADED_FIRST_STATIC_MF_OPTIONS,
+        remotes: {
+          ...LOADED_FIRST_STATIC_MF_OPTIONS.remotes,
+          remote2: {
+            name: 'remote2',
+            entry: 'http://localhost:3002/remoteEntry.js',
+            type: 'module',
+          },
+        },
+      },
+    });
+    const bootstrapAsset = output.output.find(
+      (item) => item.type === 'asset' && item.fileName.includes('mf-entry-bootstrap')
+    );
+    const bootstrapCode = (bootstrapAsset as unknown as { source: string }).source;
+
+    expect(bootstrapCode).toContain('"remote1/Module"');
+    expect(bootstrapCode).not.toContain('remote2/Module');
+    expect(bootstrapCode).not.toContain('http://localhost:3002/remoteEntry.js');
+    expect(bootstrapCode).not.toMatch(/^await /m);
+  });
+
   it('prefetches module remote entries before host init for version-first', async () => {
     const output = await buildFixture({
       fixture: 'loaded-first-static-host',
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -80,7 +80,11 @@ import {
   getResolvedLocalSharedImportMapId,
   getUsedShares,
 } from '../virtualModules/virtualRemoteEntry';
-import { getStaticRemotes, getUsedRemotesMap } from '../virtualModules/virtualRemotes';
+import {
+  getPreloadRemotes,
+  getStaticRemotes,
+  getUsedRemotesMap,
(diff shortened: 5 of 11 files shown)


gioboa (commented):

Image

PR #1167, opened 2026-09-01
fix: make dev virtual IDs deterministic

### LINKED ISSUE #1166: optimizeDeps cache is not portable: __mf_owner__<N> uses a mutable per-process counter instead of a deterministic key
# Issue draft — `@module-federation/vite`

**Repo:** https://github.com/module-federation/vite
**Affected versions:** observed on `1.20.5`; code path unchanged in `1.21.1`.

---

## Title

optimizeDeps cache is not portable/persistent: `__mf_owner__<N>` uses a mutable per-process counter instead of a deterministic key

## Summary

The runtime-init owner id embedded into shared-module virtual ids (`…__mf_owner__<N>__loadShare__…`) is assigned from a **module-level process counter** (`nextRuntimeInitOwnerId++`). Because the value depends on plugin-initialization _order within a process_ rather than on the federation config, the id differs between two processes that use the **identical** config (e.g. a standalone `vite optimize` run vs. the `vite` dev server).

Vite's dependency optimizer bakes these `__mf_owner__<N>` ids into the pre-bundled dep chunks in `node_modules/.vite*/deps`. When those chunks are produced by one process and consumed by another (a separate warmup step, a persisted/committed cache, or a **remote/CI build cache restored on another machine**), the dev server registers the runtime-init virtual modules under a _different_ owner number, so the pre-bundled chunks import virtual ids that don't exist:

```
[vite] Pre-transform error: Failed to load url
virtual:mf:__mfe_internal___mf_0_pan_mf_1_xsiam__mf_owner__1__loadShare___mf_0_spartan_mf_2_ng_mf_1_brain_mf_1_core__loadShare__.js
(resolved id: …__mf_owner__1__loadShare___…) in node_modules/.vite-federation/deps/@spartan-ng_brain_dialog.js.
Does the file exist?
```

Here the pre-bundle was produced with `__mf_owner__1` (standalone `vite optimize`) but the dev server allocated `__mf_owner__2`.

## Why this matters

Vite's dep-optimizer cache is explicitly designed to be **persistent and portable** — it is keyed by a `lockfileHash` + `configHash` and is meant to survive restarts and be shareable (this is the basis for warming/committing/remote-caching `.vite/deps`). The MF plugin silently breaks that contract for any federated app: the cache validates by hash (so Vite happily reuses it) but the _content_ references process-local owner ids, producing runtime "Failed to load url" errors.

Concretely this blocks:

- A separate **warmup task** (`vite optimize`) that pre-populates the cache before `vite`.
- **Persisting/committing** `.vite/deps` to skip cold-start pre-bundling.
- **Remote/CI build-cache** restore of the dep cache across machines (e.g. Nx/Turborepo remote cache) — the whole point of which is cross-process/cross-machine reuse.

For a large host the federated cold-start dep pre-bundle is multi-minute, so being unable to cache/share it is a significant DX cost.

## Root cause

`src/virtualModules/virtualRuntimeInitStatus.ts` (bundled at `lib/index.js`):

```js
const runtimeInitOwnerIds = new WeakMap();
let nextRuntimeInitOwnerId = 1;

function getRuntimeInitOwnerId(options) {
	let ownerId = runtimeInitOwnerIds.get(options);
	if (!ownerId) {
		ownerId = nextRuntimeInitOwnerId++; // ← process-local allocation ORDER, not config identity
		runtimeInitOwnerIds.set(options, ownerId);
	}
	return ownerId;
}

// owner id is then embedded into the virtual module id:
new VirtualModule('runtimeInit', '__mf_v__', '', `${options.internalName}${MF_OWNER_INFIX}${ownerId}`); // …__mf_owner__<N>
```

The id already contains `options.internalName`, which is **globally unique per host/remote**. The extra numeric `ownerId` only disambiguates the (rare) case of two `federation()` instances that share an `internalName` — but it does so with a value that is **not a function of the config**, which is what makes the optimizer cache non-portable.

In our single-host build there is exactly one `internalName` in the entire cache (`__mfe_internal___mf_0_pan_mf_1_xsiam`), so the numeric owner adds no disambiguation at all — it purely harms cacheability.

## Proposed fix

Make the owner key **deterministic from config** instead of a process counter, preserving per-instance namespacing while keeping the id stable across processes. Options, roughly in order of preference:

1. Drop the numeric owner entirely when `internalName` is already unique, and rely on `internalName`.
2. Derive the owner suffix from a stable hash of identity-defining config (e.g. `internalName` + `name` + resolved `shared` keys) rather than `nextRuntimeInitOwnerId++`.
3. If a counter must remain for collision handling, seed/replace it with a deterministic tiebreaker (e.g. sort instances by `internalName` and index within that stable order) so the same config always yields the same id.

Any of these keeps the virtual-id namespacing guarantee while restoring the portability/persistence contract of Vite's dep optimizer.

## Reproduction

1. Federated host with `@module-federation/vite`.
2. Run the dep optimizer in a **separate** process from the dev server, e.g. `vite optimize --config vite.config.ts` (or restore a previously produced `.vite/deps` from another machine/CI cache).
3. Start `vite`. The `configHash`/`lockfileHash` match so Vite reuses the cache, but the browser fails with `Failed to load url virtual:mf:…__mf_owner__<N>__loadShare__…` because the server allocated a different `<N>`.

Confirmable by grepping the produced cache:

```bash
grep -rho "__mf_owner__[0-9]*" node_modules/.vite*/deps/*.js | sort -u
# standalone optimize → __mf_owner__1
# dev server          → __mf_owner__2   (differs by init order)
```

## Environment

- `@module-federation/vite` 1.20.5 (code path identical in 1.21.1)
- `rolldown-vite@7.3.1`
- Angular host via `@analogjs/vite-plugin-angular`

gioboa:
Closes #1166

Derive stable owner IDs for dev optimizer cache reuse across processes, while preserving federation instance isolation and legacy compatibility.

### DIFF
--- a/src/virtualModules/__tests__/virtualSharedInstanceIsolation.test.ts
+++ b/src/virtualModules/__tests__/virtualSharedInstanceIsolation.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest';
-import VirtualModule from '../../utils/VirtualModule';
+import VirtualModule, { MF_OWNER_INFIX } from '../../utils/VirtualModule';
 import { findSharedKey } from '../../plugins/pluginProxySharedModule_preBuild';
+import { packageNameEncode } from '../../utils/packageUtils';
 import {
   normalizeModuleFederationOptions,
   type NormalizedShared,
@@ -17,10 +18,12 @@ import {
 import { generateRemoteEntry } from '../virtualRemoteEntry';
 import { getRuntimeInitStatusImportId, writeRuntimeInitStatus } from '../virtualRuntimeInitStatus';
 import { recordTreeShakingExports, setTreeShakingBuildMode } from '../../utils/treeShaking';
+import { getFederationScopeKey } from '../virtualModuleScope';
 
-function makeOptions(name = 'same-name-host') {
+function makeOptions(name = 'same-name-host', filename?: string) {
   return normalizeModuleFederationOptions({
     name,
+    filename,
     shared: {},
   });
 }
@@ -42,7 +45,7 @@ function makeShareItem(importValue: false | string): ShareItem {
 describe('shared virtual module instance isolation', () => {
   it('keeps same-package generated modules owned by same-name federation instances', () => {
     const optionsA = makeOptions();
-    const optionsB = makeOptions();
+    const optionsB = makeOptions('same-name-host', 'secondaryRemoteEntry.js');
     const shareA = makeShareItem(false);
     const shareB = makeShareItem('react');
 
@@ -66,7 +69,7 @@ describe('shared virtual module instance isolation', () => {
 
   it('uses a distinct build init barrier for each federation instance', () => {
     const optionsA = makeOptions();
-    const optionsB = makeOptions();
+    const optionsB = makeOptions('same-name-host', 'secondaryRemoteEntry.js');
     const share = makeShareItem(false);
 
     writeLoadShareModule('react', share, 'build', false, optionsA);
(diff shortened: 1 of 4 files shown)


gioboa (commented):
Since there isn’t a reproduction repository, could you try this prerelease, @vsternbach?

vsternbach:
Thanks for the quick fix! 🙏 Verified against our large Angular + Nx monorepo host (rolldown-vite 7.3.1) using the pkg.pr.new build — a separately-built `optimizeDeps` cache is now reused cleanly by the dev server (deterministic owner ids, no `__mf_owner__` mismatch errors, no re-optimization). Works as expected and fixes #1166. 🎉

PR #1168, opened 2026-09-01
fix: clean up settled pending shared loads

dmchoi77:
## Summary

This change prevents a rejected lazy shared dependency promise from remaining in the generated runtime's global `pendingShareLoads` array and poisoning later `remoteEntry.get()` calls.

The generated runtime now tracks each pending shared-load promise and removes that exact promise when it settles. The original promise is returned unchanged, so callers that are already waiting on the `Promise.all()` barrier still receive the original rejection. Once the failed promise has settled and been removed, a later retry can proceed normally.

Both the default module cache and the `react-server` module cache use the same cleanup behavior.

## Problem

`pendingShareLoads` was append-only. When a lazy shared import failed, its rejected promise stayed in the cache. Subsequent `remoteEntry.get()` and plugin entry calls awaited the same array and repeatedly observed the historical rejection, even after the shared cache had been recovered or re-evaluated. As a result, the expose factory was never called on the retry.

The `import: false` host-provided path could leave the same stale rejection behind.

## Changes

- Add the internal generated-code helper `__mfTrackPendingShareLoad` in `virtualRuntimeInitStatus.ts`.
  - Lazily creates `pendingShareLoads` only on first registration.
  - Adds and returns the original promise.
  - Removes only the matching promise by identity after resolve or reject.
  - Uses `then(cleanup, cleanup)` without consuming the rejection.
  - Is emitted separately for the default and `react-server` module caches.
- Replace the two raw `pendingShareLoads.push(...)` calls in `virtualShared_preBuild.ts`:
  - lazy local shared imports;
  - `import: false` host-provided shared imports.
- Preserve the existing `Promise.all()` barrier semantics in `virtualRemoteEntry.ts` and `pluginAddEntry.ts`.
- Add unit and integration coverage for cleanup, retry behavior, overlapping promises, `import: false`, permanent lazy `remote.get()` recovery, and `react-server` cache isolation.

## Regression coverage

- A failed first shared import rejects the current `remote.get()` and does not call the expose factory.
- After the failed promise is cleaned up and the shared cache is restored or the generated module is re-evaluated, the next `remote.get()` succeeds and calls the factory.
- Settling one of multiple pending promises does not remove another promise.
- The default and `react-server` pending-load arrays remain independent.
- The `import: false` path receives the same cleanup and retry behavior.

### DIFF
--- a/integration/shared-cache-deferred-exports.test.ts
+++ b/integration/shared-cache-deferred-exports.test.ts
@@ -1,7 +1,16 @@
 import { resolve } from 'path';
 import { describe, expect, it } from 'vitest';
 import { parseAst } from 'rollup/parseAst';
-import type { ModuleFederationOptions } from '../src/utils/normalizeModuleFederationOptions';
+import {
+  normalizeModuleFederationOptions,
+  type ModuleFederationOptions,
+} from '../src/utils/normalizeModuleFederationOptions';
+import VirtualModule from '../src/utils/VirtualModule';
+import { generateRemoteEntry } from '../src/virtualModules/virtualRemoteEntry';
+import {
+  getLoadShareModulePath,
+  writeLoadShareModule,
+} from '../src/virtualModules/virtualShared_preBuild';
 import { isRollupChunk } from './helpers/assertions';
 import { buildFixture, FIXTURES } from './helpers/build';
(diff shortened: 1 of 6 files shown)


gioboa (approved):
Thanks @dmchoi77 nice one 🍉

PR #1171, opened 2026-09-02
fix: fall back to root index.html for dev entry injection

### LINKED ISSUE #1170: [Bug]: `build.rollupOptions.input` picks the dev-time HTML entry, so a non-HTML build input skips host-init injection in `vite dev`
### Describe the bug

When `build.rollupOptions.input` is set to something other than an HTML file, `vite dev` never injects the host-init bootstrap into the page it serves. That bootstrap runs a seeding pass which populates the shared-module cache before the entry evaluates. Without it, a shared singleton's export binding is `undefined` when the entry reads it synchronously, and the page throws.

**Observed:** `typeof createRoot` is `"undefined"`, then `TypeError: createRoot is not a function`.

**Expected:** the dev bootstrap and its shared-singleton seeding are injected regardless of `build.rollupOptions.input`, since that option is build-only and has no bearing on what the dev server serves.

Deleting the `build` block from the reproduction — changing nothing else, on the same plugin version — makes the same project render correctly.

#### Mechanism

In `src/plugins/pluginAddEntry.ts`, the `configResolved(config)` hook of the `enforce: 'post'` plugin runs in both dev and build with no `command` guard, and picks the dev-time HTML entry from `config.build.rollupOptions.input` via `getBuildInput`:

```js
const inputOptions = getBuildInput(config);

if (!inputOptions) {
  htmlFilePath = path.resolve(config.root, 'index.html');
} else if (typeof inputOptions === 'string') {
  entryFiles = [resolveProjectId(inputOptions)];
} // ...array / object branches

if (entryFiles.length > 0) {
  htmlFilePath = getFirstHtmlEntryFile(entryFiles);
}

if (htmlFilePath) {
  addHtmlScriptEntries(htmlFilePath);
}
```

When the configured input resolves only to non-HTML files — e.g. a remote-only container whose input is its exposed module — `getFirstHtmlEntryFile` finds nothing, `htmlFilePath` stays `undefined`, and `addHtmlScriptEntries` never runs for the page the dev server actually serves.

Since 1.20.7 (`23cc511`, #1087, "fix: defer remote singleton fallback imports") the static local `import * as __mfLocalShare from …` fallback was removed for `servesRemoteSingletonFallback` cases, which made that seeding pass load-bearing here. To be clear, #1087 is the change that made the missing injection fatal — it is not itself the defect described here, and I am not suggesting its deferred-import behaviour should be reverted.

#### Why #1107 does not already cover this

#1104 ("Blank page on first cold dev start in 1.20.7 — `_jsxDEV is not a function`") is the same window and the same shape of failure for a different binding, fixed by #1107 in 1.20.8, which preloads the relevant shared modules before the entry import.

That preload lives inside the bootstrap that `addHtmlScriptEntries` generates and injects. In the scenario here, `htmlFilePath` is never resolved, so the bootstrap — preload block included — is never injected, and #1107's fix never gets the chance to run. #1107 fixes what happens once the right module is targeted for injection; this is about the plugin not targeting it in the first place.

#1027 / #1028 is the same general category (a deferred export shape defeating a synchronous consumer) in a different code path (`isWorkspaceSingleton` in `virtualShared_preBuild.ts`). Mentioned only as sibling context; it does not share a root cause with this.

#### Possible direction

This looks like `configResolved` conflating "what should Rollup build" with "what page does the dev server serve". One narrow option, which does not touch any case that already resolves an HTML entry: in `serve`, when none of the resolved `entryFiles` is an HTML file and a root `index.html` exists on disk, fall back to it — mirroring what already happens when `build.rollupOptions.input` is unset. Offered as a direction rather than a claim about the best shape; the same function carries several framework-specific guards (a React Router route-input filter, a Vite 8 per-environment guard, SvelteKit/Nitro SSR bail-outs) that a change here needs to leave undisturbed.

**I intend to submit a PR for this.** I have that change implemented locally with a regression test in `integration/dev-singleton-react.test.ts` (fails before, passes after); `npm test` (1151 passed) and `npm run test:integration` (93 passed) are green on it. Happy to open it, or to leave the fix to you if you would rather shape it differently.

### Version

`@module-federation/vite` 1.21.2, `vite` 7.1.5, node 24.

Confirmed on 1.21.2 by the A/B above. A downstream bisect additionally found 1.20.8, 1.20.9 and 1.21.0 affected with the same repro shape. 1.20.7 is where the change this depends on landed, so it is a reasonable suspect for the earliest affected release, but I have not independently reproduced this exact `build.rollupOptions.input` scenario on 1.20.7 itself.

### Reproduction

https://gist.github.com/Gmendesfonseca/3eb8fcc9320c4db6d341cf01f1f8be04

### Relevant log output

```shell
typeof createRoot: undefined
Uncaught TypeError: createRoot is not a function
    at http://localhost:5173/src/main.js:8:14
```

### Validations

Gmendesfonseca:
Closes #1170

## Problem

With `build.rollupOptions.input` set to a non-HTML file, `vite dev` never injects the host-init bootstrap into the page it serves. That bootstrap performs the seeding pass that populates the shared-module cache before the entry evaluates, so without it a shared singleton's export binding is `undefined` when the entry reads it synchronously.

Minimal A/B — same project, same version, one option: `typeof createRoot` is `"function"` without the `build` block and `"undefined"` with it, then `TypeError: createRoot is not a function`. Reproduction linked from the issue.

## Mechanism

`configResolved` in the `enforce: 'post'` plugin runs for both `serve` and `build`, with no `command` guard, and derives the dev-time HTML entry from `build.rollupOptions.input`:

- [`pluginAddEntry.ts#L909`](https://github.com/module-federation/vite/blob/8c58d47c8ef4706e98e9fd12b000242c09d74d34/src/plugins/pluginAddEntry.ts#L909) — reads the build input via [`getBuildInput`](https://github.com/module-federation/vite/blob/8c58d47c8ef4706e98e9fd12b000242c09d74d34/src/plugins/pluginAddEntry.ts#L235).
- [`pluginAddEntry.ts#L912`](https://github.com/module-federation/vite/blob/8c58d47c8ef4706e98e9fd12b000242c09d74d34/src/plugins/pluginAddEntry.ts#L912) — with no build input, `htmlFilePath` correctly falls back to the root `index.html`.
- [`pluginAddEntry.ts#L930`](https://github.com/module-federation/vite/blob/8c58d47c8ef4706e98e9fd12b000242c09d74d34/src/plugins/pluginAddEntry.ts#L930) — with a build input, `htmlFilePath` comes from `getFirstHtmlEntryFile(entryFiles)` instead, which yields nothing when every configured input is a non-HTML file.
- [`pluginAddEntry.ts#L934`](https://github.com/module-federation/vite/blob/8c58d47c8ef4706e98e9fd12b000242c09d74d34/src/plugins/pluginAddEntry.ts#L934) — `addHtmlScriptEntries` is then never called, so the page the dev server actually serves gets no bootstrap.

`build.rollupOptions.input` is build-only and has no bearing on what `vite dev` serves — the dev server serves `index.html` regardless — so using it to pick the dev-time HTML entry conflates "what should Rollup build" with "what page does the dev server serve".

This became load-bearing in 1.20.7. #1087 removed the static local `import * as __mfLocalShare from …` fallback for `servesRemoteSingletonFallback` cases, which is correct on its own; it just means the seeding pass is now the only thing standing between a synchronous consumer and an `undefined` binding. Nothing here changes that behaviour.

This is also why #1107 does not already cover it: that preload lives inside the bootstrap `addHtmlScriptEntries` generates, so when `htmlFilePath` is never resolved the bootstrap — preload included — is never injected at all.

## Fix

In `serve` only, when no HTML entry resolved from the configured inputs and a root `index.html` exists on disk, fall back to it — mirroring what already happens at L912 when `build.rollupOptions.input` is unset.

```diff
           htmlFilePath = getFirstHtmlEntryFile(entryFiles);
         }
 
+        if (config.command === 'serve' && !htmlFilePath) {
+          const rootIndexHtml = path.resolve(config.root, 'index.html');
+          if (fs.existsSync(rootIndexHtml)) {
+            htmlFilePath = rootIndexHtml;
+          }
+        }
+
         if (htmlFilePath) {
           addHtmlScriptEntries(htmlFilePath);
         }
```

The change is additive: it can only set `htmlFilePath` where it would otherwise have stayed `undefined`, so any config that already resolves an HTML entry is untouched. The `command === 'serve'` guard keeps it out of build, and `fs.existsSync` keeps it inert for setups with no root `index.html` — SvelteKit, Nitro and per-environment framework entries among them. The framework-specific guards above it (the React Router route-input filter, the Vite 8 client-environment check) are not modified.

## Testing

Added `boots on first load when build.rollupOptions.input targets a non-HTML file` to `integration/dev-singleton-react.test.ts`, alongside the existing #1104 coverage, extending `createDevServer` with an optional `build` option.

- Test with the source change reverted: fails, `TypeError: jsxDEV is not a function`.
- Test with the change applied: passes.
- `npm test` — 1151 passed, 1 skipped, 49 files.
- `npm run test:integration` — 93 passed, 16 files.
- `npm run typecheck` — clean.

Happy to reshape this if you would rather solve it somewhere else in the hook — the diagnosis is the part I am confident about.

### DIFF
--- a/integration/dev-singleton-react.test.ts
+++ b/integration/dev-singleton-react.test.ts
@@ -5,6 +5,7 @@ import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { createServer as createNetServer } from 'node:net';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
+import type { InlineConfig } from 'vite';
 import { createServer as createViteServer, version as viteVersion } from 'vite';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { federation } from '../src';
@@ -130,13 +131,15 @@ async function createDevServer(
   name: string,
   options: Parameters<typeof federation>[0],
   root = fixtureRoot,
-  logLevel: 'silent' | 'info' = 'silent'
+  logLevel: 'silent' | 'info' = 'silent',
+  build?: InlineConfig['build']
 ): Promise<{ origin: string }> {
   const cacheDir = await mkdtemp(path.join(tmpdir(), `mf-vite-${name}-`));
   const viteServer = await createViteServer({
     root,
     cacheDir,
     logLevel,
+    build,
     plugins: [federation(options)],
     server: {
       cors: true,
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Nice catch 👍 
Thank you @Gmendesfonseca

PR #1173, opened 2026-09-02
fix: defer default-scope singleton fallback imports

### LINKED ISSUE #1172: [Bug]: singleton: true shared dep in the default share scope still downloads its local fallback in production builds, even when the host provides a matching eager singleton
### Describe the bug

A remote built with `@module-federation/vite`, sharing a dependency as `singleton: true` (no `eager`, no `treeShaking`) in the default share scope, downloads its own bundled fallback copy of that dependency over the network in a production build — even when the host already provides an identical, matching, eager singleton, and even though the runtime identity
negotiation correctly resolves to the host's instance.

The object identity is correct, as the remote truly renders using the host's React, but the bytes for the remote's own copy are fetched and discarded every time. In our real app this is ~750KB of wasted `react-dom` transfer plus a similar `@sentry/browser` chunk, on every page load.

### Root cause

`src/virtualModules/virtualShared_preBuild.ts`, `usesDeferredSingletonFallback`:

```js
const usesDeferredSingletonFallback = hasCompleteExportCoverage && (
  isWorkspacePackage ||
  (command !== "build" && isRemoteOnlyContainer(resolvedOptions) && shareItem.shareConfig.singleton === true) ||
  (command === "build" && isRemoteOnlyContainer(resolvedOptions) &&
    (shareItem.shareConfig.singleton === true ? !isDefaultShareScope : isDefaultShareScope))
);
```

For `command === "build"`, `singleton === true`, and the default share scope, this evaluates to `false`, so the eager/always-fetch template is compiled in instead of the lazy, cache-checked-first one. The module graph decides the download before any host negotiation runs and no runtime config can prevent it, because it's a static ES import of the fallback chunk.

This is, as far as I can tell, the one remaining unfixed combination in this condition:

| singleton | share scope | status |
|---|---|---|
| `true` | non-default | fixed by #897 (closed #895) |
| `false` | default | fixed by #1021 (closed #1018) |
| `true` | default | not covered — this issue |

#895 and #1018 both describe adjacent cases of the same underlying mechanism, but neither's fix touches the `singleton === true` + default-scope branch.

### Reproduction

Minimal repro: https://github.com/cjones26/module-federation-vite-singleton-fallback-bug

```sh
cd remote && npm install && npm run build && npm run preview   # :5001
cd ../host && npm install && npm run serve                      # :5000
cd .. && npm install && node test.mjs
```

Output:

```
=== IDENTITY RESULT ===
{
  "sharedReactIsHostReact": true,
  "sharedReactDomIsHostReactDom": true,
  "hostReactVersion": "18.3.1"
}
=== loadShare react-dom chunk requests ===
http://localhost:5001/assets/..._loadShare__react_mf_2_dom_mf_1_client__loadShare__.js-*.js
http://localhost:5001/assets/..._loadShare__react_mf_2_dom__loadShare__.js-*.js   (209KB)
```

Identity matches the host; the 209KB local fallback is fetched anyway.

### Version

- `@module-federation/vite`: 1.21.2
- `@module-federation/enhanced`: 2.9.0
- `vite`: ^6.0.0
- `webpack`: ^5

### Ask

Extend the runtime-detection approach used for the non-default-scope singleton case (#897) and the default-scope non-singleton case (#1021) to cover `singleton: true` + default share scope + production build, so the default (and most common) setup doesn't include for a fallback download which isn't used.

gioboa:
Closes #1172

Avoid downloading unused local fallbacks when the host already provides a matching singleton.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -3413,6 +3413,34 @@ describe('writeLoadShareModule', () => {
     expect(generatedCode).not.toContain('await ');
   });
 
+  it('defers default-scope singleton fallback in remote builds', () => {
+    normalizeModuleFederationOptions({
+      name: 'remote',
+      exposes: { './App': './src/App.jsx' },
+      shared: { lit: { singleton: true } },
+    });
+    const pkg = 'lit';
+    const mockShareItem: ShareItem = {
+      name: pkg,
+      from: '',
+      version: '3.3.2',
+      shareConfig: {
+        singleton: true,
+        strictVersion: false,
+        requiredVersion: '^3.3.2',
+      },
+      scope: 'default',
+    };
+
+    writeLoadShareModule(pkg, mockShareItem, 'build', false);
+
+    const generatedCode = writeSyncSpy.mock.calls.at(-1)?.[0] as string;
+    expect(generatedCode).not.toContain('import * as __mfLocalShare');
+    expect(generatedCode).toContain('initPromise.then');
+    expect(generatedCode).toContain('import("mock-import-id").then((mod) => {');
+    expect(generatedCode).not.toContain('await ');
+  });
+
   it('prepends workspace singleton static import for SSR build loads only', async () => {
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1177, opened 2026-09-02
fix: prevent shared package self-import rewrites

### LINKED ISSUE #1175: [Bug]: shared-module rewrite re-routes a package's own internal imports to its share entry — build path lacks the importer guard the dev path already has
### Describe the bug

`proxyPreBuildShared:resolve-shared-loadShare`'s `resolveId` (lib/index.js:8431, key lookup at :8441) maps an import to the share's loadShare entry whenever the source lands inside `node_modules/<sharedPkg>/**` — via `findSharedKeyForSource` → `getPackageNameFromNodeModulePath(source)`. The guard chain checks many importer shapes (`localSharedImportMap`, `hostAutoInit`, tagged importers, build-config importers), but never asks whether the **importer itself is a file of that same shared package**.

So a shared package's *internal* edges get re-routed to its own share entry whenever they reach `resolveId` in a form that resolves into the package's directory. Two constellations we hit:

1. **Source-shipped package with an internal path alias** (the repro): the package's files import each other via its own `@ui/*` alias, and the consumer maps that alias to `node_modules/<pkg>/src` — the standard way to compile such a package from source. The alias expansion re-enters `resolveId` as an absolute path inside the shared package, and the internal edge is rewritten. The share entry lists only the entry's public names, so every internal name that isn't re-exported fails the build with `MISSING_EXPORT` (log below). On a production design-system package we measured 1619 such internal edges re-routed in one application build; 5 of them referenced non-public names and broke the build.

2. **Published packages whose files self-reference bare subpaths** (silent): with `shared: {jotai: {}}`, jotai's own `esm/vanilla.mjs` importing `jotai/vanilla/internals` reaches the call site as an absolute path and is re-routed to the **root** `jotai` share entry (`shareSource=jotai` — observable by logging at lib/index.js:8441). The build happens to pass, but the package's fallback copy now consumes itself through the share negotiation instead of its own files — under version skew between containers, its internals come from whichever copy won the scope.

In both cases the effect is the same: **the shared package's fallback copy is made to import itself through the federation proxy.**

### Expected

An import whose importer sits inside the shared package's own directory should stay an ordinary module edge; only *foreign* importers should be re-routed to the share.

The plugin already implements exactly this rule on its **dev path**: the esbuild `onResolve` (lib/index.js:9741) bails with

```js
if (getPackageNameFromNodeModulePath(args.importer) === getPackageName(args.path) && !isReactDomSelfReference(args.path, args.importer)) return;
```

The build-path `resolveId` has no equivalent check — dev and build disagree about the same edge.

### Suggested fix (we ship this as a patch in production)

```diff
 				const key = findSharedKeyForSource(source, shared);
 				if (!key) return;
+				// A shared package's OWN files importing its internals (e.g. an internal alias resolving
+				// into node_modules/<pkg>/src) must not be re-routed to the share entry: the entry lists
+				// only the public names, and the package's fallback copy would be importing itself.
+				if (importer && getPackageNameFromNodeModulePath(importer) === key) return;
 				if (useDirectReactImport && key === "react") return;
```

(`getPackageNameFromNodeModulePath` already exists at the call site; this mirrors the dev-path guard. Verified on the repro: pristine 1.21.2 fails with the MISSING_EXPORT below, with the guard the build succeeds and the internal edge resolves to the real file.)

### Version

### Reproduction

https://github.com/marksnode/mf-vite-repro-share-self-import

### Relevant log output

```shell
error during build:
Build failed with 1 error:

[MISSING_EXPORT] "radius" is not exported by "\0virtual:mf:__mfe_internal__host__mf_owner__1__loadShare__source_mf_2_ui__loadShare__.js".
   ╭─[ node_modules/source-ui/src/Button.js:3:9 ]
   │
 3 │ import {radius} from '@ui/tokens.js';
   │         ───┬──
   │            ╰──── Missing export
```

### Validations

dmchoi77:
## Summary

Fixes #1175

- Prevent the build resolver from proxying a shared package's own internal imports through `loadShare`.
- Compare package roots so explicit subpath and trailing-slash wildcard share keys are handled correctly.
- Add regression coverage for regular and wildcard shared keys.

## Validation

- `pnpm test` — 49 test files passed; 1,156 tests passed, 1 skipped
- `pnpm typecheck`
- `pnpm fmt.check`
- `pnpm build`
- Issue reproduction build passes without `MISSING_EXPORT`.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -1042,6 +1042,76 @@ describe('pluginProxySharedModule_preBuild', () => {
     expect(resolution).toBeUndefined();
   });
 
+  it('does not proxy internal imports from the same shared package during build', async () => {
+    hasPackageDependencyMock.mockReturnValue(false);
+
+    const plugins = proxySharedModule({ shared: makeShared() });
+    const proxyPlugin = getProxyPlugin(plugins);
+    const sharedResolvePlugin = getSharedResolvePlugin(plugins);
+    const config: MockUserConfig = { resolve: { alias: [] } };
+
+    callHook(
+      proxyPlugin.config,
+      {
+        meta: createPluginMeta(),
+        resolve: async (id: string) => ({ id: `/resolved/${id}` }),
+      } as unknown as ConfigPluginContext,
+      config,
+      { command: 'build', mode: 'production' } as ConfigEnv
+    );
+
+    const resolution = await callHook(
+      sharedResolvePlugin.resolveId,
+      {
+        resolve: async (id: string) => ({ id: `/resolved/${id}` }),
+      } as any,
+      '/repo/apps/remote/node_modules/vue/src/internal.js',
+      '/repo/apps/remote/node_modules/vue/src/index.js',
+      { isEntry: false }
+    );
+
+    expect(resolution).toBeUndefined();
+    expect(writeLoadShareModuleMock).not.toHaveBeenCalled();
+  });
+
+  it('does not proxy internal imports for wildcard shared package keys during build', async () => {
+    hasPackageDependencyMock.mockReturnValue(false);
+
+    const shared: NormalizedShared = {
+      'transitive/': {
+        ...makeShared().transitive,
+        name: 'transitive/',
+      },
+    };
+    const plugins = proxySharedModule({ shared });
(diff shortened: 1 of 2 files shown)


gioboa (approved):
it looks great to me! 
you rock @dmchoi77 💯

dmchoi77:
@gioboa Thanks for the review!

PR #1178, opened 2026-09-02
fix: keep runtime-registered shared providers

### LINKED ISSUE #1176: [Bug]: container init wipes share-scope entries registered via public registerShared, then the generated init sweep crashes the page on the container's own import:false stub
### Describe the bug

A host declares one consume-only share — the "another container provides this at runtime" configuration:

```js
federation({
    name: 'host',
    shared: {'demo-lib': {import: false, version: '0.0.0', singleton: true}},
    runtimePlugins: ['./src/provide-plugin.js'],
    dts: false,
})
```

and a runtime plugin provides exactly that share through the **public** `instance.registerShared` API — valid semver `2.0.0`, working synchronous factory — at the earliest public hooks (`beforeInit`, and again at `init`). The repro is 3 small files; the full console timeline is in the log output below. Three distinct defects stack up:

1. **The public registration is wiped.** At the `init` hook the `default` scope holds the registered `2.0.0` entry (`init: scope versions = ["2.0.0"]`); by the first `beforeLoadShare` the scope holds only the container's own `0.0.0` stub. The generated container init re-initializes the share scope and discards entries registered through the public API before it ran.

2. **The generated init sweeps then `loadShare` the own stub.** With the stub as the only surviving provider, both generated passes — the external-share bridge and the runtime-only resolve — call the `import: false` entry's `get()`, which **throws by construction**. Each pass logs a console error. On our production app (277 shared packages, consumers `import: false`, `shareStrategy: 'loaded-first'`, provider container registers asynchronously after the hosts initialize) this is a volley of 277 `Failed to bridge external shared module` errors on every container that initializes before the provider, repeated on every lazy remote load — measured with a Playwright console listener attached before navigation. This is the separate issue with a repro promised in the #952 thread (https://github.com/module-federation/vite/issues/952#issuecomment-5494021861) — the 1.21.x sweep rework and the #952/#953 selection fix did not cover this init-ordering case (also measured on 1.20.7 and 1.21.1).

3. **The throw escapes and kills the page.** The rejection propagates through the generated `mf-entry-bootstrap`, so in the minimal setup the page is dead before the first line of application code.

### Expected

- Container init merges into (or at least does not destroy) share-scope entries registered through the public API before it ran.
- The generated init sweeps skip a provider that is the container's **own** `import: false` registration: until a foreign provider registers there is nothing to bridge — only an error to log, or in the minimal case a fatal crash.

### Workaround we ship (verified on the repro and in production)

Three guards in the generated code, one per sweep pass — skip when the selected provider is the container's own consume-only registration:

```js
if (usedShare.shareConfig?.import === false && __mfMatchesSharedProvider(provider, usedShare)) return;
// …and in the pinned-runtime resolve:
if (__mfMatchesSharedProvider(provider, share)) return;
```

With these, the error volley disappears (production console goes from 277 errors per container init to zero; production works because the real provider registers into the scope *after* init and `loaded-first` resolution then finds it). In the minimal repro the fatal throw remains — the host-init `loadShare` still selects the stub because the wiped scope contains nothing else, which is defect 1.

Related: #534 reported `registerShared`/`loadShare` not meeting and was closed as invalid semver (`version: '0'`) — this is different: the registration here is valid `2.0.0` and demonstrably lands in the scope before container init discards it.

### Version

### Reproduction

https://github.com/marksnode/mf-vite-repro-registershared

### Relevant log output

```shell
beforeInit: registered demo-lib 2.0.0 via public registerShared
init: registered demo-lib 2.0.0 via public registerShared
init: scope versions = ["2.0.0"] | options.shared versions = []
[warning] [ Federation Runtime ] Version 0 from 0 of shared singleton module demo-lib does not satisfy the requirement of host which needs *)
beforeLoadShare(demo-lib): scope versions = ["0.0.0"] | options.shared versions = ["0.0.0"]
[error] [Module Federation] Failed to bridge external shared module "demo-lib" Error: [Module Federation] Shared module 'demo-lib' must be provided by host
[warning] [ Federation Runtime ] Version 0 from 0 of shared singleton module demo-lib does not satisfy the requirement of host which needs *)
beforeLoadShare(demo-lib): scope versions = ["0.0.0"] | options.shared versions = ["0.0.0"]
[error] [Module Federation] Failed to resolve runtime-only shared module "demo-lib" Error: [Module Federation] Shared module 'demo-lib' must be provided by host
[pageerror] Error: [Module Federation] Shared module 'demo-lib' must be provided by host
    at Object.get (assets/_virtual_mf-localSharedImportMap___mfe_internal__host__mf_owner__1-….js:1:178)
    at tr.loadShare (assets/dist-….js:4:41960)
    at async assets/hostInit-….js:1:2040
    at async assets/mf-entry-bootstrap-0-….js:33:3

(src/main.js — which would demonstrate loadShare vs the generated import binding — never executes.)
```

### Validations

dmchoi77:
Closes #1176 

Preserve providers registered via public registerShared across container init and skip loading the own import: false stub.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1025,6 +1025,52 @@ describe('virtualRemoteEntry', () => {
     expect(generatedCode).not.toContain('.catch(remoteEntry.init)');
   });
 
+  it('keeps shares registered on the runtime before initShareScopeMap', async () => {
+    const mod = await import('../virtualRemoteEntry');
+
+    const code = mod.generateRemoteEntry(
+      {
+        internalName: '__mfe_internal__remote',
+        name: 'remote',
+        filename: 'remoteEntry.js',
+        remotes: {},
+        runtimePlugins: [],
+        shareScope: 'default',
+        shareStrategy: 'version-first',
+      } as any,
+      'virtual:exposes',
+      'build'
+    );
+
+    expect(code).toContain('const __mfKeepRegisteredShares = (scopeName, scope) => {');
+    expect(code).toContain('Object.entries(initRes.shareScopeMap?.[scopeName] || {})');
+    expect(code).toContain('if (!provider || provider.get === usedShared[pkg]?.get) continue;');
+    expect(code).toContain('if (target[version] === undefined) target[version] = provider;');
+    expect(code).toContain(
+      'const __mfGetRuntimeShareScope = (scopeName, hostScope) => {\n      __mfKeepRegisteredShares(scopeName, hostScope);'
+    );
+    expect(code.indexOf('const initRes = runtimeInit({')).toBeLessThan(
+      code.indexOf(
+        "initRes.initShareScopeMap(\n      'default',\n      __mfGetRuntimeShareScope('default', shared)"
+      )
+    );
+  });
+
+  it('skips host init loadShare for import:false shares without a foreign provider', async () => {
+    const mod = await import('../virtualRemoteEntry');
+
+    const hostInit = mod.generateHostAutoInitCode('"virtual:remoteEntry"', 'build');
+
+    expect(hostInit).toContain('share.shareConfig?.import === false &&');
+    expect(hostInit).toContain(
+      'Object.values(runtime.shareScopeMap?.[scopeName]?.[pkg] || {}).some('
+    );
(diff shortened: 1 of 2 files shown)


gioboa:
Thanks @dmchoi77, this builds on your #1178 and adds the late-provider guard. Credit to you for the groundwork.

gioboa (commented):

Image

gioboa (approved):
Thanks @dmchoi77 for your commitment 💪

PR #1180, opened 2026-09-02
fix: respect eager shared fallback initialization

### LINKED ISSUE #1179: react/jsx-runtime crashes anything calling jsx() at module top level after #1173
Follow-up to #1172 / #1173. After I patched the plugin with that fix and rebuilt I hit a second separate bug.

Repro: https://github.com/cjones26/module-federation-vite-jsx-runtime-bug

`host/` is a webpack app using `@module-federation/enhanced`'s `ModuleFederationPlugin`. It provides `react`/`react-dom` as `singleton: true, eager: true` in the default scope.

`remote/` is a Vite app using `@module-federation/vite`. It shares `react`/`react-dom` as `singleton: true` in the default scope with no eager and no treeShaking. It also shares `icon-lib`, a MUI stand-in that uses the same `createSvgIcon(<path/>)` pattern with a top-level `jsx()` call, and `big-lib`, an unrelated singleton just there so more than one thing is shared at once.

`remote` has #1173 applied via patch-package since it isn't on npm yet.

Run it:

```sh
cd remote && npm install && npm run build
cd ../host && npm install && npm run build
cd .. && npm install && node test.mjs
```

What happens:

```
[Module Federation] Failed to bridge external shared module "icon-lib" TypeError: __mf_1 is not a function
CRASH: __mf_1 is not a function
while loading "./Exposed" from webpack/container/reference/mfRemote
```

icon-lib does this at the top of the file outside any component. MUI's own icon files do the exact same thing.

```js
import { jsx as _jsx } from "react/jsx-runtime";
export const CheckBoxSharpIcon = createSvgIcon(_jsx("path", { d: "..." }), "CheckBoxSharp");
```

I don't have react/jsx-runtime declared in remote/vite.config.ts at all. It gets pulled in automatically by `@vitejs/plugin-react`'s transform and inherits react's own config. Before #1173 its exports got assigned synchronously the moment the module ran. After #1173 it hits `usesDeferredSingletonFallback` like any other default-scope singleton in a production build so jsx/jsxs/Fragment are populated async now but icon-lib needs them synchronously the moment its own module runs during the container's own materialization step well before that resolution has any chance to finish. Any MUI icon package would hit the exact same wall.

Workaround I'm using in `remote/vite.config.ts` is that I've declared `react/jsx-runtime` on its own instead of letting it inherit from `react`.

```js
"react/jsx-runtime": { singleton: false, shareScope: "someCustomScope" },
```

jsx/jsxs/Fragment are plain functions with no shared state. There's no reason to negotiate them with the host at all. Its own scope keeps it off the deferred path and doesn't touch react/react-dom/icon-lib's own sharing.

I'd rather see `eager: true` actually respected in the branch that decides deferred vs eager (`usesDeferredSingletonFallback` in `lib/index.js`) instead of being silently ignored there, which should make this a one-line config change instead of a scope workaround.

Versions:

- `@module-federation/vite` 1.21.0, patched with #1173
- `@module-federation/enhanced` 2.9.0
- `webpack` ^5.91.0
- `react` / `react-dom` 17.0.1

gioboa:
Closes #1179

Prevents eager shared modules from using deferred fallback exports, ensuring module-scope consumers receive initialized bindings synchronously. Adds regression coverage and React JSX-runtime guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,18 @@ federation({
 
 `runtime-infer` is useful for local development and falls back to the full dependency when the required exports are not available. `server-calc` is recommended for deployments because it can use aggregated export metadata from all consumers.
 
+Use `eager: true` when a shared dependency must provide exports during module evaluation. For example, icon libraries may call `jsx()` at module scope before deferred singleton initialization completes. Configure the remote's React share as follows; discovered React subpaths such as `react/jsx-runtime` inherit this setting:
+
+```ts
+federation({
+  shared: {
+    react: { singleton: true, eager: true },
+  },
+});
+```
+
+This bundles the local fallback into the initial entry, so enable it only when synchronous evaluation requires it.
+
 Do not combine `eager: true` with `treeShaking`; eager shared dependencies are bundled into the initial entry and cannot use the on-demand tree-shaking path. Choose eager loading for small dependencies, or tree shaking for larger dependencies such as component libraries.
 
 With `server-calc`, the Vite build records the exports used by each application in its generated Module Federation metadata. A deployment service must then collect that metadata for all applications that share the same dependency and version, merge their `usedExports` lists, and use the resulting union to create one optimized secondary shared artifact. For example, if one application uses `Button` and another uses `Input`, the secondary artifact must contain both exports.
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Image

PR #1182, opened 2026-09-02
fix: skip shared dependency cycle rewrites

### LINKED ISSUE #1181: [Bug]: generated loadShare glue can be emitted above its own fallback module in the chunk — reads the namespace before it exists, and the TypeError kills the whole chunk
### Describe the bug

Three ordinary ingredients (all required — removing any one makes the build work; measured):

1. a shared package (`demo-ui` in the repro);
2. a package-level dependency cycle: `demo-ui` → `demo-utils` → `demo-ui` (bare specifiers; the second package is NOT shared — the cycle alone is enough. Components/utilities pairs do this in real design systems);
3. one import of a subpath the share table does not cover (`demo-ui/extra.js`), evaluated before the root import — the plugin rewrites only sources that map to a share key, so this edge enters the package's files directly, bypassing the loadShare glue.

The build succeeds. At runtime the page dies before the first line of application code (log below).

In the emitted chunk, the loadShare glue lands **above** the fallback module it references: the glue's eager binding read (`i = r.BUTTON_KIND` with `r = H`) is the chunk's *first* statement, while `var H = t({BUTTON_KIND: () => U, …})` — the fallback module's namespace — is its *last*. Bundled `var`s turn what real ESM would make a TDZ `ReferenceError` into a silent `undefined`; the glue throws, and the **whole chunk evaluation aborts**: the fallback never initializes, application code merged into the same chunk is dead, and the async cache seed then writes `undefined` into the share cache.

What makes it nasty is the order-sensitivity: **swap the order of the two import statements in the entry** (root import first, subpath import second) — same module graph, same config — and everything runs fine. The crash depends on which edge enters the module cycle first, which decides where the cycle is cut and whether the glue is emitted before its own fallback.

We first hit this at scale: in a 279-share production build exactly one package's chunk came out in this order, and that share plus the feature code merged into its chunk died on every startup with the same one-line `TypeError`.

### Expected

The generated loadShare glue must not evaluate before the fallback module it references — either order the emitted chunk so the fallback body comes first, or defer the glue's eager namespace read until first access, or fail loudly instead of reading `undefined` and taking the whole chunk down.

### Version

### Reproduction

https://github.com/marksnode/mf-vite-repro-share-cycle-ordering

### Relevant log output

```shell
TypeError: Cannot read properties of undefined (reading 'BUTTON_KIND')
    at assets/_virtual_mf___mfe_internal__host__mf_owner__1__loadShare__demo_mf_2_ui__loadShare__.js-DScf19v8.js:1:209

(both console.log calls in src/main.js never execute — the entry chunk chain is dead)
```

### Validations

gioboa:
Closes #1181

A dependency of a shared package that imports the shared package back (e.g. demo-ui → demo-utils → demo-ui) was proxied through the loadShare glue. That puts the glue and its prebuild wrapper inside the package's own evaluation cycle. When an untracked subpath such as demo-ui/extra.js is walked first, rolldown cuts the cycle at the wrapper's fallback edge, the wrapper's eager export reads run before the package body, and the whole chunk dies with TypeError: Cannot read properties of undefined.

Extend the #1177 same-package rule: when the importer is a node_modules package listed in the shared package's package.json dependencies, keep the ordinary module edge instead of proxying it.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -23,7 +23,7 @@ const {
   getTreeShakingSharedProviderImportIdMock,
   getTreeShakingSharedProviderNameMock,
 } = vi.hoisted(() => ({
-  hasPackageDependencyMock: vi.fn<(pkg: string) => boolean>(),
+  hasPackageDependencyMock: vi.fn<(pkg: string, cwd?: string) => boolean>(),
   existsSyncMock: vi.fn<(path: string) => boolean>(() => false),
   readFileSyncMock: vi.fn<(path: string) => string>(() => '{}'),
   addUsedSharesMock: vi.fn<(pkg: string) => void>(),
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1184, opened 2026-09-02
fix: skip satisfy-checking import:false stubs

### LINKED ISSUE #1183: [Bug]: the runtime satisfy-checks the container's own import:false stub — warning noise on every container init (thousands of lines at scale), even against a `*` requirement
### Describe the bug

A container that declares a consume-only share:

```js
shared: {'demo-lib': {import: false, version: '0.0.0', singleton: true}}
```

gets satisfy-warnings **about its own placeholder** on every init. In the reproduction (one share on the page, sharing fully working — see below) the console prints, twice:

```
[ Federation Runtime ] Version 0 from 0 of shared singleton module demo-lib does not satisfy the requirement of host which needs *)
```

The "provider" being version-checked is the container's **own** `import: false` registration — an entry that by definition offers nothing and exists only to declare the demand. Note it fails even a `*` requirement (the stub's version reaches the check as `0`, and the provider name renders as `0` too).

Easiest to observe on a build that includes the #1178 fix (`npm i https://pkg.pr.new/@module-federation/vite@1178` in the repro): the page boots cleanly, `loadShare` and the generated import binding both serve the runtime-registered provider — and the two warnings still print (see the repro README, "Bonus finding" section). On stock 1.21.2 the same two warnings precede the crash tracked in #1176 — the noise is orthogonal to that bug.

### Why it matters at scale

On our production app (277 consume-only shares, several containers, `shareStrategy: 'loaded-first'`, provider registers at runtime) this class of warning printed **~5–6.7k lines on a clean page load**, and the volley repeats on every lazy container init later in the session. The console was unusable, and the only lever was shipping `requiredVersion: false` on every consume entry — which also disables the *legitimate* version negotiation against real providers. That's a workaround we'd like to delete.

### Expected

A container's own `import: false` placeholder should not be a satisfy-check subject — the check is meaningful only against foreign/real providers. Either skip the warning when the examined provider is the consumer's own consume-only registration, or don't surface stubs as version candidates at all.

### Version

### Reproduction

https://github.com/marksnode/mf-vite-repro-registershared

### Relevant log output

```shell
[warning] [ Federation Runtime ] Version 0 from 0 of shared singleton module demo-lib does not satisfy the requirement of host which needs *)
[warning] [ Federation Runtime ] Version 0 from 0 of shared singleton module demo-lib does not satisfy the requirement of host which needs *)
[log] runtime loadShare : hello from the publicly registered copy
[log] generated binding : hello from the publicly registered copy
```

### Validations

gioboa:
Closes #1183

Excludes import: false placeholders from shared provider selection before calling runtime getRegisteredShare, and returns early when no candidates remain. Stops the runtime from treating an empty version map as version 0 and warning on every container init.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -4120,6 +4120,25 @@ describe('virtualRemoteEntry', () => {
     ).toBeUndefined();
   });
 
+  it("does not satisfy-check the container's own import:false stub", async () => {
+    const selectProvider = await getExternalSharedProviderSelector();
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const ownStub = {
+      from: 'host',
+      version: '0.0.0',
+      shareConfig: { singleton: true, import: false, requiredVersion: '^1.0.0' },
+    };
+
+    try {
+      expect(
+        selectProvider({ '0.0.0': ownStub }, 'demo-lib', ownStub, 'loaded-first')
+      ).toBeUndefined();
+      expect(warn).not.toHaveBeenCalled();
+    } finally {
+      warn.mockRestore();
+    }
+  });
+
   it('compares external and local singleton providers with version-first', async () => {
     const selectExternalProvider = await getExternalSharedProviderSelector();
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1186, opened 2026-09-03
fix: ignore type-only imports in expose preflight

tibimunteanu:
## Summary

- reuse the existing module import descriptor scanner when traversing exposed modules
- ignore type-only imports before asking Vite to resolve runtime dependencies
- preserve the existing static-versus-dynamic import behavior
- add regression coverage for a type-only dependency alongside a static remote import

## Problem

`pluginProxyRemoteEntry` traverses the raw source of exposed TypeScript modules before Vite transforms them. Its local regular expression currently treats `import type` declarations as runtime imports and passes their package specifiers to `ctx.resolve()`.

For packages that intentionally provide types without a runtime entry, this can make both development and production builds fail even though the import would normally be erased by the TypeScript transform.

The repository already has `findModuleImportDescriptors`, which distinguishes type-only imports and avoids matching import-like text in comments and strings. Reusing it here keeps the expose preflight traversal aligned with the other import-analysis paths.

## Tests

- `pnpm test`
- `pnpm run typecheck`
- `pnpm run build`
- targeted formatting check for the changed files

### DIFF
--- a/integration/fixtures/nested-remote-transitive/type-only-import.ts
+++ b/integration/fixtures/nested-remote-transitive/type-only-import.ts
@@ -1,0 +1,4 @@
+import type { TypedDocumentNode } from '@graphql-typed-document-node/core';
+import { helper } from 'remoteA/shared/helpers';
+
+export const document = helper as unknown as TypedDocumentNode;
--- a/src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
+++ b/src/plugins/__tests__/pluginProxyRemoteEntry.test.ts
@@ -143,6 +143,42 @@ describe('pluginProxyRemoteEntry', () => {
     expect(generated).not.toContain('__loadRemote__remoteB_mf_1_heavy');
   });
 
+  it('does not resolve type-only imports while collecting remote dependencies', async () => {
+    const expose = resolve('integration/fixtures/nested-remote-transitive/type-only-import.ts');
+    const resolvedSources: string[] = [];
+    const plugin = pluginProxyRemoteEntry({
+      options: getDefaultMockOptions({
+        exposes: { './document': { import: expose } as any },
+        remotes: {
+          remoteA: {
+            name: 'remoteA',
+            entry: 'http://localhost:3001/remoteEntry.js',
+            type: 'module',
+          },
+        },
+      }),
+      remoteEntryId: 'virtual:mf-remote-entry',
+      virtualExposesId: 'virtual:mf-exposes',
+    });
+    const context = {
+      resolve: async (source: string) => {
+        resolvedSources.push(source);
+        return source === expose ? { id: expose } : undefined;
+      },
+    } as any;
+
+    callHook(
+      plugin.config,
+      {} as ConfigPluginContext,
+      {},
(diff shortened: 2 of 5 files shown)


gioboa (changes requested):
Thanks @tibimunteanu for your help.
This test is showing a regression, can you check it please?
```
it('treats an imported runtime binding named type as a value import', () => {
  expect(findModuleImportDescriptors(`import { type as value } from 'remote/value';`)).toEqual([
    { kind: 'static', syntax: 'import', source: 'remote/value', typeOnly: false },
  ]);
});
```

tibimunteanu:
Thanks @gioboa 
The check was also matching type as value, which is a runtime import of the binding named type. I’ve updated the detection to exclude that form and added your test as a regression case. The full unit suite, typecheck, and formatting checks pass locally.

gioboa (changes requested):
Alias case fixed:
```
import { type as value } from 'remote/value';
```
But another runtime case remains misclassified:
```
import type from 'remote/value';
import type, { value } from 'remote/value';
```
/^type\b/ marks both type-only.

gioboa (approved):
That's great 👏 
Thank you @tibimunteanu

tibimunteanu:
Wait! That merged fix solves the reported `type` cases, but the underlying scanner is still unsafe because it approximates JavaScript/TypeScript grammar with regexes.

What works correctly now:

- `import { type as value } from 'x'`
- `import type from 'x'`
- `import type, { value } from 'x'`
- Normal `import type Foo`, `import type { Foo }`, and `import type * as Foo`

Remaining failures:

1. Comments can reverse the classification.

```ts
import type /* comment */, { value } from 'x'; // runtime, classified type-only
import { type/* comment */Foo } from 'x';      // type-only, classified runtime
```

2. The regex can consume multiple statements.

```ts
export type Local = string;
export { value } from 'x';
```

The runtime re-export is incorrectly classified as type-only and skipped.

The same happens with TypeScript import assignments followed by a normal import.

3. It invents imports from strings or comments.

```ts
export const text = " from 'fake-package'";
export const value = 1; // from 'fake-package'
```

Both can produce a fake dependency named `fake-package`, potentially causing resolution failures or incorrect remote detection.

4. It misses valid imports.

```ts
import('x', { with: { type: 'json' } });
import { value } from /* comment */ 'x';
import /* comment */ ('x');
```

5. Vue and Svelte remain an unverified risk.

The recursive scanner accepts `.vue` and `.svelte`, but reads the entire raw component as JavaScript. That can misread template/style content or cross `<script>` boundaries.

Most bugs already existed in the shared `findModuleImportDescriptors()` helper. The merged PR’s important behavioral change is that the expose preflight now uses that helper, so these bugs can affect recursive expose dependency discovery:

- A runtime remote may be silently omitted.
- A type-only dependency may still be resolved—the original problem.
- A fake dependency may be resolved.
- Re-export traversal may behave differently from before.

tibimunteanu:
I will try to open a new PR with the fix if that's ok

gioboa:
> I will try to open a new PR with the fix if that's ok

I created #1189

PR #1191, opened 2026-09-03
fix: unwrap CJS Shared Module Exports

gioboa:
Ensures shared packages return actual exports instead of default-only wrappers.


gioboa (commented):

Image

PR #1194, opened 2026-09-04
fix: handle Cyclic Shared Package Initialization

### LINKED ISSUE #1192: [Bug]: loadShare glue is emitted above its own fallback when two shared packages form a cycle and the app is a dynamic-import root (TypeError kills the merged chunk) — follow-up to #1181
### Describe the bug

Reproduction: https://github.com/marksnode/mf-vite-repro-share-cycle-merged-chunk — `npm install && npm run build && node check.mjs && npm run preview`

Two shared packages that import each other (`demo-ui` ⇄ `demo-utils`, both `singleton`), and an application behind the usual async boundary:

```js
// src/main.js
import('./app.js')
```

`npm run preview` dies with `TypeError: Cannot read properties of undefined (reading 'BUTTON_KIND')` before the first line of application code. Change `main.js` to a static `import './app.js'` — same graph, same config — and it runs fine.

### What happens
1. Both members of the cycle are shared, so every edge between them is rewritten to loadShare glue and the two glue chunks import each other. Rolldown resolves that chunk cycle by merging both glues and both fallbacks into one chunk (the plugin's `codeSplitting.groups` otherwise keep each glue in its own group). #1182 exempts a cycle member that is merely a *dependency* of a share; here both are shares, so nothing is exempt.
2. With the application as a dynamic-import root, module order is decided from the runtime-init side first: the generated `localSharedImportMap` imports every share's fallback **statically**, so a fallback is already on the traversal stack when its own glue is reached from inside the fallback's subtree. The cycle is cut on the glue → fallback edge and the glue lands **above** the fallback it reads.

`node check.mjs` prints the merged chunk's module order: `GLUE(ui) -> FALLBACK(lib-utils) -> GLUE(utils) -> FALLBACK(lib-ui)`. Each glue does `exportModule = demo_ui_exports` and reads `mod["BUTTON_KIND"]` synchronously while `var demo_ui_exports = __exportAll({...})` sits below it in the same chunk — a TDZ error in real ESM, a silent `undefined` after bundling. The throw aborts the whole chunk: neither fallback initialises and the application code merged into the chunk never runs.

`shareStrategy` does not matter; `version-first` and `loaded-first` fail the same way.

### Impact
A workspace monorepo where the shared table is the workspace itself hits this at scale: 279 shares, one strongly connected cluster of 69 glue chunks merged into a single 1.6 MB chunk, the first glue in it reads `undefined`, and every feature bundled into that chunk is dead on every startup. It is what keeps us on a provider-only architecture instead of plain `shared`.

### Expected
The glue must never evaluate before the fallback module it references. Options: order a merged cycle fallback-first (cut on a consumer → glue edge, never on glue → fallback); make the glue's namespace read lazy (resolve `exportModule` on first access); or fail the build loudly when a glue is emitted above its own fallback rather than shipping a chunk that cannot run.

### Version

@module-federation/vite 1.21.3 (Vite 8.2.x / rolldown)

### Reproduction

https://github.com/marksnode/mf-vite-repro-share-cycle-merged-chunk

### Relevant log output

```shell
$ node check.mjs
merged chunk: _virtual_mf___mfe_internal__host__mf_owner__27235446888521__loadShare__demo_mf_2_ui__loadShare__.js-BlF8-mej.js
module order: GLUE(ui) -> FALLBACK(lib-utils) -> GLUE(utils) -> FALLBACK(lib-ui)
GLUE ABOVE FALLBACK - the glue reads an undefined namespace, the chunk dies (see `npm run preview`)

$ npm run preview  (browser console)
Uncaught TypeError: Cannot read properties of undefined (reading 'BUTTON_KIND')
    at _virtual_mf___mfe_internal__host__mf_owner__27235446888521__loadShare__demo_mf_2_ui__loadShare__.js-BlF8-mej.js

# control: src/main.js with a static `import './app.js'` instead of `import('./app.js')`
Button(): [OK button]
describeDefaultButton(): default button kind: primary
```

### Validations

gioboa:
Closes #1192

Preserves normal imports across shared dependency cycles, preventing initialization errors in development and production.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,6 +22,8 @@ import pluginProxyRemotes from './plugins/pluginProxyRemotes';
 import {
   excludeSharedSubDependencies,
   findSharedKey,
+  getSharedPackageFromFile,
+  isSharedPackageDependency,
   proxySharedModule,
 } from './plugins/pluginProxySharedModule_preBuild';
 import { pluginRemoteNamedExports } from './plugins/pluginRemoteNamedExports';
@@ -691,6 +693,14 @@ function createEarlyVirtualModulesPlugin(options: NormalizedModuleFederationOpti
                 if (isSharedResolverInternalImporter(importer)) return;
                 const key = findSharedKey(source, shared);
                 if (!key) return;
+                const importerPackage = getSharedPackageFromFile(importer, shared, root);
+                const reactDomSelfReference = isReactDomSelfReference(source, importer);
+                if (
+                  !reactDomSelfReference &&
+                  (importerPackage === getPackageName(key) ||
+                    (importerPackage && isSharedPackageDependency(key, importerPackage)))
+                )
+                  return;
                 if (isAssetLikeImport(source)) return;
                 const shareItem = shared[key];
                 const isReactSingleton =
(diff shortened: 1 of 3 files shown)


gioboa (commented):

Image

PR #1195, opened 2026-09-04
fix: preserve exports after scanner failure

### LINKED ISSUE #1193: [Bug]: export scan still discards the whole export list on 1.21.3 — a semicolon-less `export const` followed by a multi-line generic (no JSX involved)
### Describe the bug

Reproduction: https://github.com/marksnode/mf-vite-repro-scanner — `npm install && npm run build && node check.mjs`

Two shared packages: `broken-lib` is the JSX case from #1164 (now clean, thanks to #1165); `broken-lib-ts` is plain semicolon-less TypeScript:

```ts
export const a = 1
export const b = 2

export function pick<
    T extends string,
    R = T,
>(value: T): R {
    return value as unknown as R
}
```

### What happens
`getNamedExportsViaRegex` matches `export const a = 1` and calls `getAdditionalTopLevelDeclaratorNames` to collect extra declarators. Without a `;` the helper keeps lexing past the statement into the function below; the `<` opening the type-parameter list is followed by a newline and is not recognised as a type-argument start, so the `,` after `T extends string` reads as a declarator separator at depth 0 with an invalid name → `undefined` → `scanState.complete = false` (`lib/index.js:2508`) → `namedExports: void 0` for the module (`:2176`).

The whole export list of the file is discarded. Adding `;` after `1` and `2` makes the scan clean — the only difference is formatting (Prettier `semi: false`).

### Impact
- `import: false` consumers get a stub with no named exports → hard `MISSING_EXPORT` build failures. Upgrading a monorepo from a patched 1.21.2 to 1.21.3 fails 8 of 13 apps with 38 such errors, all from files whose only trait is a multi-line generic somewhere below a semicolon-less `export const`.
- bundled shares: the loadShare chunk exports the fallback's constants directly (`canLiveRebind = false`), singleton unification silently off — `check.mjs` shows the shape.

### Suggested fix
As proposed in #1164: bound the failure to the extra declarators instead of invalidating the scan —

```diff
 		const additionalNames = getAdditionalTopLevelDeclaratorNames(source, exportedVariableDeclarationRegex.lastIndex, codePositions);
-		if (additionalNames === void 0) scanState.complete = false;
-		else for (const name of additionalNames) names.add(name);
+		if (additionalNames !== void 0) for (const name of additionalNames) names.add(name);
```

(`hasUnsupportedBindingPattern` on the next line keeps failing closed for genuinely unsupported patterns.) Additionally, stopping the helper at a top-level newline followed by a statement keyword (`export`, `function`, `const`, `import`, …) would keep a semicolon-less declaration from ever lexing into the next statement. We run the one-line version above as a yarn patch in production.

### Version

@module-federation/vite 1.21.3 (Vite 8.2.x / rolldown)

### Reproduction

https://github.com/marksnode/mf-vite-repro-scanner

### Relevant log output

```shell
$ node check.mjs
broken-lib: CLEAN SCAN - live re-bindable exports, waits for runtime init
broken-lib-ts: SCANNER BAILED - exports pinned to the local fallback constants (canLiveRebind=false)

# the same file with `;` after the two consts scans clean

# at monorepo scale (import: false consumers), plain 1.21.3:
[MISSING_EXPORT] "MS_IN_SECOND" is not exported by "\0virtual:mf:__mfe_internal__homePage__mf_owner__..__loadShare___mf_0_bpinc_mf_1_lib_mf_2_date_mf_2_utils__loadShare__.js"
[MISSING_EXPORT] "isRegularUserDetails" is not exported by "\0virtual:mf:...__loadShare___mf_0_bpinc_mf_1_ad_mf_2_transport_mf_2_types__loadShare__.js"
... 38 errors across 8 apps
```

### Validations

gioboa:
Closes #1193

Prevents multiline generics after semicolon-less declarations from invalidating detected named exports.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -614,11 +614,17 @@ Object.keys(dependency).forEach(function (key) {
       return "module['exports'] = { hidden: 1 };";
     }
     if (filePath.endsWith('/repo/packages/generic-initializer.ts')) {
-      return `export const definitionRefRegistry = new WeakMap<object, any>();
-export const createThing = factory<Result, Options>();
+      return `export const definitionRefRegistry = new WeakMap<object, any>()
+export const createThing = factory<Result, Options>()
 export const nestedRegistry: Map<string, WeakMap<object, any>> = new Map();
 export const identity = <T = string, U = number>(value: T) => value;
-export const assertedRegistry = <Map<object, any>>new Map();`;
+export const assertedRegistry = <Map<object, any>>new Map(), secondaryRegistry = new Map()
+export function pick<
+  T extends string,
+  R = T,
+>(value: T): R {
+  return value as unknown as R
+}`;
     }
     if (filePath.endsWith('/repo/packages/generic-and-multi.ts')) {
       return 'export const first = factory<Result, Options>(), second = 2;';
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1197, opened 2026-09-04
fix: fix Deferred Singleton Initialization Ordering

### LINKED ISSUE #1196: [Bug]: 1.21.3 defers the default-scope singleton binding assignment, breaking chunks the remote entry imports statically
### Describe the bug

`1.21.3` regressed production builds of a remote-only container that shares a plain npm
package as a default-scope singleton. When a shared package reads one of that singleton's
bindings synchronously while its own module body is evaluated, the binding is `undefined`
and the container never boots:

```text
TypeError: __mf_3 is not a function
```

In a minified application build the same failure reads as `TypeError: P is not a function`.
`1.21.2` is unaffected. Vite `8.1.5` and `8.2.2` behave identically, so the bundler version
is not involved.

### What changed

`1.21.2` compiles the local fallback into the `loadShare` chunk, so the binding is assigned
before the module body runs, whatever the evaluation order is. `1.21.3` assigns it only
inside an `initPromise` continuation:

```js
exportModule = __mfLookupShareModule(__mfModuleCache.share, descriptor);
if (exportModule === void 0) __mfTrackPendingShareLoad(initPromise.then(() => { /* import fallback, assign later */ }));
else __mfAssign(exportModule);

var MODULE_ID = __mf_3(); // reads the binding now
```

That template is only safe once the share cache is populated.

### Why the cache is still empty

The generated bootstrap gates the application graph behind `initHost()`:

```js
(async () => {
  const __mfHostInit = await __mfImport("./hostInit-*.js");
  await initHost();
  ...
})().then(() => __mfImport("./index-*.js"));
```

but `initHost()` itself begins with

```js
const runtime = await (await import('../remoteEntry.js')).init();
```

and only afterwards walks its ordered layers registering shares. Everything the remote entry
reaches **statically** is therefore evaluated before the first share is registered, with
`globalThis.__mf_module_cache__.share` still empty.

This is not a corner case: it happens as soon as a shared package imports
`@module-federation/enhanced/runtime`. The bundler then places the federation runtime in the
same chunk as that package's `loadShare` wrapper, and the generated remote entry imports the
runtime from there — giving the remote entry a static edge into a shared consumer chunk.

### Suspected cause

`writeLoadShareModule()` in `src/virtualModules/virtualShared_preBuild.ts`, changed by #1173
(commit `c58d00e`):

```diff
   const usesDeferredSingletonFallback =
     hasCompleteExportCoverage &&
     shareItem.shareConfig.eager !== true &&
     (isWorkspacePackage ||
       (command !== 'build' && isRemoteOnlyContainer(resolvedOptions) && shareItem.shareConfig.singleton === true) ||
       (command === 'build' &&
         isRemoteOnlyContainer(resolvedOptions) &&
-        (shareItem.shareConfig.singleton === true ? !isDefaultShareScope : isDefaultShareScope)));
+        (shareItem.shareConfig.singleton === true || isDefaultShareScope)));
```

`singleton: true` in the default share scope used to be excluded from the deferred template
in `build`. #1173 included it to stop production builds from downloading a fallback the host
already provides (#1172) — a real problem, but the deferred template assumes an invariant
that does not hold for chunks the remote entry imports statically.

#1173 also dropped the `command !== 'build'` guard on `usesEntryInjectedRemoteFallback`. The
reproduction uses the default `hostInitInjectLocation: 'html'`, so that path is not exercised
here.

Related: #1027 already covered a workspace singleton read synchronously at module-eval time.
This is the same hazard for a non-workspace default-scope singleton.

### Possible directions

- Keep an eager binding assignment (compiled fallback) whenever the shared package is read at
  module-evaluation time, and defer only the network fetch — this would preserve the intent
  of #1173.
- Or keep the remote entry's static graph free of shared consumer chunks, since `initHost()`
  imports it before registering any share.
- Or treat "reachable statically from the remote entry" the way `isConsumedByPeerSingleton`
  is already treated, and use the eager template for those chunks.

Version

vite 8.2.2, @module-federation/vite 1.21.3 (1.21.2 unaffected), @module-federation/enhanced 2.9.0, node 24.15.0, pnpm 11.1.1

Reproduction

https://github.com/francesc79/mf-vite-deferred-default-scope-singleton-repro

Relevant log output

$ pnpm install && pnpm build && pnpm repro
readsSharedBindingAtModuleScope: true
fallbackInlinedInChunk: false
fallbackDeferredBehindInitPromise: true
remoteEntryStaticallyImportsConsumer: true
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal

$ pnpm browser
rendered: "loading"
page errors: ["__mf_3 is not a function"]

# after pinning @module-federation/vite to 1.21.2 in packages/host/package.json
$ pnpm build && pnpm repro
readsSharedBindingAtModuleScope: true
fallbackInlinedInChunk: true
fallbackDeferredBehindInitPromise: false
remoteEntryStaticallyImportsConsumer: true

$ pnpm browser
rendered: "runtime-bridge with-instance / vendor-consumer V1StGXR8Z5jdHi6BmyT8C with-instance"
page errors: none

### Version

8.2.2

### Reproduction

https://github.com/francesc79/mf-vite-deferred-default-scope-singleton-repro

### Relevant log output

```shell

```

### Validations

gioboa:
Close #1196

Preserves deferred fallbacks while eagerly initializing runtime-linked singleton dependencies.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -428,6 +428,19 @@ vi.mock('../../utils/packageUtils', () => ({
         },
       };
     }
+    if (pkg === 'runtime-consumer') {
+      return {
+        path: '/repo/apps/remote/node_modules/runtime-consumer/package.json',
+        dir: '/repo/apps/remote/node_modules/runtime-consumer',
+        packageJson: {
+          name: 'runtime-consumer',
+          dependencies: {
+            '@module-federation/enhanced': '^2.9.0',
+            react: '^19.0.0',
+          },
+        },
+      };
+    }
   }),
   getPackageName: (packageString: string) => {
     const match = packageString.match(/^(?:@[^/]+\/)?[^/]+/);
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1199, opened 2026-09-04
fix: end type arguments at a line break in semicolon-less code

### LINKED ISSUE #1198: [Bug]: export scan misreads `<…, …>` type arguments before a line break — drops the export list or emits a phantom export (main after #1195)
### Describe the bug

Follow-up to #1193 / #1195, reproduced on current `main` (c030ee2) via pkg.pr.new. I have a fix ready and will open a PR right after this issue.

`#1195` stops the extra-declarator scan at the next `export` keyword, so the multi-line generic case from #1193 is fixed. But when plain (non-exported) code follows a semicolon-less `export const`, `getAdditionalTopLevelDeclaratorNames` still lexes through it, and `findLikelyTypeArgumentEnd` refuses a `<…, …>` range whose closing `>` is followed by a line break and an identifier — the most common shape in semicolon-less TypeScript (`hasLikelyTypeArgumentFollower` only accepts `(`, `=`, `.`, `as`, `satisfies` and a few punctuation characters). Two consequences, both reproduced with a real `vite build`:

1. **Export list dropped.** `type Shape = Pick<{x: string; y: string}, 'x'>` after an exported const → the `,` inside the type arguments is taken for a declarator separator, `'x'` is not an identifier → the helper returns `undefined` → `scanState.complete = false` → the shared module loses its whole export list (glue pinned to the local fallback, no live re-binding).
2. **Phantom export.** `let registry: Map<string, number>` after an exported const → the type argument `number` *is* an identifier, so it is recorded as an extra declarator and the generated glue exports and re-binds a `number` binding the module never had (`_=e=>{h=e.a,g=e.b,e.number,e.default}` in the loadShare chunk).

Real-world hit: a React component file `export const Controls: FC<PropsWithChildren> = ({children}) => …` followed by `type ButtonProps = Pick<JSKitButtonProps, 'onClick'>` loses all of its exports.

### Minimal sources (from the repro repo)

`lib-ts-alias/index.ts`
```ts
export const a = 1

type Shape = Pick<{x: string; y: string}, 'x'>

export const b: Shape = {x: 'b'}
```

`lib-ts-phantom/index.ts`
```ts
export const a = 1

let registry: Map<string, number>
registry = new Map()

export const b = () => registry.size
```

### Expected

Both libs report a complete scan with exports `a`, `b` — no bail, no `number` export.

### Proposed fix

In `hasLikelyTypeArgumentFollower`, accept a line break after the closing `>` as a valid follower: in semicolon-less code `a < b, c >` followed by a new line cannot continue a declarator list, so the range can only be type arguments. With that one-line change all four libs in the repro report CLEAN, and a scan over the 6 571 source files of our monorepo goes from 3 bailing files to 0 for this class.

### Version

vite 8.2.1, @module-federation/vite main @ c030ee2 (pkg.pr.new; 1.21.3 affected too), node 22.14.0

### Reproduction

https://github.com/marksnode/mf-vite-repro-scanner — npm install && npm run build && node check.mjs (libs broken-lib-ts-alias / broken-lib-ts-phantom)

### Relevant log output

```shell
$ npm run build && node check.mjs      # @module-federation/vite main @ c030ee2
broken-lib: CLEAN SCAN - live re-bindable exports, waits for runtime init
broken-lib-ts: CLEAN SCAN - live re-bindable exports, waits for runtime init
broken-lib-ts-alias: SCANNER BAILED - exports pinned to the local fallback constants (canLiveRebind=false)
broken-lib-ts-phantom: CLEAN SCAN - live re-bindable exports, waits for runtime init
broken-lib-ts-phantom:   PHANTOM EXPORT - the glue exports a binding named "number"

# glue chunk of broken-lib-ts-phantom re-binds a `number` export the module never had:
_=e=>{h=e.a,g=e.b,e.number,e.default}

# with the proposed fix applied locally:
broken-lib-ts-alias: CLEAN SCAN - live re-bindable exports, waits for runtime init
broken-lib-ts-phantom: CLEAN SCAN - live re-bindable exports, waits for runtime init
```

### Validations

marksnode:
## Description

Fixes #1198. Follow-up to #1195.

When plain code follows a semicolon-less `export const`, `getAdditionalTopLevelDeclaratorNames` keeps lexing into it and asks `findLikelyTypeArgumentEnd` about every `<`. `hasLikelyTypeArgumentFollower` only accepted `(`, `=`, `.`, `as`, `satisfies` and a few punctuation characters after the closing `>`, so a `<…, …>` range that ends a line — `type Shape = Pick<Options, 'onClick'>`, `let registry: Map<string, number>` — was read as a comparison. The `,` inside then became a declarator separator: `'onClick'` fails the identifier check and the whole export list is dropped (`scanState.complete = false`), while `number` passes it and is emitted as a phantom export.

A line break after `>` is a safe terminator: in semicolon-less code `a < b, c >` followed by a new line cannot continue a declarator list, so the range can only be type arguments.

## Changes

- `typeArgumentScanner.ts`: treat a line break between the closing `>` and the next code token as a valid follower.
- `typeArgumentScanner.test.ts`: three semicolon-less line-break cases (string-literal type argument, annotated `let`, trailing line comment).
- `virtualShared_preBuild.test.ts`: the `generic-initializer` mock now has a `type` alias and an annotated `let` between two exported declarations; the existing expectation (exports stay live, no `export *` fallback) covers both symptoms.

## Verification

- `pnpm test`: all green except the pre-existing macOS `/private/var` realpath failure in `packageUtils.test.ts`, which fails identically on `main`.
- Repro repo https://github.com/marksnode/mf-vite-repro-scanner: on `main` (c030ee2) `broken-lib-ts-alias` bails and `broken-lib-ts-phantom` exports `number`; with this branch all four libs report CLEAN.
- Our monorepo (6 571 source files scanned with the extracted scanner): 3 → 0 bailing files of this class.

### DIFF
--- a/src/utils/__tests__/typeArgumentScanner.test.ts
+++ b/src/utils/__tests__/typeArgumentScanner.test.ts
@@ -27,6 +27,14 @@ describe('findLikelyTypeArgumentEnd', () => {
     );
   });
 
+  it.each([
+    ["type Shape = Pick<Options, 'onClick'>\nconst later = 1", "<Options, 'onClick'>"],
+    ['let registry: Map<string, number>\nregistry = new Map()', '<string, number>'],
+    ['const registry = new WeakMap<object, any> // comment\nconst later = 1', '<object, any>'],
+  ])('ends the range at a line break in semicolon-less %s', (source, expected) => {
+    expect(findTypeArgumentRange(source)).toBe(expected);
+  });
+
   it('finds nested type arguments in a TypeScript angle-bracket assertion', () => {
     expect(findTypeArgumentRange('const registry = <Map<object, any>>new Map();')).toBe(
       '<Map<object, any>>'
--- a/src/utils/typeArgumentScanner.ts
+++ b/src/utils/typeArgumentScanner.ts
@@ -62,6 +62,9 @@ function hasLikelyTypeArgumentFollower(
   let next = end + 1;
   while (next < source.length && (!codePositions[next] || /\s/.test(source[next]))) next++;
   if (next >= source.length || /[([.!?=;,)\]}:|&]/.test(source[next])) return true;
+  // A line break after `>` ends the statement in semicolon-less code: `a < b, c >` followed by a
+  // new line cannot continue a declarator list, so the range can only be type arguments.
+  if (source.slice(end + 1, next).includes('\n')) return true;
 
   const followingToken = source
     .slice(next)
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -632,6 +632,9 @@ export const createThing = factory<Result, Options>()
 export const nestedRegistry: Map<string, WeakMap<object, any>> = new Map();
 export const identity = <T = string, U = number>(value: T) => value;
 export const assertedRegistry = <Map<object, any>>new Map(), secondaryRegistry = new Map()
+type Shape = Pick<Options, 'onClick'>
+let registry: Map<string, number>
+registry = new Map()
 export function pick<
   T extends string,
   R = T,


gioboa (approved):
Thanks @marksnode nice fix 👏

PR #1201, opened 2026-09-04
fix: ignore members named export in object and type literals

### LINKED ISSUE #1200: [Bug]: export scan fails closed on a property named `export` (`{ export: … }`, `{ export?: T }`) and drops the module's export list
### Describe the bug

Reproduced on current `main` (c030ee2) via pkg.pr.new; also on 1.21.3. PR with a fix follows right after this issue (independent from #1198 / #1199).

The last pass of `getNamedExportsViaRegex` walks every `\bexport\b` occurrence that no earlier regex claimed and fails closed (`scanState.complete = false`) unless the keyword is preceded by `.` or followed by `(`. A property key named `export` — `{ export: () => 'Export' }` in an object literal, or `{ export?: boolean }` in a type literal — is neither, so a perfectly ordinary module is treated as having an unrecognised export form: `namedExports` becomes `undefined`, the loadShare glue falls back to `export * from` the local copy and the module loses live re-binding for the whole export list.

`export` is a reserved word, so a property key is the only place it can legally appear bare (other than `obj.export`, which is already skipped). `export` followed by `:` or `?:` can therefore never start an export declaration and is safe to skip in the catch-all.

Real-world hit: i18n message tables in two of our shared packages (`{ export: () => 'Export' as const, … }`) — both lose their export list today.

### Minimal source (from the repro repo)

`lib-ts-export-key/index.ts`
```ts
export const messages = {
    export: () => 'Export',
    import: () => 'Import',
}

type Options = {export?: boolean}

export const b = (options: Options) => options.export ? 2 : 1
```

### Expected

Complete scan with exports `messages`, `b`; live singleton proxy, no `export *` fallback.

### Proposed fix

In the catch-all loop, after the existing `(` check, also skip the keyword when the next code character is `:` or `?` followed by `:`. With that change the repro lib reports CLEAN and the two affected files in our monorepo scan cleanly.

### Version

vite 8.2.1, @module-federation/vite main @ c030ee2 (pkg.pr.new; 1.21.3 affected too), node 22.14.0

### Reproduction

https://github.com/marksnode/mf-vite-repro-scanner — npm install && npm run build && node check.mjs (lib broken-lib-ts-export-key)

### Relevant log output

```shell
$ npm run build && node check.mjs      # @module-federation/vite main @ c030ee2
broken-lib-ts-export-key: SCANNER BAILED - exports pinned to the local fallback constants (canLiveRebind=false)

# with the proposed fix (skip `export` followed by `:` / `?:` in the catch-all):
broken-lib-ts-export-key: CLEAN SCAN - live re-bindable exports, waits for runtime init
```

### Validations

marksnode:
## Description

Fixes #1200. Independent from #1199 (both branch off `main`).

The final catch-all in `getNamedExportsViaRegex` walks every `\bexport\b` that no earlier regex claimed and fails closed unless the keyword is preceded by `.` or followed by `(`. A property named `export` — `{ export: () => 'Export' }` in an object literal, `{ export?: boolean }` in a type literal — is neither, so the module was treated as having an unrecognised export form and lost its whole export list (glue fell back to `export * from` the local copy, no live re-binding).

`export` is a reserved word, so a bare property key is the only place it can legally appear besides `obj.export` (already skipped); `export` followed by `:` or `?:` can never start an export declaration.

## Changes

- `virtualShared_preBuild.ts`: in the catch-all, also skip the keyword when the next code character is `:`, or `?` followed by `:`.
- `virtualShared_preBuild.test.ts`: new `export-property-key.ts` mock (object literal with `export:` and `'import':` members, type literal with `export?:` / `import ?:`) and a test asserting the live singleton proxy with `messages`, `visible` and no `export *` fallback. Red before the fix, green after.

## Verification

- `pnpm test`: all green except the pre-existing macOS `/private/var` realpath failure in `packageUtils.test.ts` (fails identically on `main`).
- Repro https://github.com/marksnode/mf-vite-repro-scanner (`broken-lib-ts-export-key`): SCANNER BAILED on `main` c030ee2 → CLEAN SCAN with this branch.
- Two i18n message tables in our monorepo (`{ export: () => 'Export' as const, … }`) scan cleanly with this branch.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -552,6 +552,7 @@ vi.mock('fs', () => ({
       filePath.endsWith('/repo/packages/typed-destructuring.ts') ||
       filePath.endsWith('/repo/packages/decorated-export.ts') ||
       filePath.endsWith('/repo/packages/export-import-alias.ts') ||
+      filePath.endsWith('/repo/packages/export-property-key.ts') ||
       filePath.endsWith('/repo/packages/custom-shared-source/index.ts') ||
       filePath.endsWith('/repo/packages/cached-shared-source/index.ts') ||
       filePath.endsWith('/repo/packages/cached-shared-source/leaf.ts')
@@ -575,6 +576,14 @@ export default function createDefaultOnly() {}`;
 export interface Shape { value: string }
 export declare const declaredOnly: string;
 export default function createDefaultOnly() {}`;
+    }
+    if (filePath.endsWith('/repo/packages/export-property-key.ts')) {
+      return `export const messages = {
+  export: () => 'Export',
+  'import': () => 'Import',
+}
+type Options = { export?: boolean; import ?: boolean }
+export const visible = 2`;
     }
     if (filePath.endsWith('/repo/packages/default-only-export-lookalikes.js')) {
       return `// export*from"./comment.js";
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Good improvement ⭐️
Thank you @marksnode

PR #1203, opened 2026-09-05
fix: keep walking shared dependency cycles past unresolvable dependencies

### LINKED ISSUE #1202: [Bug]: shared-cycle detection (#1194) stops at the first unresolvable dependency — an optional peer deep in the tree puts loadShare glue back above its fallback (follow-up to #1192)
### Describe the bug

Follow-up to #1192 / #1194, reproduced on current `main` (f5b16d4). A one-line fix with a test is ready; PR follows right after this issue.

#1194 keeps a dependency cycle between shared packages from turning into a cycle of glue chunks by not rewriting an import when the importer is reachable through the imported share's manifest dependencies (`isSharedPackageDependency`). The reachability walk is

```ts
for (let installed = queue.shift(); installed; installed = queue.shift()) {
```

so the **first queued dependency whose `package.json` cannot be resolved ends the walk** — everything still in the queue is never expanded. That entry is typically an optional / uninstalled `peerDependency` of some package in the tree (`use-context-selector → react-native`, `@swc/core → @swc/helpers`, `react-dnd → @types/hoist-non-react-statics`, …). Because `reachable.add` happens at enqueue time, direct children survive, but grandchildren behind the cut are lost, so a cycle of length ≥ 4 can go undetected: its edges are rewritten to glue, the glue chunks import each other through the fallbacks, rolldown merges them, and the glue is emitted above its own fallback again — exactly the #1192 failure (`Cannot read properties of undefined (reading '…')` at module evaluation).

In our monorepo (288 shared workspace packages) the walk is cut short for 226 of them and 323 cycle edges flip from "skip" to "rewrite to glue"; the naive all-shared build still dies at boot on `main` for that reason.

### Minimal graph (repro repo)

Four shared packages in a cycle, app behind `import('./app.js')`:

```
demo-ui -> demo-utils -> demo-cases -> demo-forms -> demo-ui
```

`demo-cases/package.json`
```json
"peerDependencies": {"react-native": "*", "demo-forms": "*"},
"peerDependenciesMeta": {"react-native": {"optional": true}}
```

Walking from `demo-utils`: expand `demo-utils` → `[demo-cases]`; expand `demo-cases` → `[react-native, demo-forms]`; `react-native` is not installed → the loop ends, `demo-forms` is never expanded, `demo-ui` is never reached → `demo-ui -> demo-utils` is rewritten to glue (same for `demo-utils -> demo-cases`).

### Expected

An unresolvable dependency is skipped, the walk continues, all four edges are recognised as cycle edges and no glue cycle is formed.

### Proposed fix

```ts
while (queue.length) {
  const installed = queue.shift();
  if (!installed || visited.has(installed.dir)) continue;
  …
}
```

With that change the repro's merged chunk becomes `FALLBACK(lib-utils) -> FALLBACK(lib-ui) -> FALLBACK(lib-forms) -> FALLBACK(lib-cases) -> GLUE(ui)` and the page prints its results.

### Version

vite 8.2.1, @module-federation/vite main @ f5b16d4 (pkg.pr.new), node 22.14.0

### Reproduction

https://github.com/marksnode/mf-vite-repro-share-cycle-missing-peer — npm install && npm run build && node check.mjs; npm run preview for the runtime error

### Relevant log output

```shell
$ npm run build && node check.mjs      # @module-federation/vite main @ f5b16d4
chunk _virtual_mf___mfe_internal__host__mf_owner__171068619721864__loadShare__demo_mf_2_cases__loadShare__.js-D6nNhVoN.js
  module order: GLUE(cases) -> FALLBACK(lib-utils) -> GLUE(utils) -> FALLBACK(lib-ui) -> FALLBACK(lib-forms) -> FALLBACK(lib-cases)
  GLUE ABOVE FALLBACK - the glue reads an undefined namespace, the chunk dies (see `npm run preview`)
glue chunks emitted: 2

$ npm run preview   # browser console
TypeError: Cannot read properties of undefined (reading 'caseTitle')

# with the proposed fix applied locally:
chunk _virtual_mf___mfe_internal__host__mf_owner__171068619721864__loadShare__demo_mf_2_ui__loadShare__.js-DO2yYL5M.js
  module order: FALLBACK(lib-utils) -> FALLBACK(lib-ui) -> FALLBACK(lib-forms) -> FALLBACK(lib-cases) -> GLUE(ui)
  fallback-first - runs fine
Button(): [OK button]
describeDefaultButton(): case: form with primary button / default button
```

### Validations

marksnode:
## Description

Fixes #1202. Follow-up to #1192 / #1194.

`isSharedPackageDependency` decides whether an import between two shared packages is a cycle edge (left alone) or gets rewritten to `loadShare` glue. Its reachability walk was

```ts
for (let installed = queue.shift(); installed; installed = queue.shift()) {
```

so the first queued dependency without a resolvable `package.json` — in practice an optional / uninstalled `peerDependency` somewhere in the tree (`use-context-selector → react-native`, `@swc/core → @swc/helpers`, …) — ended the walk with the rest of the queue unexpanded. Cycle edges behind that point were rewritten to glue, the glue chunks imported each other through the fallbacks, rolldown merged them and the glue landed above its own fallback again (the #1192 failure).

## Changes

- `pluginProxySharedModule_preBuild.ts`: skip an unresolvable entry and keep walking (`while (queue.length) { … if (!installed || visited.has(installed.dir)) continue; … }`).
- `pluginProxySharedModule_preBuild.test.ts`: `vue → (react-native: optional peer, not installed) → forms → react`; the edge from the `react` workspace package to `vue` must not be proxied. Red before the fix, green after.

## Verification

- `pnpm test`: green except the pre-existing macOS `/private/var` realpath failure in `packageUtils.test.ts` (same on `main`).
- Repro https://github.com/marksnode/mf-vite-repro-share-cycle-missing-peer: on `main` (f5b16d4) the merged chunk is `GLUE(cases) -> FALLBACK(lib-utils) -> GLUE(utils) -> …` and the page dies with `Cannot read properties of undefined (reading 'caseTitle')`; with this branch it is `FALLBACK(…) x4 -> GLUE(ui)` and the page prints its results.
- Our monorepo (288 shared workspace packages): the walk was cut short for 226 of them and 323 cycle edges flipped to glue; with this branch none are cut short.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -1171,6 +1171,53 @@ describe('pluginProxySharedModule_preBuild', () => {
     readFileSyncMock.mockReset().mockReturnValue('{}');
   });
 
+  it('keeps walking the dependency tree past an unresolvable optional peer when detecting cycles', async () => {
+    hasPackageDependencyMock.mockReturnValue(false);
+    getInstalledPackageEntryMock.mockImplementation((pkg) =>
+      pkg === 'react' ? '/repo/packages/react/index.js' : undefined
+    );
+    // vue -> (react-native: optional peer, not installed) -> forms -> react: the cycle closes through forms
+    existsSyncMock.mockImplementation(
+      (p: string) =>
+        p === '/repo/apps/remote/node_modules/vue/package.json' ||
+        p === '/repo/apps/remote/node_modules/forms/package.json'
+    );
+    readFileSyncMock.mockImplementation((p: string) => {
+      if (p.endsWith('/vue/package.json')) {
+        return '{"peerDependencies":{"react-native":"*","forms":"1"},"peerDependenciesMeta":{"react-native":{"optional":true}}}';
+      }
+      if (p.endsWith('/forms/package.json')) return '{"dependencies":{"react":"1"}}';
+      return '{}';
+    });
+
+    const plugins = proxySharedModule({ shared: makeShared() });
+    const proxyPlugin = getProxyPlugin(plugins);
+    const sharedResolvePlugin = getSharedResolvePlugin(plugins);
+
+    callHook(
+      proxyPlugin.config,
+      {
+        meta: createPluginMeta(),
+        resolve: async (id: string) => ({ id: `/resolved/${id}` }),
+      } as unknown as ConfigPluginContext,
+      { resolve: { alias: [] } },
+      { command: 'build', mode: 'production' } as ConfigEnv
+    );
+
+    const resolution = await callHook(
+      sharedResolvePlugin.resolveId,
+      { resolve: async (id: string) => ({ id: `/resolved/${id}` }) } as any,
+      'vue',
+      '/repo/packages/react/internal.js',
+      { isEntry: false }
+    );
+
+    expect(resolution).toBeUndefined();
+    expect(writeLoadShareModuleMock).not.toHaveBeenCalled();
+    existsSyncMock.mockReset().mockReturnValue(false);
+    readFileSyncMock.mockReset().mockReturnValue('{}');
+  });
+
   it('does not proxy internal imports for wildcard shared package keys during build', async () => {
     hasPackageDependencyMock.mockReturnValue(false);
 
--- a/src/plugins/pluginProxySharedModule_preBuild.ts
+++ b/src/plugins/pluginProxySharedModule_preBuild.ts
@@ -321,7 +321,10 @@ function getDependencyManifest(dep: string, fromDir: string): InstalledPackageJs
         found = {
           path: packageJsonPath,
           dir,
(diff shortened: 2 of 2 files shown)


marksnode:
Force-pushed a second revision (e3653c3 → eafa5dc) after measuring the fix on our monorepo.

With the walk no longer cut short, it visits every package in `node_modules` (~1 700 here), and `getInstalledPackageJson` (`require.resolve` + reading every ancestor `package.json`) costs ~12 ms per lookup — the first cycle check took minutes and the build effectively hung. So the walk now reads dependency manifests through a plain `node_modules` walk-up with a per-`(dir, dep)` cache (`getDependencyManifest`), and falls back to `getInstalledPackageJson` only when the walk-up finds nothing; the cache is cleared alongside `sharedDependencyCache` in the `config` hook.

Numbers on our monorepo: the host build went from not finishing in 10 min back to 21 s (bundling 17.4 s vs 19.3 s on `main`). The repro and the new test behave the same as with the first revision; `pnpm test` unchanged.

gioboa (approved):
That's great @marksnode 👏

PR #1205, opened 2026-09-05
fix: declare exportModule once for eager workspace singletons

### LINKED ISSUE #1204: [Bug]: `eager: true` on a workspace singleton emits `let exportModule` twice — build fails with PARSE_ERROR
### Describe the bug

Reproduced on current `main` (f5b16d4); PR with a fix and a test follows right after this issue.

A shared package that resolves to a workspace path (a monorepo package, or a `file:` dependency as in the repro) with `shareConfig.eager: true` makes `vite build` fail:

```
[PARSE_ERROR] Identifier `exportModule` has already been declared
     ╭─[ \0virtual:mf:__mfe_internal__host__mf_owner__…__loadShare__demo_mf_2_ui__loadShare__.js ]
 196 │     let exportModule = __mfReadSharedCache(__mfModuleCache.share, {"canonical":"default:demo-ui", …})
 201 │     let exportModule = __mfReadSharedCache(__mfModuleCache.share, {"canonical":"default:demo-ui", …});
```

### Why

In `writeLoadShareModule` the eager export block for a workspace singleton (`generateEagerWorkspaceSingletonExports`) begins with its own `import * as __mfLocalShare … let exportModule = …`, but `moduleBody` only skips its own `let exportModule` declaration for the deferred variants (`usesDeferredSingletonFallback || usesDeferredTreeShakingFallback`). Without `eager` a workspace singleton is also `usesDeferredSingletonFallback`, so the duplicate never showed; `eager: true` turns that flag off (`shareConfig.eager !== true`) while `usesEagerWorkspaceFallback` stays on, and both declarations land in one module. `usesEntryInjectedRemoteFallback` uses the same generator and is exposed the same way.

The existing test “emits eager local exports for an explicitly eager leaf workspace singleton” passes today because it only checks substrings of the generated code.

### Why it matters

`eager: true` is the only way to get a workspace singleton's exports bound synchronously when no other shared singleton depends on it. Otherwise the bindings are assigned after the runtime init and any module-scope read in the app — a `createBrowserRouter([{path, element: <App/>}])` table, a HOC call — captures `undefined` and React throws error #130. That is exactly what happened to our host's root route element (`@bpinc/ad-app-ui` is a leaf share nobody else depends on).

### Minimal repro

`vite.config.js`
```js
federation({name: 'host', shared: {'demo-ui': {singleton: true, eager: true}}, dts: false})
```
`src/main.js`
```js
import {Button} from 'demo-ui'
console.log('Button():', typeof Button === 'function' ? Button() : `DEAD — ${String(Button)}`)
```

### Expected

The build succeeds, the glue declares `exportModule` once, and the module-scope read prints `Button(): [button]`.

### Proposed fix

```ts
const exportLineDeclaresModule =
  usesDeferredSingletonFallback || usesDeferredTreeShakingFallback ||
  usesEagerWorkspaceFallback || usesEntryInjectedRemoteFallback;
const moduleBody = exportLineDeclaresModule ? /* body without its own declaration */ : /* body with it */;
```

With that our host (213 eager workspace singletons) builds in 19.7 s and boots.

### Version

vite 8.2.1, @module-federation/vite main @ f5b16d4 (pkg.pr.new; 1.21.3 affected too), node 22.14.0

### Reproduction

https://github.com/marksnode/mf-vite-repro-eager-workspace-singleton — npm install && npm run build

### Relevant log output

```shell
$ npm run build          # @module-federation/vite main @ f5b16d4
error during build:
[PARSE_ERROR] Identifier `exportModule` has already been declared
     ╭─[ \0virtual:mf:__mfe_internal__host__mf_owner__…__loadShare__demo_mf_2_ui__loadShare__.js:196:9 ]
 196 │     let exportModule = __mfReadSharedCache(__mfModuleCache.share, {"canonical":"default:demo-ui","aliases":["demo-ui"]})
     │         ──────┬─────
     │               ╰─────── `exportModule` has already been declared here
 201 │     let exportModule = __mfReadSharedCache(__mfModuleCache.share, {"canonical":"default:demo-ui","aliases":["demo-ui"]});
     │         ──────┬─────
     │               ╰─────── It can not be redeclared here

# with the proposed fix applied locally: build OK, `vite preview` prints
Button(): [button]
```

### Validations

marksnode:
## Description

Fixes #1204.

A workspace singleton with `shareConfig.eager: true` failed to build with `[PARSE_ERROR] Identifier \`exportModule\` has already been declared`. `generateEagerWorkspaceSingletonExports` (and the entry-injected remote fallback, which shares the generator) already declares `exportModule` in the export block, but `moduleBody` only left out its own declaration for the two deferred variants. Without `eager` a workspace singleton is also `usesDeferredSingletonFallback`, which hid the duplicate; `eager: true` clears that flag and both declarations end up in one module.

## Changes

- `virtualShared_preBuild.ts`: `exportLineDeclaresModule = usesDeferredSingletonFallback || usesDeferredTreeShakingFallback || usesEagerWorkspaceFallback || usesEntryInjectedRemoteFallback` decides which module body is emitted.
- `virtualShared_preBuild.test.ts`: the existing “emits eager local exports for an explicitly eager leaf workspace singleton” test now also asserts a single `let exportModule` and that the generated module parses (`parseAst`). Red before the fix, green after.

## Verification

- `pnpm test`: green except the pre-existing macOS `/private/var` realpath failure in `packageUtils.test.ts` (same on `main`).
- Repro https://github.com/marksnode/mf-vite-repro-eager-workspace-singleton: build fails on `main` (f5b16d4), succeeds with this branch, and `vite preview` prints `Button(): [button]` from a module-scope read.
- Our host with 213 eager workspace singletons: build 19.7 s, boots; the module-scope route table that previously captured `undefined` (React #130) now gets the synchronously bound export.

### DIFF
--- a/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
+++ b/src/virtualModules/__tests__/virtualShared_preBuild.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { readFileSync } from 'fs';
+import { parseAst } from 'vite';
 import { normalizePathForImport } from '../../utils/buildPaths';
 import { getSharedExportConditions } from '../../utils/sharedExportConditions';
 import {
@@ -3743,6 +3744,9 @@ describe('writeLoadShareModule', () => {
     expect(generatedCode).toContain('const __mfApplyEagerShareExports = (mod) => {');
     expect(generatedCode).toContain('__mfApplyEagerShareExports(exportModule);');
     expect(generatedCode).not.toContain('initPromise.then');
+    // The eager export block declares `exportModule` itself; the module body must not declare it again.
+    expect(generatedCode.match(/\blet exportModule\b/g)).toHaveLength(1);
+    expect(() => parseAst(generatedCode)).not.toThrow();
   });
 
   it('does not treat parent package.json name mismatches as workspace package matches', () => {
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -2215,9 +2215,14 @@ export function writeLoadShareModule(
         ? `;() => import(${escapeGeneratedStringLiteral(devImportSource)}).catch(() => {});`
         : '';
 
-  const moduleBody =
-    usesDeferredSingletonFallback || usesDeferredTreeShakingFallback
-      ? `
+  // These export blocks declare `exportModule` themselves; the plain body must not declare it a second time.
+  const exportLineDeclaresModule =
+    usesDeferredSingletonFallback ||
+    usesDeferredTreeShakingFallback ||
+    usesEagerWorkspaceFallback ||
+    usesEntryInjectedRemoteFallback;
+  const moduleBody = exportLineDeclaresModule
+    ? `
     ${prebuildImportLine}
     ${devDynamicImportLine}
     ${importLine}
--- a/src/virtualModules/virtualShared_preBuild.ts
+++ b/src/virtualModules/virtualShared_preBuild.ts
@@ -2230,7 +2230,7 @@ export function writeLoadShareModule(
     ${normalizeLocalShareModuleCode}
     ${exportLine}
   `
-      : `
+    : `
     ${prebuildImportLine}
     ${devDynamicImportLine}
     ${importLine}


gioboa (approved):
Thank you @marksnode nice fix

PR #1206, opened 2026-09-05
fix: harden scanner and dependencies

gioboa:
Handle commented export properties and upgrade vulnerable transitive dependencies.


gioboa (commented):

Image

PR #1208, opened 2026-09-05
fix: seed shares behind an unresolved runtime-only share before the entry

### LINKED ISSUE #1207: [Bug]: init() stops seeding local fallbacks at the first unresolved `import: false` share, leaving every later share unseeded (module-scope reads get `undefined` under `loaded-first`)
### Describe the bug

In the generated container `init()` (`generateRemoteEntry`), local fallbacks are seeded in dependency order. An uncached `import: false` (or tree-shaking) share is a "pre-init seed barrier": it and everything after it in `__mfSeedOrder` go to `__mfDeferredSeedKeys`, and the deferred pass ends with

```js
if (__mfIsRuntimeOnlySharePending(pkg)) break;
```

When that share never gets a provider (nothing in the layout registers it), the loop stops there and **every later share stays unseeded — including shares that do not depend on it at all**. The `catch` branch does the same `break` after logging.

With `shareStrategy: 'loaded-first'` nothing else seeds those shares: the `hostInit` preload loop is generated only for `version-first` (`shouldPreloadShares`). Their deferred `loadShare` wrappers then evaluate inside the entry's own import graph against an empty cache, register their pending load *after* the bootstrap's `pendingShareLoads` barrier has already passed, and assign their exports only after `initPromise` + a dynamic import. Any module-scope read of such a share in the app (`<App/>` in a route table, a HOC call) captures `undefined`.

The stop is silent when the share is merely pending (no `console.error`), which makes this very hard to diagnose.

### Reproduction

https://github.com/marksnode/mf-vite-repro-seed-stops-at-unresolved-share

```bash
npm install && npm run build && npm run preview   # console: Button(): DEAD — undefined
```

Host, `shareStrategy: 'loaded-first'`, two shares: `aaa-host-only` (`import: false`, nobody provides it) and `demo-ui` (a `file:` workspace singleton with a local fallback, no dependency on the first). `src/main.js` reads `Button` from `demo-ui` at module scope. Rename `aaa-host-only` to `zzz-host-only` (so it sorts *after* `demo-ui`) and the page prints `[button]`. With the default `version-first` it also works, because the `hostInit` loop seeds every share after `init()`.

Real-world case (monorepo host, 315 shares): `keyboard-key` — `import: false`, not installed — sits at index 47 of the seed order; the 267 shares after it were never seeded and the app died with React error #130 on an element created in the route table. `eager: true` on every workspace share "fixed" it, which is how we noticed.

### Expected behavior

An unresolved share blocks only itself and its dependents (the shares that would capture an undefined singleton — the case the existing comment guards). Shares that do not depend on it keep seeding, and whatever is still unseeded when the bootstrap reaches the entry is bound before the entry evaluates.

### Two halves

1. The `break` is too broad: only the unresolved share and its dependents need to wait; unrelated shares should keep seeding. `getShareBatches` already knows every share's shared prerequisites.
2. Even a correctly blocked share still ends up with a deferred wrapper that the bootstrap's `pendingShareLoads` barrier never sees, because in build mode nothing evaluates the wrappers before the entry. In the real-world case above the `import: false` share is a `dependency` of the UI kit, so almost every share is (transitively) its dependent — blocking them precisely still leaves the whole app unbound. The bootstrap should evaluate the still-unseeded wrappers after the remote preloads (a provider may have registered by then) and before the entry, so the existing barrier covers them.

PR follows with both.

### System Info

- `@module-federation/vite` main @ `0e67309` (also 1.21.3)
- Vite 8.2.1, rolldown
- Node 22 / 20, macOS

### Version

main @ 0e67309 (and 1.21.3)

### Reproduction

https://github.com/marksnode/mf-vite-repro-seed-stops-at-unresolved-share

### Relevant log output

```shell
Button(): DEAD — undefined

(main.js reads `Button` from the workspace singleton `demo-ui` at module scope; the loadShare wrapper assigned it after initPromise. Real host: `Minified React error #130 … args[]=undefined` from an element created in the route table.)
```

### Validations

marksnode:
## Description

Fixes #1207.

In the generated `init()`, the deferred seeding pass stopped at the first runtime-only share (`import: false` / tree-shaking) that stayed unresolved:

```js
if (__mfIsRuntimeOnlySharePending(pkg)) break;
```

(and the same `break` after a logged resolution error). Every share after it in `__mfSeedOrder` was left unseeded, whether or not it depended on the unresolved one. Under `loaded-first` nothing else seeds them — `hostInit` only preloads shares for `version-first` — so their deferred `loadShare` wrappers assign their exports after the entry has already read them at module scope (`undefined` bindings, React #130). In a 315-share host an uninstalled `import: false` share at index 47 of the order left 267 shares unseeded.

Two things fix it, and both are needed:

1. **`init()` blocks only the unresolved share and its dependents.** The comment's intent — an unresolved provider must not let its consumers capture an undefined or provisional singleton — is kept, but applied to the consumers only; unrelated shares keep seeding. `getShareBatches` already computes every share's shared prerequisites, so the seed code now emits them.
2. **The build bootstrap evaluates still-unseeded wrappers before the entry.** A blocked share (in the host above the `import: false` package is a `dependency` of the UI kit, so nearly every share is its transitive dependent) still has a deferred wrapper, and `pendingShareLoads` can only wait for loads that were registered. `hostInit` now exports `preloadPendingShares()`: after the remote preloads — so a provider registered by a remote still wins — it imports the wrapper of every materialized share whose cache entry is still empty; the deferred ones register their load, and the existing barrier awaits it. Version-first already evaluates every share right after `init()`, so this only brings `loaded-first` to the same point, one step later.

## Changes

- `virtualRemoteEntry.ts`
  - `getSharePrerequisites(ordered)` factored out of `getShareBatches` (same computation); `generateRuntimeSharedCacheSeedCode` emits `__mfSeedPrerequisites` for the shares that have any, as positions in `__mfSeedOrder` (9.8 kB in the 315-share host, 265 entries).
  - Deferred pass: `__mfBlockedSeedKeys` instead of `break` — an unresolved or failed share and every share with a blocked prerequisite are skipped, the rest are seeded.
  - `generateHostAutoInitCode` (build only): `__mfPendingShareImports` = `[pkg, () => import(<loadShare wrapper>)]` for materialized shares with a local fallback (`import !== false`, no tree-shaking), and `preloadPendingShares()` that imports the ones still missing from the share cache after `initHost()`.
- `pluginAddEntry.ts`: the bootstrap calls `preloadPendingShares()` after the remote preloads and before the `pendingShareLoads` await (guarded, so a `hostInit` without it — serve, older builds — is unaffected).
- Tests
  - `virtualRemoteEntry.test.ts`: `seeds shares that do not depend on an unresolved runtime-only share` (fails on `main`), `lets the build bootstrap evaluate still-unseeded share wrappers before the entry`; `does not reject init when a runtime-only provider fails to load` now expects the unrelated share to be attempted too.
  - `pluginAddEntry.test.ts`: the build bootstrap orders `initHost()` → `preloadPendingShares()` → `pendingShareLoads`.

## Verification

- Repro https://github.com/marksnode/mf-vite-repro-seed-stops-at-unresolved-share: `DEAD — undefined` on main → `[button]` with this branch.
- Monorepo host (315 shares, `loaded-first`): React #130 on the route table → boots clean; the index-encoded `__mfSeedPrerequisites` adds 9.8 kB to the container chunk.
- `pnpm test`: 1190 passed; the one failure (`packageUtils > ignores copied package metadata beside the resolved entry`) also fails on untouched `main` on macOS (`/var` vs `/private/var` tmpdir realpath) and is unrelated.


gioboa (commented):
It looks great to me ⭐️

gioboa (approved):

PR #1210, opened 2026-09-05
fix: keep cycle edges from unshared workspace packages off the loadShare glue

### LINKED ISSUE #1209: [Bug]: the shared-cycle guard skips importers that are unshared workspace packages, so their cycle edge becomes loadShare glue inside the share's own evaluation cycle
### Describe the bug

#1194 / #1203 keep an import between two members of a package-level dependency cycle as an ordinary module edge instead of rewriting it to `loadShare` glue — otherwise the glue lands inside its own share's evaluation cycle, rolldown merges the chunks, and the glue can run above its fallback and read an undefined namespace (#1192).

The guard resolves the *importer's* package with `getSharedPackageFromFile`, which only knows two kinds of files:

- files inside `node_modules` → the package name (shared or not), via `getPackageNameFromNodeModulePath`;
- files of **shared** workspace packages → the share's package name (`sharedPackageDirectoryCache` is built from `Object.keys(shared)` only).

A workspace package that is a member of the cycle but **not shared** — it is inlined into the fallback of the share that depends on it — resolves to `undefined`, so `isSharedPackageDependency(key, importerPackage)` is never consulted and its import of the share becomes glue. The identical import from an unshared `node_modules` dependency is kept as a direct edge.

### Reproduction

https://github.com/marksnode/mf-vite-repro-cycle-guard-skips-unshared-workspace-importer

```bash
npm install && npm run build && npm run check
```

```
aaa-ui (shared) ──► zzz-bridge (workspace, NOT shared) ──► aaa-ui   (cycle)
demo-cases (shared) ──► aaa-ui                                     (a peer that depends on aaa-ui → eager wrapper)
```

On `main` the check prints

```
zzz-bridge reads BUTTON_KIND: through the aaa-ui loadShare GLUE (cycle guard skipped)  <<<
bridge body: var bridgeTitle = () => `bridge(${__mf_0})`;
```

`__mf_0` is the wrapper's rebinding variable, assigned only when `__mfApplyEagerShareExports` runs — the glue is now part of `aaa-ui`'s own evaluation cycle. Add `zzz-bridge` to `shared` and the same line reads `directly from the aaa-ui fallback (cycle guard kept the edge)`.

Whether the merged chunk dies depends on where rolldown breaks the cycle. In this 3-package graph it orders the fallback first and `npm run preview` runs. In our monorepo host (315 shares, a 61-module SCC) the same kind of edge — `@bpinc/lib-chat-ui`, unshared, inlined into `cases-ui`'s fallback, importing the shared `@bpinc/lib-ad-ui-kit` — put the `lib-ad-ui-kit` wrapper at position 50 of the merged chunk and its fallback at position 87, so the eager wrapper ran `__mfApplyEagerShareExports(undefined)`:

```
TypeError: Cannot read properties of undefined (reading 'ChatSeparatorDate')
```

The only workaround was to share every workspace package in the app's transitive dependency closure so that no cycle member stays unshared (+28 shares in our case).

### Expected behavior

An unshared workspace importer is judged like an unshared `node_modules` importer: if the imported share's dependency tree reaches the importer's package, the edge is a cycle edge and stays an ordinary import.

### Fix

PR follows: `getSharedPackageFromFile` falls back to the `name` from the nearest `package.json` above the importer (outside `node_modules`, cached per directory), so `isSharedPackageDependency` gets to decide. With it the repro reads `BUTTON_KIND` directly and the monorepo host boots without the closure workaround.

### System Info

- `@module-federation/vite` main @ `0e67309` (also 1.21.3)
- Vite 8.2.1, rolldown
- Node 22 / 20, macOS

### Version

main @ 0e67309 (and 1.21.3)

### Reproduction

https://github.com/marksnode/mf-vite-repro-cycle-guard-skips-unshared-workspace-importer

### Relevant log output

```shell
npm run check (main):
  zzz-bridge reads BUTTON_KIND: through the aaa-ui loadShare GLUE (cycle guard skipped)  <<<
  bridge body: var bridgeTitle = () => `bridge(${__mf_0})`;

Monorepo host, same class of edge inside a 61-module merged chunk:
  TypeError: Cannot read properties of undefined (reading 'ChatSeparatorDate')
      at __mfApplyEagerShareExports (…loadShare___mf_0_bpinc_mf_1_ad_mf_2_form_mf_2_builder__loadShare__.js…chunk.js)
```

### Validations

marksnode:
## Description

Fixes #1209. Follow-up to #1194 / #1203.

The shared-cycle guard in `proxyPreBuildShared:resolve-shared-loadShare` asks `getSharedPackageFromFile(importer, shared)` which package the importing file belongs to and then checks `isSharedPackageDependency(key, importerPackage)`. That lookup knew only `node_modules` files (any package) and files of *shared* workspace packages. A workspace package that sits in the cycle but is not shared — inlined into the fallback of the share that depends on it — came back as `undefined`, the guard was skipped, and its import of the share was rewritten to `loadShare` glue inside the share's own evaluation cycle: the #1192 hazard, now behind an asymmetry (the same import from an unshared `node_modules` dependency is kept direct).

In a 315-share monorepo host this put the eager `lib-ad-ui-kit` wrapper at position 50 of a 61-module merged chunk and its fallback at position 87 → `__mfApplyEagerShareExports(undefined)` → `TypeError: Cannot read properties of undefined (reading 'ChatSeparatorDate')`. Sharing the whole workspace closure was the only workaround.

## Changes

- `pluginProxySharedModule_preBuild.ts`
  - `getSharedPackageFromFile`: when the importer is neither in `node_modules` nor inside a shared package, return the `name` of the nearest `package.json` above it (`getWorkspacePackageNameFromFile`, cached per directory, cleared with the other caches). `isSharedPackageDependency` then decides exactly as for `node_modules` importers.
- `pluginProxySharedModule_preBuild.test.ts`
  - `does not proxy shared dependency cycle edges from unshared workspace packages` — `vue → bridge → vue` with `bridge` an unshared workspace package: the import from `/repo/packages/bridge/src/title.js` is not proxied. Fails on `main`.
  - `still proxies imports from unshared workspace packages outside the shared package cycle` — a workspace package the share does not depend on is proxied as before.

## Verification

- Repro https://github.com/marksnode/mf-vite-repro-cycle-guard-skips-unshared-workspace-importer: `npm run check` goes from `through the aaa-ui loadShare GLUE` to `directly from the aaa-ui fallback`.
- Monorepo host (315 shares): with this fix on top of #1208 it boots without the "share the whole closure" workaround (28 fewer shares) and without `eager: true`.
- `pnpm test`: 1190 passed; the one failure (`packageUtils > ignores copied package metadata beside the resolved entry`) also fails on untouched `main` on macOS (`/var` vs `/private/var` tmpdir realpath).

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -1171,6 +1171,98 @@ describe('pluginProxySharedModule_preBuild', () => {
     readFileSyncMock.mockReset().mockReturnValue('{}');
   });
 
+  it('does not proxy shared dependency cycle edges from unshared workspace packages', async () => {
+    hasPackageDependencyMock.mockReturnValue(false);
+    getInstalledPackageEntryMock.mockImplementation((pkg) =>
+      pkg === 'react' ? '/repo/packages/react/index.js' : undefined
+    );
+    // vue -> bridge -> vue: `bridge` is a workspace package that is not shared, so it gets inlined into vue's
+    // fallback. Its import of vue is a cycle edge exactly like one from a node_modules dependency.
+    existsSyncMock.mockImplementation(
+      (p: string) =>
+        p === '/repo/apps/remote/node_modules/vue/package.json' ||
+        p === '/repo/packages/bridge/package.json'
+    );
+    readFileSyncMock.mockImplementation((p: string) => {
+      if (p.endsWith('/vue/package.json')) return '{"dependencies":{"bridge":"workspace:*"}}';
+      if (p.endsWith('/bridge/package.json')) return '{"name":"bridge"}';
+      return '{}';
+    });
+
+    const plugins = proxySharedModule({ shared: makeShared() });
+    const proxyPlugin = getProxyPlugin(plugins);
+    const sharedResolvePlugin = getSharedResolvePlugin(plugins);
+
+    callHook(
+      proxyPlugin.config,
+      {
+        meta: createPluginMeta(),
+        resolve: async (id: string) => ({ id: `/resolved/${id}` }),
+      } as unknown as ConfigPluginContext,
+      { resolve: { alias: [] } },
+      { command: 'build', mode: 'production' } as ConfigEnv
+    );
+
+    const resolution = await callHook(
+      sharedResolvePlugin.resolveId,
+      { resolve: async (id: string) => ({ id: `/resolved/${id}` }) } as any,
+      'vue',
+      '/repo/packages/bridge/src/title.js',
+      { isEntry: false }
+    );
+
+    expect(resolution).toBeUndefined();
+    expect(writeLoadShareModuleMock).not.toHaveBeenCalled();
+    existsSyncMock.mockReset().mockReturnValue(false);
+    readFileSyncMock.mockReset().mockReturnValue('{}');
+  });
+
+  it('still proxies imports from unshared workspace packages outside the shared package cycle', async () => {
+    normalizeModuleFederationOptions({ name: 'host', shared: {} });
+    hasPackageDependencyMock.mockReturnValue(false);
+    getInstalledPackageEntryMock.mockImplementation((pkg) =>
+      pkg === 'react' ? '/repo/packages/react/index.js' : undefined
+    );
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Nice fix @marksnode 👏

PR #1211, opened 2026-09-05
fix: inject federation init into attributed head tags

tonoizer:
## Summary

Fallback `injectEntryScript` only matched exact `<head>`. Tags like `<head lang="en">` or `<HEAD>` left federation init out of the HTML.

## Fix

Case-insensitive match of the first `<head…>` tag, keep the original tag text, inject the module script after it. `<header>` is not treated as a head tag.

## Test plan

- [x] Unit tests for attributed / uppercase head
- [x] Exact `<head>` still becomes `<head><script…>`


gioboa (approved):
Thanks @tonoizer great fix 👏

PR #1212, opened 2026-09-05
fix: serialize method shorthand and skip native functions in runtime options

tonoizer:
## Summary

`serializeRuntimeOptions` emitted unloadable JS for method shorthand (`onError() { … }` → invalid object literal) and for native functions (`function parse() { [native code] }`).

## Fix

Rewrite method shorthand as a function expression. Native / non-reconstructable functions serialize as `undefined` instead of breaking the generated module.

## Test plan

- [x] Unit tests for method shorthand and native functions
- [x] Existing arrow/`function` forms still serialize


gioboa (commented):
Thanks that's great :clap:

PR #1214, opened 2026-09-05
fix: reset alias regex lastIndex before shared-key matches

tonoizer:
## Summary

Global `RegExp` objects in alias-conflict / shared-key checks kept `lastIndex`, so a later key like `react-dom` could be skipped after `/react/g` matched `react`.

## Fix

Reset `lastIndex` around `.test` so each key is matched from the start.

## Test plan

- [x] Unit test: `/react/g` still matches `react-dom` afterward


gioboa (approved):
Nice fix @tonoizer 🚀

PR #1215, opened 2026-09-05
fix: align SSR and wildcard shared-key matching

tonoizer:
## Summary

Three matcher bugs left after #1148 / #1158:

1. SSR `findVmSharedKey` still auto-mapped `COMMON_SHARED_SUBPATHS` for `import: false` parents; the browser matcher did not.
2. Trailing-slash wildcards used first-key-wins instead of longest prefix (`@scope/ui/` swallowed `@scope/ui/forms/`).
3. `getMatchingNodeModuleSubpath` treated `react-dom/server.browser.js` as `react-dom/server` via a naive `.` prefix.

Also drops a stale `pluginModuleParseEnd` comment that still treated `react-dom/` as a dynamic prefix after #1158.

## Fix

Shared `sharedKeyMatcher` used by browser and SSR. Longest wildcard prefix. Path matcher requires a real module extension or path boundary.

## Test plan

- [x] FAIL→PASS tests for import:false SSR, longest wildcard, server.browser path
- [x] Typecheck / vitest on touched files

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -353,6 +353,35 @@ describe('pluginProxySharedModule_preBuild', () => {
       expect(findSharedKey('react-dom/client', shared)).toBe('react-dom');
     });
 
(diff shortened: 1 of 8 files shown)


gioboa (approved):
Nice improvement 🍒

PR #1216, opened 2026-09-05
fix: scope SSR init remotes and strip hash from SSR exposes filename

tonoizer:
## Summary

`initVirtualModules` registered bare remote names while `generateRemotes` used scoped runtime aliases — first writer of runtime init could leave `loadRemote("…__remote/App")` without a matching runtime entry.

Default hashed `filename` (`remoteEntry-[hash]`) also left a literal `[hash]` in SSR `.exposes.js` URLs (entry was already stripped in #1157).

## Fix

Both paths use `getSsrRuntimeRemotes` / `getRuntimeRemoteAlias`. `getSsrExposesFileName` strips `[hash]` / `[hash:N]` like the SSR entry.

## Test plan

- [x] Isolation test for scoped remotes in SSR init
- [x] Hashed exposes middleware / filename unit tests


gioboa (approved):
Thanks @tonoizer 👏

PR #1217, opened 2026-09-05
fix: complete solid-js and zustand common shared subpaths

tonoizer:
## Summary

`COMMON_SHARED_SUBPATHS` for `solid-js` / `zustand` was incomplete. Apps that share the package root still got separate copies of `solid-js/jsx-runtime` or `zustand/middleware` / `shallow`.

## Fix

Add `solid-js/jsx-runtime`, `solid-js/jsx-dev-runtime`, `zustand/middleware`, and `zustand/shallow`. Collapse policy for `solid-js/` and `zustand/` is unchanged. `import: false` still skips auto-map. Does not reopen `react/` prefix policy.

## Test plan

- [x] Auto-map tests for new subpaths
- [x] import:false does not auto-map


gioboa (approved):
That's great 💯

PR #1218, opened 2026-09-05
fix: guard untyped SSR JSON and remove unused shared helpers

tonoizer:
## Summary

SSR manifest and ModuleRunner invoke responses were trusted via `JSON.parse(…) as …`, so arrays / primitives / non-string `name` could become bad URLs or throw on `'error' in null`.

Also removes unused `getPreBuildLibPath` (duplicate of `getPreBuildLibImportId`), unused `getRemoteFromId`, and unused `include`/`exclude` on `proxySharedModule`.

## Fix

Parse as `unknown` with plain-object / field guards. Delete the unused helpers and options.

## Test plan

- [x] SSR entry loader tests for malformed JSON
- [x] Typecheck / vitest


gioboa (approved):
It looks perfect to me 🥇

PR #1220, opened 2026-09-06
fix: materialize the automatic JSX runtime share on a warm dev cache

### LINKED ISSUE #1219: [Bug]: a warm optimizeDeps cache materializes no shares, so init() seeds nothing and the entry reads undefined shared bindings (#1104 on the warm path)
### Describe the bug

On a dev server for a remote-only container, whether shared modules work depends on whether `node_modules/.vite` is cold or warm.

- Cold (`rm -rf node_modules/.vite`, then `vite`): the page renders.
- Warm (start `vite` a second time): the app throws before mounting.

Every generated module is byte-identical between the two runs. The only difference is the `materialize` flags in the served `virtual:mf-localSharedImportMap`:

| share | cold | warm |
| --- | --- | --- |
| `react` | `materialize: true` | `materialize: false` |
| `react/jsx-dev-runtime` | `materialize: true` | `materialize: false` |
| `react/jsx-runtime` | `materialize: false` | `materialize: false` |
| `react/compiler-runtime` | `materialize: false` | `materialize: false` |

`__mfSeedKeys` filters on `materialize !== false`, so on the warm run `init()` seeds **nothing**. Every `__loadShare__` wrapper then takes the browser branch, which assigns its exported bindings only inside `initPromise.then(...)` — strictly after the wrapper module's own evaluation has finished. The entry's top-level code runs with `undefined` bindings.

Reading `globalThis.__mf_module_cache__` at the moment the entry module evaluates confirms it:

- cold: the share cache holds `react`, `react/jsx-dev-runtime`, …; `pendingShareLoads` is `0`
- warm: those keys are absent; `pendingShareLoads` is `1`

The request order shows the same thing. On the cold run `__prebuild__react_jsx-dev-runtime__prebuild__.js` is requested during `init()`; on the warm run it is never requested at all, and `node_modules/.vite/deps/react_jsx-dev-runtime.js` is fetched dead last — long after the entry has already thrown.

Which binding is the casualty depends on evaluation order: `_jsxDEV is not a function`, `createRoot is not a function`, or a crash inside the prebundled `react-dom_client.js`.

**Cause.** `materialize` is only ever set by `addUsedShares`, and for the automatic JSX runtime both callers miss on a warm cache:

1. `registerEntryImports` (`src/index.ts`) crawls the entry graph, but `findModuleImportDescriptors` is a **textual** scan. `react/jsx-dev-runtime` is injected by the JSX transform and never written in user source, so this scan structurally cannot see it — cold or warm.
2. The resolver hook the plugin installs into `optimizeDeps.esbuildOptions.plugins` / `rolldownOptions.plugins` (both call `addUsedShares`) does see it — but Vite invokes that hook only while it **scans dependencies**, and it skips scanning entirely once `node_modules/.vite` is warm.

So on a warm start nothing calls `addUsedShares`, `getMaterializedShares` emits `materialize: false`, and the share is never seeded. In this reproduction `react` itself is affected too, because no file textually imports `react` either — the only import of it is the compiler-injected one.

The comment already in `generateRuntimeSharedCacheSeedCode` names the assumption that breaks here:

> A share is normally skipped here until the dev scanner has observed a real import and set materialize.

The dev scanner does not run on a warm cache, so "nothing was observed" is being read as "nothing is used".

**Relation to #1104.** #1104 reported the same `_jsxDEV is not a function` symptom with the same 1.20.6 → 1.20.7 bisect, for the *cold* first load, and was closed by #1107 ("preload remote-only dev singleton wrappers before entry import"). The cold path does work now — this reproduction renders on a cold cache on every version from 1.20.7 through `main`, and `integration/dev-singleton-react.test.ts`'s `boots on first load when a remote-only singleton imports jsx-dev-runtime (#1104)` passes on `main`. The warm path is not covered: on a warm cache nothing is materialized at all, so there is nothing for a preload to have warmed.

**Two controls**, to pin the cause rather than the symptom:

- Replacing all JSX with `createElement` calls — no automatic-runtime import anywhere — makes the warm run **pass**.
- Declaring `'react/jsx-dev-runtime'` explicitly in `shared` does **not** help: declared shares are registered with `addConfiguredShare`, which does not materialize either.

**Workaround.** A textual import of the runtime in an entry file makes the crawl see it, and the warm run passes:

```js
import 'react/jsx-dev-runtime'
```

I have a fix and will submit a PR.

### Version

vite 8.2.2, @module-federation/vite 1.21.3 (also reproduced on `main` @ fe9770b)

### Reproduction

https://github.com/Gmendesfonseca/mf-vite-repro-warm-optimizedeps-unmaterialized-shares

```bash
npm install
npm run cold   # clears node_modules/.vite, then starts vite -> renders "mounted"
# stop the server, then:
npm run warm   # starts vite again, cache now warm -> blank page, console throws
```

The same six files, inline. No workspace, no framework beyond React. Deterministic: 3/3 warm runs fail, and clearing `node_modules/.vite` makes it pass again every time.

`package.json`

```json
{
  "name": "mf-vite-repro-warm-optimizedeps-unmaterialized-shares",
  "private": true,
  "type": "module",
  "scripts": { "dev": "vite", "cold": "node ./clear-cache.mjs && vite", "warm": "vite" },
  "dependencies": { "react": "^19.2.0", "react-dom": "^19.2.0" },
  "devDependencies": {
    "@module-federation/vite": "1.21.3",
    "@vitejs/plugin-react": "^6.0.5",
    "vite": "^8.2.0"
  }
}
```

`vite.config.js`

```js
import { defineConfig } from 'vite'
import react from '@vitejs/plugin-react'
import { federation } from '@module-federation/vite'

export default defineConfig({
  plugins: [
    react(),
    federation({
      name: 'remote-a',
      filename: 'remoteEntry.js',
      dts: false, // unrelated to the bug; the dts worker wants a tsconfig and this repro has none
      // A remote-only container: it exposes a module and declares no `remotes`.
      exposes: { './export-app': './src/export-app.jsx' },
      shared: {
        react: { singleton: true, requiredVersion: '^19.2.0' },
      },
    }),
  ],
  server: { port: 5301, strictPort: true },
})
```

`index.html`

```html
<!doctype html>
<html lang="en">
  <head><meta charset="UTF-8" /><title>repro</title></head>
  <body>
    <div id="root"></div>
    <script type="module" src="/src/main.jsx"></script>
  </body>
</html>
```

`src/main.jsx`

```jsx
import { createRoot } from 'react-dom/client'
import ExportApp from './export-app'

createRoot(document.getElementById('root')).render(<ExportApp />)
```

`src/export-app.jsx`

```jsx
export default function ExportApp() {
  return <div id="ok">mounted</div>
}
```

`clear-cache.mjs` just removes `node_modules/.vite` so the next `vite` run is a cold start.

Expected: the page renders identically both times.

**Affected versions** — bisected on this reproduction, warm run:

| version | warm run |
| --- | --- |
| 1.20.6 | passes |
| 1.20.7 | fails |
| 1.21.0 | fails |
| 1.21.2 | fails |
| 1.21.3 | fails |
| `main` @ fe9770b | fails |

Present since 1.20.7, so not a 1.21.3 regression, and still present on `main` including the recent share-seeding fixes (#1194, #1197, #1205, #1208).

### Relevant log output

```shell
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'S')
    at node_modules/.vite/deps/react-dom_client.js:12766:1325
    at node_modules/.vite/deps/react-dom_client.js:14373:4
    at node_modules/.vite/deps/rolldown-runtime-CXvrFHJf.js:14:49

# with react-dom/client also declared shared, the same run instead throws:
Uncaught (in promise) TypeError: _jsxDEV is not a function
```

### Validations

Gmendesfonseca:
## Problem

On a dev server for a remote-only container, shared modules work or don't depending on whether
`node_modules/.vite` is cold or warm. Cold: the page renders. Warm: a shared package's generated
`__loadShare__` wrapper hands the entry an `undefined` binding and the app throws before mounting
(`_jsxDEV is not a function`, `createRoot is not a function`, or a crash inside the prebundled
`react-dom_client.js`, depending on evaluation order).

Fixes #1219.

Same symptom family as #1104, which #1107 fixed for the cold first load. The warm path was not
covered: on a warm cache nothing is materialized at all, so `init()` seeds nothing and there is
nothing for a preload to have warmed.

The generated modules are byte-identical in both cases. The only difference is the `materialize`
flags in `virtual:mf-localSharedImportMap`:

| share | cold | warm |
| --- | --- | --- |
| `react` | `materialize: true` | `materialize: false` |
| `react/jsx-dev-runtime` | `materialize: true` | `materialize: false` |

`__mfSeedKeys` filters on `materialize !== false`, so on the warm run `init()` seeds nothing and
every wrapper falls through to the deferred `initPromise.then(...)` branch — which assigns the
exported bindings strictly after the entry's own module evaluation has finished.

## Cause

`materialize` is only ever set by `addUsedShares`, and for the automatic JSX runtime both callers
miss on a warm cache:

- `registerEntryImports` crawls the entry graph textually. `react/jsx-dev-runtime` is injected by
  the JSX transform and never appears in user source, so this scan structurally cannot see it.
- The resolver hook installed into `optimizeDeps.esbuildOptions.plugins` /
  `rolldownOptions.plugins` does see it, but Vite invokes that hook only while it scans
  dependencies, and it skips scanning entirely once `node_modules/.vite` is warm.

The existing comment in `generateRuntimeSharedCacheSeedCode` states the assumption that breaks:
"A share is normally skipped here until the dev scanner has observed a real import and set
materialize." On a warm cache that scanner never runs, so "nothing observed" was being treated as
"nothing used".

## Change

Two narrow changes, both dev-only in effect:

1. `registerEntryImports` records `react/jsx-dev-runtime` for any crawled `.jsx`/`.tsx` file when
   it is a configured share. This fills the textual scan's blind spot at the point where the blind
   spot is, rather than disabling the optimization.
2. `getMaterializedShares` materializes a subpath share's root package alongside it. A subpath
   resolves through its root's module graph, and the seed batches already order the root first, so
   seeding one without the other leaves half a package seeded.

A configured share that nothing imports still stays lazy — the existing
`materializes shares imported by the entry graph but leaves configured-only shares lazy` test
covers exactly that and still passes.

## Result

The warm run now emits the same `materialize` flags the cold run does — and no more than it does:

| share | cold (before & after) | warm before | warm after |
| --- | --- | --- | --- |
| `react` | `true` | `false` | `true` |
| `react/jsx-dev-runtime` | `true` | `false` | `true` |
| `react/jsx-runtime` | `false` | `false` | `false` |
| `react/compiler-runtime` | `false` | `false` | `false` |

## Verification

- New unit test `materializes the automatic JSX runtime a JSX entry graph imports, and its root
  package` asserts the subpath is materialized, that its root comes with it, and that an unrelated
  configured share (`vue`) stays lazy. It fails on `main` and passes with this change.
- `pnpm test`: 1194 passed / 0 failed (baseline 1193 / 0 — the difference is the new test).
- `pnpm test:integration`: 94 passed / 0 failed, unchanged from baseline. This includes
  `boots on first load when a remote-only singleton imports jsx-dev-runtime (#1104)`, which passes
  before and after.
- Verified against a minimal reproduction (remote-only container, one shared `react`, JSX entry):
  3/3 warm runs fail on `main` and 3/3 pass with this change, cold runs unaffected.
- Also verified against a larger real remote (multiple exposes, workspace shared packages,
  react-router): same result.

## Affected versions

Bisected on the minimal reproduction: 1.20.6 passes, 1.20.7 through 1.21.3 and current `main` all
fail. Not a 1.21.3 regression.


gioboa (approved):
Nice catch 👏

PR #1221, opened 2026-09-06
fix: harden cyclic share resolution

gioboa:
Handles nameless workspace manifests and transitively blocks unresolved cyclic share dependencies.


gioboa (commented):

Image

PR #1222, opened 2026-09-06
fix: preserve moduleParseTimeout 0

dmchoi77:
## Summary

Fixed an issue where moduleParseTimeout: 0 was normalized to the default value of 10.

- Replaced || with ?? to preserve 0
- Added a regression test for moduleParseTimeout: 0


gioboa (approved):
Great catch 👀 @dmchoi77 👍

PR #1223, opened 2026-09-06
fix: macOS realpath portability in package entry test

dmchoi77:
## Summary

Make the `getInstalledPackageEntry()` unit test compare filesystem paths using `realpathSync()` so it passes consistently on macOS.

## Problem

On macOS, temporary directories are commonly created under `/var/folders`, which is a symlink to `/private/var/folders`. The test expected the lexical `/var/...` path, while `getInstalledPackageEntry()` returned the equivalent filesystem-resolved `/private/var/...` path.

The implementation was correct; the test compared two equivalent paths as raw strings.

## Changes

Updated the copied-package-metadata test to normalize its expected entry with `realpathSync()`.


gioboa (approved):
Thanks @dmchoi77 that's great 👏

PR #1224, opened 2026-09-06
fix: isolate SSR temp cache per process

dmchoi77:
## Summary

Isolate SSR temp files per Node process.

## Root Cause

All processes shared `<cwd>/node_modules/.ssr-cache`, while each process removed the entire directory on exit. One process could therefore delete another process's in-flight or cached SSR modules.

## Changes

- Store temp files under `<cwd>/node_modules/.ssr-cache/<process.pid>`.
- Remove only the current process's cache directory on exit.
- Add a regression test for the process-scoped path and cleanup target.


gioboa (approved):
it looks a solid fix 💯 Thank you

PR #1227, opened 2026-09-06
fix: decide unshared workspace cycle edges from runtime imports instead of manifests

### LINKED ISSUE #1226: [Bug]: unshared workspace importers are glued to the remote's local copy of a shared singleton when the singleton's manifest closure reaches them
### Describe the bug

Since #1210 `getSharedPackageFromFile` also returns the name of an *unshared* workspace package that owns an importing file, so the package-level cycle rule in `proxyPreBuildShared:resolve-shared-loadShare`

```ts
if (importerPackage && isSharedPackageDependency(key, importerPackage)) return;
```

applies to every workspace file. `isSharedPackageDependency` walks the shared package's whole manifest closure (`dependencies` + `peerDependencies` + `optionalDependencies`, transitively). In a monorepo that closure is not the module graph: type-only packages, tooling helpers and page packages all sit in `dependencies`, so on paper a shared singleton "reaches" most of the workspace while its code evaluates none of it. Every unshared workspace importer of that singleton then gets an ordinary edge to the **remote's local fallback** — even when the host provides the singleton and the rest of the remote uses the host copy through the loadShare proxy. Result: two copies of a `singleton: true` share on one page.

**Reproduction** (repo linked below)

```
host   ── shares aaa-runtime (singleton), registers a service in it, loads remote/widget
remote ── shares aaa-runtime (singleton, local fallback), exposes ./widget
             widget ──► zzz-hooks (workspace, NOT shared, bundled into the widget) ──► aaa-runtime
aaa-runtime ──(dependencies)──► mid-types ──(dependencies)──► zzz-hooks     manifest-only: mid-types never imports zzz-hooks
```

`npm install && npm run build && npm run check` on `main` @ `82013b2`:

```
host registered localSettings in aaa-runtime copy 8qgktw
RESULT FAIL: Service "localSettings" is not registered in the container (copy dbciyj)
```

The widget chunk imports `resolve` straight from `lib-runtime-*.js` instead of the `aaa-runtime` loadShare wrapper. Controls: remove the `zzz-hooks` entry from `lib-types/package.json` → `RESULT OK`; `@module-federation/vite@1.21.3` (before #1210) → `RESULT OK`.

**Real-world hit**

Agent Desktop (Bright Pattern, ~290 workspace packages): a calendar remote embedded in the host shell fails with `Service "localSettings" is not registered in the container`. The IoC-container singleton `@bpinc/platform-runtime` is reached by the unshared `@bpinc/lib-formatting-hooks` through an ordinary edge because the singleton's manifest closure covers 175 of 289 workspace packages (`platform-runtime → state-factory → ad-local-settings-types → records-page-ui → lib-formatting-hooks`, all type-only imports at runtime). The same layout worked before #1210.

**Expected behavior**

An import of a shared singleton from an unshared workspace package resolves through the loadShare proxy, exactly like the same import from the remote's own source files, so the remote uses the host-provided copy.

**Proposed fix** (PR follows)

Keep recognizing unshared workspace importers, but judge the cycle by the imports the shared package actually evaluates — its source files and those of the workspace packages they import, `import type` dropped, `node_modules` as leaves — instead of the manifest closure. The genuine evaluation cycle of #1209 still keeps its ordinary edge (rolldown needs the hoisted local binding there); a manifest-only chain stays on the proxy so the remote uses the host's copy.

### Version

Vite 8.2.1; @module-federation/vite main @ 82013b2 (pkg.pr.new) — 1.21.3 is not affected (before #1210)

### Reproduction

https://github.com/marksnode/mf-vite-repro-unshared-importer-glued-to-local-singleton

### Relevant log output

```shell
$ npm run check
host registered localSettings in aaa-runtime copy 8qgktw
RESULT FAIL: Service "localSettings" is not registered in the container (copy dbciyj)

# remote/dist/assets/widget-*.js on main @ 82013b2 — the unshared importer bypasses the loadShare wrapper:
import { r as resolve, t as instanceId } from "./lib-runtime-Cwnb4jIA.js";
```

### Validations

marksnode:
Fixes #1226

**What**

- `proxyPreBuildShared:resolve-shared-loadShare`: an importer that is an *unshared workspace package* (the case #1210 added) now keeps its ordinary edge only when the shared package reaches it through the imports its code evaluates — `isSharedPackageRuntimeDependency` walks the shared package's source files and those of the workspace packages they import (`import type` / `export type` dropped, Node builtins and non-package specifiers ignored, `node_modules` packages treated as leaves, files over 256 KB skipped as bundled artifacts). Shared and `node_modules` importers keep the manifest-based `isSharedPackageDependency` as before.
- The eager workspace singleton wrapper is hardened for the merged-chunk case: `__mfApplyEagerShareExportsWhenReady` skips the synchronous copy while `__mfLocalShare` is `undefined` or in TDZ, and the deferred cache write (`__mfInitializedLocalShare`) never publishes an uninitialized namespace — writing `undefined` fanned out to every subscriber as `TypeError: Cannot read properties of undefined (reading '…')`.

**Why**

`isSharedPackageDependency` walks the shared package's manifest closure. In a monorepo that closure is not the module graph: type-only packages, page packages and tooling all sit in `dependencies`, so a singleton "reaches" most of the workspace on paper (175 of 289 packages in our host) while its code evaluates a handful. Since #1210 every unshared workspace importer of such a singleton got an ordinary edge to the **remote's local fallback**, so a remote embedded in a host that provides the singleton ran two copies of it (`Service "localSettings" is not registered in the container`). The ordinary edge is still required for genuine evaluation cycles (#1209: a merged chunk places the wrapper above its fallback, and module-scope calls need the hoisted local binding), so the cycle test has to look at what the fallback actually evaluates.

**Tests**

- `pluginProxySharedModule_preBuild.test.ts`: the #1210/#1221 cases now provide the shared package's source (`import 'bridge'`) and still expect the ordinary edge; new cases cover a manifest-only chain (`vue → types-only → hooks`, proxied) and `getRuntimeImportSpecifiers` (type-only statements, builtins and minified string fragments dropped).
- `virtualShared_preBuild.test.ts`: two runtime tests evaluate the generated eager wrapper with `__mfLocalShare` undefined; on `main` they fail with the #1209 signature.
- `vitest run --dir src`: 1246 passed (the one failing `packageUtils` case fails on clean `main` on macOS too); `vitest run integration`: 93/94 (same pre-existing failure as `main`, `dev-singleton-react` green); `oxfmt --check`, `tsc` clean.
- Repro https://github.com/marksnode/mf-vite-repro-unshared-importer-glued-to-local-singleton: `RESULT OK`, the widget chunk imports through the `aaa-runtime` wrapper. The #1209 repro still prints `directly from the aaa-ui fallback` and boots. Agent Desktop (315 shares, 8 apps): full layout builds and boots; the embedded calendar reads the host container again.

**Note** — while wiring the scan I hit an unrelated hang: `getInstalledPackageJson('stream')` (a Node builtin name) loops forever because `require.resolve` returns the bare name and the `path.dirname` walk never reaches a root. The scanner filters builtins so it cannot trigger it; the loop itself is left for a separate change.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -13,6 +13,8 @@ const {
   hasPackageDependencyMock,
   existsSyncMock,
   readFileSyncMock,
+  readdirSyncMock,
+  statSyncMock,
   addUsedSharesMock,
   refreshHostAutoInitMock,
   writeLoadShareModuleMock,
@@ -26,6 +28,8 @@ const {
   hasPackageDependencyMock: vi.fn<(pkg: string, cwd?: string) => boolean>(),
   existsSyncMock: vi.fn<(path: string) => boolean>(() => false),
   readFileSyncMock: vi.fn<(path: string) => string>(() => '{}'),
+  readdirSyncMock: vi.fn<(dir: string, opts?: unknown) => unknown[]>(() => []),
+  statSyncMock: vi.fn<(file: string) => { size: number }>(() => ({ size: 128 })),
   addUsedSharesMock: vi.fn<(pkg: string) => void>(),
   refreshHostAutoInitMock: vi.fn<() => void>(),
   writeLoadShareModuleMock: vi.fn(),
@@ -49,6 +53,8 @@ vi.mock('fs', async (importOriginal) => {
     ...actual,
     existsSync: existsSyncMock,
     readFileSync: readFileSyncMock,
+    readdirSync: readdirSyncMock,
+    statSync: statSyncMock,
   };
 });
 
@@ -164,8 +170,11 @@ vi.mock('../../utils/VirtualModule', () => ({
 import {
   excludeSharedSubDependencies,
   findSharedKey,
+  getRuntimeImportSpecifiers,
   proxySharedModule,
 } from '../pluginProxySharedModule_preBuild';
+
+const sourceFile = (name: string) => ({ name, isDirectory: () => false, isFile: () => true });
 import {
   getResolvedLocalSharedImportMapId,
   getUsedShares,
@@ -1229,6 +1238,7 @@ describe('pluginProxySharedModule_preBuild', () => {
     expect(writeLoadShareModuleMock).toHaveBeenCalled();
     existsSyncMock.mockReset().mockReturnValue(false);
     readFileSyncMock.mockReset().mockReturnValue('{}');
+    readdirSyncMock.mockReset().mockReturnValue([]);
   });
 
   it('does not proxy shared dependency cycle edges from workspace packages', async () => {
@@ -1269,6 +1279,7 @@ describe('pluginProxySharedModule_preBuild', () => {
     expect(writeLoadShareModuleMock).not.toHaveBeenCalled();
     existsSyncMock.mockReset().mockReturnValue(false);
     readFileSyncMock.mockReset().mockReturnValue('{}');
+    readdirSyncMock.mockReset().mockReturnValue([]);
   });
 
   it('does not proxy shared dependency cycle edges from unshared workspace packages', async () => {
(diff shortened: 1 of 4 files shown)


gioboa (approved):
That's a great one 👏 Thanks @marksnode

PR #1228, opened 2026-09-06
fix: stop the installed-package lookup from looping forever on Node c…

### LINKED ISSUE #1225: 1.21.3: build hangs forever — infinite loop in resolveInstalledPackageJson() when a shared package depends on a Node core module name
## Summary

`@module-federation/vite@1.21.3` makes `vite build` hang forever (100 % CPU, no output after `transforming...`, event loop fully blocked so `Ctrl+C` and timers no longer fire).

Root cause is an **infinite loop in `resolveInstalledPackageJson()`** (`src/utils/packageUtils.ts`) that is newly *reachable* in 1.21.3 through the new `isSharedPackageDependency()` (`src/plugins/pluginProxySharedModule_preBuild.ts`). It triggers when the transitive dependency graph of a shared package contains a package **named after a Node core module** (`readline`, `util`, `punycode`, `events`, `process`, `buffer`, …) that is not installed as a real package.

1.21.2 builds the same project in 9 s; 1.21.3 never finishes.
Still present on `main` (checked at 82013b2) — see "Why it is still reachable on main" below.

## Minimal reproduction

Verified end to end. Hangs on `1.21.3`, builds in ~0.8 s on `1.21.2`.

```bash
mkdir mf-hang-repro && cd mf-hang-repro
echo '{ "name": "mf-hang-repro", "private": true, "type": "module", "version": "1.0.0" }' > package.json
npm i vite@8.2.2 @module-federation/vite@1.21.3

# create the two fake packages AFTER npm install (npm prunes them otherwise)
mkdir -p node_modules/pkg-a node_modules/pkg-b src

# pkg-a is the shared package; its manifest lists a dependency whose name is a Node core module
cat > node_modules/pkg-a/package.json <<'EOF'
{ "name": "pkg-a", "version": "1.0.0", "type": "module", "main": "index.js",
  "dependencies": { "readline": "^1.3.0" } }
EOF
echo 'export const a = 1;' > node_modules/pkg-a/index.js

# pkg-b is an importer inside node_modules, so `importerPackage` is set in the resolveId hook
cat > node_modules/pkg-b/package.json <<'EOF'
{ "name": "pkg-b", "version": "1.0.0", "type": "module", "main": "index.js",
  "dependencies": { "pkg-a": "^1.0.0" } }
EOF
echo 'import { a } from "pkg-a"; export const b = a + 1;' > node_modules/pkg-b/index.js

echo '<!doctype html><html><body><script type="module" src="/src/main.js"></script></body></html>' > index.html
echo 'import e from "./e.js"; console.log(e);' > src/main.js
echo 'import { b } from "pkg-b"; export default b;' > src/e.js

cat > vite.config.mjs <<'EOF'

export default {
  plugins: [
    federation({
      name: 'reproRemote',
      filename: 'remoteEntry.js',
      exposes: { './E': './src/e.js' },
      remotes: {},
      shared: { 'pkg-a': {} },
      dts: false,
    }),
  ],
  build: { outDir: 'build' },
};
EOF

npx vite build     # 1.21.3: hangs at "transforming..." forever
```

Then `npm i @module-federation/vite@1.21.2`, re-create `node_modules/pkg-a` and `node_modules/pkg-b`, and `npx vite build` completes in ~0.8 s.

Note that `readline` does **not** have to be installed — only the *name* in the manifest matters.

### The infinite loop

`src/utils/packageUtils.ts`, `resolveInstalledPackageJson()`:

```ts
let currentDir = path.dirname(resolvedPath);
const rootDir = path.parse(currentDir).root;
let matchingPackage: InstalledPackageJson | undefined;

while (true) {
  const packageJsonPath = path.join(currentDir, 'package.json');
  if (existsSync(packageJsonPath)) { /* readFileSync + JSON.parse */ }
  if (currentDir === rootDir) break;
  currentDir = path.dirname(currentDir);
}
```

For a Node core module, `require.resolve()` does not return a path — it returns the bare specifier:

```js
require.resolve('readline')        // 'readline'      (not a path!)
path.dirname('readline')           // '.'
path.parse('.').root               // ''
path.dirname('.')                  // '.'
// => currentDir stays '.', rootDir is '', `currentDir === rootDir` is never true
```

The loop spins forever, doing `existsSync('./package.json')` + `readFileSync` + `JSON.parse` on every iteration. Instrumented run: 122 000 iterations after 57 s and still climbing. This is platform independent (`path.posix` behaves identically); observed on Windows 11 with Node 22.23.2.

CPU profile taken over the inspector protocol while hanging (the event loop is blocked, so `--cpu-prof` cannot be flushed):

```
48.4 %  readFileUtf8
37.7 %  existsSync
 8.5 %  resolveInstalledPackageJson   packageUtils.ts

stack:
  readFileSync
  resolveInstalledPackageJson
  getInstalledPackageJson
  isSharedPackageDependency          pluginProxySharedModule_preBuild.ts
  resolveId  "proxyPreBuildShared:resolve-shared-loadShare"
```

### Why it became reachable in 1.21.3

`isSharedPackageDependency()` is new in 1.21.3 (it does not exist anywhere in the 1.21.2 bundle). It BFS-walks the **entire transitive `dependencies` / `peerDependencies` / `optionalDependencies` graph** of every shared package and calls `getInstalledPackageJson()` for every name it finds — including names that are not resolvable packages.

Measured on our real project (`@iobroker/gui-components` shared):

| | calls to `resolveInstalledPackageJson` | build |
|---|---|---|
| 1.21.2 | 9 (only the 8 shared packages) | ✓ 9 s |
| 1.21.3 | 194, then hangs | ✗ never finishes |

The chain there is: shared `@iobroker/gui-components` → `@iobroker/js-controller-common-db` → its manifest lists `readline` → boom. Any dependency named after a core module anywhere in a shared package's transitive manifest graph is enough.

### Why it is still reachable on `main`

`getDependencyManifest()` (added after 1.21.3) walks `node_modules/<dep>/package.json` upwards with a correct termination guard, but ends with:

```ts
found ??= getInstalledPackageJson(dep, { cwd: fromDir, packageName: dep });
```

For `readline` no `node_modules/readline/package.json` exists at any level, so it falls through to `getInstalledPackageJson()` and hits the same infinite loop.

### Side note: same function, same release

1.21.3 also replaced the early `return` on the first name match with

```ts
if (currentDir.endsWith(path.join('node_modules', packageName))) return packageInfo;
matchingPackage ??= packageInfo;
```

so every match that is *not* located at `node_modules/<name>` now keeps walking to the filesystem root instead of returning immediately. Not the cause of the hang, but it multiplies the fs cost of every lookup.

## Suggested fix

1. **Bail out when `require.resolve()` did not return a path** (core modules) in `resolveInstalledPackageJson()`:

```ts
import { isBuiltin } from 'node:module';
// ...
if (isBuiltin(pkg) || isBuiltin(packageName)) return undefined;
// or, defensively, after resolving:
let currentDir = path.dirname(resolvedPath);
if (!path.isAbsolute(currentDir)) return undefined;
```

2. **Make the upward walks structurally terminating** — all three `while (true)` loops in `resolveInstalledPackageJson()` should use the same guard `getDependencyManifest()` already uses, instead of comparing against `path.parse().root`:

```ts
const parentDir = path.dirname(currentDir);
if (parentDir === currentDir) break;
currentDir = parentDir;
```

(2) alone turns the hang into a fast no-op; (1) additionally avoids the pointless fs traffic.

## Environment

- `@module-federation/vite` 1.21.3 (regression vs 1.21.2)
- `vite` 8.2.2 (rolldown)
- Node 22.23.2
- Windows 11 (path logic is platform independent)

marksnode:
Fixes #1225

## What

`resolveInstalledPackageJson()` walked up from `path.dirname(require.resolve(pkg))` and terminated on `currentDir === path.parse(currentDir).root`. For a Node core module `require.resolve()` returns the bare specifier (`readline`, `node:readline`), so `path.dirname()` yields `'.'`, whose root is `''` — the loop never terminates and the build hangs with no output.

- When the resolved path is not absolute, the lookup now skips the walk-up and falls through to the `node_modules` / pnpm-store lookup (an npm package can legitimately shadow a core module name — `util`, `buffer`, `events`, `punycode` are all real packages — and `require.resolve()` prefers the core module, so it has to be found on disk instead).
- All three upward walks terminate structurally (`path.dirname(dir) === dir`) instead of comparing against the filesystem root, so no other odd input can loop them either.

## Why

Since 1.21.3, `isSharedPackageDependency()` walks a shared package's whole transitive manifest graph and asks `getInstalledPackageJson()` about every dependency the `node_modules` walk-up misses. A dependency named after a core module that is not installed (e.g. a peer `buffer` or a `readline` optional dependency) is enough to hang the build — #1225 has the numbers and a minimal repro.

## Tests

- `returns undefined for a Node core module instead of walking up from its bare specifier` — `readline` and `node:readline` return `undefined` promptly (on `main` this test never finishes).
- `finds an installed package that shares its name with a Node core module` — a `node_modules/util` package is found even though `require.resolve('util')` resolves to the core module.

`pnpm test`: 1243 passed, 1 skipped, 1 failed (`ignores copied package metadata beside the resolved entry` fails the same way on a clean `main` on macOS). `pnpm fmt.check`, `pnpm typecheck` clean.

### DIFF
--- a/src/utils/__tests__/packageUtils.test.ts
+++ b/src/utils/__tests__/packageUtils.test.ts
@@ -55,6 +55,42 @@ describe('getInstalledPackageJson', () => {
     );
   });
 
+  it('returns undefined for a Node core module instead of walking up from its bare specifier', () => {
+    const root = mkdtempSync(path.join(tmpdir(), 'mf-vite-core-module-'));
+    tempDirs.push(root);
+
+    const hostDir = path.join(root, 'apps/host');
+    mkdirSync(hostDir, { recursive: true });
+    writeFileSync(path.join(hostDir, 'package.json'), JSON.stringify({ name: 'host' }));
+
+    // `require.resolve('readline')` returns 'readline' itself, not a file path
+    expect(getInstalledPackageJson('readline', { cwd: hostDir })).toBeUndefined();
+    expect(getInstalledPackageJson('node:readline', { cwd: hostDir })).toBeUndefined();
+  });
+
+  it('finds an installed package that shares its name with a Node core module', () => {
+    const root = mkdtempSync(path.join(tmpdir(), 'mf-vite-core-module-shadow-'));
+    tempDirs.push(root);
+
+    const hostDir = path.join(root, 'apps/host');
+    const packageDir = path.join(hostDir, 'node_modules', 'util');
+    mkdirSync(packageDir, { recursive: true });
+    writeFileSync(path.join(hostDir, 'package.json'), JSON.stringify({ name: 'host' }));
+    writeFileSync(
+      path.join(packageDir, 'package.json'),
+      JSON.stringify({ name: 'util', version: '0.12.5', main: './util.js' })
+    );
+    writeFileSync(path.join(packageDir, 'util.js'), 'module.exports = {};');
+
+    // `require.resolve('util')` still prefers the core module, so the lookup must fall back to node_modules
+    const installed = getInstalledPackageJson('util', { cwd: hostDir });
+
+    expect(installed?.packageJson.version).toBe('0.12.5');
+    expect(normalizePathForImport(installed?.path || '')).toContain(
+      '/apps/host/node_modules/util/package.json'
+    );
+  });
+
   it('caches lookups per (cwd, pkg) instead of re-reading the filesystem every call', () => {
     const packageName = 'mf-test-cache-pkg';
     const root = mkdtempSync(path.join(tmpdir(), 'mf-vite-cache-'));
--- a/src/utils/packageUtils.ts
+++ b/src/utils/packageUtils.ts
@@ -443,7 +443,6 @@ function resolveInstalledPackageJson(
   };
   const findPackageInPnpmStore = (startDir: string): InstalledPackageJson | undefined => {
     let currentDir = startDir;
-    const rootDir = path.parse(currentDir).root;
 
     while (true) {
       const pnpmStoreDir = path.join(currentDir, 'node_modules', '.pnpm');
@@ -458,15 +457,15 @@ function resolveInstalledPackageJson(
           }
         } catch {}
       }
-      if (currentDir === rootDir) break;
-      currentDir = path.dirname(currentDir);
+      const parentDir = path.dirname(currentDir);
+      if (parentDir === currentDir) break;
+      currentDir = parentDir;
     }
   };
 
+  let resolvedPath: string | undefined;
   try {
     const projectRequire = createRequire(pathToFileURL(path.join(cwd, 'package.json')));
-    let resolvedPath: string | undefined;
-
     if (opts?.fromResolvedEntry) {
       resolvedPath = opts.fromResolvedEntry;
     } else {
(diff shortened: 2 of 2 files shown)


gioboa (approved):
Thanks @marksnode nice fix 👏

PR #1230, opened 2026-09-06
fix: skip another container's consume-only stub when bridging a shared provider

### LINKED ISSUE #1229: [Bug]: provider init logs "Failed to bridge external shared module … must be provided by host" when a host's `import: false` stub is the same-version scope entry
### Describe the bug

Three parts on one page: a Vite **host** that consumes `react` / `react-dom` with `import: false` (it bundles no copy of its own), a Vite **provider** remote that really provides both (`singleton: true`), and a webpack remote that uses them. The webpack widget renders fine in both load orders, but every provider init logs, deterministically:

```
[Module Federation] Failed to bridge external shared module "react-dom" Error: [Module Federation] Shared module 'react-dom' must be provided by host
    at Object.get (…/_virtual_mf-localSharedImportMap___mfe_internal__host__mf_owner__1-….js)
```

What happens:

1. The host registers first, so `scope.default['react-dom']['18.3.1']` is the host's consume-only stub — a scope entry whose `get()` throws `must be provided by host` by construction (`shareConfig.import === false`).
2. The provider's init runs `__mfBridgeExternalSharedProvider('react-dom', …)`. `__mfSelectExternalSharedProvider` picks that stub as the same-version provider (first registrant wins the tie), `__mfLoadPinnedRuntimeShare` calls its `get()`, and the throw lands in the `catch` → `console.error`.
3. The only guard on that path checks whether the stub is the **container's own** import:false share (`usedShare.shareConfig?.import === false && __mfMatchesSharedProvider(provider, usedShare)`); a stub owned by *another* container passes through. The same gap is in `__mfBridgeMaterializedProvider`.

It is only noise (the host's `__mfResolveImportFalseShared` later bridges the stub onto the provider's real module), but it is a `console.error` on every page load for the most ordinary consume-only host + provider topology, and it looks exactly like the real failures from #952 / #1064 / #1176.

**Expected:** the bridge skips a foreign consume-only stub the same way it already skips its own — nothing to bridge onto, nothing to log.

**Steps** (repro linked below):

```
npm install
npm run build
npm run check     # headless, both load orders: exits 1 and prints RESULT FAIL while the console carries "Failed to bridge"
npm run preview   # http://localhost:4180/  and  ?order=wremote-first
```

`RESULT FAIL` on 1.21.3 and on main @ 82013b2; `RESULT OK` with the one-line guard `if (provider?.shareConfig?.import === false) return;` in both bridge passes (PR follows).

### Version

1.21.3 (also main @ 82013b2)

### Reproduction

https://github.com/marksnode/mf-vite-repro-foreign-consume-only-stub-bridge

### Relevant log output

```shell
$ npm run check

=== http://localhost:4180/
page: webpack widget, clicks: 1
error: [Module Federation] Failed to bridge external shared module "react-dom" Error: [Module Federation] Shared module 'react-dom' must be provided by host

=== http://localhost:4180/?order=wremote-first
page: webpack widget, clicks: 1
error: [Module Federation] Failed to bridge external shared module "react-dom" Error: [Module Federation] Shared module 'react-dom' must be provided by host

RESULT FAIL: provider init logged "Failed to bridge"
```

### Validations

marksnode:
Fixes #1229

## What

Both bridge passes in the generated remote entry — `__mfBridgeMaterializedProvider` and `__mfBridgeExternalSharedProvider` — now return before touching a selected provider whose `shareConfig.import === false`. Such a scope entry is another container's consume-only stub: its `get()` throws `must be provided by host` by construction, so there is nothing to bridge onto and the attempt can only end in the `catch` → `console.error('Failed to bridge external shared module …')`.

The guard sits next to the existing one for the container's **own** stub (`usedShare.shareConfig?.import === false && __mfMatchesSharedProvider(provider, usedShare)`), which is the same rule restricted to one owner; `__mfResolveImportFalseShared` already treats any `shareConfig.import === false` entry as a stub (`__mfIsOwnStub`).

## Why

A Vite host that consumes `react` / `react-dom` with `import: false` registers first, so the same-version scope entry is its stub. When a Vite provider remote that really provides both initialises, `__mfSelectExternalSharedProvider` picks that stub (first registrant wins the tie), `__mfLoadPinnedRuntimeShare` calls its `get()`, and every page load logs

```
[Module Federation] Failed to bridge external shared module "react-dom" Error: [Module Federation] Shared module 'react-dom' must be provided by host
```

The app keeps working — the host's `__mfResolveImportFalseShared` later bridges the stub onto the provider's module — but it is a deterministic `console.error` for the most ordinary consume-only host + provider topology, indistinguishable at a glance from the real failures in #952 / #1064 / #1176. Details and a three-part reproduction (Vite host with `import: false`, Vite provider, webpack consumer) in #1229.

## Tests

- `never bridges onto another container's consume-only stub` (virtualRemoteEntry.test.ts): both generated bridge functions carry the guard, ahead of the provider-entry version read and of `__mfLoadPinnedRuntimeShare` respectively.
- Reproduction (https://github.com/marksnode/mf-vite-repro-foreign-consume-only-stub-bridge): `npm run check` prints `RESULT FAIL` (bridge error in the console) on 1.21.3 and on `main` @ 82013b2, `RESULT OK` with this branch packed and installed; the webpack widget renders in both load orders throughout.

`pnpm test`: 1242 passed, 1 skipped, 1 failed (`ignores copied package metadata beside the resolved entry` fails the same way on a clean `main` on macOS). `pnpm fmt.check`, `pnpm typecheck` clean.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -2900,6 +2900,48 @@ describe('virtualRemoteEntry', () => {
     expect(materializedBridgeCode).toContain('if (usedShare.canLiveRebind === false) return;');
   });
 
+  it("never bridges onto another container's consume-only stub", async () => {
+    optionsMock.shareStrategy = 'loaded-first';
+    const mod = await import('../virtualRemoteEntry');
+
+    const code = mod.generateRemoteEntry(
+      {
+        internalName: '__mfe_internal__provider',
+        name: 'provider',
+        filename: 'remoteEntry.js',
+        exposes: {},
+        remotes: {},
+        shared: {},
+        runtimePlugins: [],
+        shareScope: 'default',
+        shareStrategy: 'loaded-first',
+      } as any,
+      'virtual:exposes',
+      'serve'
+    );
+
+    // A host's `import: false` share registers a scope entry whose get() throws "must be provided by
+    // host". Selected as the same-version provider (first registrant wins), it must be skipped before
+    // its get() runs, in both bridge passes — otherwise every provider init logs "Failed to bridge".
+    const guard = 'if (provider?.shareConfig?.import === false) return;';
+    const materializedBridgeCode = code.slice(
+      code.indexOf('const __mfBridgeMaterializedProvider ='),
+      code.indexOf('const __mfBridgeExternalSharedProvider =')
+    );
+    const externalBridgeCode = code.slice(
+      code.indexOf('const __mfBridgeExternalSharedProvider ='),
+      code.indexOf('for (const batch of __mfMaterializedShareBatches) await Promise.all(')
+    );
+    expect(materializedBridgeCode).toContain(guard);
+    expect(materializedBridgeCode.indexOf(guard)).toBeLessThan(
+      materializedBridgeCode.indexOf('const { version } = providerEntry;')
+    );
+    expect(externalBridgeCode).toContain(guard);
+    expect(externalBridgeCode.indexOf(guard)).toBeLessThan(
+      externalBridgeCode.indexOf('await __mfLoadPinnedRuntimeShare(')
+    );
+  });
+
   it('keeps the remote registry intact while pre-seeding version-first shares', async () => {
     optionsMock.shareStrategy = 'version-first';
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Good fix 🥇  Thanks @marksnode

PR #1233, opened 2026-09-07
fix: follow directory imports to their index file when walking a shared package's runtime imports

### LINKED ISSUE #1232: [Bug]: 1.21.4: a shared package whose entry re-exports a directory loses its runtime-import walk — unshared workspace cycle edges become loadShare glue again (#1209 crash is back)
### Describe the bug

Since #1227 (1.21.4) the shared-cycle guard decides an **unshared workspace importer's** edge from the shared package's runtime imports. The follow-up commit in that PR (409e91a) walks them from the package **entry**, following relative imports with `resolveLocalRuntimeImport`. Its candidate list starts with the bare resolved path (`${resolved}` + `''`), and `existsSync` accepts a **directory**:

```js
// packages/cases-ui/index.ts
export * from './src/components/Case'      // -> a directory; its index.ts is the real module
```

`resolveLocalRuntimeImport(entry, './src/components/Case')` returns the directory, `readFileSync` throws `EISDIR`, the walk `continue`s and drops the branch. Because the entry itself *was* read (`scanned = true`), the whole-directory fallback never runs either. Measured on our monorepo: `collectReachableRuntimeImports` from `cases-ui/index.ts` reaches **0** bare specifiers; `isSharedPackageRuntimeDependency('@bpinc/lib-ad-ui-kit', '@bpinc/activities-ui')` is `false` where it was `true` on the #1227 branch.

Consequence: every unshared workspace importer inside a real module-level cycle is rewritten to loadShare glue again, the glue lands above its fallback in the merged chunk, and the host dies at boot with the #1209 crash (`TypeError: dz is not a function` — `withErrorBoundary(memo(...))` at module level reading an unassigned rebinding). `export * from './dir'` is the default barrel layout of a TS monorepo, so this hits any such host on 1.21.4; 1.21.3 is unaffected (no runtime-import walk).

**Repro** (linked below) is the #1209 graph with one change — `aaa-ui/index.js` is `export * from './components'`:

```
npm install && npm run build && npm run check
```

prints `RESULT FAIL: cycle edge became loadShare glue` on 1.21.4 (the same graph with a flat entry prints OK).

**Fix:** `resolveLocalRuntimeImport` must return only files (`statSync(candidate).isFile()`), so `./components` resolves to `components/index.js`. With it the repro prints `directly from the aaa-ui fallback (cycle guard kept the edge)` and the monorepo host boots. PR follows.

### Version

1.21.4 (main @ 485ffc7)

### Reproduction

https://github.com/marksnode/mf-vite-repro-reachable-imports-directory-index

### Relevant log output

```shell
$ npm run check   # 1.21.4
chunk _virtual_mf___mfe_internal__host__mf_owner__193216886484743__loadShare__aaa_mf_2_ui__loadShare__.js-CuzevA-R.js
  module order: FALLBACK(lib-bridge) -> FALLBACK(lib-ui) -> GLUE(ui)
  zzz-bridge reads BUTTON_KIND: through the aaa-ui loadShare GLUE (cycle guard skipped)  <<<
  bridge body: var bridgeTitle = () => `bridge(${__mf_0})`;
RESULT FAIL: cycle edge became loadShare glue

# In a 300-package monorepo host (every packages/*/index.ts is `export * from './src/components/Foo'`), at boot:
Uncaught (in promise) TypeError: dz is not a function
    at Activity.tsx:24:44        # withErrorBoundary(memo(...)) at module level, read through the loadShare glue above its fallback
```

### Validations

marksnode:
Fixes #1232

## What

`resolveLocalRuntimeImport` (the relative-import step of the runtime-import walk added in #1227 / 409e91a) now accepts a candidate only when `statSync(candidate).isFile()`. Its first candidate is the bare resolved path, and for `./components` that is a directory that exists: `existsSync` accepted it, `readFileSync` threw `EISDIR`, the walk dropped the branch — and since the entry itself had been read, the whole-directory fallback never ran. With the file check the directory is skipped and `components/index.{js,ts,…}` is taken, as intended.

## Why

`export * from './src/components/Foo'` is the default barrel layout of a TypeScript monorepo. On 1.21.4 every such shared package reaches **zero** runtime imports, so every unshared workspace importer inside a real module-level cycle is rewritten back to loadShare glue, the glue evaluates above its fallback in the merged chunk, and the host dies at boot with the #1209 crash (`TypeError: dz is not a function`). Measured on our host: `isSharedPackageRuntimeDependency('@bpinc/lib-ad-ui-kit', '@bpinc/activities-ui')` flipped from `true` (the #1227 branch) to `false` (1.21.4). Details and the repro in #1232.

## Tests

- `follows a directory import to its index file when walking the shared package runtime imports` — vue's entry is `export * from './components'`, the directory exists, only its index imports `bridge`; the bridge → vue edge must stay ordinary. Fails on 1.21.4, passes with the fix. The default `statSync` mock now also reports `isFile()`.
- Repro https://github.com/marksnode/mf-vite-repro-reachable-imports-directory-index: `npm run check` prints `RESULT FAIL` on 1.21.4 and `RESULT OK` with this branch packed and installed; our 300-package host boots again on the packed build.

`pnpm test`: 1252 passed, 1 skipped. `pnpm fmt.check`, `pnpm typecheck` clean.

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -29,7 +29,10 @@ const {
   existsSyncMock: vi.fn<(path: string) => boolean>(() => false),
   readFileSyncMock: vi.fn<(path: string) => string>(() => '{}'),
   readdirSyncMock: vi.fn<(dir: string, opts?: unknown) => unknown[]>(() => []),
-  statSyncMock: vi.fn<(file: string) => { size: number }>(() => ({ size: 128 })),
+  statSyncMock: vi.fn<(file: string) => { size: number; isFile: () => boolean }>(() => ({
+    size: 128,
+    isFile: () => true,
+  })),
   addUsedSharesMock: vi.fn<(pkg: string) => void>(),
   refreshHostAutoInitMock: vi.fn<() => void>(),
   writeLoadShareModuleMock: vi.fn(),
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks for this fix @marksnode

PR #1234, opened 2026-09-07
fix: use portable pnpm filter syntax

dmchoi77:
## Summary

- This is a very very small change that replaces `pnpm -filter` with the documented `pnpm --filter` syntax in all 10 affected root scripts.
- It covers the dev, preview, external-runtime preview, mixed-mode, and multi-example CI scripts so they work in pnpm environments with strict argument parsing.


gioboa (approved):
Thanks @dmchoi77

PR #1236, opened 2026-09-07
fix: include wildcard shared subpaths in the pending preload list

dmchoi77:
## Summary

Fix concrete subpaths configured through wildcard sharing, such as `shared: { 'react/': ... }`, being omitted from the build bootstrap's pending preload list.

This regression was introduced by [PR #1208](https://github.com/module-federation/vite/pull/1208), which added pending-share preloading but looked up only exact shared keys.

## Changes

- Use `getShareItemForPreload()` so pending-share generation handles both exact and wildcard keys.
- Add a regression test for the `react/` configuration with `react/jsx-runtime`.

## Validation

- `virtualRemoteEntry` tests: 108 passed
- Full test suite: 1253 passed, 1 skipped
- `pnpm typecheck`: passed

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1013,6 +1013,26 @@ describe('virtualRemoteEntry', () => {
     expect(hostInit).not.toContain('preloadPendingShares');
   });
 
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Great help as usual @dmchoi77 🫶

PR #1237, opened 2026-09-07
fix(ssr): preserve temp-file module identity in cyclic ESM graphs

dmchoi77:
## Summary

Make the SSR `temp-file` loader use the same `?v=<versionKey>` query for the root and nested imports. Previously, a cyclic import could make Node evaluate the root as two ESM modules and split the container state.

## Changes

- Added a helper for temporary-file import URLs
- Applied the version query to recursively rewritten HTTP ESM imports
- Strengthened the circular-graph regression test

## Validation

- Unit: 50 files, 1252 passed, 1 skipped
- Integration: 16 files, 94 passed
- `pnpm typecheck`, `pnpm build`, and `pnpm fmt.check`: passed
- Real HTTP reproduction: root evaluation changed from 2 to 1, with the same module instance observed


gioboa (approved):
Nice fix 👏

PR #1238, opened 2026-09-07
fix: preserve shared cycles when runtime dependency scanning is incomplete

dmchoi77:
## Summary

Fix the runtime dependency scanner from incorrectly treating a large shared entry as having no dependency and inserting a `loadShare` proxy inside a cycle.

This regression was introduced by [PR #1227](https://github.com/module-federation/vite/pull/1227), which added runtime-import-based cycle detection. Files larger than 256KB were skipped by both the reachable scan and its fallback, so actual runtime dependencies could be missed.

## Changes

- Track whether runtime import scanning completed successfully.
- Preserve the ordinary edge when an entry or reachable local module cannot be analyzed instead of assuming the dependency is absent.
- Add a regression test for a 256KB+ shared entry importing an unshared workspace package in a cycle.

## Validation

- Related plugin tests: 79 passed
- Full test suite: 1253 passed, 1 skipped
- `pnpm typecheck`: passed

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -1484,6 +1484,59 @@ describe('pluginProxySharedModule_preBuild', () => {
     readdirSyncMock.mockReset().mockReturnValue([]);
   });
 
+  it('keeps ordinary edges when a shared entry exceeds the runtime scan limit', async () => {
+    normalizeModuleFederationOptions({ name: 'host', shared: {} });
+    hasPackageDependencyMock.mockReturnValue(false);
+    getInstalledPackageEntryMock.mockImplementation((pkg) =>
+      pkg === 'vue' ? '/repo/apps/remote/node_modules/vue/index.js' : undefined
+    );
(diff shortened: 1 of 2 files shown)


gioboa (changes requested):
This fallback seems too broad:

```ts
return !complete || reachable.has(dependency);
```

Once any reachable file exceeds the scan limit or cannot be read, every unshared workspace importer is treated as part of the cycle. Unrelated imports then bypass `loadShare` and bind to the remote fallback—potentially restoring the duplicate-singleton bug fixed by #1227/#1226.

Could we instead make the reachable scan complete by scanning oversized entry files, or use a narrower fallback for unknown results? Please also add a regression test where scanning is incomplete but an unrelated workspace package imports the share; that import should still use the proxy.

dmchoi77:
@gioboa Thanks for catching this! 
I narrowed the fallback so incomplete scans no longer classify every workspace importer as a cycle. Large package entries are scanned directly, while oversized nested modules remain bounded. Added a regression test for unrelated workspace imports and verified the full test suite.

gioboa (approved):
Nice one 🚀 Thank you @dmchoi77

PR #1239, opened 2026-09-07
fix: respect export conditions during runtime dependency scans

dmchoi77:
## Summary

This regression originated in the runtime-import-based cycle edge detection introduced by [#1227](https://github.com/module-federation/vite/pull/1227). Make shared package runtime dependency scans use the same conditions as Vite's client and SSR export resolution to prevent conditional exports and wildcard shared subpaths from misclassifying dependency cycles and inserting an incorrect `loadShare` proxy.

[#1233](https://github.com/module-federation/vite/pull/1233) was a follow-up that improved directory-import/index handling in the same scanner; the regression was originally introduced by #1227.

## Problem

- The scanner introduced by #1227 did not use Vite's actual client/SSR resolver conditions, so it could inspect a different entry and import graph from the one that would execute for a conditional export.
- A scan calculated for the client could be reused during SSR, mixing runtime dependency graphs across environments.
- For wildcard shared keys such as `react/`, the scanner received the configured key prefix instead of the actual import, such as `react/jsx-runtime`, and could miss the subpath entry.

This could incorrectly remove the cycle edge between an unshared workspace package and a shared package, causing an ordinary import to be handled through a `loadShare` proxy.

## Fix

- Pass Vite's client/SSR export conditions into runtime dependency scans and subpath resolution.
- Include conditions in the cache key so client and SSR scan results remain isolated.
- For wildcard shared entries, scan the actual imported subpath instead of the configured `pkg/` prefix and disable the `require` fallback for that subpath lookup.

## Review focus

- Verify that runtime scan conditions match the `resolveId` environment, including the SSR target and user-provided `resolve.conditions`.
- Verify that scan results from different conditions cannot be reused from the same cache entry.
- Verify both ordinary and wildcard shared keys, including directory/index resolution and unresolved-entry fallback.
- Verify that `loadShare` is not injected when the runtime graph requires the ordinary cycle edge to be preserved.

## Changes

- Separate scan caches by client/SSR conditions and pass export conditions to subpath resolution.
- Scan the actual import subpath for wildcard shared keys instead of the configured `pkg/` prefix.
- Add regression coverage for SSR conditions and wildcard subpaths.

## Validation

- `pnpm build`
- `pnpm test` (50 files, 1255 passed, 1 skipped)
- `pnpm test:integration` (16 files, 94 passed)
- `pnpm e2e` (27 passed)
- `pnpm typecheck`
- `pnpm fmt.check`

### DIFF
--- a/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
+++ b/src/plugins/__tests__/pluginProxySharedModule_preBuild.test.ts
@@ -38,7 +38,17 @@ const {
   writeLoadShareModuleMock: vi.fn(),
   writeLocalSharedImportMapMock: vi.fn<() => void>(),
   writePreBuildLibPathMock: vi.fn(),
-  getInstalledPackageEntryMock: vi.fn<(pkg: string) => string | undefined>(() => undefined),
+  getInstalledPackageEntryMock: vi.fn<
+    (
+      pkg: string,
+      opts?: {
+        cwd?: string;
+        packageName?: string;
+        conditions?: string[];
+        resolveSubpathWithRequire?: boolean;
+      }
+    ) => string | undefined
+  >(() => undefined),
   hasTreeShakingSharedProviderMock: vi.fn<(pkg: string, shareItem?: unknown) => boolean>(
     () => false
   ),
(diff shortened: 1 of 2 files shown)


gioboa (approved):
LGTM 👏 
Thanks @dmchoi77

PR #1240, opened 2026-09-08
fix(ssr): bust unversioned entry cache on revalidation

dmchoi77:
## Summary
SSR entries without manifest version data always used the same `unversioned` temp-file/import identity. After a remote changed from `v1` to `v2` at the same URL, explicit `revalidate()` could still return the old module from Node's ESM/temp caches, so the loader now adds a process-local generation; this does not detect deployments by itself and only makes explicit revalidation effective.

## Changes
- Advance the generation on explicit per-remote/global `revalidate()` to create a new temp-file/import identity.
- Add HTTP regression coverage that returns `v2` after `v1` for the `remoteEntry.js` convention fallback under explicit revalidation.

## Validation
- `pnpm exec vitest run src/utils/__tests__/ssrEntryLoader.unversioned-revalidate.test.ts`: passed (1 test).
- `pnpm test`: passed (51 files, 1,259 passed, 1 skipped).
- `pnpm typecheck`: passed.
- `pnpm build`: passed.

### DIFF
--- a/src/utils/__tests__/ssrEntryLoader.test.ts
+++ b/src/utils/__tests__/ssrEntryLoader.test.ts
@@ -1327,7 +1327,9 @@ type RunnerOptions = {
 };
 
 async function freshLoaderWithRunner(
-  runnerFactory: (options: RunnerOptions) => { import: (id: string) => Promise<unknown> } | null
+  runnerFactory: (
+    options: RunnerOptions
+  ) => { import: (id: string) => Promise<unknown>; clearCache?: () => void } | null
 ) {
   vi.resetModules();
   const runnerModule = {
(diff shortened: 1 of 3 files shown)


gioboa (changes requested):
revalidate() does not invalidate runnerCache. For /__mf_ssr__/, loadSSRRemoteEntry() calls runner.import(urlObj.pathname) without versionKey, potentially returning stale modules. Call ModuleRunner.clearCache() for the matching origin and add a regression test covering this path.

dmchoi77:
@gioboa Thanks! I’ve cleared the matching ModuleRunner cache in revalidate() and added a regression test for the dev SSR path. Fixed in 197af36f4.

gioboa (approved):
That's awesome 🚗 
TY @dmchoi77

PR #1246, opened 2026-09-09
fix: derive SSR entry filenames like the browser ones

Gmendesfonseca:
## Summary
`[hash]` placeholder stripping is implemented four times. Three copies are byte-identical under two different names — `resolveDevHashEntryFileName` in `pluginProxyRemoteEntry` and `pluginAddEntry`, `resolveDevRemoteEntryFileName` in `pluginMFManifest`. The fourth, `stripSsrFilenameHashPlaceholder` in `virtualRemoteEntrySSR`, has drifted: it tests `/\.[^.]+$/` against the whole path instead of the basename, so a dotted output directory is mistaken for a file extension and the SSR companion names come out wrong.

| `filename` | SSR entry (before) | exposes (before) |
| --- | --- | --- |
| `assets/v1.2/remoteEntry-[hash]` | `assets/v1.ssr.2/remoteEntry` | `assets/v1.exposes.js` |
| `static/js.v2/mf-[hash:8]` | `static/js.ssr.v2/mf` | `static/js.exposes.js` |

The entry path is mangled and the exposes name loses its directory. This is reachable rather than theoretical: `pluginMFManifest` composes both variants as `getSsrRemoteEntryFileName(resolveDevRemoteEntryFileName(...))`, so the browser and SSR names for the same `filename` can disagree — which is what #1157 and #1216 were about. The logic living in four places is why they can fall out of step again.

Behaviour is unchanged for any `filename` that has an extension or an undotted directory, so all existing tests pass untouched; the only change is that the SSR path now agrees with the browser path in the dotted-directory case.

## Changes
- Add `resolveHashPlaceholderFileName` to `utils/buildPaths` and use it from `pluginProxyRemoteEntry`, `pluginAddEntry` and `pluginMFManifest`, removing the three duplicate definitions.
- Derive the SSR companions through the same helper so `getSsrRemoteEntryFileName` and `getSsrExposesFileName` no longer treat a dotted directory as an extension.
- Name the two patterns involved — `HASH_PLACEHOLDER_RE` and `FILE_EXTENSION_RE` — following the existing `EXTERNAL_URL_RE` / `ASSET_LIKE_IMPORT_RE` convention.
- Cover the helper in `buildPaths.test.ts` (separators, sized placeholders, `.js` fallback, dotted directories, and repeated calls against the `/g` `lastIndex` hazard from #1214) and add dotted-directory regression cases to `virtualRemoteEntrySSR.test.ts`.

## Validation
- `pnpm test`: passed (51 files, 1,272 passed, 1 skipped).
- `pnpm test:integration`: passed (16 files, 94 passed).
- `pnpm typecheck`: passed.
- `pnpm build`: passed.
- `pnpm fmt.check`: passed.

## Follow-ups and open questions
This came out of a static-analysis pass over `src` that I went through by hand, keeping only findings that stand up as bug risk on their own. I'd rather land maintainability work in small reviewable pieces than drop a large one on you, and I'd like your steer before going further — you know which files are about to move.

Written and green, happy to open if useful:
1. **Share a single `escapeRegExp`.** It is defined three times with identical bodies (`utils/bundleHelpers`, `utils/VirtualModule`, `plugins/pluginProxyRemotes`) and inlined a fourth time in `utils/controlChunkSanitizer`. No behaviour change.
2. **Name inline regexes and use `Object.hasOwn`.** Hoist the two patterns in `getRuntimeHelpersImplementation`, and replace `Object.prototype.hasOwnProperty.call` at all 15 sites — target is ES2022, engines require Node >= 20.19, and the test suite already uses it.

Larger, and I would not start either without a ruling from you, since both touch the highest-churn files here and would conflict with in-flight fixes:

3. **Lifting types out of implementation files.** `index.ts` declares a run of local `type`/`interface` blocks (`MutableBundlerOutput`, `RolldownOptionsLike`, `BuilderLike`, `ModulePreloadResolveContext`, `ResolveAliasEntry`, `BundleChunkLike`) inline between functions. A colocated `types.ts` would make the plugin bodies easier to read, but it moves code you may be editing right now.
4. **Splitting the largest files** — `virtualRemoteEntry.ts` (2387 lines), `virtualShared_preBuild.ts` (2273), `index.ts` (2207). There are natural seams, but these change most often, so a split would hurt `git blame` and rebase every open PR. Only worth attempting if you want it and can point at seams you would accept.

For clarity on what I am *not* proposing: no style sweep. The nested ternaries read fine as house dialect, and the `//` comments carry the *why* behind specific bundler and framework quirks — turning those into JSDoc parameter boilerplate would throw that away. Happy to be told otherwise on any of it.

### DIFF
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -2,7 +2,11 @@ import * as fs from 'fs';
 import { createHash } from 'node:crypto';
 import * as path from 'node:path';
 import type { Plugin, ResolvedConfig, Rollup } from 'vite';
-import { normalizePathForImport, rebaseImport } from '../utils/buildPaths';
+import {
+  normalizePathForImport,
+  rebaseImport,
+  resolveHashPlaceholderFileName,
+} from '../utils/buildPaths';
 import { findRemoteEntryFile } from '../utils/bundleHelpers';
 import { mapCodeToCodeWithSourcemap } from '../utils/mapCodeToCodeWithSourcemap';
 
@@ -222,15 +226,6 @@ function stripQueryAndHash(file: string) {
 
 function isReactRouterClientRouteInput(file: string) {
   return /[?&]__react-router-build-client-route(?:[=&]|$)/.test(file);
-}
-
-function resolveDevHashEntryFileName(fileName: string) {
-  if (!fileName.includes('[hash')) return fileName;
-
-  const normalized = fileName.replace(/(?:[._-]?\[hash(?::\d+)?\])/g, '');
-  const baseName = path.basename(normalized);
-
-  return path.extname(baseName) ? normalized : `${normalized}.js`;
 }
 
 export function getBuildInput(config: any) {
@@ -803,7 +798,7 @@ for (const __mfRemoteEntryPrefetchUrl of __mfRemoteEntryPrefetchUrls) {
             next();
             return;
           }
-          const devFileName = resolveDevHashEntryFileName(fileName);
+          const devFileName = resolveHashPlaceholderFileName(fileName);
           if (
             devFileName !== fileName &&
             req.url?.startsWith((viteConfig.base + devFileName).replace(/^\/?/, '/'))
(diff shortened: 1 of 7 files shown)


gioboa (approved):
Thanks @Gmendesfonseca nice one 🚀

PR #1249, opened 2026-09-09
fix: re-negotiate version-first singletons before host preload skip

### LINKED ISSUE #1247: [Bug]: version-first singleton static imports resolve to the host's own version instead of the negotiated highest version
### Describe the bug

With `shareStrategy: 'version-first'` and a `singleton: true` shared package, a static `import` of that package resolves to the host's own declared version instead of the negotiated highest version across host + remote(s).

Per the docs, this should not happen:

> If singleton mode is enabled, the shared dependencies between the producer components and the consumer will only be loaded once, and a higher version will be loaded if the versions are inconsistent. A warning will be given for the party with the lower version.
>
> — https://module-federation.io/configure/shared.html#singleton

> `'version-first'`: Version priority, ensuring that the highest version of shared dependencies is used. All remotes entry files will be automatically loaded during initialization to register shared dependencies and ensure version compatibility. This strategy is recommended when there are strict version requirements.
>
> — https://module-federation.io/configure/shareStrategy.html

### Version

1.21.5

### Reproduction

https://github.com/arshabh-agarwal/mf-vite-version-first-singleton-repro

```bash
git clone https://github.com/arshabh-agarwal/mf-vite-version-first-singleton-repro
cd mf-vite-version-first-singleton-repro
./run.sh
```

Then open http://localhost:5001.

### Relevant log output

```shell
host saw: 1.0.0   (expected 1.5.0)
remote saw: 1.0.0 (expected 1.5.0)
```

Zero console warnings/errors — no `strictVersion` mismatch, versions are semver-compatible.

### Validations

- [x] Read the docs.

arshabh-agarwal:
Fixes #1247

## Bug

With `shareStrategy: 'version-first'` and a `singleton: true` shared package, a static `import` of that package resolved to the host's own declared version instead of the negotiated highest version across host + remote(s), with no console warning to surface the mismatch.

## Root cause

`hostAutoInit`'s shared-cache preload loop (`generateHostAutoInitCode`) skips calling `runtime.loadShare()` for a package whenever the cache already holds a value owned by the current container, to avoid redundant work:

```js
if (
  __mfReadSharedCache(...) !== undefined &&
  __mfReadSharedCacheOwner(...) === cacheOwner   // build mode
) return;
```

Separately, the container's own `init()` (called by `hostAutoInit` before this loop runs) unconditionally seeds the shared cache with its **local** fallback the moment it's needed, before remotes have been checked — this seed always writes with `owner: <this container>`.

For a `version-first` singleton, by the time the preload loop above runs, the cache is therefore *always* already populated and self-owned, so the skip condition is always true. The one code path that actually performs full, hook-respecting negotiation (`runtime.loadShare()`) never gets a chance to run and correct the premature local guess — even though the runtime's share scope has, by then, already registered the remote's higher version.

## Fix

A singleton negotiates against every remote under any strategy other than `'loaded-first'`, so the pre-init seed can only ever be a provisional guess for it. This bypasses the self-owned skip specifically for singleton shares, so `loadShare()` always gets to confirm or upgrade that guess once remotes have registered. Non-singleton shares, and the dev/serve-mode skip condition, are unchanged.

## Testing

- New integration test (`integration/version-first-singleton-browser.test.ts`) builds a real host + remote pair with `shareStrategy: 'version-first'` and a `singleton: true` shared package declared at two different versions, serves them, and asserts both sides' static imports resolve to the higher version in a real browser. Verified this test fails without the fix and passes with it.
- Full existing unit (1265 passed) and integration (95 passed) suites pass.
- Full e2e Playwright suite (`vite-vite-host`, `vite-vite-remote`, `multi-example`, `vite-runtime-register` — 27 tests, including the singleton-specific "shared-lib is initialized exactly once" and "does not download shared dependency payloads when dynamic remote hits cache" tests) passes, confirming no duplicate-load regression from removing the self-owned skip for singletons.
- Manually reproduced and verified against the minimal repro linked in the issue.

### DIFF
--- a/integration/fixtures/version-first-singleton-host/entry.js
+++ b/integration/fixtures/version-first-singleton-host/entry.js
@@ -1,0 +1,10 @@
+import sharedLib from 'shared-lib';
+
+window.__host_saw_version__ = sharedLib.getVersion();
+
+const { REMOTE_SHARED_VERSION } = await import('remote1/Module');
+window.__remote_saw_version__ = REMOTE_SHARED_VERSION;
+
+document.querySelector('#app').textContent =
+  `host:${window.__host_saw_version__} remote:${window.__remote_saw_version__}`;
+console.log('__mf_host_entry_evaluated__');
--- a/integration/fixtures/version-first-singleton-host/entry.js
+++ b/integration/fixtures/version-first-singleton-host/entry.js
@@ -2,9 +2,10 @@ import sharedLib from 'shared-lib';
 
 window.__host_saw_version__ = sharedLib.getVersion();
 
-const { REMOTE_SHARED_VERSION } = await import('remote1/Module');
-window.__remote_saw_version__ = REMOTE_SHARED_VERSION;
+import('remote1/Module').then(({ REMOTE_SHARED_VERSION }) => {
+  window.__remote_saw_version__ = REMOTE_SHARED_VERSION;
 
-document.querySelector('#app').textContent =
-  `host:${window.__host_saw_version__} remote:${window.__remote_saw_version__}`;
-console.log('__mf_host_entry_evaluated__');
+  document.querySelector('#app').textContent =
+    `host:${window.__host_saw_version__} remote:${window.__remote_saw_version__}`;
+  console.log('__mf_host_entry_evaluated__');
+});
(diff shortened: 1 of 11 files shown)


gioboa (approved):
Thanks @arshabh-agarwal nice fix 👏

PR #1250, opened 2026-09-09
refactor: share a single escapeRegExp helper

Gmendesfonseca:
## Summary
`escapeRegExp` is defined three times with byte-identical bodies — in `utils/bundleHelpers`, `utils/VirtualModule` and `plugins/pluginProxyRemotes` — and inlined a fourth time in `utils/controlChunkSanitizer`, plus twice more in `pluginAddEntry`'s test. Each copy is the same `value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')`.

The escape set matters because federation interpolates user-controlled strings into generated matchers — package names, chunk file names, virtual module ids — and those routinely contain `.`, `+`, `[` and `\`. Having the set written out in four places means a correction to one would silently miss the others.

No behaviour change: the pattern is unchanged at every call site, and a unit test now pins it.

## Changes
- Add `utils/regexEscape` exporting `escapeRegExp`, with the character class as a named `REGEXP_SPECIAL_CHARS_RE` constant.
- Import it from `utils/bundleHelpers`, `utils/VirtualModule`, `utils/controlChunkSanitizer` and `plugins/pluginProxyRemotes`, deleting the three local definitions and the inline copy.
- Use it for the two copies in `pluginAddEntry`'s test as well, so no duplicate of the class remains in `src`.
- Add `utils/__tests__/regexEscape.test.ts` covering the full metacharacter set and the three real interpolation shapes: a scoped package name, a hashed chunk file name, and a null-byte-prefixed virtual module id.

## Validation
- `pnpm test`: passed (52 files, 1,279 passed, 1 skipped).
- `pnpm typecheck`: passed.
- `pnpm build`: passed.
- `pnpm fmt.check`: passed.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -1,6 +1,7 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import { escapeRegExp } from '../../utils/regexEscape';
 import type {
   ConfigEnv,
   ConfigPluginContext,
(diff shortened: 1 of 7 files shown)


gioboa (approved):
Thanks 💯 @Gmendesfonseca

PR #1251, opened 2026-09-09
refactor: name inline regexes and use Object.hasOwn

Gmendesfonseca:
## Summary
Two small readability changes, kept out of #1246 and the `escapeRegExp` PR so those stayed single-concern.

Being upfront about the value: unlike #1246, this one fixes no defect. It is mechanical and cheap to review, but if you would rather not take cosmetic churn in files that see this much fix traffic, close it without ceremony — no work lost on my side.

## Changes
- Hoist the two regexes in `getRuntimeHelpersImplementation` into named constants (`RUNTIME_INDEX_ENTRY_RE`, `TRAILING_SLASH_RE`), following the existing `EXTERNAL_URL_RE` / `ASSET_LIKE_IMPORT_RE` convention, and use `.exec()` so the pattern is compiled once rather than per call.
- Replace `Object.prototype.hasOwnProperty.call` with `Object.hasOwn` at all 15 sites. `target` is ES2022, `engines.node` is `^20.19.0 || >=22.12.0`, and the test suite already used `Object.hasOwn`.

All 15 sites are build-time code; none sit inside generated browser output, so nothing in the emitted runtime changes.

## Validation
- `pnpm test`: passed (51 files, 1,274 passed, 1 skipped).
- `pnpm typecheck`: passed.
- `pnpm build`: passed.
- `pnpm fmt.check`: passed.

## Note
The two larger questions from #1246 are still open, and I have not started either — lifting the inline `type`/`interface` blocks out of `index.ts` into a colocated `types.ts`, and splitting `virtualRemoteEntry.ts` / `virtualShared_preBuild.ts` / `index.ts`. Both would fight the churn in exactly those files, so I would only attempt them if you want them and can point at seams you would accept. No rush from my side.

### DIFF
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -2161,7 +2161,7 @@ describe('vite:module-federation-early-init', () => {
         g: getAugmentedNamespace
       });
       function getDefaultExportFromCjs(x) {
-        return x && x.__esModule && Object.prototype.hasOwnProperty.call(x, "default") ? x["default"] : x;
+        return x && x.__esModule && Object.hasOwn(x, "default") ? x["default"] : x;
       }
       function getAugmentedNamespace(n) {
         return n;
--- a/src/index.ts
+++ b/src/index.ts
@@ -252,8 +252,13 @@ function appendResolveAlias(config: UserConfig, alias: ResolveAliasEntry): void 
   ];
 }
 
+// `<dir>/index.js` runtime entry, captured as (directory, extension) so the
+// sibling `helpers` module can be addressed with the same extension.
+const RUNTIME_INDEX_ENTRY_RE = /^(.*[\\/])index(\.[cm]?js)$/;
+const TRAILING_SLASH_RE = /\/$/;
+
 function getRuntimeHelpersImplementation(runtimeImplementation: string): string {
-  const indexEntryMatch = runtimeImplementation.match(/^(.*[\\/])index(\.[cm]?js)$/);
+  const indexEntryMatch = RUNTIME_INDEX_ENTRY_RE.exec(runtimeImplementation);
   if (indexEntryMatch) {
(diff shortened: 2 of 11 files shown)


gioboa (commented):
I did a minor change.

Object.hasOwn() is ES2022 and may be unavailable in older browsers. This code is injected into consumer browser bundles; Vite cannot transpile/polyfill that API.

Object.prototype.hasOwnProperty.call() provides identical safe behavior—including null-prototype objects—across older browsers.

gioboa (approved):
Nice refactor 👍 
TY @Gmendesfonseca

Gmendesfonseca:
Thanks for catching that @gioboa, and you're right — that one was on me.

I'd claimed all 15 sites were build-time, but `sharedCacheHelperCode` is a template literal that gets interpolated into the generated remote entry and shared pre-build modules at 8 call sites, so it ships to consumer browsers and Vite won't transpile it against their targets. I checked each site's surrounding code but not whether it sat inside an emitted string — which is exactly the case where it matters.

One thing that might be worth a follow-up: the emitted code currently sits at roughly an ES2021 baseline — `??=` / `||=` appear in 8 places across `virtualRemotes`, `virtualRuntimeInitStatus`, `virtualRemoteEntry` and `packageUtils` — and `Object.hasOwn` (ES2022 / Safari 15.4) was the first thing to cross it. `AggregateError` is in there too, but only on the SSR path, so Node-only.

Happy to send a small PR that makes that boundary explicit, if useful — either a note on the emitted-code templates, or a unit test that fails when post-ES2021 syntax shows up inside them. Or neither, if you'd rather not add the ceremony.

gioboa:
Yeah, thanks, feel free to drop a PR. Let's try to preserve the compatibility 💪⭐️

PR #1252, opened 2026-09-09
test: guard the emitted browser runtime against post-ES2021 APIs

Gmendesfonseca:
## Summary
Follow-up to #1251, per @gioboa's "let's try to preserve the compatibility".

The virtual modules are generated as source strings and interpolated into the consumer's browser bundle — `sharedCacheHelperCode` alone reaches 8 call sites across `virtualRemoteEntry` and `virtualShared_preBuild`. Vite does not transpile those strings against the consumer's build targets, so an API that is too new breaks their app at runtime with nothing failing at build time. That is exactly how `Object.hasOwn` slipped through review in #1251.

This adds a test that catches the next one. Running it against the reverted change reproduces the catch:

```
utils/packageUtils.ts:~264 uses Object.hasOwn (ES2022)
```

No source changes — one new test file.

## Changes
- Add `src/__tests__/emittedRuntimeCompat.test.ts`, scanning the template literals of the six browser-emitting generators (`packageUtils`, `virtualExposes`, `virtualRemoteEntry`, `virtualRemotes`, `virtualRuntimeInitStatus`, `virtualShared_preBuild`) and failing on APIs newer than the baseline.
- Set the baseline at **ES2021**, which is what the emitted code already relies on (`??=` / `||=` appear in 8 places). It is a floor rather than an aspiration — nothing existing had to change to make it pass.
- List the two SSR generators as an explicit exemption. Their output is imported by Node, never a browser, which is why `AggregateError` in the SSR entry is fine and should not trip this.
- Skip `${...}` interpolations, which are build-time TypeScript rather than emitted output, and pin the scanner with its own tests so a broken detector cannot pass silently.

## Notes
Detection matches on specific APIs rather than on syntax, which keeps false positives low but means the list is a floor, not a proof — a genuinely new API has to be added to `POST_BASELINE_APIS` to be caught. I have seeded it with the reachable ones (`Object.hasOwn`, `.at()`, `structuredClone`, `findLast`, `toSorted`/`toReversed`/`toSpliced`, `Object.groupBy`/`Map.groupBy`, `Array.fromAsync`).

The scan is also a source-level heuristic: it reads template literals rather than executing the generators, so it covers every code path regardless of config, but it cannot tell an emitted string from a log message that happens to sit in one of those six files. In practice that direction of error is safe — it surfaces as a loud test failure, never a silent break.

Happy to drop it, narrow it, or move the baseline if you would rather have less ceremony here.

## Validation
- `pnpm test`: passed (53 files, 1,288 passed, 1 skipped).
- `pnpm typecheck`: passed.
- `pnpm build`: passed.
- `pnpm fmt.check`: passed.
- Negative check: reintroducing the `Object.hasOwn` reverted in #1251 fails the guard with the message above; restoring it passes.

### DIFF
--- a/src/__tests__/emittedRuntimeCompat.test.ts
+++ b/src/__tests__/emittedRuntimeCompat.test.ts
@@ -1,0 +1,220 @@
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * Guard for the JavaScript baseline of the code we *emit*.
+ *
+ * The virtual modules below are generated as source strings and shipped into
+ * the consumer's browser bundle. Vite does not transpile them against the
+ * consumer's build targets, so an API that is too new here breaks their app at
+ * runtime with no build-time warning — see the `Object.hasOwn` revert in #1251.
+ *
+ * The emitted code currently sits at ES2021 (`??=` / `||=`), so anything newer
+ * than that is rejected. This is a floor, not an aspiration: if the baseline is
+ * ever raised deliberately, move BASELINE and the pattern list together.
+ *
+ * SSR generators are exempt — their output is imported by Node on the server,
+ * never by a browser, which is why `AggregateError` is fine in the SSR entry.
+ */
+const BASELINE = 'ES2021';
+
+const REPO_SRC = path.resolve(__dirname, '..');
+
+// Modules whose template literals become browser-executed code.
+const BROWSER_EMITTING_MODULES = [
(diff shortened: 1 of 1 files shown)


gioboa (commented):
Expanded guard covers all browser runtime emitters, scans template and quoted strings, detects dynamic .at() calls and Promise.withResolvers, avoids regex-backtick false positives, and retains SSR exemptions.

gioboa (approved):
Thanks @Gmendesfonseca that's a great guard 🔒

PR #1256, opened 2026-09-10
fix: preserve equal-version singleton identity

gioboa:
Avoid unnecessary singleton renegotiation when versions match, preventing split React, Preact, and Solid runtimes.


gioboa (commented):

Image

PR #1257, opened 2026-09-10
fix: do not preload bare remote aliases as root exposes

### LINKED ISSUE #1255: Vite: version-first preloads bare remote aliases as root exposes
### Describe the bug

When using `@module-federation/vite` with `shareStrategy: "version-first"`, the Vite plugin eagerly preloads every configured remote alias as a module.

For a remote configured as `"remote"`, the generated bootstrap contains:

```js
__mfPreloadRemote("remote", "remote")
```

The runtime resolves the bare remote alias as the root expose `"."` and calls:

```js
container.get(".")
```

This fails for valid remotes that expose only subpaths, for example:

```js
exposes: {
  "./App": "./src/App",
  "./AppWebComponent": "./src/AppWebComponent"
}
```

The browser then reports:

```text
Module "." does not exist in container.
```

The actual import of `"remote/AppWebComponent"` still works, but the host starts with a false federation error.

The problem is caused by conflating:

1. a configured remote registration (`"remote"`), and
2. an actually requested root module (`"remote"`).

The Vite plugin currently adds every configured remote with:

```js
addUsedRemote(key, key)
```

and later treats that entry as a module preload. A configured remote alias does not imply that the remote exposes `"."`.

### Suspected fix direction

Registration of configured remotes and actually used modules should be separated.

For `version-first`:

- every configured remote should still be registered for `initializeSharing()` / `runtimeInit({ remotes })`
- the bootstrap should only preload actually imported exposes
- a bare remote should only be loaded as `"."` when the application really does `import("remote")`
- `runtime.loadRemote("remote")` for pure remote initialization should not be required — the runtime already loads remote entries for version-first via `initializeSharing()` / `getEntry()`, and the Vite host already warms module-type entry URLs via the remote-entry prefetch block

### Regression tests

- Host with `shareStrategy: "version-first"`
- Remote without root expose, only `./AppWebComponent`
- No request to `container.get(".")`
- No browser console / page errors from a missing `"."`
- `remote/AppWebComponent` still loads successfully
- Separate case: a real root expose `"."` must still work when the app imports the bare remote

### Related code (current main)

- `src/index.ts` — eager `addUsedRemote(key, key)` for configured remotes
- `src/plugins/pluginAddEntry.ts` — bootstrap `__mfPreloadRemote` from `getUsedRemotesMap()` (comment notes bare ids may represent root `.`)
- `src/plugins/__tests__/pluginAddEntry.test.ts` — currently asserts bare remote preload
- `@module-federation/runtime-core` `packages/runtime-core/src/utils/manifest.ts` — bare id → expose `"."`

Related discussion: #998 (unused expose preload under version-first) — this issue is specifically the bare-alias → `"."` false error for remotes without a root expose.

### Version

`@module-federation/vite` current `main`

### Validations

- [x] Read the docs
- [x] Read the common issues list
- [x] Check that there isn't already an issue that reports the same bug
- [x] The provided reproduction is a minimal reproducible example of the bug

tonoizer:
## Description

`version-first` hosts were calling `__mfPreloadRemote("remote", "remote")` for every **configured** remote alias. The runtime maps a bare remote id to expose `"."`, so remotes that only expose subpaths (e.g. `./AppWebComponent`) logged a false `Module "." does not exist in container` error even when `import("remote/AppWebComponent")` worked.

Configured remotes still need to appear in `localSharedImportMap` / `runtimeInit({ remotes })` for share negotiation. That registration must not be treated as an imported root module.

Fixes #1255

## Changes Made

- Add `ensureUsedRemote(key)` so a configured remote is registered with an **empty** module set (init only).
- Keep `addUsedRemote(key, module)` for modules the app actually imported.
- Version-first bootstrap `__mfPreloadRemote` / `loadRemote` uses the imported-module list only; bare `remote` is preloaded only when the host imports the bare specifier.
- Remote-entry URL prefetch uses registered remote keys, so entry warming still happens without `loadRemote(".")`.

## Testing

On current `main`, new tests failed as expected (bare alias treated as used module). After fix: `vitest run --dir src` → **1312 passed**, 1 skipped; `host-build.test.ts` → **14 passed**; related integration → **12 passed**.

Coverage: version-first + subpath-only remote → no bare `__mfPreloadRemote`; real bare `import("remote")` / root `"."` still preloads; unused configured remotes still in `usedRemotes` / `localSharedImportMap`.


gioboa (commented):

Image

gioboa (approved):
Thanks @tonoizer for fixing it 👏

PR #1269, opened 2026-09-11
feat: expose first-class ssrEntryLoader strategy option

### LINKED ISSUE #1264: Feature: expose a first-class SSR entry loader strategy option
## Problem

The Vite plugin injects `@module-federation/vite/ssrEntryLoader` during `configResolved` and computes `resolvedShared` there. Consumers that need the `vm` strategy must currently post-process the mutable runtime plugin tuple after the upstream hook runs, while preserving the generated `resolvedShared` map.

## Proposal

Expose a first-class Vite federation option such as `ssrEntryLoader: { strategy: "vm" | "temp-file" }` (or an equivalent nested option). The plugin should apply the requested strategy to its injected SSR entry loader while retaining the computed `resolvedShared` values and the existing temp-file default.

The option should keep the existing behavior when omitted and retain the single warning plus temp-file fallback when Node is not started with `--experimental-vm-modules`.

This would let integrations remove post-processing adapters that depend on the current `runtimePlugins` mutation behavior.

tonoizer:
## Description

Exposes a first-class `ssrEntryLoader: { strategy?: 'temp-file' | 'vm' }` option so consumers do not need to post-process the auto-injected `@module-federation/vite/ssrEntryLoader` runtime plugin tuple after `configResolved` while preserving `resolvedShared`.

Fixes #1264

## Changes Made

- Add/normalize `ssrEntryLoader.strategy` in `ModuleFederationOptions` / `normalizeModuleFederationOptions` (omitted, empty, or invalid → `undefined`; `'vm'` / `'temp-file'` preserved)
- Auto-inject pushes `[specifier, { resolvedShared, strategy }]` when strategy is set; keeps `[specifier, { resolvedShared }]` when omitted (default temp-file)
- Existing single warning + temp-file fallback when Node lacks `--experimental-vm-modules` is unchanged

## Testing

- `pnpm test` (src unit): **1318 passed**, 1 skipped
- Targeted normalize + inject cases for omitted / invalid / `vm` / `temp-file`: **5 passed**

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,27 @@ You can specify the place the host initialization file is injected with the **ho
 The **moduleParseTimeout** option allows you to configure the maximum time to wait for module parsing during the build process.
 The **moduleParseIdleTimeout** option is an alternative that resets the timer on every parsed module. It only fires when there has been no module activity for the configured duration, making it suitable for large codebases where the total build time exceeds the fixed timeout.
 
+## SSR entry loading strategy
+
+SSR hosts can choose how HTTP ESM remote entries are evaluated during build and preview:
+
+```ts
+federation({
+  name: "host",
+  remotes: {
+    // ...
+  },
+  ssrEntryLoader: {
+    strategy: "vm",
+  },
+});
+```
+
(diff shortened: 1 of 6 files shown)


gioboa (commented):
Thanks for this improvement @tonoizer 🚀

PR #1270, opened 2026-09-11
fix: defer non-eager share wrappers in expose asset closures

### LINKED ISSUE #1267: [Bug]: non-eager `__loadShare__` wrappers classified as `exposes[*].assets.js.sync` since 1.20.8
### Describe the bug

Since **1.20.8**, a remote whose shares are all `import: false` advertises the per-share `__loadShare__` wrapper chunks as **`exposes[*].assets.js.sync`**. They are then unconditionally modulepreloaded, so the host fetches share providers it is going to supply itself.

The cause is [`expandExposeAssets`](https://github.com/module-federation/vite/blob/main/src/plugins/pluginMFManifest.ts) (added in #1092), which folds each expose's entire static-import closure into `sync`:

```ts
const syncChunks = collectStaticChunks(bundle, assets.js.sync);
const sync = Array.from(
  new Set([...bootstrapAssets, ...syncChunks.map((chunk) => chunk.fileName)])
);
```

That closure contains the `__loadShare__` wrappers. They *are* statically imported, so the walk is correct as written — but a wrapper for a non-eager share resolves through the host at runtime. Classifying it `sync` defeats the point of `eager: false`.

`#1243` does not cover this. With `import: false` the remote bundles no provider, so `filesMap[shareKey]` is empty — every entry in the emitted `shared[]` is `{"sync": [], "async": []}` — and that loop has nothing to act on.

### Evidence

One application tree, Vite 8.2.2 throughout, only the plugin version changing:

| `@module-federation/vite` | `expandExposeAssets` | `./router` js.sync | `./menu` js.sync | `__loadShare__` chunks on disk |
|---|---|---|---|---|
| 1.16.12 | absent | **1** | **1** | 9 |
| 1.21.5 | present | 23 | 12 | 9 |
| 1.21.6 | present | 23 | 12 | 9 |

The wrapper chunks exist in every build. Only their classification changes, so this is a manifest question rather than a chunking one.

Of the 23 entries on `./router`, 7 are `__loadShare__` wrappers:

```
assets/_virtual_mf__..._loadShare__react__loadShare__.js-<hash>.js
assets/_virtual_mf__..._loadShare__react_mf_2_dom__loadShare__.js-<hash>.js
assets/_virtual_mf__..._loadShare__react_mf_2_router__loadShare__.js-<hash>.js
assets/_virtual_mf__..._loadShare___mf_0_tanstack_mf_1_react_mf_2_query__loadShare__.js-<hash>.js
assets/_virtual_mf__..._loadShare___mf_0_sentry_mf_1_react__loadShare__.js-<hash>.js
assets/_virtual_mf__..._loadShare___mf_0_radix_mf_2_ui_mf_1_react_mf_2_tooltip__loadShare__.js-<hash>.js
assets/_virtual_mf__..._loadShare___mf_0_netboxlabs_mf_1_ui_mf_2_platform_mf_2_shared_mf_2_react_mf_2_utils__loadShare__.js-<hash>.js
```

**Expected:** those 7 in `async`, the expose's own chunks in `sync`.
**Actual:** all 23 in `sync`, and modulepreloaded.

### Fix sketch

In `expandExposeAssets`, partition the static closure: a chunk whose modules include `getLoadShareModulePath(shareKey, …)` for a share with `shareConfig.eager !== true` goes to `async` rather than `sync`. `eager: true` shares keep their current classification, and the expose's own chunks are untouched.

One detail worth flagging for anyone implementing it: `chunk.moduleIds` carry Rollup's `\0` virtual-module prefix while `getLoadShareModulePath` returns the unprefixed id, so the comparison needs `normalizeVirtualModuleId` — the same helper `isContainerBootstrapChunk` already uses a few lines above.

### Not a duplicate of

- **#1241 / #1243** — same symptom, different path. That fix reclassifies `filesMap[shareKey]`, which is empty for an `import: false` consumer.
- **#1265** — that is *eager* shares producing too many wrapper chunks and `manualChunks` being unable to coalesce them. This is *non-eager* wrappers being misclassified, with the chunk count unchanged.

### Version

1.21.6 (regression introduced in 1.20.8)

### Reproduction

Any Vite remote whose shares are all `import: false`:

1. Build a remote with `exposes` and several `shared` entries, all `import: false`.
2. `vite build`, then read `dist/mf-manifest.json`.
3. `exposes[*].assets.js.sync` contains the `__loadShare__` wrappers, while `shared[]` is `{"sync": [], "async": []}`.
4. Downgrade only `@module-federation/vite` to 1.20.7 or earlier and rebuild — `sync` drops to the expose's own chunks.

### Relevant log output

```shell
# 1.21.6 — exposes[*].assets.js
./router     js.sync=23  js.async=28
./menu       js.sync=12  js.async=0

# 1.16.12, same tree, same Vite 8.2.2
./router     js.sync=1   js.async=8
./menu       js.sync=1   js.async=0
```

DanBeckDev:
Fixes #1267.

## Problem

`expandExposeAssets` (`src/plugins/pluginMFManifest.ts`, added in #1092, first released in **1.20.8**) folds each expose's entire static-import closure into `assets.js.sync`. That closure carries the per-share `__loadShare__` wrapper chunks.

The wrappers *are* statically imported, so the walk is correct as written. But a wrapper for a share with `eager !== true` resolves through the host at runtime, so advertising it as `sync` gets it modulepreloaded and the remote fetches a provider the host is going to supply. That defeats the point of `eager: false`.

`#1243` does not cover this. With `import: false` the remote bundles no provider, so `filesMap[shareKey]` is empty — every entry in the emitted `shared[]` is `{"sync": [], "async": []}` — and that loop has nothing to act on.

Measured on one real application, Vite 8.2.2 throughout, only the plugin version changing:

| `@module-federation/vite` | `expandExposeAssets` | `./router` js.sync | `./menu` js.sync |
|---|---|---|---|
| 1.16.12 | absent | 1 | 1 |
| 1.21.5 | present | 23 | 12 |
| 1.21.6 | present | 23 | 12 |

The 9 `__loadShare__` chunks exist on disk in every one of those builds. Only their classification changes.

## Change

Partition the static closure in `expandExposeAssets`: a chunk that exists to load a non-eager share goes to `async`, everything else keeps its current classification.

```diff
-    const syncChunks = collectStaticChunks(bundle, assets.js.sync);
+    const allSyncChunks = collectStaticChunks(bundle, assets.js.sync);
+    const syncChunks = allSyncChunks.filter(
+      (chunk) => !isDeferredShareChunk(chunk, deferredShareModules)
+    );
+    const deferredChunks = allSyncChunks.filter((chunk) =>
+      isDeferredShareChunk(chunk, deferredShareModules)
+    );
     const sync = Array.from(
       new Set([...bootstrapAssets, ...syncChunks.map((chunk) => chunk.fileName)])
     );
     const syncSet = new Set(sync);
-    const asyncChunks = collectStaticChunks(bundle, assets.js.async);
-    const async = asyncChunks
-      .map((chunk) => chunk.fileName)
-      .filter((fileName) => !syncSet.has(fileName));
+    const asyncChunks = [...collectStaticChunks(bundle, assets.js.async), ...deferredChunks];
+    const async = Array.from(
+      new Set(asyncChunks.map((chunk) => chunk.fileName))
+    ).filter((fileName) => !syncSet.has(fileName));
```

Two small helpers do the identification. `collectDeferredShareModules` builds the id set from `getUsedShares` + `getLoadShareModulePath`, skipping any share with `shareConfig.eager === true`. `isDeferredShareChunk` tests a chunk's `facadeModuleId` and `moduleIds` against it.

One detail worth calling out: `chunk.moduleIds` carry Rollup's `\0` virtual-module prefix while `getLoadShareModulePath` returns the unprefixed id, so both sides go through `normalizeVirtualModuleId` — the same comparison `isContainerBootstrapChunk` makes a few lines above. Without that normalization the match silently never fires, which is how the first version of this patch passed its unit test and changed nothing in a real build.

- Scoped to non-eager shares — `eager: true` wrappers are untouched
- The expose's own chunks are untouched
- Asset totals are conserved; entries move between arrays rather than disappearing

## Test

Two regression tests in `src/plugins/__tests__/pluginMFManifest.test.ts`:

- `'classifies a non-eager share wrapper in an expose closure as async'` — an expose statically importing a wrapper chunk whose `moduleIds` carry the `\0`-prefixed loadShare id. Asserts the wrapper leaves `sync`, lands in `async`, and the expose's own chunk stays in `sync`.
- `'keeps an eager share wrapper in an expose closure synchronous'` — the same fixture with `eager: true`, guarding against over-correction.

The mock for `../../virtualModules` gains `getLoadShareModulePath`.

Verified locally:

- Full suite: **1315 passed | 1 skipped**
- Same suite with the source reverted and the tests kept: the non-eager test fails as expected, the other **1314 pass** — the test catches the bug and nothing existing regresses
- `pnpm typecheck` and `pnpm fmt --check` clean
- Empirical end-to-end: built `lib/` patched into a real Vite 8 remote's `node_modules`

  | expose | before | after |
  |---|---|---|
  | `./router` | sync 23 / async 28 | **sync 16 / async 35** |
  | `./menu` | sync 12 / async 0 | **sync 9 / async 3** |

  All 7 and 3 `__loadShare__` wrappers respectively move out of `sync`; totals conserved (51 and 12). The remaining `sync` entries are the container bootstrap chunks and the expose's own code, which is what #1092 intends.

### DIFF
--- a/e2e/vite-vite/tests/eager-manifest.spec.ts
+++ b/e2e/vite-vite/tests/eager-manifest.spec.ts
@@ -1,0 +1,17 @@
+import { expect, test } from '@playwright/test';
+
+test('classifies eager and deferred share wrappers in the built manifest', async ({ request }) => {
+  const response = await request.get('/mf-manifest.json');
+  expect(response.ok()).toBe(true);
+
+  const manifest = await response.json();
+  const exposed = manifest.exposes.find(
+    ({ name }: { name: string }) => name === 'EagerManifestFixture'
+  );
+  expect(exposed).toBeDefined();
+
+  const { sync, async } = exposed.assets.js;
+  expect(sync.some((asset: string) => asset.includes('loadShare-eager'))).toBe(true);
+  expect(async.some((asset: string) => asset.includes('__loadShare__vue__loadShare__'))).toBe(true);
+  expect(sync.some((asset: string) => asset.includes('__loadShare__vue__loadShare__'))).toBe(false);
+});
--- a/e2e/vite-vite/tests/host-preview.spec.ts
+++ b/e2e/vite-vite/tests/host-preview.spec.ts
@@ -8,6 +8,28 @@ import { expect, test } from '@playwright/test';
  * shared deps, default imports, named imports, CJS interop, etc.
  */
 test.describe('vite-vite host preview', () => {
+  test('classifies eager and deferred share wrappers in the built manifest', async ({
+    request,
+  }) => {
+    const response = await request.get('/mf-manifest.json');
+    expect(response.ok()).toBe(true);
+
+    const manifest = await response.json();
+    const exposed = manifest.exposes.find(
+      ({ name }: { name: string }) => name === 'EagerManifestFixture'
+    );
+    expect(exposed).toBeDefined();
+
+    const { sync, async } = exposed.assets.js;
+    expect(sync.some((asset: string) => asset.includes('loadShare-eager'))).toBe(true);
+    expect(async.some((asset: string) => asset.includes('__loadShare__vue__loadShare__'))).toBe(
+      true
+    );
+    expect(sync.some((asset: string) => asset.includes('__loadShare__vue__loadShare__'))).toBe(
+      false
+    );
+  });
+
   test('renders host app with React shared dep', async ({ page }) => {
     await page.goto('/');
     const heading = page.getByRole('heading', { name: 'MF HOST Demo', exact: true });
--- a/e2e/vite-vite/tests/host-preview.spec.ts
+++ b/e2e/vite-vite/tests/host-preview.spec.ts
@@ -8,28 +8,6 @@ import { expect, test } from '@playwright/test';
  * shared deps, default imports, named imports, CJS interop, etc.
  */
 test.describe('vite-vite host preview', () => {
-  test('classifies eager and deferred share wrappers in the built manifest', async ({
-    request,
-  }) => {
-    const response = await request.get('/mf-manifest.json');
-    expect(response.ok()).toBe(true);
-
-    const manifest = await response.json();
-    const exposed = manifest.exposes.find(
-      ({ name }: { name: string }) => name === 'EagerManifestFixture'
-    );
-    expect(exposed).toBeDefined();
-
-    const { sync, async } = exposed.assets.js;
-    expect(sync.some((asset: string) => asset.includes('loadShare-eager'))).toBe(true);
-    expect(async.some((asset: string) => asset.includes('__loadShare__vue__loadShare__'))).toBe(
-      true
-    );
-    expect(sync.some((asset: string) => asset.includes('__loadShare__vue__loadShare__'))).toBe(
-      false
-    );
-  });
-
   test('renders host app with React shared dep', async ({ page }) => {
     await page.goto('/');
     const heading = page.getByRole('heading', { name: 'MF HOST Demo', exact: true });
--- a/examples/vite-vite/vite-host/src/EagerManifestFixture.jsx
+++ b/examples/vite-vite/vite-host/src/EagerManifestFixture.jsx
@@ -1,0 +1,7 @@
+import { Button } from 'antd';
+import React from 'react';
+import { version as vueVersion } from 'vue';
+
+export default function EagerManifestFixture() {
+  return React.createElement(Button, null, `eager manifest fixture ${vueVersion}`);
+}
--- a/examples/vite-vite/vite-host/vite.config.js
+++ b/examples/vite-vite/vite-host/vite.config.js
@@ -6,6 +6,9 @@ const treeShakingMode = ['server-calc', 'runtime-infer'].includes(process.env.TR
   ? process.env.TREE_SHAKING_MODE
   : undefined;
 const eagerShared = process.env.EAGER_SHARED === 'true';
+const eagerManifestPort = process.env.EAGER_MANIFEST_PORT
+  ? Number(process.env.EAGER_MANIFEST_PORT)
+  : undefined;
 const externalRuntime = process.env.EXTERNAL_RUNTIME === '1';
 const antdShared = {
   singleton: true,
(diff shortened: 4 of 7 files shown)


gioboa (approved):
Thanks @DanBeckDev nice fix 🫶

DanBeckDev:
> Thanks @DanBeckDev nice fix 🫶

No problem at all, I see one of the e2e test are failing...taking a look now.

DanBeckDev:
Thanks for adding the coverage — the new `classifies eager and deferred share wrappers in the built manifest` test is a better assertion than my unit tests, and it passes against this fix.

The 10 failures alongside it are caused by `5138be9`, not by the fix, and I have bisected them locally.

**`EAGER_SHARED=true` in `playwright.config.ts` breaks the example host for the whole suite.** It is set on the shared `preview-vv:ci` web server, so every `vite-vite-host` spec runs against a host that no longer renders — each failure is `getByRole('heading', { name: 'MF HOST Demo' })` not found.

Bisected on `main` + this PR, `--project=vite-vite-host`:

| Tree | Result |
|---|---|
| this PR, without `5138be9` | **10 passed** |
| this PR + `5138be9` | 10 failed, 1 passed (the new test) |
| `5138be9` with only `EAGER_SHARED=true` reverted in `playwright.config.ts` | **10 passed**, 1 failed (the new test, which needs the env) |
| `5138be9` with `EAGER_SHARED=true` kept but `antd`'s `eager: true` disabled | 10 failed |

So the trigger is narrowed to one line:

```js
vue: eagerShared ? { import: false } : {},
```

not the `antd` eager branch.

That reads like the root cause: `import: false` says "do not bundle a provider, get it from the host" — but this *is* the host, and `EagerManifestFixture.jsx` does `import { version as vueVersion } from 'vue'`. With nothing above it in the share graph there is no provider for vue at all, so the host fails to boot and every spec that loads `/` goes down with it.

Two options, whichever you prefer:

1. Give the fixture its own Playwright project and web server, so the env change does not reach the other specs.
2. Put the deferred share on the **remote** rather than the host, where `import: false` is the normal configuration, and assert the remote's manifest.

Happy to push either — say which and I will do it. I have left the branch as-is so you can decide.

gioboa (commented):
Thanks for the report @DanBeckDev I pushed a small fix, let's see

PR #1271, opened 2026-09-11
fix: skip manifest entry prefetch

### LINKED ISSUE #1266: [Bug]: Nested federation hosts can fetch the same manifest more than once
## Description

When a Module Federation remote is also a host, two runtime instances can
resolve the same remote manifest independently:

1. the parent runtime resolves the nested host and its manifest; and
2. the nested host initializes and resolves its own providers.

Without a page-level request cache shared by those runtime instances, the same
`mf-manifest.json` URL can be requested more than once. If the manifest is
served with revalidation-oriented cache headers such as
`Cache-Control: max-age=0, must-revalidate`, the browser still performs a
network round trip even when the response body is unchanged.

## Environment

- `@module-federation/vite`: `1.21.6`
- `@module-federation/runtime`: `2.9.0`
- Vite/Rolldown: `8.2.0`
- Browser: Chromium-based browser
- Topology: host -> mid-shell host -> leaf remote
- Manifest mode: `manifest: true`

## Steps to Reproduce

1. Create three applications: `host`, `mid-shell`, and `leaf`.
2. Configure `host` to load `mid-shell` as a remote.
3. Configure `mid-shell` as both a remote and a host with its own child
   remote, `leaf`.
4. Enable `manifest: true` for the host containers and register the nested
   providers through the standard Module Federation runtime APIs.
5. Serve the manifests with a revalidation policy, for example:

```http
Cache-Control: max-age=0, must-revalidate
```

6. Do not add an application-level manifest cache or early manifest warmup.
7. Open a host route that loads `mid-shell`, then allow `mid-shell` to
   initialize and resolve `leaf`.
8. Record requests for each `mf-manifest.json` URL, including conditional
   requests that receive `304 Not Modified`.

## Actual Behavior

The same manifest URL can be requested by both the parent and nested runtime
instances. Even when the second response is a fast `304`, it still adds a
network round trip and response parsing/coordination work. Under higher
latency, the duplicate request can extend the time before the nested route is
ready.

In one multi-level integration, a page-level response handoff was required to
prevent this behavior. With that handoff enabled, the live resource timeline
shows one request per warmed manifest, so the workaround masks the duplicate
at the application level.

## Expected Behavior

The runtime should deduplicate same-origin manifest requests across nested
runtime instances for the lifetime of a page, or expose a documented cache
interface that parent and child instances can share.

The behavior should also support explicit invalidation so independently
deployed remotes can refresh their manifest without requiring a full page
reload.

## Impact

Nested shells pay avoidable manifest network latency during startup and route
transitions. The impact is most visible when manifests are not strongly cached,
when the topology is more than two levels deep, or when the nested host is
loaded over a high-latency connection.

## Technical Details

A manifest response is a one-shot `Response` body. A shared implementation
therefore needs to either:

- cache parsed immutable manifest data; or
- retain the in-flight response and return a clone to each consumer.

The cache key should include the absolute URL and request properties that can
change the response, such as credentials/mode where applicable. Failed
requests should not permanently poison the cache.

## Suggested Implementation

- Add a page-level manifest request registry in the runtime.
- Deduplicate both in-flight and completed same-origin requests.
- Return cloned responses or parsed immutable manifests to each caller.
- Provide an explicit force-refresh/invalidation API.
- Add an integration test for host -> mid-shell -> leaf that asserts one
  request per manifest, including under `max-age=0, must-revalidate`.

## Acceptance Criteria

- The nested-host integration test passes without application-specific boot
  code.
- Concurrent parent and child lookups share one request/promise.
- A failed request can be retried normally.
- Explicit refresh bypasses the cached result and retrieves the current
  manifest.

### Version

1.21.6

### Reproduction

NA

### Relevant log output

```shell

```

### Validations

gioboa:
Closes #1266

Prevents duplicate manifest requests by excluding JSON entries from module prefetching.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -1289,11 +1289,17 @@ describe('pluginAddEntry', () => {
           type: 'var',
           shareScope: 'default',
         },
+        manifest: {
+          entry: 'http://localhost:5003/mf-manifest.json',
+          name: 'manifest',
+          type: 'module',
+        },
       },
     } as any;
     addUsedRemote('remote', 'remote/Route', federationOptions);
     addUsedRemote('remote', 'remote/Button', federationOptions);
     addUsedRemote('legacy', 'legacy/Widget', federationOptions);
+    addUsedRemote('manifest', 'manifest/App', federationOptions);
 
     const plugins = addEntry({
       entryName: 'hostInit',
@@ -1319,6 +1325,7 @@ describe('pluginAddEntry', () => {
       'const __mfRemoteEntryPrefetchUrls = ["http://localhost:5001/remoteEntry.js"];'
     );
     expect(code).not.toContain('http://localhost:5002/remoteEntry.js');
+    expect(code).not.toContain('http://localhost:5003/mf-manifest.json');
     // A prefetch failure must stay silent; loadRemote reports real errors later.
     expect(code).toContain(
       'import(/* @vite-ignore */ __mfRemoteEntryPrefetchUrl).catch(() => {});'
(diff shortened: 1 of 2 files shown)


gioboa (commented):

Image

PR #1273, opened 2026-09-11
fix: leave eager shares out of the pending-share seed barrier

### LINKED ISSUE #1272: [Bug]: pendingShares seeds eager shares that are already static imports, so each one becomes a dynamic entry
### Describe the bug

`generatePendingSharesCode()` lists every materialized share as `[pkg, () => import(<loadShare module>)]` and `preloadPendingShares()` awaits them before the entry runs. The filter skips `import: false` and tree-shaken shares, but not `eager: true` ones.

For an eager share the seed is redundant: `generateLocalSharedImportMap()` already emits `import * as __mfEagerShare_N from "<local shared package path>"` for it, so the module is evaluated before the entry either way. What the seed does add is a dynamic import per share — and a dynamic-import target has to be kept reachable as an entry of its own, so the bundler gives every eager share a chunk.

The repro prints what the build emitted: with 1.21.6 the barrier imports all ten eager shares, with the one-line fix it imports none.

Whether dropping those dynamic entries also collapses the wrapper chunks depends on the shape of the graph — in the ten-module repro it does not, in a real app it does. Measured on our Agent Desktop host (275 shares, no provider containers, every app bundling its own fallback; Vite 8 / Rolldown):

| | index.html | modulepreload | first-paint bytes | files < 2 KB |
|---|---|---|---|---|
| `eager: true` on the share table | 91,511 B | 435 | 9.84 MB | 272 |
| the same, with the eager seeds skipped | 27,163 B | 128 | 9.61 MB | 6 |

Same bytes, a third of the requests: once the shares are no longer dynamic entries, Rolldown merges the wrappers on its own.

I intend to submit a PR for this — the fix is one condition in the `generatePendingSharesCode` filter, plus a unit test. If you would rather keep the barrier symmetric, the alternative is to import the eager shares statically in `pendingShares` too; that removes the dynamic entries just as well.

One thing I could not settle from the outside: whether any `loadShareSync`-style or SSR consumer relies on the seeded cache even for an eager share. Our app (8 entry points, ~275 shares) runs clean with the filter, but you know those paths better.

### Version

vite 8.2.1, @module-federation/vite 1.21.6 (same in 1.21.5)

### Reproduction

https://github.com/marksnode/mf-vite-repro-eager-pending-shares

### Relevant log output

```shell

```

### Validations

marksnode:
Fixes #1272.

`generatePendingSharesCode()` seeds every materialized share with `() => import(<loadShare module>)`. For an `eager` share that seed is redundant — `generateLocalSharedImportMap()` already imports it statically, so it is evaluated before the entry runs either way — and it is what makes the share a dynamic entry, which the bundler then has to give a chunk of its own.

The change is one condition in the filter; the unit test asserts both halves of the invariant: the eager share still appears as a static `__mfEagerShare_` import in the local shared import map, and no longer appears in `__mfPendingShareImports`.

Repro: https://github.com/marksnode/mf-vite-repro-eager-pending-shares — ten eager shares; `npm run report` prints `shares the seed barrier imports: 10` on 1.21.6 and `0` with this change.

Measured on a real host (275 shares, no provider containers, Vite 8 / Rolldown): `index.html` 91,511 → 27,163 B, modulepreload links 435 → 128, first-paint bytes unchanged at ~9.6 MB, files under 2 KB 272 → 6. Once the shares stop being dynamic entries, Rolldown merges the wrappers on its own.

If you would rather keep the barrier symmetric, the alternative is to import the eager shares statically inside `pendingShares` as well — that removes the dynamic entries too; say the word and I will switch the PR to that shape.

One open question I could not settle from outside the project: whether any `loadShareSync`-style or SSR consumer relies on the seeded cache even when the share is eager. Full unit suite passes (70 files, 1411 tests), and our own app runs clean across 8 entry points.

### DIFF
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -1033,6 +1033,38 @@ describe('virtualRemoteEntry', () => {
     );
   });
 
+  it('leaves eager shares out of the build bootstrap: they are already static imports', async () => {
+    const share = (name: string, eager = false) => ({
+      name,
+      version: '1.0.0',
+      scope: 'default',
+      shareConfig: {
+        singleton: true,
+        strictVersion: false,
+        ...(eager ? { eager: true } : {}),
+      },
+    });
+    normalizedSharedMock.mockReturnValue({
+      react: share('react'),
+      'eager-shared': share('eager-shared', true),
+    });
+    const mod = await import('../virtualRemoteEntry');
+    mod.getUsedShares().clear();
+    mod.addUsedShares('react');
+    mod.addUsedShares('eager-shared');
+
+    const localMap = mod.generateLocalSharedImportMap();
+    const buildCode = mod.generatePendingSharesCode('build');
+
+    // The import map already imports the eager share statically, so it is evaluated before the entry
+    // either way; seeding it as well would only make it a dynamic entry, and so a chunk of its own.
+    expect(localMap).toContain('from "virtual:prebuild:eager-shared";');
+    expect(buildCode).toContain(
+      'const __mfPendingShareImports = [["react", () => import("virtual:loadShare:react")]];'
+    );
+    expect(buildCode).not.toContain('virtual:loadShare:eager-shared');
+  });
+
   it('orders React package roots before their subpath shares', async () => {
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks for streamlining eager shares @marksnode 👏

PR #1275, opened 2026-09-11
fix: keep React DOM fallbacks coherent

### LINKED ISSUE #1274: [Bug]: production singleton fallbacks eagerly evaluate local React DOM against shared React
## Reproduction

With `@module-federation/vite@1.21.6`, `@module-federation/runtime@2.9.0`, and Vite 8, a generated nested Vite React deployment has:

- host on React/React DOM 19.2.8;
- nested host `remote_a` on 19.2.4;
- sibling `remote_b` on 19.2.6;
- leaf `remote_c` on 19.2.8;
- `react`, `react/jsx-runtime`, `react/jsx-dev-runtime`, `react/compiler-runtime`, and `react-dom/client` all configured as singleton shares with `strictVersion: false`, `shareScope: "default"`, and `requiredVersion: "^19.2.4"`;
- `shareStrategy: "version-first"`.

The generated production apps are served with `pnpm preview`. The nested route `/remote-a/remote-c/` and the sibling route `/remote-b/` fail when the remotes mount.

## Actual result

The page-wide Module Federation cache is coherent after bootstrap. A dump of `globalThis.__mf_module_cache__.share`, with owners from `Symbol.for("module-federation.shared-cache-owners")`, is:

```
default:react                 19.2.8  owner=host
default:react/jsx-runtime     19.2.8  owner=host
default:react/jsx-dev-runtime 19.2.8  owner=host
default:react/compiler-runtime 19.2.8  owner=host
default:react-dom/client       19.2.8  owner=host
```

Nevertheless, loading the lower-patch remotes throws React error #527 before Bridge mount:

```
Minified React error #527 ... args[]=19.2.8&args[]=19.2.4
Minified React error #527 ... args[]=19.2.8&args[]=19.2.6
```

The first version is the shared React seen by the local React DOM fallback; the second is that fallback's installed React DOM version. The failure is not a per-key share negotiation split.

A prototype that lazily imports only `react-dom/client` removes #527 and allows the standalone remote pages to render, but the nested leaf then produces an invalid-hook-call failure:

```
Cannot read properties of null (reading 'useContext')
```

The renderer stack shows the leaf's local `react-dom/client` prebuild rendering components whose React bindings have already rebound to the host's shared React. This is the second manifestation of the same eager-fallback identity problem.

## Cause

For remote-only containers with `hostInitInjectLocation: "entry"` (including a nested exposed leaf), the generated `__loadShare__` module follows the eager fallback shape:

```js
import * as __mfLocalShare from "<local prebuild>";
let exportModule = __mfReadSharedCache(__mfModuleCache.share, descriptor);
  Promise.resolve().then(() => {
    if (__mfReadSharedCache(__mfModuleCache.share, descriptor) !== undefined) return;
    const localShare = __mfInitializedLocalShare(__mfLocalShare);
    if (localShare !== undefined) __mfWriteSharedCache(__mfModuleCache.share, descriptor, localShare, "remote_c");
  });
  exportModule = __mfLocalShare;
}
__mfSubscribeSharedCache(__mfModuleCache.share, descriptor, __mfApplyEagerShareExports);
__mfApplyEagerShareExportsWhenReady(exportModule);
```

The static import evaluates the local `react-dom/client` prebuild before the cache is populated. The prebuild imports the local `react` share wrapper during that evaluation, so React DOM captures that fallback's React object. A later cache write only re-applies the exported React DOM bindings; it cannot change the React object captured by the already evaluated CJS renderer. The result is either #527 (different versions) or #321 (same version, different React identity).

## Expected result

A production singleton fallback must not evaluate a local React family member against a different shared family member. The fix should preserve standalone remote operation and the complete local family fallback.

Possible upstream fixes:

1. Make each local React-family prebuild self-consistent: local `react-dom/*` must resolve its React peer to the same container-local `react` prebuild, and the local family must remain the fallback as one identity.
2. Defer non-workspace singleton fallbacks for exposed and nested remote-only containers using the existing deferred bootstrap. The cache must be checked after federation initialization, the cached provider must be bound before a local fallback is evaluated, and an entry-injected wrapper must not bind the local namespace first or overwrite a cached provider later. The root `react` cannot be implemented with top-level await because its CJS consumers trigger Rolldown `REQUIRE_TLA`.

Please add a regression covering a nested host and leaf, not only a standalone remote. The relevant previous changes are #1173, #1197, #1243, #1247/#1249, and #1256. Related React DOM sharing report: https://github.com/module-federation/core/issues/4727.

This reproduction came from EnBWAG/smo-architektur.dev-web PR #899 and is being tracked in https://github.com/EnBWAG/smo-architektur.dev-web/issues/298.

gioboa:
Closes #1274
Preserve #1173, #1177;

Defers local singleton-family materialization, reuses host cache, and preserves shared-package self-imports.

### DIFF
--- a/src/plugins/__tests__/pluginAddEntry.test.ts
+++ b/src/plugins/__tests__/pluginAddEntry.test.ts
@@ -907,6 +907,14 @@ describe('pluginAddEntry', () => {
 
     expect(internalResult).toBeUndefined();
 
+    const localMapResult = await runTransform(
+      buildPlugin,
+      "const requiredExport = 'createRoot';",
+      '\0virtual:mf-localSharedImportMap:__mfe_internal__host__mf_owner__1'
+    );
+
+    expect(localMapResult).toBeUndefined();
+
     const entryResult = (await runTransform(
       buildPlugin,
       'import { hydrateRoot } from "react-dom/client";\nhydrateRoot(document, app);',
--- a/src/plugins/pluginAddEntry.ts
+++ b/src/plugins/pluginAddEntry.ts
@@ -679,7 +679,9 @@ for (const __mfRemoteEntryPrefetchUrl of __mfRemoteEntryPrefetchUrls) {
   function isFederationInternalVirtualId(id: string) {
     const normalized = decodeViteId(id).replace(/^\0+/, '');
     return (
-      normalized.includes('virtual:mf:') || /__(?:loadShare|prebuild|loadRemote)__/.test(normalized)
+      normalized.includes('virtual:mf:') ||
+      normalized.startsWith('virtual:mf-') ||
+      /__(?:loadShare|prebuild|loadRemote)__/.test(normalized)
     );
   }
 
--- a/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
+++ b/src/virtualModules/__tests__/virtualRemoteEntry.test.ts
@@ -673,6 +673,83 @@ describe('virtualRemoteEntry', () => {
     );
   });
 
+  it('reuses a cached singleton family before importing its local fallback', async () => {
+    const mod = await import('../virtualRemoteEntry');
+    mod.getUsedShares().clear();
+    mod.addUsedShares('react-dom/client');
+
+    const hostRenderer = { marker: 'host-react-dom', createRoot: () => undefined };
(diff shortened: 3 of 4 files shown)


gioboa (commented):

Image

PR #1279, opened 2026-09-11
fix: coherent __prebuild__ sibling resolve for React family (patch skew #527)

### LINKED ISSUE #1278: [Bug]: production eager local react-dom fallbacks bind shared react and throw #527 on patch skew
### Describe the bug

When federated Vite apps share the React family as singletons with a caret range such as `requiredVersion: '^19.2.4'` but install **different React 19.2.x patches**, production builds crash with minified React error **#527** and/or `Cannot read properties of null (reading 'useState')`.

This happens even when share negotiation is coherent (one host-owned `react` / `react-dom` in the runtime cache). The crash is caused by **eager local fallback evaluation**, not by a split share registry.

I intend to submit a PR for this.

### Mechanism

Eager singleton fallback templates statically evaluate the container-local `react-dom` (or `react-dom/client`) copy. Files in that `__prebuild__` graph import `react`. `proxyPreBuildShared:resolve-shared-loadShare` rewrites those sibling imports to `__loadShare__react`, which is the **host** React instance.

Result: local `react-dom@19.2.4` (or `19.2.6`) initializes against host `react@19.2.8`. React then throws #527 (`19.2.8` vs `19.2.4`). A later share-cache rebind cannot fix the React object already captured by the evaluated CJS renderer.

`shouldSkipTaggedImporterProxy` only skips a wrapper's **own** fallback import. Cross-package shared imports from a `__prebuild__` graph (for example `react-dom` → `react`) still go through `__loadShare__`.

Related: #1173, #1197, #1243, #1247, #1249, #1256, #1274, #1275.

#1275 reused a compatible cached React-family singleton in the import-map getter. That does not keep the **local fallback graph** version-coherent: an eager `__prebuild__react-dom` still binds `__loadShare__react` during static evaluation.

### Expected

If the importer is inside the `__prebuild__` graph of shared package P, and the source is another shared package that P depends on, resolve to the sibling `__prebuild__` (same container-local family) instead of `__loadShare__`. Keep provider rebinding on the loadShare wrappers. Do not introduce package-level cycles and do not undo #1173 deferred fallbacks.

### Version

`@module-federation/vite@1.21.6` (repro) / current `main` after #1275. Vite 6.3.x / 8. React 19.2.x patch skew. `shareStrategy: 'version-first'`.

### Reproduction

https://github.com/tonoizer/react-mf-patch-skew-repro

| App | React | Role |
| --- | --- | --- |
| host | 19.2.8 | remotes: `remote_a`, `remote_b` |
| remote_a | 19.2.4 | nested host, remotes `remote_c` |
| remote_b | 19.2.6 | remote-only |
| remote_c | 19.2.8 | nested under `remote_a` |

Shared as `singleton: true`, `strictVersion: false`, `requiredVersion: '^19.2.4'`: `react`, `react/jsx-runtime`, `react/jsx-dev-runtime`, `react/compiler-runtime`, `react-dom/client`.

### Relevant log output

```
Minified React error #527 ... args[]=19.2.8&args[]=19.2.4
Minified React error #527 ... args[]=19.2.8&args[]=19.2.6
Cannot read properties of null (reading 'useState')
```

Evidence from the repro (`evidence/force-527.json`, `evidence/share-owners.json`): the share cache is host-owned 19.2.8 throughout; #527 is thrown while evaluating a lower-patch local `react-dom` fallback against that shared `react`.

tonoizer:
## Summary
- Production eager local `react-dom` / `react-dom/client` fallbacks were evaluating against shared `__loadShare__` `react` when patches differ under caret `^19.2.x`, causing React **#527**.
- When an importer is inside shared package P's `__prebuild__` graph and `source` is another shared package P depends on, resolve to the sibling `__prebuild__` instead of `__loadShare__` so the unused local fallback stays self-consistent. Provider rebinding via loadShare is unchanged.
- Closes https://github.com/module-federation/vite/issues/1278

## Repro
https://github.com/tonoizer/react-mf-patch-skew-repro

## Related
#1173, #1197, #1243, #1247/#1249, #1256, #1275

## Test plan
- [x] Unit tests in `pluginProxySharedModule_preBuild.test.ts` (FAIL→PASS): `react-dom` / `react-dom/client` → `__prebuild__react`; app imports still `__loadShare__`; `import:false` stays loadShare
- [x] Full unit suite: 1332 passed, 1 skipped
- [ ] Maintainers: optional check against https://github.com/tonoizer/react-mf-patch-skew-repro after a canary/plugin bump


gioboa (changes requested):
This PR introduces a production-build regression on Vite 6 and 7.

Added integration test: `integration/shared-deps.test.ts` with a fixture that:

- imports `react` and `react-dom/client`;
- uses React 19.2.4 CommonJS packages;
- shares the React family as singletons;
- uses `shareStrategy: 'version-first'`;
- uses `hostInitInjectLocation: 'entry'`.

During bundling, `react-dom/client` imports `react`. This PR redirects that dependency to the virtual `__prebuild__react` module. Vite's CommonJS transform then creates a `?commonjs-proxy` importer whose bare `react` import is unresolved:

```text
[vite:load-fallback] Could not load react
imported by ...__prebuild__react__prebuild__.js?commonjs-proxy
ENOENT: no such file or directory, open 'react'
```

Results:

- Vite 6.4.3: fails
- Vite 7.3.6: fails
- Vite 8.2.0: passes

Expected: all supported Vite versions build successfully.

gioboa (commented):
That's amazing 👏 
Can you double check why React it's failing in the [frameworks pipeline](https://github.com/gioboa/module-federation-vite-examples/actions/runs/34676809764/job/103508003595#step:10:206)?

tonoizer:
@gioboa Reproduced — this was our regression, not the examples config / flaky Playwright.

Vite 8 DEV on the react example died before paint:

```
The requested module '/@fs/.../react@19.2.4/node_modules/react/jsx-runtime.js' does not provide an export named 'jsx'
```

`shouldKeepSharedImportLocal` treated **every** shared package with a `react` peerDep as a coherent fallback. `@mf-vite-example/shared-ui` (tsc emit: `import { jsx } from "react/jsx-runtime"`) was rewritten to the raw CJS entry. Vite 8 DEV then served it via `/@fs/` without named ESM exports → empty `#root` → `expect(page.getByText("I'm the host app")).toBeVisible()` timeout. Host text is independent of the remote, so this is a host boot crash, not a remote load flake. Vue/Solid/Svelte never take this compiled-shared-UI path; TanStack skips the DEV e2e step.

Fix in `5aa2ec3`: keep the sibling-local edge **only** for `react-dom → react` (patch-skew #527 + Vite 6/7 `?commonjs-proxy` ENOENT). Workspace shared packages stay on `__loadShare__`.

Locally validated against the failing examples path (`pnpm react:dev` / Vite 8.1.0 / artifact `@269031f` + this patch):
- host + remote render, counters increment independently, no #527 / dual React
- unit suite: 1330 passed, 1 skipped
- react-dom local + `?commonjs-proxy` tests green
- `integration/shared-deps` + React identity / react-skew browser tests green

Please re-run the frameworks pipeline against `npm i https://pkg.pr.new/@module-federation/vite@1279` once pkg.pr.new picks up `5aa2ec3`.

gioboa (commented):

Image

gioboa (approved):
Thanks @tonoizer great PR 👏

PR #1282, opened 2026-09-13
fix: leave consume-only loadShare wrappers to the bundler's own chunking

### LINKED ISSUE #1281: [Bug]: every `import: false` share gets an isolated loadShare chunk — one request per share — though its wrapper has no fallback to isolate
### Describe the bug

`mfChunkName` gives every loadShare wrapper a chunk of its own (`codeSplitting.groups` on Rolldown, `manualChunks` on Rollup). The isolation exists so that a wrapper's fallback — the share's own source, statically imported — cannot close a top-level-await cycle across wrappers.

A consume-only share (`import: false`) has no fallback in its wrapper: the generated module is a few `let` bindings filled from the runtime's shared cache (`generateDeferredHostProvidedExports`). Nothing in it can take part in such a cycle, yet it is isolated all the same — so a container that consumes N shares from a host or a provider ships N extra ~1.2 KB chunks and pays N extra requests on every load.

At our scale (a host consuming ~110 shares from a provider container, seven embedded remotes consuming the same table) that is ~110 requests per container, all of them for files under 2 KB.

### What the repro shows

Three Vite containers on one page: a provider (14 eager shares with fallbacks), a host and an embedded widget, both `import: false` for the same 14 shares. `npm run check` serves `dist/` and counts what a headless Chromium fetches per container:

| | 1.21.6 | with the fix |
|---|---|---|
| host | 21 requests | 17 |
| widget | 11 requests | 8 |

The files that disappear are exactly the `_virtual_mf___loadShare__<pkg>__loadShare__` chunks of the shares each container uses, ~1.2 KB each. Same bytes, same page, no console output, every part still renders (`RESULT OK`).

### Expected behavior

A wrapper with nothing to isolate is left to the bundler's own chunking, as any other tiny module would be: `mfChunkName` returns `null` for a share whose `shareConfig.import === false`, on both the Rolldown and the Rollup path, and only the wrappers that carry a fallback keep a chunk each. The `loadShare-eager` group from #1268 is unaffected.

PR follows.

### Version

@module-federation/vite 1.21.6 (and main @ 48e9a33), Vite 8.3.0 / Rolldown, Node 22.14, macOS

### Reproduction

### Relevant log output

```shell
$ npm run check      # 1.21.6
js fetched:
  host       21 requests     119 KB
  provider   10 requests     257 KB
  widget     11 requests      93 KB
  wremote     2 requests       6 KB

$ npm run check      # with the fix
js fetched:
  host       17 requests     119 KB
  provider   10 requests     257 KB
  widget      8 requests      93 KB
  wremote     2 requests       6 KB
RESULT OK
```

### Validations

marksnode:
Closes #1281.

`mfChunkName` isolates every loadShare wrapper into a chunk of its own so a fallback's static imports cannot close a top-level-await cycle across wrappers. An `import: false` share has no fallback in its wrapper — the generated module is a few live bindings filled from the shared cache — so there is nothing to isolate, and the isolation only turns each consumed share into a request of its own (~110 per container in the layout that prompted this, all under 2 KB).

This returns `null` for a wrapper whose share is `import: false`, on both the Rolldown (`codeSplitting.groups`) and the Rollup (`manualChunks`) path, so the bundler places it as it would any small module. Wrappers that carry a fallback are untouched, and so is the `loadShare-eager` group from #1268.

Measured on the repro (https://github.com/marksnode/mf-vite-repro-shared-provider, `npm run check`): host 21 → 17 requests, embedded widget 11 → 8, same bytes, no console output, every part still renders. In the application behind it a consume-only host went from ~130 to ~29 emitted chunks.

The integration test `applies host-provided exports synchronously when cache is populated` located the wrapper by chunk name; it now finds it by the code it generates, since that wrapper is no longer a chunk of its own.

- [x] unit test: `leaves consume-only loadShare wrappers to the bundler instead of a chunk each`
- [x] full suite green (the one failure, `cleans a failed lazy import before a successful retry`, fails on `main` as well)

### DIFF
--- a/integration/shared-cache-deferred-exports.test.ts
+++ b/integration/shared-cache-deferred-exports.test.ts
@@ -195,15 +195,14 @@ describe('remote loadShare (import: false — host-provided)', () => {
       mfOptions: REMOTE_IMPORT_FALSE,
     });
 
+    // An import:false wrapper has no fallback to isolate, so it is not a chunk of its
+    // own — find it by the code it generates rather than by a chunk name.
     const loadShareChunk = output.output
       .filter((c) => c.type === 'chunk')
-      .find((c) => c.fileName.includes('__loadShare__'));
+      .find((c) => (c as { code: string }).code.includes('__mfApplyHostProvidedExports'));
 
     expect(loadShareChunk).toBeDefined();
     const code = (loadShareChunk as { code: string }).code;
-
-    // Must use __mfApplyHostProvidedExports
-    expect(code).toContain('__mfApplyHostProvidedExports');
 
     // The else branch (cache hit) must apply synchronously
     // Structure: if (exportModule === void 0) initPromise.then(...) else { __mfApplyHostProvidedExports(exportModule) }
@@ -212,7 +211,7 @@ describe('remote loadShare (import: false — host-provided)', () => {
     expect(elseMatch).not.toBeNull();
 
     // Must NOT register a pending share load in the else branch
-    const elseIndex = code.lastIndexOf('else');
+    const elseIndex = elseMatch!.index!;
     const afterElse = code.slice(elseIndex, elseIndex + 200);
     expect(afterElse).not.toContain('pendingShareLoads');
   });
--- a/src/__tests__/index.test.ts
+++ b/src/__tests__/index.test.ts
@@ -925,6 +925,32 @@ describe('module-federation-esm-shims', () => {
     expect(mfWarn).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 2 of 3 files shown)


marksnode:
@gioboa thanks for the manifest fix and the spec — and sorry for the noise: I amended my own spec fix for the same failure and force-pushed a minute before I saw your commit, which briefly dropped it. The branch is back to your `a89a768` on top of `af07e3d`, nothing of mine added on top; your version of the spec supersedes what I had.

gioboa (commented):

Image

gioboa (approved):
Thank you for your fix @marksnode 💯

PR #1283, opened 2026-09-14
fix: exclude every instance's remote entry from parse tracking

### LINKED ISSUE #1285: [Bug]: two `federation()` instances in one build always wait for `moduleParseIdleTimeout` (parse trackers wait on each other's remote entry)
## Describe the bug

When a single Vite build contains more than one `federation()` plugin instance, `vite build` always stalls until `moduleParseIdleTimeout` fires and logs `forcing resolve`, even though the module graph is complete and nothing is actually stuck. The pending module each instance reports is always the **other** instance's remote entry.

This affects a host that consumes remotes packaged as independent libraries, where each library owns its Module Federation configuration and contributes its own `federation({...})` plugin (different `remotes` / `shared` / `filename`) and the host config spreads both plugin arrays. Since #926 scoped the virtual module ids per owner, that configuration builds correctly — it just pays the idle timeout on every build and emits a warning that reads like the remote entry may be incomplete.

I have a fix and regression tests ready: #1283.

### Root cause

`pluginModuleParseEnd` receives an `excludeFn` from `federation()` that only excludes the instance's **own** ids:

```ts
// src/index.ts
const moduleParsePlugins = pluginModuleParseEnd(
  (id: string) => {
    return (
      id.includes(getHostAutoInitPath(options)) ||
      id.includes(getPendingSharesPath(options)) ||
      id.includes(remoteEntryId) ||            // <- only this instance's remote entry
      id.includes(virtualExposesId) ||
      ...
    );
  },
  ...
);
```

`remoteEntryId` is `virtual:mf-REMOTE_ENTRY_ID:<internalName>__<filename>`, so instance A's exclude function does not match instance B's remote entry id and vice versa. That closes the following cycle:

1. Every instance emits its own remote entry chunk (also pure hosts, for the shared scope init), so both remote entry ids are loaded during the build.
2. `pluginModuleParseEnd`'s `parseStart.load` hook runs for every id in the build, in every instance. Instance A's `load` sees B's remote entry, `excludeFn` returns `false`, and A adds it to its `parseStartSet`.
3. `pluginProxyRemoteEntry`'s `load` for B's remote entry returns `getParsePromise().then(generateRemoteEntry)`, i.e. B's remote entry is only generated once **B's** parse promise resolves.
4. B's parse promise cannot resolve, because B is tracking A's remote entry, which is waiting on A's parse promise, which is waiting on B's remote entry being parsed.

This is a circular wait rather than a slow build, so only the idle timer ends it; `moduleParseIdleTimeout` just decides how long the build stalls. With equal timeouts on both instances, both warn. With different timeouts only the shorter one warns, and the build ends after the shorter timeout: once that instance force-resolves, its remote entry is generated and parsed, which unblocks the other instance so it completes normally with `graph-complete`. I verified both cases against the reproduction below (1s/1s produced two warnings, 1s/3s produced one warning from the 1s instance and still finished in about 1.2s).

`vite dev` is unaffected, because `pluginModuleParseEnd` is `apply: 'build'`.

## Version

Reproduced on `main` (a88137c), with Vite 8.2.0 (Rolldown), Node 24.11.1, pnpm 10.28.2.

The exclude function shown above is unchanged in the released 1.21.6, and `pluginProxyRemoteEntry` still returns `getParsePromise().then(generateRemoteEntry)` from `load`, so released versions should be affected in the same way.

## Reproduction

Two `federation()` instances in one config are enough. The remotes do not need to be reachable, because the stall happens at build time before any network access; `moduleParseIdleTimeout` is lowered only to keep the reproduction short.

```ts
// vite.config.ts

export default defineConfig({
  plugins: [
    federation({
      name: 'host',
      filename: 'remoteEntry-a.js',
      dts: false,
      moduleParseIdleTimeout: 1,
      remotes: {
        remote_a: {
          type: 'module',
          name: 'remote_a',
          entry: 'https://example.com/remote-a/remoteEntry.js',
        },
      },
    }),
    federation({
      name: 'host',
      filename: 'remoteEntry-b.js',
      dts: false,
      moduleParseIdleTimeout: 1,
      remotes: {
        remote_b: {
          type: 'module',
          name: 'remote_b',
          entry: 'https://example.com/remote-b/remoteEntry.js',
        },
      },
    }),
  ],
});
```

```sh
vite build
```

Expected: the build finishes as soon as the module graph is parsed, with no warning.

Actual: the build stalls for `moduleParseIdleTimeout` seconds after the last real module is parsed, then each instance force-resolves and warns about the other instance's remote entry.

As an executable reproduction, #1283 adds `integration/multi-instance-parse-deadlock.test.ts`, which runs exactly this build through `vite build` and asserts that no `moduleParseIdleTimeout` warning is logged. Against `main` without the fix it fails with two warnings; with the fix it passes.

## Relevant log output

Captured from that integration build on `main` without the fix, with `moduleParseIdleTimeout: 1` on both instances:

```shell
[Module Federation] moduleParseIdleTimeout: no module activity for 1s, forcing resolve. Some shared/remote dependencies may be missing. Consider increasing moduleParseIdleTimeout. Tracked modules: 7/7. Last loaded:  vite/modulepreload-polyfill.js. Last parsed:  vite/modulepreload-polyfill.js. Pending modules: virtual:mf-REMOTE_ENTRY_ID:__mfe_internal__host__remoteEntry-b_js
[Module Federation] moduleParseIdleTimeout: no module activity for 1s, forcing resolve. Some shared/remote dependencies may be missing. Consider increasing moduleParseIdleTimeout. Tracked modules: 7/6. Last loaded:  vite/modulepreload-polyfill.js. Last parsed:  vite/modulepreload-polyfill.js. Pending modules: virtual:mf-REMOTE_ENTRY_ID:__mfe_internal__host__remoteEntry-a_js
```

Each instance reports the other instance's remote entry. On this tiny fixture the build itself takes about 1.2s instead of about 0.1s, i.e. essentially all of it is the idle timeout; on a real application the cost is whatever `moduleParseIdleTimeout` is set to, on every build.

Note that the counters can read `7/7` (or even `7/6`) while a module is still pending: `parseEndSet` also receives ids that never went through the tracked `load` hook, so `parseEndSet.size >= parseStartSet.size` does not imply that `parseStartSet` is drained.

## Proposed fix

Exclude every remote entry virtual module from parse tracking, not only the instance's own one. A remote entry's content is generated *from* the parse result rather than contributing to it, so tracking another instance's remote entry can only create this cycle. The existing unit and integration suites pass with it excluded.

```diff
 const moduleParsePlugins = pluginModuleParseEnd(
   (id: string) => {
     return (
       id.includes(getHostAutoInitPath(options)) ||
       id.includes(getPendingSharesPath(options)) ||
+      id.includes(REMOTE_ENTRY_ID) ||
       id.includes(remoteEntryId) ||
       id.includes(virtualExposesId) ||
```

A broader variant would be to exclude every `virtual:mf-` id owned by a different federation instance — host auto-init, virtual exposes and pending shares are instance-scoped in the same way — but the remote entry is the only one observed to block the graph, since it is the only one whose generation waits on the parse promise.

Raising `moduleParseIdleTimeout` does not help; it only lengthens the stall. Disabling both timers (possible since #1222) would leave the cycle unbroken, so the build should hang on `Unfinished hook action(s) on exit: (proxyRemoteEntry) load ...`, as in #132.

Closest related issues: #926 (multi-instance isolation), #437 / #443 (idle timeout), #132 / #245 (hang when the parse promise never resolves).

kmkzt:
Fixes #1285

Excludes remote entries from parse tracking by their shared `virtual:mf-REMOTE_ENTRY_ID` prefix instead of the instance-scoped id, matching how `virtual:mf-localSharedImportMap` is already excluded. Both virtual modules have the same `${PREFIX}:${owner}` id shape and both are generated behind their owner's `parsePromise`, but only the local shared map was excluded for every instance — so with two `federation()` instances in one build, each tracked its sibling's remote entry and neither could generate it, and the build stalled until `moduleParseIdleTimeout` force-resolved it.

Since `getRemoteEntryId()` returns `${REMOTE_ENTRY_ID}:${scopeKey}`, the previous check was fully subsumed by the prefix and is removed rather than kept alongside.

The regression tests also stop racing a wall-clock budget to decide whether the barrier resolved, which is what failed CI on this branch; they advance the barrier's completion check with fake timers instead. That changes the existing local shared map test too, since it shares the same helper.


gioboa (approved):
Thanks @kmkzt nice catch and fix 👏

gioboa (commented):

Image

kmkzt:
@gioboa Thank you for reviewing and merging the PR 🙇

PR #1286, opened 2026-09-14
fix: defer singleton fallbacks in consuming remotes

### LINKED ISSUE #1284: [Bug]: a container that consumes any remote bundles its singleton fallbacks statically, so embedded in a host it downloads shares the host already provides
### Describe the bug

`writeLoadShareModule` decides per share whether the wrapper imports its local fallback statically or lazily (`usesDeferredSingletonFallback`). For a non-workspace singleton in a production build the lazy form is taken only when `isRemoteOnlyContainer(options)` holds — exposes present and **no** remotes at all.

A remote that also consumes a remote (an independently delivered app that lazy-loads a sub-remote, or one using the common self-reference pattern `import('<self>/index')` as its own async boundary) therefore gets the static form for every singleton it shares: the wrapper chunk carries the fallback, the fallback is fetched together with the wrapper, and it is fetched even when a host's copy has already won the singleton — the negotiation decides which module is *used*, not which bytes are *downloaded*.

In the application behind this report an independently delivered calendar app shares a design-system package with its host (`singleton: true, requiredVersion: '*'`, with a fallback so it can also run standalone). Embedded in the host, which provides that package, the calendar still downloads its own 3.8 MB copy on every page load — the largest duplicate left on the page.

### What the repro shows

`sibling/` is a remote with exposes, a remote table of its own and four singletons with a fallback each; the host embeds it after loading a provider that has all four. `npm run check` lists what the sibling fetched:

| wrapper | with its remote table | rebuilt with `SIBLING_NO_REMOTES=1` |
|---|---|---|
| `__loadShare__dayjs__loadShare__.js` | 9.3 KB (fallback inside) | 2.7 KB |
| `__loadShare__react__loadShare__.js` | 11.3 KB | 4.0 KB |
| `__loadShare__uuid__loadShare__.js` | 12.4 KB | 2.8 KB |

Same container, same shares, same host; `sibling ok` renders in both builds. The only difference is that the remote table flips `isRemoteOnlyContainer`, and with it the fallbacks from lazy to static.

### Expected behavior

For a production build the fallback of a singleton share stays lazy whether or not the container consumes remotes: the `pendingShares` barrier already loads a lazy fallback before the container's own code evaluates when nobody else provides the share, so the static form buys nothing there, and when a host does provide the share the lazy fallback is simply never fetched. If the remote-only restriction guards a case I am not seeing, an opt-in per share (say `shareConfig.deferredFallback`) would let an app take the lazy form knowingly. Happy to send a PR for whichever shape you prefer.

### Version

@module-federation/vite 1.21.6 (and main @ 48e9a33), Vite 8.3.0 / Rolldown, Node 22.14, macOS

### Reproduction

### Relevant log output

```shell
$ npm run check         # sibling with its remote table
sibling fetched:
  LS:dayjs__loadShare__.js-C6faI5N8.js 9.3 KB
  LS:react__loadShare__.js-CNFPVPOV.js 11.3 KB
  LS:uuid__loadShare__.js-DzMSLwbd.js 12.4 KB
sibling text: sibling ok 2026

$ SIBLING_NO_REMOTES=1 npx vite build --config sibling/vite.config.js && npm run check
sibling fetched:
  LS:dayjs__loadShare__.js-B8t7Lunj.js 2.7 KB
  LS:react__loadShare__.js-DkFA2Ux7.js 4.0 KB
  LS:uuid__loadShare__.js-CAZ3qxXN.js 2.8 KB
sibling text: sibling ok 2026
```

### Validations

gioboa:
Closes #1284

A remote that also consumes remotes bundled singleton fallbacks statically, downloading shares its host already provides. Gate the build-time lazy fallback on having exposes instead of remote-only.


gioboa (commented):
@marksnode Could you test this pre-release?

marksnode:
@gioboa Yes, I'm getting started.

marksnode:
Tested the pre-release (`@module-federation/vite@1286`, `8bf2086`) on the app that prompted #1284; a full production build (10 containers: a provider with 275 eager singletons, eight consume-only apps, and the independently delivered calendar that shares 9 singletons with a fallback each and consumes remotes of its own), served locally and driven logged in.

**Calendar embedded in the host:** 28 requests / 1.8 MB, down from 20 / 4.3 MB on 1.21.5 — its 3.8 MB design-system fallback is no longer fetched, the host's copy is used. **Calendar standalone** (nobody provides the shares): renders correctly on three runs, no page errors, the fallbacks load through the entry. Whole logged-in page: 120 requests / 18.5 MB vs 112 / 20.9 MB before. Every other entry point (host, embedded widget, seven sub-apps) unchanged, no `[Module Federation]` output anywhere.

One note for the record: I had tried the narrower change alone (only the `usesDeferredSingletonFallback` clause) and the standalone calendar broke with `TypeError: o is not a function` in `useSyncExternalStoreWithSelector` — `use-sync-external-store/shim/with-selector` inside one lazily loaded fallback snapshots React's hooks at evaluation, and the lazy fallbacks load concurrently. Your version also widens `usesEntryInjectedRemoteFallback`, so those singletons take the entry-injected form and load in dependency order — and the crash does not reproduce. The repro's `sibling/` now includes that CJS shim as one of its shares and passes standalone as well.

While testing this I also ran into an unrelated webpack-interop problem (a webpack remote consuming `react` with `eager: true`), filed separately as #1287 with a PR. Looks good from here — thank you for the quick turnaround.

gioboa (commented):
Thanks for your check @marksnode 🙏 appreciated

gioboa (commented):

Image

marksnode:
One data point from lifting this onto a release: applied on top of 1.21.5 alone (the lib/index.js diff of this commit, everything else 1.21.5), the deferred fallback breaks the standalone page of the remote that consumes: `TypeError: o is not a function` at boot, a CommonJS shim read before its snapshot. On main (06cd0b4) the same remote is fine, so the fix depends on something that landed in between (#1275 / #1279 / #1282 / #1283 are the candidates). Nothing to change here, just a note for anyone backporting: take it with the release, not as a single hunk. The win for us is 4.4 MB -> 0.8 MB of calendar per page load, so a 1.21.7 carrying this would be very welcome.

gioboa:
Let's wait the resolution of all the issues and then we can drop a new release.

PR #1288, opened 2026-09-14
fix: hand a loaded share back synchronously from get()

### LINKED ISSUE #1287: [Bug]: a webpack remote that consumes a share eager: true cannot take it from a Vite container once the share is loaded (Shared module is not available for eager consumption)
### Describe the bug

Every share a Vite container registers in the scope has an `async get()`, so `get()` always returns a promise. A webpack remote built with `eager: true` for that share calls `entry.get()` synchronously during module evaluation and needs the factory back; older webpack codegen throws

```
Error: Shared module is not available for eager consumption: 52668
```

webpack would fall back to its own copy, but its `register` only replaces a scope entry that nobody has `loaded` (`!o || !o.loaded && (...)`), and the Vite container marks its entry `loaded = true` the moment a host resolves the share — so a webpack remote loaded after that point has no way out: the entry it must use is the Vite one, and that one answers with a promise.

This surfaces as soon as the Vite side holds the shares that a webpack remote consumes eagerly — for us a provider container whose singletons the host resolves at boot, then a webpack QM app (built with our previous webpack setup, `eager: true` for `react`) loaded on navigation. It does not surface while the host's own entry stays unloaded (webpack then wins the register race and uses its own copy), which is why a fallback-per-app layout happened to work.

### What the repro shows

The public repro's webpack remote (`wremote/`, webpack 5.110) does not show it: current webpack resolves eager consumes during `init`. The remote that does is an older-webpack build I cannot publish (146 KB entry, `eager: true` for react, uniqueName `@bpinc/omni-qm`); the README describes the setup (`local/omni-qm`, `OMNIQM=1`). The key facts are in the generated code either way: `usedShared[pkg].get` is `async`, and `loaded` is set on the first call.

### Expected behavior

Once a share has been loaded, `get()` returns its factory synchronously — the same factory the promise resolved to — so a webpack `eager: true` consumer gets a function. The runtime awaits `get()` anyway, so a plain factory is as good to it as a promise; nothing changes for a share that has not loaded yet.

### Version

@module-federation/vite 1.21.6 / main @ f253ff4, @module-federation/runtime 2.9.0, Vite 8.3.0 / Rolldown, Node 22.14, macOS; consumer: webpack 5 ModuleFederationPlugin (older codegen with the eager-consumption throw)

### Reproduction

### Relevant log output

```shell
Error: Shared module is not available for eager consumption: 52668
    at forEach.o.m.<computed> (https://localhost:8080/omni-qm/remoteEntry.js:2:145193)
    at o (https://localhost:8080/omni-qm/remoteEntry.js:2:139653)
    at 73335 (https://localhost:8080/omni-qm/231.89fa1d16.chunk.js:2:619526)
    ...
    at https://localhost:8080/agent/static/js/dist.4Ro5OGwX.chunk.js:4:28543   <- the host's federation runtime loading the remote

// the webpack side (omniQm remoteEntry.js, deminified):
// 52668: () => h("default", "react", true /* eager */, [1,18,3,1], () => () => o(14041) /* own fallback */)
// register: (!o || !o.loaded && (!r != !o.eager ? r : u > o.from)) && (l[n] = {get: t, from: u, eager: !!r})
// scope entry from the Vite provider at that moment: {version: '18.3.1', from: 'sharedProvider', loaded: 1, get: async () => ...}
```

### Validations

marksnode:
Closes #1287.

A share's scope entry had `async get()`, so `get()` always returned a promise. A webpack remote built with `eager: true` calls `get()` synchronously and needs the factory; older webpack codegen throws `Shared module is not available for eager consumption` when it gets a promise, and webpack cannot fall back to its own copy because it only replaces an entry nobody has `loaded` — which a Vite entry is from the host's first resolution on.

`get()` is now a plain method. Once the share is loaded it returns `lib`, the factory runtime-core itself stores on the entry, synchronously. While the load is in flight it keeps the pending promise on `loading` and hands the same promise to every caller, so concurrent calls share one import and one factory. The React-family fast path sets `lib` the same way. The runtime awaits `get()` regardless, so nothing changes for it.

Note: an `import: false` entry now throws synchronously from `get()` instead of returning a rejected promise. That is what an eager consumer needs, and `await share.get()` callers still see the error as a rejection.

Verified against the real older-webpack remote that hit this (a QM app consuming `react` eagerly from a Vite provider, described in #1287): with this change `import('omniQm')` resolves and the app mounts; on `main` it throws. Unit tests: `hands a loaded share back synchronously from get() once it has loaded` (promise on the first call, the same factory synchronously on the second, `lib` set) and `hands every get() caller the same pending promise while the share is loading` (two calls before the first resolves: one import, one promise, one factory).


gioboa (commented):
Thanks, the direction is right. Three small things before merging:

1. get() still returns a Promise while the import is in flight (src/virtualModules/virtualRemoteEntry.ts:296)

loaded = true is set before pkgDynamicImport() resolves, but the factory is only cached afterwards. A webpack/rspack eager: true consumer that calls get() in that window still gets a Promise, and each concurrent call starts its own import with its own {...res} copy. Cache the pending promise on loading and return it on re-entry:
```
const share = usedShared[${toSafeJsLiteral(key)}]
if (share.lib) return share.lib
if (share.loading) return share.loading
// ...cachedSingleton branch sets share.lib and returns it...
share.loading = (async () => {
  // existing load body
  share.lib = function () { return exportModule }
  share.loaded = true
  share.loading = undefined
  return share.lib
})()
return share.loading
```
2. Prefer lib over __mfLoadedFactory. Runtime-core stores the resolved factory on lib, and this file already reads provider.lib (:850,
:1471). No need for a second private field.

3. import: false now throws synchronously instead of rejecting. That is correct for eager consumers and fine for await share.get() callers, but please note it in the PR description.

No caller guard is needed: returning the factory once-scope contract that webpack, rspack and runtime-core all handle.

A test for two get() calls before the first resolves (same promise, same factory) would cover point 1.

marksnode:
Thanks, all three addressed in 95a174f:

1. `get()` returns `lib` once it is set and the pending promise from `loading` on re-entry, so concurrent callers share one import and one `{...res}` copy; `loading` is cleared when the factory lands.
2. `__mfLoadedFactory` is gone, the factory lives on `lib`.
3. The description now notes that `import: false` throws synchronously from `get()`.

Added `hands every get() caller the same pending promise while the share is loading`: two `get()` calls before the first resolves, same promise, same factory, one import.

gioboa (changes requested):
Nit: if pkgDynamicImport() rejects, loading keeps the rejected promise and loaded stays true, so get() never retries. Wrap the load in try/finally resetting loading, and set loaded only after success.

marksnode:
Good point, fixed in 0afd468: the load runs inside try/finally that clears `loading`, and `loaded` is set only after the factory is in place, so a rejected import leaves the entry as it was and the next get() imports again. Test added: `retries the import after a failed load instead of pinning the rejection` (first import rejects, `loading`/`loaded`/`lib` are back to their initial state, the second get() resolves to the factory, two imports in total).

gioboa (approved):
Thanks @marksnode nice one 🚀

PR #1290, opened 2026-09-14
fix: always proxy a consume-only share, even from its own dependency

### LINKED ISSUE #1289: [Bug]: consume-only share (import: false) is inlined when imported from one of its own dependencies
### Describe the bug

A share configured with `import: false` is meant to come from the share scope only: the consumer must never bundle its own copy. The `proxyPreBuildShared:resolve-shared-loadShare` resolver honours that for ordinary importers, but it keeps an "ordinary edge" when the importer is one of the shared package's own dependencies (the package-level cycle rule from #1203 / #1210 / #1227):

```ts
const importerPackage = getSharedPackageFromFile(importer, shared);
if (importerPackage === getPackageName(key)) return;
if (importerPackage) {
  // isSharedPackageRuntimeDependency / isSharedPackageDependency
  if (keepsOrdinaryEdge) return;
}
```

That rule exists so the loadShare glue never lands inside a package's own evaluation cycle, and it is right when a local fallback exists. With `import: false` there is no fallback the edge could stay coherent with: the ordinary edge resolves to the package source on disk and the bundler inlines it. The consumer ends up with a private copy of a share it declared it would only consume.

In our monorepo (a provider container that bundles ~275 shares eagerly, every other app `import: false`) the agent host inlined 71 shared packages this way, including the DI container package. The host then ran two containers, and the first modal to open died with `State "modalsState" is not registered in the container`. The plugin gave no warning: the build is green, the copy is just there.

### Expected behavior

A consume-only share is always proxied through the loadShare wrapper, whatever the importer. The cycle rule only applies when the share has a local fallback (`import !== false`).

### Version

1.21.5 (rule unchanged on main, f253ff4); vite 8 / rolldown; node 22.14.0

### Reproduction

Unit test in the linked PR: the existing vue -> bridge -> vue cycle fixture with shared.vue.shareConfig.import = false (returns undefined on main, the loadShare module with the fix)

### Relevant log output

```shell
Runtime, the first modal opened in the host:
Error: State "modalsState" is not registered in the container

Build-level check (consumer app, share is import: false), platform-runtime source files inside the consumer's chunks:
  main:  71 shared packages inlined, 'not registered' string present in the host chunk
  fixed: 0 copies, host chunk 1.09 MB instead of 2.4 MB
```

### Validations

marksnode:
## Description

Fixes #1289.

The `proxyPreBuildShared:resolve-shared-loadShare` resolver keeps an ordinary edge when a shared package is imported from one of its own dependencies (the package-level cycle rule from #1203 / #1210 / #1227), so the loadShare glue never lands inside the package's own evaluation cycle. That rule presumes a local fallback. For a consume-only share (`import: false`) there is none: the ordinary edge resolves to the package source on disk and the bundler inlines a private copy of a share the consumer declared it would only consume. In a provider-container layout that gave the host two DI containers and `State "modalsState" is not registered in the container`.

This PR skips the cycle rule for `shareConfig.import === false`: a consume-only share is always proxied.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Tests added: `always proxies a consume-only share, even from a dependency that would keep an ordinary edge` (the `vue -> bridge -> vue` fixture with `import: false`; fails on `main`, passes here)
- [x] `pnpm test`: 1431 passed, 1 pre-existing failure on `main` too (`shared-cache-deferred-exports > cleans a failed lazy import before a successful retry`)
- [x] `pnpm typecheck`, `pnpm format:check`


gioboa (approved):
Nice catch 👏 @marksnode 🙏

PR #1293, opened 2026-09-14
fix: run the plugin's preload lists through build.modulePreload.resolveDependencies

### LINKED ISSUE #1291: [Bug]: host-init preload seeds every __loadRemote__ wrapper, one request per remote on the first paint
### Describe the bug

`injectHostInitPreloads` writes a `<link rel="modulepreload">` into the host page for every chunk reachable from its seeds, and the seeds (`HOST_INIT_PRELOAD_CHUNKS`) include `isPreloadableVirtualMfChunk`, i.e. every chunk whose name contains `virtual_mf`. That takes in every `__loadRemote__` wrapper the build produced: one per remote, whether or not the page renders that remote.

On our host (10 remotes, `React.lazy(() => import('calendar'))`-style loaders behind routes and panels) that is 13 preloaded wrappers on the first paint; 8 of them belong to remotes the landing page never renders (supervision, omniQm, the exposes of calendar / knowledgeBase / …). Each wrapper is ~3 KB, so the cost is requests rather than bytes: 21 preload links, 13 of them wrappers.

`build.modulePreload.resolveDependencies` cannot filter them: it runs for Vite's own preload list, while these tags are written by the plugin in `generateBundle` after Vite's HTML plugin. We currently strip them with a `post` plugin that edits the html asset, keeping the wrappers our bootstrap imports right away (`./index`, `./register`).

A wrapper that is only reachable through a dynamic import does not need preloading: when the remote is first rendered, the wrapper is fetched one round trip ahead of the remote's entry, which dwarfs it.

### Expected behavior

Only wrappers statically reachable from the entry / `hostInit` are preloaded (the walk in `collectPreloadChunkFiles` already follows static imports, so dropping `__loadRemote__` chunks from the seeds would do it), or the plugin runs its list through `build.modulePreload.resolveDependencies` so a host can decide. Happy to send a PR for whichever you prefer.

### Version

1.21.5 (same code on main, 06cd0b4: `isPreloadableVirtualMfChunk` / `HOST_INIT_PRELOAD_CHUNKS`)

### Reproduction

Any host with a remote imported only through `React.lazy(() => import('remote'))`: build, open dist/index.html, count the `modulepreload` links whose href contains `__loadRemote__`

### Relevant log output

```shell
dist/index.html of the host (21 modulepreload links, 13 of them wrappers):
<link rel="modulepreload" crossorigin href="/agent/static/js/virtual_mf___mfe_internal__agent__loadRemote__calendar__mf_owner__1__loadRemote__.D4e0JGa7.chunk.js">
<link rel="modulepreload" crossorigin href="/agent/static/js/virtual_mf___mfe_internal__agent__loadRemote__campaigns__mf_owner__1__loadRemote__.BptCviuD.chunk.js">
<link rel="modulepreload" crossorigin href="/agent/static/js/virtual_mf___mfe_internal__agent__loadRemote__chatCenter__mf_owner__1__loadRemote__.dLbSPyCf.chunk.js">
<link rel="modulepreload" crossorigin href="/agent/static/js/virtual_mf___mfe_internal__agent__loadRemote__omniQm__mf_owner__1__loadRemote__.DkDBM2aZ.chunk.js">
<link rel="modulepreload" crossorigin href="/agent/static/js/virtual_mf___mfe_internal__agent__loadRemote__supervision__mf_owner__1__loadRemote__.7IcyQQv5.chunk.js">
... (8 more)
```

### Validations

marksnode:
## Description

Fixes #1291.

The plugin writes two preload lists of its own: the `<link rel="modulepreload">` tags `injectHostInitPreloads` puts into the host page, and the warmup list `appendRemoteEntryWarmup` appends to the remote entry and replays at runtime. Both are seeded from every `virtual_mf` chunk, so every `__loadRemote__` wrapper is preloaded — one request per remote on the first paint, whether or not the page renders that remote (13 wrappers on our host, 8 of them for remotes the landing page never shows).

Rather than the plugin deciding for everyone which wrappers deserve a preload, both lists now go through the user's `build.modulePreload.resolveDependencies`, the hook Vite already uses for its own preloads: `hostType: 'html'` with the page as `hostId` for the tags, `hostType: 'js'` with the remote entry as `hostId` for the warmup. Nothing changes for a build that does not configure the hook.

The plugin's own `resolveDependencies` wrapper (`src/index.ts`) is tagged with the user function it wraps, so a second federation config in the same build sees through it to the user's function instead of treating the first wrapper as the user's setting.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Tests: `modulePreloadPolicy.test.ts` (3) and two in `pluginAddEntry.test.ts`, one per list, with a user function that drops `__loadRemote__` wrappers
- [x] README: a short section under `codeSplitting` with the hook and an example
- [x] `pnpm test`: 1440 passed; the one failure (`shared-cache-deferred-exports > cleans a failed lazy import before a successful retry`) fails on `main` too
- [x] `pnpm typecheck`, `oxfmt --check`


gioboa (changes requested):
Thanks for the PR. I'd prefer the first option from the issue: drop __loadRemote__ wrappers from the seeds. Fixes both lists by default, no global state. Could you rework it this way?
```
const isPreloadableVirtualMfChunk = (name: string) =>
  name.includes('virtual_mf') &&
  !name.includes('__prebuild__') &&
  !name.includes('__loadShare__') &&
  !name.includes('__loadRemote__');
```

gioboa:
Vite already preloads every __loadRemote__ wrapper a page imports, statically or via its preload helper. Seeding them all only fetches unused remotes, so dropping them fixes both lists without global state.

@marksnode are you ok with this change?

gioboa (commented):

Image

gioboa (approved):
Thanks @marksnode

gioboa:
> @marksnode are you ok with this change?

If that turns out not to be the case, we can always go back.

marksnode:
Yes, that is the better cut: the wrapper of a remote the page renders is fetched the moment its import runs, so seeding it from the plugin only ever helped the unused ones. The one thing we lose is the head start for the two wrappers our bootstrap imports right away (the shared provider's ./index and the ./register exposes): they were in the page's preload list and now cost one round trip of 3 KB before the remote entry request starts. That is well below noise next to the provider entry itself, so fine by us. Thanks for taking it over, and for #1295.

Issue #1292, opened 2026-09-14
[Bug]: a remote pays ~340 KB and 9 requests of federation glue before its first module: runtime bundled per container, 99 KB consume-only share map, sub-kilobyte facade chunks


marksnode:
### Describe the bug

Measured on our agent desktop (Vite 8.2.1 / Rolldown 1.2.4, plugin 1.21.5 with #1273 and #1290 lifted; a provider container bundles 282 shares, every other app is consume-only, `shareStrategy: 'loaded-first'`). Loading one remote (`homePage`, one exposed page of 200 KB) costs, before its first application module arrives:

```
   0 ms   remoteEntry.js                              0.4 KB   var shim -> esmRemoteEntry
 126 ms   esmRemoteEntry.js                            91 KB   @module-federation/runtime/dist/helpers.js + the virtual entry
 248 ms   vite-preload-helper.*.js                      1 KB
 250 ms   dist.*.chunk.js                              65 KB   runtime-core (33 files) + sdk + error-codes + runtime
 386 ms   virtualExposes.*.js                           1 KB
 399 ms   _virtual_mf-localSharedImportMap_*.js        99 KB   282 consume-only stubs
 558 ms   register.*.chunk.js                           0 KB   facade re-exporting ./register
 861 ms   mf-shares.*.chunk.js                         78 KB   the loadShare wrappers (regrouped into one chunk on our side)
1055 ms   rolldown-runtime.*.js, defineProperty.*.js    0 KB   two helper chunks
1685 ms   exposes.*.chunk.js                            0 KB   facade re-exporting the exposes
1698 ms   HomePage.*.chunk.js                         200 KB   the first application module
```

9 requests and 340 KB of the remote's 13 requests / 568 KB are federation glue, in five sequential waves (1.7 s on localhost). Over the whole page (host + provider + 6 remotes) 70 of the 107 script requests are glue. Three things stand out.

[...]

PR #1294, opened 2026-09-14
perf: build consume-only share entries through one helper

marksnode:
## Description

Part of #1292 (item 2, the consume-only share map).

A consume-only share (`import: false`) has no local module, so its `usedShared` entry carries nothing key-specific beyond name, version, scope, `materialize` and the four `shareConfig` flags — yet `generateLocalSharedImportMap` emitted the full literal for each one (the `get()` body, `from`, `canLiveRebind`, `loaded`, the `shareConfig` object), ~350 bytes a key. A consumer with a few hundred shares pays 100 KB for that table, once per container on the page.

The map now defines two helpers once (`__mfConsumeOnly` for the entry, `__mfHostOnly` for the `importMap` thunk) and emits one call per consume-only key. The entry has the same shape and behaviour as before: same fields, `canLiveRebind: true` (which `import: false` already forced), `get()` still rejects with the `must be provided by host` error, `shareConfig.import: false`. A consume-only share with tree-shaking metadata keeps the literal, since its entry does carry key-specific data.

On our host layout (282 consume-only shares): `_virtual_mf-localSharedImportMap` 102,392 -> 40,274 bytes (5.1 -> 3.7 KB gzipped); the saving is mostly parse and memory, gzip already folds most of the repetition.

Note: #1288 changes the `get()` of shares *with* a local module in the same template; the two do not touch the same lines, whichever lands second rebases trivially.

## Type of change

- [x] Performance improvement (non-breaking)

## Checklist

- [x] Test `builds a consume-only entry through one helper with the shape the literal had` evaluates the generated map and checks every field and the rejection; the `canLiveRebind` test adjusted for the helper form
- [x] `pnpm test`: 1436 passed, the one failure (`shared-cache-deferred-exports > cleans a failed lazy import before a successful retry`) fails on `main` too
- [x] `pnpm typecheck`, `oxfmt --check`


gioboa (approved):
Nice improvement 🚀 
TY @marksnode

gioboa (commented):

Image

PR #1295, opened 2026-09-14
fix: allow provideExternalRuntime on exposing containers

gioboa:
Part of #1292 (item 1).

`provideExternalRuntime: true` threw on any container with exposes, so a host that also exposes modules could not publish the shared runtime-core and every remote bundled its own copy.
Drop the guard: the injected plugin is unchanged and externalRuntime remotes read the global lazily.

On the issue's repro: widget 90 KB → 29 KB, page 586 KB → 461 KB, all parts render. Tests, JSDoc and README updated.


gioboa (commented):

Image

marksnode:
Tried this on the layout behind #1292 (host with exposes providing, the shared-provider remote and five embedded remotes with externalRuntime), lib built from this PR: each externalRuntime remote drops its 65 KB runtime-core chunk to a 2 KB shim, the logged-in landing page goes 18.24 -> 17.67 MB and 147 -> 143 requests; the host, the provider and the embedded remotes render. Two notes from that run. (1) What is left per remote is `esmRemoteEntry.js` at 91 KB, which is `@module-federation/runtime/dist/helpers.js` inlined; externalRuntime covers runtime-core only, so the larger half of the per-container runtime stays duplicated. (2) An externalRuntime remote cannot open its own standalone page any more (`_FEDERATION_RUNTIME_CORE is missing`), which our remotes still do (a Home or Chat page reached directly). A lazy fallback when the global is absent - `import('@module-federation/runtime-core')` from a chunk of its own - would let one build serve both the embedded and the standalone case; without it externalRuntime is an all-or-nothing choice per container. Happy to open both as follow-ups to #1292 if you agree.

gioboa:
Thanks for the numbers. (1) 91 KB isn't the runtime, it's 3.5 KB total, please dump the chunk contents first. (2) Agreed, fallback belongs in the remoteEntry.js facade. Open both. :+1:

PR #1296, opened 2026-09-15
perf: drop redundant entry imports

gioboa:
Skip the static remoteEntry/virtualExposes imports in build: 
- Rollup already correct
- Rolldown removed two extra requests per remote


gioboa (commented):

Image

PR #1300, opened 2026-09-15
fix: check for `environment.config.consumer`

gu-stav:
Vite environment names are user-defined, but the load hook treated every named environment except `client` as server. Custom browser environments received an SSR-only static workspace-singleton import and lost lazy fallback behavior.

For example, this Vite config defines a browser environment named `federation`:

```ts
import {defineConfig} from 'vite'

export default defineConfig({
  environments: {
    federation: {consumer: 'client'},
  },
})
```

The old name check classified it as server because its name was not literally `client`.

The hook now checks `environment.config.consumer`, which identifies whether an environment runs in the browser or on the server. Client environments keep lazy fallback behavior; server environments receive the static SSR import. When the consumer is missing, the hook falls back to the environment name for older or incomplete contexts. When `this.environment` is absent, it uses root `build.ssr` to preserve Vite 5-7 SSR behavior.

The regression test uses `federation` with both client and server consumers and checks that only the server receives the static import.


gioboa (changes requested):
Thanks. Could you move the consumer check into remoteConsumerTarget.ts as a shared helper and reuse it in resolveRemoteConsumer too?

gioboa (approved):
Thanks, that's great 👏

PR #1301, opened 2026-09-15
fix: respect Vite environment consumer roles when resolving remote targets

gu-stav:
`resolveRemoteConsumer` used an environment's user-defined name to choose the client or server remote wrapper. A custom browser environment named `federation` was treated as server even with `consumer: 'client'`:

```ts
export default {
  environments: {
    federation: {
      consumer: 'client',
    },
  },
};
```

Module Federation generated the server remote wrapper instead of the client wrapper.

This change prefers Vite's `environment.config.consumer` when present because it defines the environment's semantic role. Contexts without `consumer` keep the existing name-based fallback, and disabling multi-environment mode still selects the unified wrapper.

Regression coverage uses the same custom name, `federation`, with both client and server consumers.


gu-stav:
@gioboa Sorry I closed this one because I wanted to share the utility function like you asked in the first PR. I think this fix is self-contained and I will open a refactoring PR on top of this one instead.

gioboa (commented):

Image

gioboa (approved):
Thanks @gu-stav you rock 💪

PR #1302, opened 2026-09-15
fix: shorten chunk names to fix ENAMETOOLONG

simonihmig:
## Problem

`packageNameEncode` turns shared-module specifiers (npm package names,
optionally with a deep import subpath, e.g. `@scope/pkg/deep/sub/export`)
into filesystem/module-id-safe segments via a 1:1, length-*increasing*
character substitution. For a long specifier, the result can exceed 255
bytes — `NAME_MAX` on macOS/Linux — once used as a filesystem path segment,
throwing `ENAMETOOLONG` and failing the build/dev-server outright.

Observed in production:

```
ENAMETOOLONG: name too long, stat '.../node_modules/virtual:mf:...__loadShare__..._mf_0_...__loadShare__.js/package.json'
```

The same overflow risk applies anywhere else `packageNameEncode` is called
directly: `loadShare` chunk names, `getTreeShakingSharedProviderName`'s
emitted asset path, and `virtualRemoteEntry.ts`'s owner-name encoding.

## Fix

`packageNameEncode`/`packageNameDecode` (`src/utils/packageUtils.ts`) now
have a length-safe fallback built in, so every caller gets it for free:

- If the substituted result fits under a 100-char threshold, it's returned
  unchanged — no behavior change for the common case.
- Otherwise, fall back to a **readable-prefix + content-hash** id: the
  first 84 characters of the substituted name plus a 16-hex-char sha256
  hash (no separator, so the full 100 chars go to the readable prefix). The
  hash-to-original mapping is kept in memory so `packageNameDecode` can
  recover the exact original specifier later.

100 isn't arbitrary: `VirtualModule#getImportId()` concatenates *two*
independently hashed, threshold-capped segments plus fixed tag/suffix
overhead into one id, and that must stay under 255 bytes even if both
segments hit the cap at once — which bounds the threshold to ~110; 100
keeps a margin under that ceiling.

## Testing

New unit tests in `packageUtils.test.ts` (threshold boundary, hashed-format
shape, round-trip decode, hash determinism/distinctness) and
`VirtualModule.test.ts` (end-to-end overlong shared-module specifier,
asserting the import id stays bounded and round-trips via `findName`).


gioboa (changes requested):
Long owner-scoped `mfName` gets hashed here, dropping the tag sentinel so `findName` fails. Encode `mfName` and `tag` separately.

Regression test (fails on this branch, add to `VirtualModule.test.ts`):

```ts
it('keeps the tag sentinel intact for a long owner-scoped mfName', () => {
  normalizeModuleFederationOptions({ name: 'analytics' });

  const scopeName = `@my-company/very-long-application-name-remote-shell-v2${MF_OWNER_INFIX}281474976710655`;
  const vm = new VirtualModule('react', '__loadShare__', '.js', scopeName);
  vm.writeSync('export default 1;');

  const importId = vm.getImportId();
  expect(importId).toContain('__loadShare__react__loadShare__');
  expect(VirtualModule.findName('__loadShare__', importId)).toBe('react');
  expect(VirtualModule.findModule('__loadShare__', importId)).toBe(vm);
});
```

simonihmig:
Thanks @gioboa for the quick feedback! This should have been addressed now. Please double check, as the whole PR has been AI assisted, so possible I might have missed other edge cases.

gioboa (commented):
Thanks @simonihmig nice fix 👍

PR #1303, opened 2026-09-15
fix: initialize federation in custom Vite client environments via host-init injection

gu-stav:
Custom client entries could skip the host-init bootstrap, allowing application code to run before Module Federation initialization. The plugin checked whether the environment was named `client`, so a client consumer named `federation` was excluded.

For example, this Vite configuration has 2 client consumers: the normal client environment with its implicit `index.html` entry, and a custom `federation` environment with a TypeScript entry:

```ts
import {federation} from '@module-federation/vite';
import {defineConfig} from 'vite';

export default defineConfig({
  plugins: [
    federation({
      name: 'host',
      remotes: {
        catalog: {
          type: 'module',
          name: 'catalog',
          entry: 'http://localhost:5001/remoteEntry.js',
        },
      },
    }),
  ],
  environments: {
    client: {
      consumer: 'client',
    },
    federation: {
      consumer: 'client',
      build: {
        rolldownOptions: {
          input: 'src/federation.ts',
        },
      },
    },
  },
});
```

`configResolved` is global and has no environment context, so the fix collects resolved inputs from every client consumer there. During `transform`, it uses Vite's `consumer` role to identify client code. This lets the custom entry receive the bootstrap that waits for Module Federation initialization before application code runs, while server consumers remain excluded.

The fix preserves the Vite 5-7 fallback when environment context is unavailable, the top-level build-input fallback, and the implicit `index.html` entry when another client consumer has an explicit input.


gioboa (approved):
nice fix @gu-stav 🚀

PR #1304, opened 2026-09-15
fix: defer non-singleton fallbacks in custom share scopes

gu-stav:
Non-singleton dependencies in a custom share scope load their local fallback before the remote container finishes initializing.

```ts
federation({
  name: 'remote',
  shareScope: 'custom',
  shared: {
    lit: {
      shareScope: 'custom',
      singleton: false,
      requiredVersion: '^3.3.2',
    },
  },
})
```

When 2 compatible remotes use this configuration, each remote can evaluate this static import before the shared scope is ready:

```js
import * as __mfLocalShare from 'lit'
```

The page then loads 2 copies of the dependency instead of reusing the compatible version already registered in `custom`.

This change removes the default-scope and singleton checks from remote fallback generation. Non-eager fallbacks that aren't consumed by a peer now wait for `initPromise` and use a dynamic import in every share scope. That gives the runtime time to initialize the configured scope and resolve a compatible shared module before loading the local fallback.


gioboa (approved):
Thanks @gu-stav 👏

PR #1310, opened 2026-09-16
fix: collect nested dynamic imports transitively

### LINKED ISSUE #1309: bundleAllCSS: option to record all CSS as async metadata without eager runtime injection
@module-federation/vite currently exposes two CSS behaviours, and neither covers federated SSR with lazy client-side injection:

- With `bundleAllCSS: false`, `expandExposeAssets`, `collectStaticChunks`, and `analyzeChunkAssets` only include one level of dynamic imports, so CSS for lazy routes nested two or more hops behind an expose root is missing from the manifest.
- With `bundleAllCSS: true`, `addCssAssetsToAllExports` records every bundle stylesheet as `css.sync` for every expose and shared entry, and the runtime eagerly injects all of it on every remote load.

Our SSR asset collector builds server stylesheet links from manifest `css.sync` plus `css.async`, so the first mode can make a server-rendered lazy route flash unstyled. The second mode preserves complete metadata but causes eager CSS injection and has no ownership information for shared entries.

Proposal: add a mode such as `bundleAllCSS: 'metadata'` or `cssInjection: 'none'` that keeps complete `addCssAssetsToAllExports` metadata, classifies dynamically reachable stylesheets as `css.async`, and leaves the runtime `cssAssetMap` empty. That would preserve SSR metadata while allowing Vite's preload path to deliver lazy CSS on the client.

gioboa:
Closes #1309

Manifest asset analysis stopped at the first dynamic import, so CSS and JS for lazy routes nested deeper were missing.


gioboa (commented):

Image

PR #1313, opened 2026-09-16
fix: keep shared providers out of expose preloads

gu-stav:
A host can load 2 remotes that provide the same compatible non-eager dependency. Here, non-eager means `eager: false`: the runtime should load the remote's fallback only when it can't reuse a compatible provider. When the first provider is already loaded, preloading the second remote before navigation with:

```ts
await instance.preloadRemote([
  {
    nameOrAlias: '@namespace/viteViteRemoteSecondary',
    exposes: ['InstanceMarker'],
    resourceCategory: 'all',
  },
])
```

currently fetches the second remote's fallback provider through its expose preload list. This bypasses runtime share selection, so the runtime can't reuse the compatible provider that's already loaded.

[#1092](https://github.com/module-federation/vite/pull/1092) added expose closure collection, [#1243](https://github.com/module-federation/vite/pull/1243) deferred lazy provider assets in shared entries, [#1270](https://github.com/module-federation/vite/pull/1270) moved non-eager wrappers to expose async assets, and [#1310](https://github.com/module-federation/vite/pull/1310) completed nested dynamic closure traversal. Together, those paths can still place deferred provider assets in expose preloads.

This patch removes non-eager provider JS and CSS assets from both synchronous and asynchronous expose preload lists. The assets remain on each remote's shared entry as its fallback, which lets the runtime select and reuse an already-loaded compatible provider. Eager shares are unchanged.


gioboa (approved):
Thanks @gu-stav nice fix

PR #1314, opened 2026-09-16
fix: preserve compatible import:false shared providers

dmchoi77:
## Summary
Fix host auto-init checking only whether a provider is registered in the current share scope, so an incompatible provider could prevent adoption of a compatible provider from a sibling federation instance. For example, an `import:false` consumer requiring `^1.0.0` can now select a sibling `1.0.0` provider when the current share scope only has `2.0.0`. Failed `loadShare()` results are not written to the shared cache.

## Changes
- Add provider validation using the runtime `getRegisteredShare` selection rules and exclude `import:false` stubs. Non-singletons are checked against `requiredVersion` and `shareStrategy`, while singletons retain the existing runtime negotiation behavior when a real provider is present in the scope.
- Inspect sibling instances only when the current share scope has no usable provider. A sibling map must pass the same compatibility check before adoption; only real providers are copied, and existing real providers at the same version are preserved.
- Stop processing and caching a `loadShare()` result when it is `false` instead of normalizing it as a module.
- Add regression coverage for an incompatible provider in the current share scope and a compatible provider in a sibling scope, missing providers, existing sibling adoption, and protection against provider re-registration.

## Validation
- `pnpm exec vitest run src/virtualModules/__tests__/virtualRemoteEntry.test.ts`: 124 passed
- `pnpm test`: 53 files, 1,374 passed, 1 skipped
- `pnpm typecheck`: passed
- `pnpm fmt.check`: passed
- `pnpm build`: passed
- `pnpm test:integration`: 18 files, 99 passed
- `pnpm exec playwright test e2e/vite-vite/tests/host-preview.spec.ts`: 11 passed


gioboa (commented):
Thanks. A few reuse requests before merging:

1. **Reuse `__mfSelectSharedProvider`.** `__mfHasUsableProvider` duplicates logic already in `sharedProviderSelectionHelperCode`: the import:false filtering, the `getRegisteredShare` call, and the `{ emit: (params) => params }` hook (`__mfResolveShareHook`). Please emit `sharedProviderSelectionHelperCode` into hostAutoInit and reduce to:

   ```js
   const __mfHasUsableProvider = (versions, pkg, share, scopeName) => {
     const providers = Object.fromEntries(
       Object.entries(versions || {}).filter(([, p]) => p?.shareConfig?.import !== false)
     );
     if (Object.keys(providers).length === 0) return false;
     if (share.shareConfig?.singleton) return true;
     return Boolean(__mfSelectSharedProvider(providers, pkg, { ...share, scope: [scopeName] }, strategy));
   };
   ```

2. **Align the import alias.** remoteEntry imports `share as runtimeShare`; this PR uses `share as __mfRuntimeShare`. Use one name.

3. **`hasConfiguredShared`** re-inlines `Object.keys(options.shared ?? {}).length > 0`, now the 5th copy (`virtualRemoteEntry.ts:1224`, `index.ts:710/727/1965`). Please extract a small util and use it here.

Nit: `Array.isArray(share.scope) ? share.scope : [share.scope || 'default']` now appears twice in hostAutoInit plus once in `__mfSelectSharedProvider`; a shared scope-names helper would tidy this too.

dmchoi77:
@gioboa Looks like it’s already been fixed and merged. Thanks for the review!

gioboa:
> @gioboa Looks like it’s already been fixed and merged. Thanks for the review!

Yeah I did that. Thanks for this great improvement @dmchoi77

PR #1315, opened 2026-09-16
fix: preserve singleton sharing with commented exports

gu-stav:
Comments between export tokens can cause a shared dependency to bypass singleton negotiation. The export scanner treats valid JavaScript as incomplete and generates a wrapper that uses the local module instead of the negotiated provider.

For example, both host and remote can share a dependency with:

```ts
federation({
  name: 'host', // use 'remote' in the remote configuration
  shared: {
    'shared-lib': { singleton: true },
  },
})
```

If the dependency contains comments between its export tokens:

```js
// shared-lib/index.js
export/* version */const version = '1.0.0'
export/* provider */default { getVersion: () => version }
```

then a host using version 1.0.0 can keep its local copy even when the remote provides version 1.5.0. Removing the comments makes both use the higher version again.

This patch uses the existing code-position lexer to identify actual comments and replace them with whitespace before named and mutable export analysis. It preserves source offsets and line breaks, and leaves strings, template literals, and regular expressions intact, including module specifiers in re-exports.

Comment-separated exports can then use the generated shared bindings and participate in singleton negotiation. Syntax the scanner still cannot resolve retains the existing conservative fallback.


gioboa (approved):
Nice catch 👍

PR #1316, opened 2026-09-16
fix: infer shared versions from the configured project and import

gu-stav:
Shared version inference can advertise the wrong provider version and accept a different range than the application declared. It builds `requiredVersion` from the installed version, resolves replacement imports using the original shared key, and runs before Vite supplies `root`.

For example, a child project's `apps/host/package.json` can declare:

```json
{
  "dependencies": {
    "shared-lib": "~1.2.0",
    "replacement-lib": "~3.4.0"
  }
}
```

and share those dependencies with:

```ts
export default defineConfig({
  root: 'apps/host',
  plugins: [
    federation({
      name: 'host',
      shared: {
        'shared-lib': { singleton: true },
        'shared-alias': { import: 'replacement-lib', singleton: true },
      },
    }),
  ],
})
```

With `shared-lib` 1.2.3 installed, the required range becomes `^1.2.3` instead of `~1.2.0`. The alias gets its metadata from `shared-alias` instead of `replacement-lib`. If the dependencies are only installed in the child project, the provider version can be undefined because normalization looked in the parent project.

This patch reads the required range from the application's dependency declarations and the provider version from the installed package, using the configured import when present. It retains the original shared settings and refreshes inferred metadata when Vite supplies the project root, before generating shared modules, and again with the resolved root. Explicit version settings and the existing protocol fallbacks are preserved.


gioboa (approved):
👍 it looks great to me

PR #1318, opened 2026-09-17
fix: preserve shared module identity for absolute imports

gu-stav:
Absolute imports of a shared package's internal files can be replaced by the package-root wrapper. The resolver reduces the path to a package name, so an internal file with different exports can cause a production `MISSING_EXPORT` failure.

For example, an application can share a dependency with:

```ts
federation({
  name: 'host',
  shared: {
    'audit-lib': { singleton: true },
  },
})
```

If an alias or another plugin resolves an import to an internal file:

```js
// node_modules/audit-lib/index.js
export const value = 'public'

// node_modules/audit-lib/internal.js
export const secret = 'internal'

// Application import after resolution
import { secret } from '/project/node_modules/audit-lib/internal.js'
```

the plugin substitutes the wrapper for `audit-lib`, which exports `value` but not `secret`.

This patch uses Vite's resolver to compare the requested file with configured shared entry points. Internal files keep their original imports, and matching subpaths keep their own wrapper identity. Entry lookup follows Vite's resolution settings, skips external or unresolved results, and avoids recursively entering federation resolution. Private package exports are checked before asking Vite to resolve them. Tree-shaking analysis awaits the same lookup before completing export collection.

Matching equivalent entries across installations remains supported without a new option. The comparison retains the full suffix after the last `node_modules/`, so a different installation directory does not turn an internal file into a package-root import.

A single resolver in `src/utils/sharedSource.ts` owns entry matching, safe Vite resolution, query handling, and caching. The plugin supplies its current root and conditions and clears the resolver on rebuilds or watched-file changes. Hook types are derived from Vite's `Plugin` interface. Existing package-path and Vite file-URL/query helpers are reused, and the unused filename-based matchers are removed.

Completed lookups, including misses, are cached per federation instance and Vite environment. Requests with options that can change resolution bypass the cache; pending lookups remain independent to avoid deadlocks when Vite re-enters resolution. Vite cache queries remain supported, while resource queries such as `?raw` stay with Vite's loaders.

As part of consolidating path utilities, `normalizeNodeModulePath` now strips hash fragments as well as queries. Shared-source matching still excludes fragment-bearing imports before normalization.


gu-stav:
Looking into the test failure.

gioboa (commented):
Thanks, the entry-identity approach is right and the regression tests are solid. Two requests before merging:

1. No new public option. Please drop allowNodeModulesSuffixMatch from ShareItem, normalizeShareItem and the option types in normalizeModuleFederationOptions.ts. Cross-installation entry matching was the pre-PR behavior, so in getSharedSource (pluginProxySharedModule_preBuild.ts) make the getNodeModulesSuffix(entry) === suffix branch unconditional, and merge the first two cases in shared-absolute-imports.test.ts accordingly.
2. Reuse utils. Move getSharedSource, getNodeModulesSuffix, the /@fs/ strip and the query allow-list into src/utils (see getPackageNameFromNodeModulePath, VITE_FS_PREFIX, stripQueryAndHash), and delete the now-unused getMatchingNodeModuleSubpath / getCommonSharedSubpathFromNodeModulePath in pathNormalization.ts.

gioboa:
Thanks, three notes:

1. normalizeNodeModulePath now strips # too (via stripQueryAndHash). It has six callers; looks safe, but it's a silent behavior change worth a line in the description.
2. Resolve cost. resolveSharedSource issues 1 + N this.resolve() calls per node_modules import, in both resolveId and transform, with no memoization. A per-build cache keyed by normalizedSource in getSharedSource would keep large dev graphs fast.
3. collectTreeShakingImports returns Promise on one path and undefined on the early returns. The caller awaits so it works, but returning Promise.resolve() on all paths would make the signature honest.

gioboa (approved):
That's awesome to me 👏 @gu-stav

gioboa (commented):
Can I mark it as ready @gu-stav ?

PR #1320, opened 2026-09-17
fix: preserve shared metadata for nested prefixes

gu-stav:
Shared imports matched by a nested prefix can lose their configuration before runtime registration. The import matcher recognizes the prefix, but the metadata lookup only checks the exact import, the package name, and the package name with a trailing slash. It misses the nested prefix that holds the inferred version and shared settings.

For example, both host and remote can share a package's feature entrypoints with:

```ts
federation({
  name: 'host', // use 'remote' in the remote configuration
  shareStrategy: 'version-first',
  shared: {
    'audit-lib/features/': {
      singleton: true,
      requiredVersion: '*',
    },
  },
})
```

and import an entrypoint with:

```js
import { Button } from 'audit-lib/features/button'
```

If the host has `audit-lib@2.0.0` installed and the remote has `audit-lib@1.0.0`, each entrypoint should inherit its installation's package version. With this configuration, both imports should use the host's higher-version provider.

Instead, the metadata lookup checks `audit-lib/features/button`, `audit-lib`, and `audit-lib/`, but never `audit-lib/features/`. Without a matching configuration, the generated runtime registration omits the provider. If a package-level configuration also exists, it can be selected instead of the nested prefix's settings. The browser reproduction selected the lower-version remote implementation in the host.

This patch makes metadata lookup reuse the existing shared-key matcher used for import interception. Exact entries retain precedence, and nested prefixes resolve to the most specific matching configuration before the existing package-level fallbacks. Matching entrypoints retain the inferred package version and settings such as `singleton`, `requiredVersion`, `shareScope`, and `import` when registered with the runtime.

The browser regression verifies that the host and remote both use the host's `2.0.0` implementation.


gioboa (approved):
Thanks @gu-stav nice fix

PR #1321, opened 2026-09-17
fix: harden shared version inference

gioboa:
Infer versions from the share key for path imports, reject non-semver package.json ranges, cache package.json reads, keep eager version on root miss.


gioboa (commented):

Image

PR #1322, opened 2026-09-17
fix: only seed the package root when importable

simonihmig:
`expandSharedPrefixKey` unconditionally seeded the bare package name as a materialized loadShare wrapper for trailing-slash shared prefixes. That base is not always importable: packages that only publish subpaths (no "." export) fail to resolve when the plugin emits `import * as x from "<base>"` for an eager host.


src/utils/packageUtils.ts
```diff
@@ -164,14 +164,18 @@ function getPackageExportsTarget(pkg: string, packageName: string, exportsField:
 export function isPackageExportAvailable(pkg: string, opts: PackageEntryConditions = {}): boolean {
   const packageName = getPackageName(pkg);
   const installed = getInstalledPackageJson(packageName, opts);
-  if (installed?.packageJson.exports == null) return true;
+
+  if (!installed) return false;
```
simonihmig: This is changing the behaviour of this function. I assume returning false here if the pjson cannot be found is the right call? But this could have an unexpected effect maybe on the only existing consumer [here](https://github.com/module-federation/vite/blob/d3fb556e99a382950a67d1ff188bc5954fd0fdd7/src/utils/sharedSource.ts#L117), so please double check!

src/virtualModules/virtualRemoteEntry.ts
```diff
@@ -13,20 +13,23 @@ import {
   getPackageName,
   getSharedCacheDescriptor,
   hasPackageDependency,
+  isPackageExportAvailable,
   packageNameEncode,
   sharedCacheHelperCode,
 } from '../utils/packageUtils';
 import { serializeRuntimeOptions, toSafeJsLiteral } from '../utils/serializeRuntimeOptions';
 import { SSR_ONLY_RUNTIME_PLUGINS } from '../utils/ssrCapabilities';
+import { getTreeShakingExportUsage } from '../utils/treeShaking';
```
simonihmig, line 23: There is a bit of unrelated noise here due to my VS Code reordering imports on save, but I assume this is expected given [this config](https://github.com/module-federation/vite/blob/main/.vscode/settings.json#L3)? I can revert if not wanted.

gioboa (commented):
Thanks @simonihmig 
Core fix is correct and the exports tests are thorough. One concern: changing isPackageExportAvailable to return false for missing packages also affects the sharedSource.ts pre-filter, turning "can't verify, let context.resolve() decide" into a hard gate. That silently drops candidates Vite would resolve, e.g. aliased shares (react mapped to preact/compat, react uninstalled) or non-hoisted pnpm deps. Suggest two separate functions instead: keep isPackageExportAvailable lenient as before, add isPackageInstalled, and use isPackageInstalled(base) && isPackageExportAvailable(base) in expandSharedPrefixKey. That keeps sharedSource.ts untouched and makes each check explicit. Nit: rename mf-test-legacy-subpath-reject, since that test now asserts acceptance.

src/virtualModules/virtualRemoteEntry.ts
gioboa: that's fine. TY

simonihmig:
Yeah, that was the thing I was not really confident with, hence the other comment by me. I'll change that. 

What made me try this more simplified approach first was the fact that it didn't break any tests. I haven't digged really deep into what's happening in sharedSource.ts, but do you think it would be feasible to add a (failing on my branch) test that covers the case you described?

gioboa (commented):
>  do you think it would be feasible to add a (failing on my branch) test that covers the case you described?

Yes, let me do that

gioboa:
Let me know what you think of [a780f01](https://github.com/module-federation/vite/pull/1322/commits/a780f01182cb46393b335e247d17f816604ef2c1) :pray:

src/utils/packageUtils.ts
gioboa: Fixed

simonihmig:
LGTM, thank you for taking this over!

gioboa (approved):
Thanks @simonihmig for your help

Issue #1265, opened 2026-09-11
[Bug]: Eager shared dependencies generate excessive `__loadShare__` request fan-out


ak274:
## Description

In a Vite/Rolldown host with many eager singleton dependencies, the generated
browser graph can contain one federation `__loadShare__` wrapper chunk per
shared key. A multi-scope configuration amplifies the effect: the browser must
request and evaluate many small federation wrappers before the first exposed
remote can render.

The plugin also manages federation chunk grouping for Rolldown. In the current
behavior, user-defined `output.manualChunks` cannot be used as a supported way
to coalesce these wrappers, which makes it difficult to tune startup behavior
for a latency-sensitive shell.

## Environment

- `@module-federation/vite`: `1.21.6`
- `@module-federation/runtime`: `2.9.0`
- Vite/Rolldown: `8.2.0`
- Share strategy: `loaded-first`
- Share scopes: `default` and a second deferred scope such as `heavy-ui`
- Browser: Chromium-based browser

## Steps to Reproduce

1. Create a Vite host and a dynamically loaded Vite remote.
2. Configure the host with multiple share scopes, for example:

   ```ts
   shareScope: ['default', 'heavy-ui'];
   ```

3. Configure at least ten eager singleton packages in the `default` scope and
   at least one deferred singleton in `heavy-ui`.
4. Configure the remote to expose an `App` module and consume several of the
   shared packages.
5. Build the host with Vite 8/Rolldown and serve it over HTTP.
6. Open the host's initial route and record the resource timeline before and
   after loading the remote.
7. Count generated files containing `__loadShare__` and measure their encoded
   sizes and start/end times.

## Actual Behavior

The host generates separate `__loadShare__` wrapper resources for the shared
keys. In a production-like multi-scope application, one initial route produced:

- 34 federation wrapper resources;
- approximately 1.26 MB of combined encoded wrapper content;
- an approximately 872 ms span from the first to the last wrapper resource;
- a largest UI-foundation wrapper of approximately 248 KB encoded.

The deferred heavy scope was not requested on the initial route, so the
measurement is primarily the cost of the eager core shared set rather than the
heavy UI provider.

The generated plugin behavior is equivalent to assigning federation-owned
chunks to IDs containing `__loadShare__`. On the Rolldown path, federation
chunk groups take precedence over user grouping and `output.manualChunks` is
removed or ignored for the federation-managed portion of the graph.

## Expected Behavior

The plugin should preserve singleton and share-scope semantics while providing
[...]

PR #1323, opened 2026-09-18
fix: coalesce eager shared wrappers on the Rollup path too

simonihmig:
mfChunkName coalesced eager loadShare wrappers into a single loadShare-eager chunk only when output.codeSplitting.groups was in play (Rolldown / Vite 8+). On the Rollup path (Vite 5-7) every eager share still got its own wrapper chunk, causing the request fan-out that #1265 reported for hosts that mark every share eager -- an eager share is loaded during init by definition, so isolating its wrapper buys nothing on either bundler.


src/index.ts
```diff
@@ -1604,7 +1604,7 @@ function federation(mfUserOptions: ModuleFederationOptions): any[] {
             if (id.includes(LOAD_SHARE_TAG)) {
               const pkg = getCachedLoadSharePkg(id);
               const key = pkg && findSharedKey(pkg, shared);
-              if (useCodeSplitting && key && shared[key].shareConfig.eager === true) {
+              if (key && shared[key].shareConfig.eager === true) {
```
simonihmig, line 1607: I am not sure what the reason was to guard this on the `useCodeSplitting` (rolldown) flag, but I had this change here in a patch for our app (still on Vite 7) and it was working fine AFAICT.

gioboa: Thanks

gioboa (approved):
Nice catch 👏 TY @simonihmig

gioboa (commented):

Image

PR #1325, opened 2026-09-18
fix: defer shared bridging to getExposes to prevent init deadlock

### LINKED ISSUE #1326: [Bug]: Vite remote `init()` deadlocks webpack host when bridging shared modules of the same version
### Summary

When a webpack host loads Vite remotes that share the **same React version** as the host, the page intermittently renders nothing (blank page / infinite hang). The root cause is a circular await chain between `loadShare`, `initializeSharing`, and the Vite remote's `init()`.

### Deadlock chain

1. Webpack host's `import('./bootstrap.tsx')` compiles to `Promise.all([__webpack_require__.e("webpack_sharing_consume_react-19_react_react"), ...])` — the react sharing-consume chunk is a **hard dependency** that blocks bootstrap loading
2. The consume chunk handler calls `loadShare("react")` on the host
3. `loadShare` unconditionally awaits `initializeSharing`, which includes promises for **all remote inits** (including the Vite remotes)
4. Vite remote's `init()` eagerly runs `await __mfBridgeSharedProviders()` which calls `__mfBridgeExternalSharedProvider` → `__mfLoadPinnedRuntimeShare` → `initRes.loadShare(pkg)` back on the host
5. That `loadShare` finds the host's webpack `get()` function, which loads the same react sharing-consume chunk
6. The consume handler calls `loadShare("react")` on the host again → `initializeSharing` → returns the same cached promise that includes the Vite remote's still-pending `init()` → **DEADLOCK**

The deadlock only affects Vite remotes sharing the **same version** of React as the host. Different versions take the local seed path via `__mfNeedsPreInitSeedBarrier` and are immune.

### Where the bug lives

In `virtualRemoteEntry.ts`, the generated remote entry eagerly awaits `__mfBridgeSharedProviders()` during `init()`:

```js
await __mfBridgeSharedProviders();
if (__mfUsesWebpackShareScope) {
  __mfLateBridgeShared = __mfBridgeSharedProviders;
}
// ... global instances bridging also calls __mfBridgeExternalSharedProvider eagerly ...
```

The `__mfLateBridgeShared` deferral mechanism already exists and runs in `getExposes()`, but it is only assigned conditionally when `__mfUsesWebpackShareScope` is true. The eager `await __mfBridgeSharedProviders()` on the line before always runs and triggers the re-entrant `loadShare` that causes the deadlock.

### Fix

Always defer bridging to `getExposes()` instead of running it during `init()`. By the time `getExposes()` runs, all remote inits have completed and `initializeSharing` resolves without circular waits.

```js
__mfLateBridgeShared = async () => {
  await __mfBridgeSharedProviders();
  await __mfBridgeGlobalSharedProviders();
};
```

### Note on `@module-federation/runtime-core`

There is also a deeper architectural concern in the core runtime: `loadShare()` unconditionally awaits `initializeSharing()` which awaits **all** remote init promises. Any remote that calls `loadShare` during its `init()` can trigger this cycle — not just Vite remotes. The vite plugin fix breaks the cycle from the Vite side, but the core runtime could independently be made more resilient by not blocking `loadShare` on unrelated remote inits (e.g. splitting "ensure scope exists" from "ensure all remotes initialized").

### Version

v1.22.1

### Reproduction

https://github.com/RalphK66/mf-vite-init-deadlock-repro

### Relevant log output

```shell

```

### Validations

RalphK66:
closes #1326

## Summary

Fixes a deadlock where a webpack host loading Vite remotes that share the same React version hangs indefinitely — `import('./bootstrap')` never resolves.

### Deadlock chain

1. Webpack host's `import('./bootstrap')` → sharing-consume chunk → `loadShare("react")` → `initializeSharing` → awaits all remote init promises
2. Vite remote's `init()` eagerly runs `await __mfBridgeSharedProviders()` → `__mfBridgeExternalSharedProvider` → `loadShare(pkg)` back on the host
3. Host's `loadShare` → `initializeSharing` → same cached promise → awaits the Vite remote's still-pending `init()` → **circular deadlock**

### Fix

The `__mfLateBridgeShared` deferral mechanism already exists and runs in `getExposes()`. This change assigns it **unconditionally** instead of eagerly awaiting `__mfBridgeSharedProviders()` during `init()`. Both the materialized-batch bridging and the global-instances bridging are wrapped into a single deferred function. By the time `getExposes()` runs, all remote inits have completed and `initializeSharing` resolves without circular waits.

**Before (deadlocks):**
```js
await __mfBridgeSharedProviders();
if (__mfUsesWebpackShareScope) {
  __mfLateBridgeShared = __mfBridgeSharedProviders;
}
// ... global instances bridging runs eagerly too ...
```

**After (defers):**
```js
__mfLateBridgeShared = async () => {
  await __mfBridgeSharedProviders();
  await __mfBridgeGlobalSharedProviders();
};
```

### Reproduction

Minimal repro: https://github.com/RalphK66/mf-vite-init-deadlock-repro

1 webpack host + 2 Vite remotes, all sharing React 19 at the same version. The host page deadlocks on every load — `import('./bootstrap')` never resolves.

### Note on runtime-core

`loadShare()` in `@module-federation/runtime-core` unconditionally awaits `initializeSharing()` which includes ALL remote init promises. Any remote that calls `loadShare` during its `init()` can trigger this cycle. This PR breaks the cycle from the Vite side; the core runtime could independently be made more resilient.

## Test plan

- [x] All 1398 existing tests pass (one test updated to reflect the unconditional deferral)
- [x] Verified fix against minimal repro (1 webpack host + 2 Vite remotes)
- [x] Verified fix against larger topology (webpack host + 6 Vite remotes + 6 Vite relays with mixed React versions)


gioboa (commented):
Skip the eager bridge in init() for webpack hosts; their loadShare() awaits this init, so bridging re-entered it and hung. Regression from #1115. Verified on the #1326 and #1064 repros, unit 1398, e2e 29/29, integration 28/28.

@RalphK66 thanks for the precise diagnosis and the repro, they made this easy to pin down. We're landing a narrower version of your fix: same idea, but the deferral only applies when the host's share scope is webpack's, so Vite hosts keep bridging inside init() as before.

gioboa (approved):
Thanks @RalphK66

gioboa:
@RalphK66 two more commits on top. While building an e2e for this I found that deferring the bridge alone leaves the remote on its own React copy with a Vite 6/7 dev remote (add a useState to your widget to see it), because a subpath share seeded during init() evaluates the react proxy before the bridge runs. The first commit skips that seeding in dev when the provider is still pending; the second adds a webpack + rspack host fixture so both the deadlock and single-React rendering stay covered.

